PR #129, opened 2014-12-23
Allow tree property to be changed later

pgiraud:
User may want to change the tree.


pgiraud:
I agree that it's a good option to let the user decide if he wants to have one-time binding.

elemoine:
I did not expect that you would merge both commits without discussions. I think we should revert the second commit and have more discussion. Sorry for the confusion.

PR #131, opened 2015-01-06
Revert "Make map directive watch map expr"

elemoine:
This pull request reverts commit 2dd3937b290f177339bbef292743fcfdacb5b8a7, which was added as part as https://github.com/camptocamp/ngeo/pull/129.

2dd3937b290f177339bbef292743fcfdacb5b8a7 may make sense, but I think it requires more thinking. And if we decide to make the map directive watch map changes we need to do the same in the other directives that use a map, namely `ngeoControl`, `ngeoLayertree`, `ngeoLayertreenode`, and `ngeoResizemap`.

The question is: do we want to support changing the map during the lifetime of the application?

### DIFF
--- a/examples/animation.js
+++ b/examples/animation.js
@@ -33,7 +33,7 @@ app.mapDirective = function() {
     controller: function() {},
     controllerAs: 'ctrl',
     bindToController: true,
-    template: '<div ngeo-map="::ctrl.map"></div>'
+    template: '<div ngeo-map="ctrl.map"></div>'
   };
 };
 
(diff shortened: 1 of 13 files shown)


asaunier:
> The question is: do we want to support changing the map during the lifetime of the application?

Would you (or anyone) have use cases that would require such a feature?

elemoine:
> ?
> 
> Would you (or anyone) have use cases that would require such a feature?

I don't. This is why I'm wondering if it's worth changing all our directives for that.

asaunier:
Then it's probably no use making such changes if there is no hint they could be necessary.

elemoine:
> Then it's probably no use making such changes if there is no hint they could be necessary.

Yep, agree.

PR #132, opened 2015-01-06
Assign a uid to tree and child nodes

pgiraud:


elemoine:
Looks good to me.

But perhaps this needs explanations here in the PR. The user of the `ngeoLayertree` and `ngeoLayertreenode` directives can use the `uid`s in his partials, if using `id`'s in HTML elements is required (because of Bootstrap for example).

asaunier:
> The user of the ngeoLayertree and ngeoLayertreenode directives can use the uids in his partials, if using id's in HTML elements is required (because of Bootstrap for example).

Would you have an example? In Lux maybe?

pgiraud:
Yes, see https://github.com/Geoportail-Luxembourg/geoportailv3/commit/55c07d3bd99bc690418c726f8534208fe39ff2c0

elemoine:
> Would you have an example? In Lux maybe?

Adding ids to HTML elements in directive partials is not a good practice, because directives may be used multiple times in a page. But, unfortunately, it looks like some Bootstrap components require adding ids to elements. See the [Bootstrap collapse component](http://getbootstrap.com/javascript/#collapse) for example.

PR #133, opened 2015-01-06
Add a backgroundlayer example

elemoine:
This PR adds a `backgroundlayer` example demonstrating how to implement a background layer selector directive. The directive is based on Angular's `select`, `ngOptions`, `ngModel` and `ngChange` directives, and on the new `ngeoBackgroundLayerMgr` service. That service, added to the core of ngeo, eases the addition and removal of background layers to/from maps.


elemoine:
The example is online at http://erilem.net/ngeo/backgroundlayer/backgroundlayer.html.

asaunier:
Looks good, thanks!

I just don't really understand how the mapUid thing works?!

elemoine:
> Looks good, thanks!


Thanks a lot for the review.

> I just don't really understand how the mapUid thing works?!

This is to track the maps that have/don't have a background layer. A background layer is always set at index 0 in a map's layers collection. But the layer at index 0 may not be a background layer. If `mapUids_` is true for a given map then it means that the layer at index 0 is a background layer. If `mapUids_` is `undefined` for a given map then it means that layer at index 0 is not a background layer. And when you set a background layer in a map which does not get have a background layer then the `insertAt` function is used, otherwise the `setAt` function is used to replace the background layer by the new one.

I hope this clarifies.

asaunier:
> I hope this clarifies.

I does! Thanks for the explanations. For me it's OK to merge.

elemoine:
My plan is to merge this after some testing in the Lux project. 

Thanks again for the review.

elemoine:
I've just added new commits to add a new example: a background layer selector directive based on Angular UI Bootstrap's `dropdown` directive. The new example can be seen here: erilem.net/ngeo/backgroundlayer/backgroundlayerdropdown.html.

elemoine:
I finally decided against using Angular UI Bootstrap's in the `backgroundlayerdropdown` example. The example is now just based on Bootstrap's `dropdown` jQuery plugin.

PR #134, opened 2015-01-06
Use one-time binding in layermanager example

elemoine:
This PR simplifies the layermanager example and uses one-time binding where possible for that example.


pgiraud:
Looks good to me. Thanks. Please merge.

PR #139, opened 2015-01-09
Fix bug in layertree example

elemoine:


asaunier:
Do you mean that in partial `layertreenode.html`, params "node" and "layer" were not actually set when passed to `ctrl.onButtonClick(node, layer)`?

If yes, the PR looks good to me.

elemoine:
> Do you mean that in partial layertreenode.html, params "node" and "layer" were not actually set when passed to ctrl.onButtonClick(node, layer)?

Yes that's it. But in fact, `node` was set, but by chance. It is actually set by the ng-repeat directive on the parent scope. If the variable name was for example `n` instead than `node` then `node` would not have been set. `layer` was not set at all.

> 
> If yes, the PR looks good to me.

PR #140, opened 2015-01-13
Replace the layer tree layer factory by a layer expression

elemoine:
Currently, when using the `ngeoLayertree` and/or `ngeoLayertreenode` directives the `ngeoLayertreeLayerFactory` service should be defined by the application (this service is a function responsible for returning a layer for a tree node). This works well when only one layer tree is used in the page, but it is inflexible when multiple layer trees are used, because the layer factory function is the same for all the layer trees.

This PR fixes the problem by adding a new attribute to the `ngeoLayertree` and `ngeoLayertreenode` directives. This new attribute allows specifying a a layer expression, which, in most cases, will be a function called on the application controller. This does add some complexity to the user of the directive, but it is also more flexible.

I also took the opportunity to improve the documentation of the `ngeoLayertree` and `ngeoLayertreenode` directives.

### DIFF
--- a/examples/layertree.js
+++ b/examples/layertree.js
@@ -50,7 +50,9 @@ app.layertreeDirective = function() {
     // use "::ctrl.tree" for the "tree" expression as we know the
     // layer tree won't change
     template: '<div ngeo-layertree="::ctrl.tree" ' +
-        'ngeo-layertree-map="ctrl.map"></div>'
+        'ngeo-layertree-map="ctrl.map" ' +
+        'ngeo-layertree-layer="ctrl.getLayer(node)">' +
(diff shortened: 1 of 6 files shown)


asaunier:
The part in the example files is quite clear.
I had to struggle a little bit more to understand the "layerExpr" part in the partials/layertreenode.html because of the recursion (it's a bit confusing all those layerExpr) but it makes sense now :)
For me it's OK to merge the PR. Thanks!

elemoine:
Thanks @asaunier.

@pgiraud, note that this requires changes to applications. I'll look at it tomorrow for the Lux project.

PR #141, opened 2015-01-14
Layertree levels

pgiraud:
This pull request addresses two goals. It gives the user the opportunity to get for each node:
 - the level depth,
 - the uid of the parent element. This corresponds to tree uid for the nodes of the first level.

One purpose of this is for example to get only one node expanded at a time for a given level when using bootstrap and collapse/accordion.

Please review.


asaunier:
For me, it makes sense.

elemoine:
+1 from me as well. Please merge.

PR #143, opened 2015-01-15
Make control directive rely on control.setTarget

elemoine:
This is WIP depending on https://github.com/openlayers/ol3/pull/3139.

### DIFF
--- a/examples/control.html
+++ b/examples/control.html
@@ -18,7 +18,7 @@
   </head>
   <body ng-controller="MainController as ctrl">
     <div id="map" ngeo-map="ctrl.map"></div>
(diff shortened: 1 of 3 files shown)


elemoine:
This is now ready for review. With this the control directive relies on a control being made available in the parent scope, as opposed to a function that returned a control.

Please review.

pgiraud:
It makes things simpler. I like it.
This looks good. Please merge.

elemoine:
And the problem was that the function was executed with `window` as its scope, instead of the controller as one would have expected.

asaunier:
I guess this implies changes in projects that use the ngeo-control directive?

elemoine:
> I guess this implies changes in projects that use the ngeo-control directive?

Yes, the user just needs to provide an expression that evaluates to a control (with no target and not already added to the map).

PR #144, opened 2015-01-16
Add note about closure externs file

tsauerwein:
... when using `ngeo.GetBrowserLanguage`.

### DIFF
--- a/src/services/getbrowserlanguage.js
+++ b/src/services/getbrowserlanguage.js
@@ -2,6 +2,8 @@
  * @fileoverview Provides a function that returns the most appropriate 2-letter
  * language code depending on the list of available languages and the browser
  * languages settings.
+ * If you compile your code with the Closure Compiler, the externs file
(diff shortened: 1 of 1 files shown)


src/services/getbrowserlanguage.js
```diff
@@ -2,6 +2,8 @@
  * @fileoverview Provides a function that returns the most appropriate 2-letter
  * language code depending on the list of available languages and the browser
  * languages settings.
+ * To use this module, the externs file `externs/closure-compiler.js` has to
```
elemoine: I'll add "if you compile your code with the Closure Compiler", because using ngeo does not require using the compiler. I'll also specify that this externs is in ngeo, as its name may suggest it's in the compiler.

tsauerwein:
Ok, I changed the wording.

elemoine:
Thanks. Good to merge to me.

PR #145, opened 2015-01-16
Improvements to layertree directive

elemoine:
As discussed with @pgiraud this PR brings further improvements to the layertree directive. In particular the `ngeo-layertree-layer` attr is renamed to `ngeo-layertree-nodelayer`.

### DIFF
--- a/examples/layertree.js
+++ b/examples/layertree.js
@@ -51,7 +51,7 @@ app.layertreeDirective = function() {
     // layer tree won't change
     template: '<div ngeo-layertree="::ctrl.tree" ' +
         'ngeo-layertree-map="ctrl.map" ' +
-        'ngeo-layertree-layer="ctrl.getLayer(node)">' +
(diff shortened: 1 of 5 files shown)


asaunier:
Why are the one-time bindings useless here?

elemoine:
> Why are the one-time bindings useless here?

Because the corresponding expressions are not watched within the directive. One-time binding is useful when there's some binding, while there's no binding at all in this case. The expressions here are just used to specify where to find the objects (the node and the map) in the parent scope.

elemoine:
And fyi I verified (using `countWatchers()`) that removing the `::` in these expressions does not increase the number of watchers.

asaunier:
Thanks for the explanations. OK for merging!

elemoine:
@pgiraud, ok for you too?

pgiraud:
OK

PR #146, opened 2015-01-16
Pass a number to layer.setOpacity

elemoine:
This PR fixes a bug where the decorate layer service passed a string rather than a number to `layer.setOpacity` function.


asaunier:
Makes sense.
Funny because the "main" commit changes only one character while the "secondary" commit impacts much more lines :P

elemoine:
> Funny because the "main" commit changes only one character while the "secondary" commit impacts much more lines :P

Don't worry, the compiler will optimize this away :-)

PR #147, opened 2015-01-21
Attempt to simplify tree node templates

pgiraud:
WIP!

This pull request is the result of a try at simplifying the layer tree. For now the main enhancement is that we have a ng-repeat loop only in one partial instead of two.
This is not ready for review and may not be pushed further.


elemoine:
Thanks for working on this, and creating this PR. This is good to have.

elemoine:
I'm closing this for now. We can reopen if we feel the need to.

PR #148, opened 2015-01-21
Sortable layers

pgiraud:
With this pull request I want to add a new sortable directive. It allows user to change the order of elements in a list using drag&drop. A new example is added to show how to use this directive to sort the layers on the map.
Please review.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -161,6 +161,8 @@ dist/ngeo.css: node_modules/openlayers/css/ol.css .build/node_modules.timestamp
 
 node_modules/angular/angular.min.js node_modules/angular-animate/angular-animate.min.js: .build/node_modules.timestamp
 
+bower_components/jquery-ui-sortable/jquery-ui-sortable.min.js: .build/bower_components.timestamp
+
 .PRECIOUS: .build/examples-hosted/%.html
 .build/examples-hosted/%.html: examples/%.html
 	mkdir -p $(dir $@)
--- a/Makefile
+++ b/Makefile
@@ -45,7 +45,7 @@ test: .build/ol-deps.js .build/ngeo-deps.js .build/node_modules.timestamp
 	./node_modules/karma/bin/karma start karma-conf.js --single-run
 
 .PHONY: serve
-serve:
+serve: .build/node_modules.timestamp .build/bower_components.timestamp
 	node buildtools/serve.js
 
 .PHONY: gh-pages
@@ -78,6 +78,7 @@ dist/ngeo.js: buildtools/ngeo.json \
 	          .build/externs/angular-1.3.js \
 		  .build/externs/angular-1.3-q.js \
 	          .build/externs/angular-1.3-http-promise.js \
+	          .build/externs/jquery-1.9.js \
 		  $(SRC_JS_FILES) \
 		  .build/templatecache.js \
 		  $(EXPORTS_JS_FILES) \
@@ -89,6 +90,7 @@ dist/ngeo-debug.js: buildtools/ngeo-debug.json \
 	          .build/externs/angular-1.3.js \
 		  .build/externs/angular-1.3-q.js \
 	          .build/externs/angular-1.3-http-promise.js \
+	          .build/externs/jquery-1.9.js \
 		  $(SRC_JS_FILES) \
 		  .build/templatecache.js \
 		  $(EXPORTS_JS_FILES) \
@@ -103,12 +105,13 @@ dist/ngeo.css: node_modules/openlayers/css/ol.css .build/node_modules.timestamp
 	./node_modules/.bin/cleancss $< > $@
 
 .build/examples/%.min.js: .build/examples/%.json $(SRC_JS_FILES) $(EXPORTS_JS_FILES) .build/externs/angular-1.3.js .build/externs/angular-1.3-q.js \
-	                      .build/externs/angular-1.3-http-promise.js examples/%.js .build/node_modules.timestamp
+	                      .build/externs/angular-1.3-http-promise.js .build/externs/jquery-1.9.js examples/%.js .build/node_modules.timestamp
+	           \
 	mkdir -p $(dir $@)
 	node buildtools/build.js $< $@
 
 .build/examples/all.min.js: buildtools/examples-all.json $(SRC_JS_FILES) $(EXPORTS_JS_FILES) .build/externs/angular-1.3.js .build/externs/angular-1.3-q.js \
-	                        .build/externs/angular-1.3-http-promise.js .build/examples/all.js .build/node_modules.timestamp
+	                        .build/externs/angular-1.3-http-promise.js .build/externs/jquery-1.9.js .build/examples/all.js .build/node_modules.timestamp
 	mkdir -p $(dir $@)
 	node buildtools/build.js $< $@
 
@@ -144,6 +147,10 @@ dist/ngeo.css: node_modules/openlayers/css/ol.css .build/node_modules.timestamp
 	mkdir -p $(dir $@)
 	cp $< $@
 
+.build/examples-hosted/jquery-ui-sortable.min.js: bower_components/jquery-ui-sortable/jquery-ui-sortable.min.js
+	mkdir -p $(dir $@)
+	cp $< $@
+
 .build/examples-hosted/partials: examples/partials
 	mkdir -p $@
 	cp examples/partials/* $@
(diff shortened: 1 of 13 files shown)


src/directives/sortable.js
```diff
@@ -0,0 +1,91 @@
...
+         */
+        function(scope, element, attrs) {
+          var attr = 'ngeoSortable';
+          var prop = attrs[attr];
+
+          var sortable = /** @type {Array} */ (scope.$eval(prop));
+
+          if (!angular.element.fn || !angular.element.fn.jquery) {
+            $log.error('ui.sortable: jQuery should be included before' +
+                'AngularJS!');
```
elemoine: We don't use the `$log` service elsewhere. Please use `goog.asserts.assert` instead. The advantage is that the compiler will remove the assertion.

```js
goog.asserts.assert(goog.isDef(angular.element.fn) && goog.isDef(angular.element.fn.jquery),
    'jQuery is required');
```

src/directives/sortable.js
```diff
@@ -0,0 +1,91 @@
...
+    scope: true,
+    link:
+        /**
+         * @param {angular.Scope} scope Scope.
+         * @param {angular.JQLite} element Element.
+         * @param {angular.Attributes} attrs Attributes.
+         */
+        function(scope, element, attrs) {
+          var attr = 'ngeoSortable';
+          var prop = attrs[attr];
```
elemoine: I've renamed `prop` to `expr` elsewhere. I'd recommend this:

```js
  var sortableExpr = attrs['ngeoSortable'];
```

src/directives/sortable.js
```diff
@@ -0,0 +1,91 @@
...
+            });
+            startIndex = undefined;
+          };
+
+
+          element.sortable();
+
+          angular.forEach(callbacks, function(value, key) {
+            element.sortable('option', key, value);
+          });
```
elemoine: This won't work if the `callbacks` object's `start` and `update` keys are renamed. It's just two options, so I don't think we need to use a loop.

Makefile
```diff
@@ -195,6 +205,11 @@ node_modules/angular/angular.min.js node_modules/angular-animate/angular-animate
 	mkdir -p $(dir $@)
 	touch $@
 
+.build/bower_components.timestamp: bower.json
+	bower install
```
elemoine: This did not work in my case. I think you need to use `./node_modules/bower/bin/bower`. Maybe it works for you because you already have bower globally installed.

elemoine: And this target should depend on `.build/node_modules.timestamp` I guess.

examples/layerorder.js
```diff
@@ -0,0 +1,97 @@
...
+    ],
+    view: new ol.View({
+      center: [16339075, -5194965],
+      zoom: 7
+    })
+  });
+
+  // watch any change on layers array to refresh the map
+  $scope.layers = this['map'].getLayers().getArray();
+  $scope.$watchCollection('layers', goog.bind(function() {
```
pgiraud: this doesn't work with the hosted examples since goog is unavailable.

elemoine: Yes, we can't use `goog` in the examples. See https://github.com/camptocamp/ngeo/blob/master/docs/developer-guide.md#authoring-examples.

src/directives/sortable.js
```diff
@@ -0,0 +1,79 @@
...
+            $timeout(function() {
+              element.sortable('refresh');
+            }, 0, false);
+          });
+
+          element.sortable();
+
+          element.sortable('option', 'start', function(e, ui) {
+            // Save the starting position of dragged item
+            startIndex = ui.item.index();
```
elemoine: `item` may be renamed here.

examples/layerorder.js
elemoine: Instead of adding `layers` to the scope you could use following:

```js
  var map = this['map'];
  $scope.$watchCollection(function() {
    return map.getLayers().getArray();
  }, function() {
    map.render();
  });
```

src/directives/sortable.js
```diff
@@ -0,0 +1,79 @@
...
+          var sortable = /** @type {Array} */ (scope.$eval(sortableExpr));
+
+          goog.asserts.assert(goog.isDef(angular.element.fn) &&
+              goog.isDef(angular.element.fn.jquery), 'jQuery is required');
+
+          var startIndex;
+
+          // When we add or remove elements, we need the sortable to 'refresh'
+          // so it can find the new/removed elements.
+          scope.$watch('sortable.length', function() {
```
elemoine: `sortable` in not in the `scope` so this `$watch` has no effect.

examples/layerorder.html
```diff
@@ -0,0 +1,55 @@
...
+  <body ng-controller="MainController as ctrl">
+    <div id="map" ngeo-map="ctrl.map"></div>
+    <p>
+      Change order of layers on the map by changing the order of items in the
+      list below.
+      <br>
+      Hint: drag-drop elements using the red handle.
+    </p>
+    <ul ngeo-sortable="ctrl.map.getLayers().getArray()">
+      <li ng-repeat="layer in ctrl.map.getLayers().getArray()">
```
elemoine: I find the `ctrl.map.getLayers().getArray()` expression too complex to be used in the page. We could have the main controller expose a `layers` property:

```js

  /** @type Array.<ol.layer.Layer> */
  this['layers'] = map.getLayers().getArray();
```

and use `ngeo-sortable="ctrl.layers"` in the template.

examples/layerorder.js
```diff
@@ -0,0 +1,97 @@
...
+app.MainController = function($scope, ngeoDecorateLayer) {
+
+  /** @type {ol.layer.Tile} */
+  var boundaries = new ol.layer.Tile({
+    source: new ol.source.TileWMS({
+      url: 'http://demo.opengeo.org/geoserver/wms',
+      params: {'LAYERS': 'topp:tasmania_state_boundaries'},
+      serverType: 'geoserver'
+    }),
+    name: 'Boundaries'
```
elemoine: I am picky but please use `set` for user properties:

```js
boundaries.set('name', 'Boundaries');
```

Although using a `name` option in the constructor options works here because the example code is not compiled, it won't work in a compiled application.

See https://github.com/camptocamp/ngeo/blob/master/docs/developer-guide.md#custom-olobject-properties to understand the reason.

examples/layerorder.js
elemoine: So, taking https://github.com/camptocamp/ngeo/pull/148/files#r23315083 into account, I'd use:

```js
  var map = this['map'];

  /** @type {Array.<ol.layer.Layer>} */
  this['layers'] = map.getLayers().getArray();
  var layers = this['layers'];

  $scope.$watchCollection(function() {
    return layers;
  }, function() {
    map.render();
  });
```

src/directives/sortable.js
```diff
@@ -0,0 +1,79 @@
...
+    link:
+        /**
+         * @param {angular.Scope} scope Scope.
+         * @param {angular.JQLite} element Element.
+         * @param {angular.Attributes} attrs Attributes.
+         */
+        function(scope, element, attrs) {
+          var sortableExpr = attrs['ngeoSortable'];
+
+          var sortable = /** @type {Array} */ (scope.$eval(sortableExpr));
```
elemoine: You can more strict with the typing and use `@type {Array.<ol.layer.Base>}`.

src/directives/sortable.js
```diff
@@ -0,0 +1,79 @@
...
+
+          goog.asserts.assert(goog.isDef(angular.element.fn) &&
```
elemoine: Adding `goog.require('goog.asserts')` would make sense.

src/directives/sortable.js
elemoine: I think we need to add more extern definitions.

elemoine:
We need to think about the case where only a subset of layers should be sortable. But we can probably merge this and address that case later, possibly with a new PR.

pgiraud:
This is ready for review now.

elemoine:
Please merge.

pgiraud:
What's missing:
[] reverse option,
[] call refresh on sortable when items are added,
[] improvements in ArraySync (indices)

elemoine:
For everyone's information, I've started working on a branch that uses Closure Library (`goog.fx.DragListGroup`) rather than jQuery UI Sortable in the `ngeoSortable` directive: https://github.com/elemoine/ngeo/tree/dragdrop.

Using Closure Library rather that jQuery UI would remove a dependency, which would be a good thing.

My branch includes code that kinda works, but more testing is required. And there's also a missing functionality: right now the entire item is draggable. Instead, it should be possible to define a specific element as the drag handle, as with the jQuery UI Sortable plugin.

I don't think I'm going to work on it again, but we may want/need to continue this work in the future.

PR #150, opened 2015-01-29
Add a scale selector example

elemoine:
This PR adds a scale selector example. This example creates a "scale selector" directive based on Bootstrap's `dropdown` jQuery plugin. The view resolution changes when a scale is selected in the scale selector. And the selected scale changes in the scale selector when the view resolution changes.

See http://erilem.net/ngeo/scaleselector/scaledropdown.html.

### DIFF
--- a/examples/backgroundlayerdropdown.html
+++ b/examples/backgroundlayerdropdown.html
@@ -23,7 +23,6 @@
     <div>
       This example shows how to write an application-specific directive (<code>app-backgroundlayer</code>) to create a background layer selector. The directive is based on Bootstrap's <code>dropdown</code> jQuery plugin, and on the <code>ngeoBackgroundLayerMgr</code> service.
     </div>
-
     <script src="../node_modules/jquery/dist/jquery.js"></script>
     <script src="../node_modules/angular/angular.js"></script>
     <script src="../node_modules/bootstrap/dist/js/bootstrap.js"></script>
--- a/examples/backgroundlayerdropdown.js
+++ b/examples/backgroundlayerdropdown.js
@@ -22,8 +22,8 @@ app.module = angular.module('app', ['ngeo']);
 /**
  * The application-specific background layer directive.
  *
- * The directive is based on Angular UI Bootstrap's dropdown and dropdownToggle
- * directives. It also uses Angular's ngRepeat and ngClick directives.
+ * The directive is based on Bootstrap's dropdown jQuery plugin and on
+ * the ngeoBackgroundLayerMgr service.
  *
  * @return {angular.Directive} Directive Defintion Object.
  */
--- a/examples/scaleselector.html
+++ b/examples/scaleselector.html
@@ -1,0 +1,38 @@
+<!DOCTYPE>
+<html ng-app='app'>
+  <head>
+    <title>Scale selector example</title>
+    <meta charset="utf-8">
+    <meta name="viewport"
+          content="initial-scale=1.0, user-scalable=no, width=device-width">
+    <meta name="apple-mobile-web-app-capable" content="yes">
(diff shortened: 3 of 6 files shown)


asaunier:
Thanks, it looks good.

What about adding a scaleselector directive in ngeo, not only an example? This would save people from writing their own directive, even if an example is provided.

Also I have found the scaleselector controller code difficult to read because it uses many functions that call each other along with event handlers. Isn't it possible to simplify it a bit? :P

examples/scaledropdown.js
```diff
@@ -0,0 +1,184 @@
...
+    '2': $sce.trustAsHtml('1&nbsp;:&nbsp;50\'000\'000'),
+    '3': $sce.trustAsHtml('1&nbsp;:&nbsp;25\'000\'000'),
+    '4': $sce.trustAsHtml('1&nbsp;:&nbsp;12\'000\'000')
+  };
+
+  /**
+   * @type {ol.Map}
+   */
+  var map = this['map'];
+  window.map = map;
```
elemoine: Remove this.

examples/scaledropdown.html
```diff
@@ -0,0 +1,38 @@
...
+          content="initial-scale=1.0, user-scalable=no, width=device-width">
+    <meta name="apple-mobile-web-app-capable" content="yes">
+    <link rel="stylesheet" href="../node_modules/openlayers/css/ol.css" type="text/css">
+    <link rel="stylesheet" href="../node_modules/bootstrap/dist/css/bootstrap.css" type="text/css">
+    <style>
+      #map {
+        height: 400px;
+      }
+      .dropdown button {
+        min-width: 100px;
```
elemoine: Make it bigger.

elemoine:
> What about adding a scaleselector directive in ngeo, not only an example? This would save people from writing their own directive, even if an example is provided.

It's a bit too specific to my taste at this point, because it relies on Bootstrap's dropdown plugin. See the background layer directive for examples. There are two examples, one based on the native `select` element, and one based on Bootstrap's `dropdown` plugin. Which one would you add to ngeo?

elemoine:
> Also I have found the scaleselector controller code difficult to read because it uses many functions that call each other along with event handlers. Isn't it possible to simplify it a bit? :P

I declare today that anyone requesting more simplicity should go ahead and try to implement the thing by himself or herself :) More seriously, I did it this way to please the compiler, and I don't find it too complex personally. Do you have concrete suggestions?

asaunier:
> Do you have concrete suggestions?

Of course not! :)

elemoine:
> It's a bit too specific to my taste at this point, because it relies on Bootstrap's dropdown plugin. See the background layer directive for examples. There are two examples, one based on the native select element, and one based on Bootstrap's dropdown plugin. Which one would you add to ngeo?

That being said I agree with you that having something in the library would be better. I'll think more about it.

elemoine:
PR updated, I created an `ngeoScaleselector` directive, as suggested by @asaunier.

asaunier:
> I created an ngeoScaleselector directive

Great, thanks!

elemoine:
Thanks @asaunier.

I've just modified the directive to make it possible to use the dropup variation of the Bootstrap dropdown plugin.

CC @pgiraud.

elemoine:
> I've just modified the directive to make it possible to use the dropup variation of the Bootstrap dropdown plugin.

Also demonstrate how we can pass pass options objects to directives. We may want to apply that principle to other directives in the future (for optional options).

asaunier:
> Also demonstrate how we can pass pass options objects to directives.

Yes, that's a good example!

elemoine:
I've made another change: the options object, instead of just the `dropup` option value, is now made available to the partial. In this way, user-specific options can be passed and used in a user-specific partial.

Thanks for the review. I'll merge this when the Travis build passes.

PR #151, opened 2015-01-29
Rename ngeoArraySync to ngeoSyncArrays

elemoine:
As discussed with @pgiraud this PR renames the `ngeoArraySync` service to `ngeoSyncArrays`. This service is a function so we name it after a verb, just like `ngeoDecorateLayer` and friends.


asaunier:
This makes sense!

PR #153, opened 2015-02-02
Popup

pgiraud:
OK, this pull request adds support for showing a popup window over the page. This popup is currently place at a top-center position and is not draggable.
More options can be added later if required.
It can be tested by clicking the `"i"` link in the `layertree` example.
Please review.

### DIFF
--- a/examples/layertree.js
+++ b/examples/layertree.js
@@ -5,9 +5,11 @@
 
 goog.provide('layertree');
 
+goog.require('ngeo.CreatePopup');
 goog.require('ngeo.layertreeDirective');
 goog.require('ngeo.layertreenodeDirective');
 goog.require('ngeo.mapDirective');
+goog.require('ngeo.popupDirective');
 goog.require('ol.Map');
 goog.require('ol.View');
 goog.require('ol.layer.Tile');
@@ -31,6 +33,10 @@ app.module.value('ngeoLayertreeTemplateUrl',
 // Use an application-specific "layer tree node" template.
 app.module.value('ngeoLayertreenodeTemplateUrl',
     'partials/layertreenode.html');
+
+// Use the default "popup" template.
+app.module.value('ngeoPopupTemplateUrl',
+    '../src/directives/partials/popup.html');
 
 
 /**
@@ -64,20 +70,46 @@ app.module.directive('appLayertree', app.layertreeDirective);
 /**
  * @constructor
  * @param {angular.$http} $http Angular http service.
+ * @param {angular.$sce} $sce Angular sce service.
  * @param {function(Object):ol.layer.Layer} appGetLayer Get layer service.
+ * @param {ngeo.CreatePopup} ngeoCreatePopup Popup service.
  * @ngInject
  * @export
  */
-app.LayertreeController = function($http, appGetLayer) {
+app.LayertreeController = function($http, $sce, appGetLayer, ngeoCreatePopup) {
   $http.get('data/tree.json').then(angular.bind(this, function(resp) {
     this['tree'] = resp.data;
   }));
 
   /**
    * @private
+   * @type {angular.$http}
+   */
+  this.http_ = $http;
+
+  /**
+   * @private
+   * @type {angular.$sce}
+   */
+  this.sce_ = $sce;
+
+  /**
+   * @private
    * @type {function(Object):ol.layer.Layer}
    */
   this.getLayer_ = appGetLayer;
+
+  /**
+   * @private
+   * @type {ngeo.Popup}
+   */
+  this.infoPopup_ = ngeoCreatePopup();
+
+  /**
+   * @type {Object.<string, !angular.$q.Promise>}
+   * @private
+   */
+  this.promises_ = {};
 };
 
 
@@ -100,7 +132,20 @@ app.LayertreeController.prototype.getLayer = function(node) {
  * @export
  */
 app.LayertreeController.prototype.onButtonClick = function(node, layer) {
-  window.alert(node['name'] + ', ' + layer.get('type'));
+  var layerType = node['layerType'];
+  if (!(layerType in this.promises_)) {
+    this.promises_[layerType] = this.http_.get('data/metadata.html').then(
+        angular.bind(this, function(resp) {
+          var html = this.sce_.trustAsHtml(resp.data);
+          return html;
+        }));
+  }
+  var infoPopup = this.infoPopup_;
+  this.promises_[layerType].then(function(html) {
+    infoPopup.setTitle(node['name']);
+    infoPopup.setContent(html);
+    infoPopup.show();
+  });
 };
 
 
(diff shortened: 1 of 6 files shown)


elemoine:
Note that you can put examples online, using the `gh-pages` target:

```
$ make gh-pages GITHUB_USERNAME=pgiraud
```

See https://github.com/camptocamp/ngeo/blob/master/docs/developer-guide.md#publish-examples-to-github-pages.

elemoine:
The build fail but that may not be related to your changes. I suppose `make check` passed on your local machine.

src/directives/partials/popup.html
```diff
@@ -0,0 +1,5 @@
+<h4 class="popover-title ngeo-popup-title">
+  <span translate>{{options.title}}</span>
```
elemoine: We're not supposed to use `translate` in ngeo at this stage.

elemoine: Also, I'm wondering whether we should start using `ng-bind` instead of `{{…}}` to avoid displaying ugly expressions in the UI at load time.

src/directives/popup.js
```diff
@@ -0,0 +1,65 @@
...
+         * @param {angular.Scope} scope Scope.
+         * @param {angular.JQLite} element Element.
+         * @param {angular.Attributes} attrs Attributes.
+         */
+        function(scope, element, attrs) {
+          scope.open = scope.open || false;
+          scope.options = scope.options || {title: ''};
+
+          element.addClass('popover');
+          var bodyWidth = goog.style.getSize(document.body).width;
```
elemoine: Add `goog.require('goog.style')`.

elemoine:
This looks good to me, but I think there are issues related to property renaming by the compiler.

examples/layertree.js
```diff
@@ -100,7 +126,12 @@ app.LayertreeController.prototype.getLayer = function(node) {
  * @export
  */
 app.LayertreeController.prototype.onButtonClick = function(node, layer) {
-  window.alert(node['name'] + ', ' + layer.get('type'));
+  this.http_.get('data/metadata.html').then(angular.bind(this, function(resp) {
+    var html = this.sce_.trustAsHtml(resp.data);
+    this.infoPopup_.setContent(html);
+  }));
```
elemoine: In that specific case it does not make sense to trigger the same request on a each click. You should rely on a the promise returned by `then` instead. Something like this:

```js
app.LayertreeController.prototype.onButtonClick = function(node, layer) {
  if (goog.isNull(this.metadataPromise_)) {
    this.metadataPromise_ = this.http_.get('data/metadata.html').then(function(resp) {
      var html = this.sce_.trustAsHtml(resp.data);
      return html;
    });
  }
  var infoPopup = this.infoPopup_;
  this.metadataPromise_.then(function(html) {
    infoPopup.setTitle(node['name']);
    infoPopup.setContent(html);
    infoPopup.show();
  });
};
```

Does it make sense?

src/services/popup.js
```diff
@@ -0,0 +1,123 @@
+/**
+ * @fileoverview Provides a factory to create a popup in the page.
+ * The factory returns a ngeo.Popup object.
+ *
+ * Example:
+ *
+ * var popup = ngeoCreatePopup({});
+ * popup.setTitle = "A title";
+ * popup.setContent = "Some content";
```
elemoine: `setTitle` and `setContent` are functions.

pgiraud:
http://pgiraud.github.io/ngeo/popup/layertree.html

examples/layertree.js
```diff
@@ -100,7 +126,19 @@ app.LayertreeController.prototype.getLayer = function(node) {
  * @export
  */
 app.LayertreeController.prototype.onButtonClick = function(node, layer) {
-  window.alert(node['name'] + ', ' + layer.get('type'));
+  if (!angular.isDefined(node.metadataPromise_)) {
```
elemoine: You should store the promise on the controller (`this`), no?

pgiraud: This would work in the example since the url is the same for every layer. However, in a real application one will want to call a different service.

elemoine: Yeah… If you really want to show how to do it if the metadata URL is not the same for every layer then I'd rather use a new object rather that storing the promises in the nodes themselves. And anyway, it would be one metadata URL per **layer**, as opposed to per **node**.

So, in the controller constructor:

```js
  /**
   * @type {Object.<string, !angular.$q.Promise>}
   * @private
   */
  this.promises_ = {}

  …
```

And in the `onButtonClick` function:

```js
  var layerType = node['layerType'];
  if (!(layerType in this.promises_)) {
    this.promises_[layerType] = $http.get(…).then(function(resp) {
      // …
      return html;
    });
  }
  var infoPopup = this.infoPopup_;
  this.promises_[layerType].then(function(html) {
    // …
  });
```

Makes sense?

elemoine:
@pgiraud, FYI, I'm looking at the build failure right now.

src/directives/popup.js
```diff
@@ -0,0 +1,65 @@
...
+        function(scope, element, attrs) {
+          element.addClass('popover');
+          var bodyWidth = goog.style.getSize(document.body).width;
+          var elementWidth = goog.style.getSize(element[0]).width;
+          element.css({
+            left: bodyWidth / 2 - elementWidth / 2,
+            top: 50
+          });
+
+          scope.close = function(evt) {
```
elemoine: What calls that function?

src/directives/popup.js
```diff
@@ -0,0 +1,65 @@
+/**
+ * @fileoverview Provides a directive used to show a popup over the page with
+ * a title and content.
+ * The directive doesn't create any scope but relies on its parent scope.
+ * Properties like 'content', 'title' or 'open' come from the parent scope.
+ *
+ * This directive is intented to be used along with the popup service.
```
elemoine: You probably want to mention the templateUrl, like we do for other widget directives.

src/services/popup.js
```diff
@@ -0,0 +1,116 @@
...
+
+/**
+ * @constructor
+ * @param {angular.$compile} $compile The compile provider.
+ * @param {angular.Scope} $rootScope The rootScope provider.
+ */
+ngeo.Popup = function($compile, $rootScope) {
+
+  /**
+   * The scope the element is compile with.
```
elemoine: compile -> compiled

elemoine:
I've added a few more comments, but it otherwise looks very good.

This will need to be rebased on master after #155 is merged.

examples/layertree.js
pgiraud: Makes perfect sense. I agree that it's much better.

src/directives/popup.js
pgiraud: Indeed. It's never used.

elemoine:
`Popup WIP` is a strange commit message for something merged into master. Too late :-)

PR #155, opened 2015-02-02
Fix build by adding paths to the ol.ext dir

elemoine:
This PR should fix the build failures we're currently seeing in https://github.com/camptocamp/ngeo/pull/153.


elemoine:
The build is broken w/o this. Merging…

PR #157, opened 2015-02-02
Change param type for ngeo.Popup#setContent

elemoine:
This is to handle the cases where the popup content is trusted with `$sce.trustAsHtml`.


pgiraud:
Please merge.

PR #158, opened 2015-02-03
Better styling for the drag handle

pgiraud:


examples/layerorder.html
```diff
@@ -7,6 +7,7 @@
           content="initial-scale=1.0, user-scalable=no, width=device-width">
     <meta name="apple-mobile-web-app-capable" content="yes">
     <link rel="stylesheet" href="../node_modules/openlayers/css/ol.css" type="text/css">
+    <link rel="stylesheet" href="//maxcdn.bootstrapcdn.com/font-awesome/4.3.0/css/font-awesome.min.css">
```
elemoine: It would be better to rely on a local installation of `font-awesome` (using a node package), but we can do that later.

elemoine:
Looks good. I'll update my branch when this is merged.

asaunier:
Nice :)

PR #159, opened 2015-02-04
Somehow deal with ngeoLocation/$location incompatibility

elemoine:
This documents the incompatibility and shows a possible workaround in the `permalink` example.

@asaunier, could you please test the workaround in your application?

### DIFF
--- a/examples/permalink.js
+++ b/examples/permalink.js
@@ -18,6 +18,49 @@ app.module = angular.module('app', ['ngeo']);
 
 
 /**
+ * Note: the "config" and "run" function defined below are used to work around
+ * a bug in Angular that makes the ngeoLocation service incompatible with the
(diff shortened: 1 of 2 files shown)


asaunier:
I have tested the workaround in Schweizmobil: it works like a charm! Thanks a lot!
+1 for merging.

elemoine:
@asaunier, do you test on IE 9? This is where I'd like to make sure that things still work as expected.

asaunier:
> do you test on IE 9? This is where I'd like to make sure that things still work as expected.

Argh!
When loading http://example.com/alex/, IE9 redirects to http://example.com/#/alex/

elemoine:
This `$location` service is really annoying, and we're forced to use it when using `ng-include` for example. That sucks.

elemoine:
Does not work in IE 9. Closing.

PR #160, opened 2015-02-05
Document incompat between ngeoLocation and $location

elemoine:
@asaunier, I added the "mock location provider" function to ngeo, next to the ngeoLocation service. What do you think?


asaunier:
OK, this way the workaround is easily enabled using
```
goog.require('ngeo.MockLocationProvider');
app.module.config(ngeo.mockLocationProvider);
```

elemoine:
Yep.

PR #162, opened 2015-02-12
datepicker

ger-benjamin:
To use a jquery-ui datepicker

### DIFF
--- a/externs/jqueryui.js
+++ b/externs/jqueryui.js
@@ -23,3 +23,10 @@ jQueryUI.prototype.item;
 * @return {jQuery}
 */
 jQuery.prototype.sortable = function(opt_options) {};
+
+
+/**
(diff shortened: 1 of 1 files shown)


asaunier:
Perhaps you could make your commit comment more explicit?

ger-benjamin:
You're right, thanks (and done)

asaunier:
Looks good to me.
(or maybe ``jQuery Datepicker`` => ``jQuery UI Datepicker``?)

ger-benjamin:
Thanks. Note: I think this is ``jQuery.prototype`` and not ``jQueryUI.prototype`` because jQueryUI functions seem to be added directly at jQuery. Only some "jquery-ui" objects are not in the jQuery object (but are in object ``jQuery.ui``)

fredj:
please merge

PR #163, opened 2015-02-12
Measure interactions

pgiraud:


src/services/measure/measure.js
```diff
@@ -0,0 +1,270 @@
...
+   */
+  this.measureTooltip_;
+
+
+  /**
+   * Pointer move listener key
+   * @type {goog.events.Key}
+   * @private
+   */
+  this.pointerMoveHandlerKey_;
```
tsauerwein: As you are inheriting from `ol.interaction.Pointer`, there is already a method `handleMoveEvent` that you could overwrite.

src/services/measure/measure.js
```diff
@@ -0,0 +1,270 @@
...
+
+
+/**
+ * @inheritDoc
+ */
+ngeo.interaction.Measure.prototype.setMap = function(map) {
+  goog.base(this, 'setMap', map);
+  this.overlay.setMap(map);
+
+  map.addInteraction(this.drawInteraction);
```
tsauerwein: Not sure about this, but would you also have to call `this.drawInteraction.setMap(map)`?

src/services/measure/measure.js
```diff
@@ -0,0 +1,270 @@
...
+};
+
+
+/**
+ * Handle draw interaction `drawend` event.
+ * @param {ol.DrawEvent} evt
+ * @private
+ */
+ngeo.interaction.Measure.prototype.onDrawEnd_ = function(evt) {
+  // FIXME we shouldn't manipulate DOM style here
```
tsauerwein: I think it's not a problem to change the style in an interaction, but usually we use the Closure methods:
http://docs.closure-library.googlecode.com/git/namespace_goog_dom.html
http://docs.closure-library.googlecode.com/git/namespace_goog_dom_classlist.html

src/services/measure/measure.js
```diff
@@ -0,0 +1,270 @@
...
+};
+
+
+/**
+ * Creates a new help tooltip
+ * @private
+ */
+ngeo.interaction.Measure.prototype.createHelpTooltip_ = function() {
+  this.removeHelpTooltip_();
+  this.helpTooltipElement_ = document.createElement('div');
```
tsauerwein: Here you also could use `goog.dom.createDom`, like here: 
https://github.com/openlayers/ol3/blob/master/src/ol/control/overviewmapcontrol.js#L68

src/services/measure/measure.js
```diff
@@ -0,0 +1,270 @@
...
+};
+
+
+/**
+ * Destroy the help tooltip
+ * @private
+ */
+ngeo.interaction.Measure.prototype.removeHelpTooltip_ = function() {
+  this.getMap().removeOverlay(this.helpTooltip_);
+  if (this.helpTooltipElement_) {
```
tsauerwein: This should be `!goog.isNull`.

tsauerwein: And the same at some other places. Usually we explicitly specify what we are checking for, e.g. `isDef` or `isNull`.

src/services/measure/measure.js
```diff
@@ -0,0 +1,270 @@
...
+ * Handle pointer move.
+ * @param {ol.MapBrowserEvent} evt
+ * @private
+ */
+ngeo.interaction.Measure.prototype.pointerMoveHandler_ = function(evt) {
+  if (evt.dragging) {
+    return;
+  }
+  /** @type {string} */
+  var helpMsg = 'Click to start drawing';
```
tsauerwein: The messages should be changeable via an option property.

src/services/measure/measurearea.js
```diff
@@ -0,0 +1,82 @@
...
+
+
+/**
+ * Format measure output.
+ * @param {ol.geom.Polygon} polygon
+ * @return {string}
+ * @private
+ */
+ngeo.interaction.MeasureArea.prototype.formatMeasure_ = function(polygon) {
+  var area = polygon.getArea();
```
tsauerwein: Do you want to add an option to do geodesic calculations? See https://github.com/openlayers/ol3/pull/3222/commits

tsauerwein:
Some minor comments. But besides I think it was a good decision to have the Draw/DrawAzimut interaction as a composite in the measure/area/azimut interactions.

elemoine:
I could not resist a quick look at this PR :)

I'd just have one comment: I think the `services` directory should only include Angular services. The objects that this PR adds are ol extensions (which currently do not have in ngeo), so I'd rather use an `ol-ext` dir, with an `interactions` subdir.

(I won't be able to follow up on this, so feel free to ignore me if you decide to do differently.)

pgiraud:
@elemoine your comment makes a lot of sense. I myself felt unconfident when I put the interactions in the services directory.

asaunier:
Excellent!

What about having a "measureTools" directive directly in ngeo and not only in the examples? This way it could be used out of the box with only a few lines to write?! Do you expect having very different measuring tools from one project to another?

src/services/measure/measure.js
pgiraud: This should be done automatically.
https://github.com/openlayers/ol3/blob/bfcceac68f321a308979f897e2d51b68b289dd58/src/ol/map.js#L435

tsauerwein: Ok, you're right.

src/services/measure/measurearea.js
pgiraud: Maybe later.

src/services/measure/measure.js
pgiraud: Indeed.

src/services/measure/measure.js
elemoine: `setMap` is passed `null` when the interaction is removed from the map. This case should be handled.

src/services/measure/measurearea.js
```diff
@@ -0,0 +1,84 @@
...
+ * Interaction dedicated to measure length.
+ *
+ * @constructor
+ * @extends {ngeo.interaction.Measure}
+ * @param {{
+ *    startMsg: (string|undefined),
+ *    continueMsg: (string|undefined),
+ *    style: (ol.style.Style|Array.<ol.style.Style>|ol.style.StyleFunction|undefined),
+ *    sketchStyle: (ol.style.Style|Array.<ol.style.Style>|ol.style.StyleFunction|undefined)
+ * }=} opt_options Options
```
elemoine: We need to create an `ngeox.js` externs file for the interaction options, as `olx.js` in ol3.

src/services/measure/measure.js
```diff
@@ -0,0 +1,278 @@
...
+
+/**
+ * @classdesc
+ * Interaction that allows measuring (length, area, ...).
+ *
+ * @constructor
+ * @extends {ol.interaction.Pointer}
+ * @param {{startMsg: (string|undefined),
+ *          style:(ol.style.Style|Array.<ol.style.Style>|ol.style.StyleFunction|undefined)}
+ *         =} opt_options Options
```
elemoine: Using a `@typedef` would be cleaner. (But this `@typedef` need no go in `ngeox.js` as this class is just a base class that won't be used directly in user code.)

src/services/measure/measurearea.js
```diff
@@ -0,0 +1,84 @@
...
+ * @export
```
elemoine: We don't want to use `@export` here, because we don't want/need the constructor to be exported when compiling ngeo and app together. Add a file to the `exports` dir instead, with explicit `goog.export*` statements.

src/services/measure/measurearea.js
```diff
@@ -0,0 +1,84 @@
...
+  var options = goog.isDef(opt_options) ? opt_options : {};
+
+  goog.base(this, options);
+
+  this.drawInteraction = new ol.interaction.Draw(
+      /** @type {olx.interaction.DrawOptions} */ ({
+        type: 'Polygon',
+        features: this.overlay.getFeatures(),
+        style: options.sketchStyle
+      }));
```
elemoine: I think it would be cleaner to pass the draw interaction to the parent constructor, and have `drawInteraction` be a private member of the parent class.

src/services/measure/measure.js
```diff
@@ -0,0 +1,278 @@
...
+  }
+};
+
+
+/**
+ * @inheritDoc
+ */
+ngeo.interaction.Measure.prototype.setActive = function(active) {
+  goog.base(this, 'setActive', active);
+  if (!goog.isNull(this.drawInteraction)) {
```
elemoine: I think `drawInteraction` should never be `null`. See my other comment regarding having specific classes pass the interaction to the parent constructor.

src/services/measure/measure.js
```diff
@@ -0,0 +1,278 @@
...
+};
+
+
+/**
+ * Handle pointer move.
+ * @param {ol.MapBrowserEvent} evt
+ * @private
+ */
+ngeo.interaction.Measure.prototype.handlePointerMoveEvent_ = function(evt) {
+  if (!this.getActive()) {
```
elemoine: I guess this case cannot occur. If the interaction is not active then handlers are not called.

src/services/measure/measure.js
```diff
@@ -0,0 +1,278 @@
...
+    this.measureTooltipElement_ = null;
+    this.measureTooltip_ = null;
+  }
+};
+
+
+/**
+ * @inheritDoc
+ */
+ngeo.interaction.Measure.prototype.setActive = function(active) {
```
elemoine: You could also listen to change event on the `active` property instead of overloading `setActive`, just like you do for the `DrawAzimut` interaction.

examples/measure.js
```diff
@@ -0,0 +1,130 @@
...
+      stroke: new ol.style.Stroke({
+        color: 'rgba(0, 0, 0, 0.7)'
+      }),
+      fill: new ol.style.Fill({
+        color: 'rgba(255, 255, 255, 0.2)'
+      })
+    })
+  });
+
+  /** @type {ngeo.interaction.MeasureLength} */
```
elemoine: You provide a type for `measureLength` but not the other ones.

src/services/measure/measure.js
```diff
@@ -0,0 +1,278 @@
...
+  /**
+   * The message to show when user is about to start drawing.
+   * @type {string}
+   */
+  this.startMsg = goog.isDef(options.startMsg) ? options.startMsg :
+      'Click to start drawing';
+
+  goog.base(this, {
+    handleMoveEvent: this.handlePointerMoveEvent_
+  });
```
elemoine: I'd move the call to the parent constructor to the beginning of the function. It is strange to have it in the middle.

examples/measure.js
```diff
@@ -0,0 +1,130 @@
...
+ * @return {angular.Directive} The directive specs.
+ */
+app.measureToolsDirective = function() {
+  return {
+    restrict: 'A',
+    scope: {
+      'map': '=appMeasureToolsMap'
+    },
+    controller: 'AppMeasureToolsController',
+    controllerAs: 'measureCtrl',
```
elemoine: Minor but you can just use `ctrl` here. The directive uses an isolate scope so the property is very "local".

examples/measure.js
```diff
@@ -0,0 +1,130 @@
...
+  ngeoDecorateInteraction(measureLength);
+  this['map'].addInteraction(measureLength);
+  this['measureLength'] = measureLength;
+
+  var measureArea = new ngeo.interaction.MeasureArea({
+    sketchStyle: style
+  });
+  measureArea.setActive(false);
+  ngeoDecorateInteraction(measureArea);
+  this['map'].addInteraction(measureArea);
```
elemoine: I'd create a local `map` variable instead of repeating `this['map']`. That's minor.

examples/measure.html
```diff
@@ -0,0 +1,64 @@
...
+        #map {
+          width: 800px;
+          height: 400px;
+        }
+
+        [ngeo-btn-group] {
+          padding: 5px;
+        }
+
+
```
elemoine: You could remove empty lines here.

examples/measure.html
```diff
@@ -0,0 +1,64 @@
...
+        .tooltip-static:before {
+          border-top-color: #ffcc33;
+        }
+
+    </style>
+  </head>
+  <body ng-controller="MainController as ctrl">
+    <div id="map" ngeo-map="ctrl.map"></div>
+
+    <div app-measure-tools app-measure-tools-map="ctrl.map"></div>
```
elemoine: You can use a one-time binding expression for the map.

elemoine: For consistency reasons I'd rename the directive `app-measuretools`.

src/services/measure/measure.js
```diff
@@ -0,0 +1,278 @@
...
+goog.require('ol.MapBrowserEvent');
+goog.require('ol.Observable');
+goog.require('ol.Overlay');
+goog.require('ol.interaction.Pointer');
+goog.require('ol.style.Style');
+
+
+
+/**
+ * @classdesc
```
elemoine: Super minor, but we don't use `@classdesc` elsewhere in ngeo.

elemoine:
Again, I could not resist a few more comments. I hope they make sense.

src/services/measure/measure.js
```diff
@@ -0,0 +1,278 @@
...
+   * Overlay to show the measurement.
+   * @type {ol.Overlay}
+   * @private
+   */
+  this.measureTooltip_ = null;
+
+
+  /**
+   * The sketch feature.
+   * @type {ol.Feature}
```
elemoine: You can use `@protected` for properties that are only used in child classes.

src/services/measure/measurearea.js
elemoine: I now understand that creating the draw interaction requires the feature overlay, which is created by the parent constructor. Another option would be to use a `getDrawInteraction` method, which is an `abstractMethod` in the parent class and implemented in child classes. It's up to you.

src/services/measure/measure.js
```diff
@@ -0,0 +1,278 @@
...
+   * @type {ol.interaction.Draw|ngeo.interaction.DrawAzimut}
+   */
+  this.drawInteraction = null;
+
+  /**
+   * The message to show when user is about to start drawing.
+   * @type {string}
+   */
+  this.startMsg = goog.isDef(options.startMsg) ? options.startMsg :
+      'Click to start drawing';
```
elemoine: Do you already know how we'll handle internationalization?

If the user messages are of type `(string|Element)` we may be able to pass `$compile`'d messages to the interaction.

Something for a separate PR maybe, but worth considering I think.

pgiraud: I haven't payed attention to translations yet.
Using an Element seems a good option.

externs/ngeox.js
```diff
@@ -0,0 +1,60 @@
...
+/**
+ * Interactions for measure tools.
+ * @typedef {{
+ *    startMsg: (string|undefined),
+ *    continueMsg: (string|undefined),
+ *    style: (ol.style.Style|Array.<ol.style.Style>|ol.style.StyleFunction|undefined),
+ *    sketchStyle: (ol.style.Style|Array.<ol.style.Style>|ol.style.StyleFunction|undefined)
+ * }}
+ */
+ngeox.interaction.MeasureOptions;
```
elemoine: This typedef need not be in ngeox.js. It could be in measure.js, before the ngeo.interaction.Measure constructor.

src/services/measure/measurearea.js
```diff
@@ -0,0 +1,88 @@
...
+
+
+/**
+ * @classdesc
+ * Interaction dedicated to measure length.
+ *
+ * @constructor
+ * @extends {ngeo.interaction.Measure}
+ * @param {ngeox.interaction.MeasureOptions=} opt_options Options
+ * @export
```
elemoine: `@export` to be removed.

pgiraud:
@tsauerwein can you please give this pull request a final review?
Note: I squashed everything in one single commit.

tsauerwein:
Some minor comments. Besides looking very good!

PR #168, opened 2015-02-24
Changes to the popup service

elemoine:
This PR makes a number of changes to the ngeo popup service:

* Use an isolate scope
* Replace the `show` method by `setOpen` (making it possible to hide the popup using `setOpen(false)`)
* Add a `getOpen` method to know whether the popup is currently open

Please review.


pgiraud:

PR #169, opened 2015-02-24
Use autoPan: false for the measure tooltips

elemoine:


pgiraud:
Please merge once Travis is happy.

PR #170, opened 2015-02-24
Base the sortable directive on Closure

elemoine:
This PR changes the `ngeo-sortable` directive to rely on Closure's `goog.fx.DragListGroup` rather than jQuery UI's sortable plugin.

This removes the dependency on jQuery UI and it will make the `ngeo-sortable` directive work on touch devices when https://github.com/google/closure-library/pull/405 is fixed.

Please review.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -45,7 +45,7 @@ test: .build/ol-deps.js .build/ngeo-deps.js .build/node_modules.timestamp
 	./node_modules/karma/bin/karma start karma-conf.js --single-run
 
 .PHONY: serve
-serve: .build/node_modules.timestamp .build/bower_components.timestamp
+serve: .build/node_modules.timestamp
 	node buildtools/serve.js
 
 .PHONY: gh-pages
@@ -147,10 +147,6 @@ dist/ngeo.css: node_modules/openlayers/css/ol.css .build/node_modules.timestamp
 	mkdir -p $(dir $@)
 	cp $< $@
 
-.build/examples-hosted/jquery-ui-sortable.min.js: bower_components/jquery-ui-sortable/jquery-ui-sortable.min.js
-	mkdir -p $(dir $@)
-	cp $< $@
-
 .build/examples-hosted/partials: examples/partials
 	mkdir -p $@
 	cp examples/partials/* $@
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -182,19 +175,17 @@ bower_components/jquery-ui-sortable/jquery-ui-sortable.min.js: .build/bower_comp
...
-			  .build/examples-hosted/angular-animate.min.js \
-			  .build/examples-hosted/bootstrap.min.js \
-			  .build/examples-hosted/bootstrap.min.css \
-			  .build/examples-hosted/jquery.min.js \
-			  .build/examples-hosted/jquery-ui-sortable.min.js \
-			  .build/examples-hosted/data \
-			  .build/examples-hosted/partials \
-			  .build/node_modules.timestamp \
-			  .build/bower_components.timestamp
+						.build/examples-hosted/%.js \
```
asaunier: The new indenting is a bit weird, isn't it?

src/directives/sortable.js
```diff
@@ -1,33 +1,55 @@
...
+ *   <li ng-repeat="layer in ctrl.layers">
+ *     <span class="sortable-handle">handle</span>{{layer.get('name')}}
  *   </li>
  * </ul>
+ *
+ * The value of the "ngeo-sortable" attribute is an expression which evaluates
+ * to an array (an array of layers in the above example). This is the array
+ * that is re-ordered after a drag-and-drop.
+ *
+ * The element with the class "handle" is the "drag handle". It is required.
```
asaunier: "handle" or "sortable-handle"?

Makefile
elemoine: I'll check it.

src/directives/sortable.js
elemoine: "sortable-handle". I'll fix it.

src/directives/sortable.js
```diff
@@ -1,33 +1,55 @@
...
+
+
+/**
+ * @param {angular.$log} $log Angular log service.
  * @param {angular.$timeout} $timeout Angular timeout service.
  * @return {angular.Directive} The directive specs.
  * @ngInject
  */
-ngeo.sortableDirective = function($timeout) {
+ngeo.sortableDirective = function($log, $timeout) {
```
elemoine: No need for the `$log` service I guess.

elemoine:
I added a commit that fixes the indentation in the Makefile. I also addressed @asaunier's other comments.

@asaunier other comments?

asaunier:
> @asaunier other comments?

Nope. It's OK from my point of view.

elemoine:
Ok, merging this then.

PR #171, opened 2015-02-24
ol.Overlay autoPan now false by default

elemoine:
The default value for ol.Overlay autoPan is now  false. So there's no need to set autoPan to false for the measure tooltips.


asaunier:
Makes sense!

PR #172, opened 2015-02-25
Compute length and area using a wsg84 sphere (geodesic)

pgiraud:
This pull request's goal is to get more realistic values for length and area measurements by converting coordinates to wsg84 and using geodesic measurements.
I don't think it needs to be optional since it will work in all cases no matter what the map projection is.
Please review.

### DIFF
--- a/src/ol-ext/interaction/measurearea.js
+++ b/src/ol-ext/interaction/measurearea.js
@@ -3,6 +3,7 @@ goog.provide('ngeo.interaction.MeasureArea');
 goog.require('ngeo.interaction.Measure');
 goog.require('ol.geom.Polygon');
 goog.require('ol.interaction.Draw');
+goog.require('ol.sphere.WGS84');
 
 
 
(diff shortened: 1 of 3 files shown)


src/ol-ext/interaction/measurearea.js
```diff
@@ -74,7 +75,12 @@ ngeo.interaction.MeasureArea.prototype.handleMeasure = function(callback) {
  * @private
  */
 ngeo.interaction.MeasureArea.prototype.formatMeasure_ = function(polygon) {
-  var area = polygon.getArea();
+  var map = this.getMap();
+  var sourceProj = map.getView().getProjection();
+  var geom = /** @type {ol.geom.Polygon} */(polygon.clone().transform(
```
elemoine: Please add a space after `*/`.

elemoine:
I just added a very minor comment. Please merge when addressed.

PR #173, opened 2015-02-25
Profile

pgiraud:
The goal of this pull request is to provide a profile directive to display an elevation profile in the page.
It's really simple for now and it's not configurable via options. For example, it expects the elevation data to be given by a "mnt" key in the data.

It relies on D3js for which angular externs have been added. Those externs have been automatically created and don't include @types.

Please review.


examples/data/profile.json
```diff
@@ -0,0 +1,682 @@
+{
+    "profile": [{
+        "y": 199340,
+        "values": {
+            "mnt": 788.7
+        },
+        "dist": 0.0,
+        "x": 541620
+    }, {
```
gberaudo: The format should be customizable, at least to a reasonable extent.

examples/profile.html
```diff
@@ -0,0 +1,33 @@
...
+
+        .profile {
+          width: 450px;
+          height: 250px;
+        }
+
+    </style>
+  </head>
+  <body ng-controller="MainController as ctrl">
+    <div class="profile" ngeo-profile="ctrl.profileData"
```
elemoine: You can use one-time binding here.

src/d3-ext/profile.js
```diff
@@ -0,0 +1,320 @@
...
+
+  /**
+   * D3 y scale.
+   */
+  var y;
+
+  var profile = function(selection) {
+    selection.each(function(data) {
+
+      var width = this.getBoundingClientRect().width -
```
gberaudo: Why not use this.clientWidth and this.clientHeight?

src/directives/profile.js
```diff
@@ -0,0 +1,51 @@
...
+              selection.datum(newVal['profile']).call(profile);
+            }
+          });
+
+          profile.hover(function(point) {
+            scope.$eval(attrs['ngeoProfileOnhover'], {'point': point});
+          });
+
+          profile.out(function() {
+            scope.$eval(attrs['ngeoProfileOnhover']);
```
elemoine: Wrong attribute?

src/d3-ext/profile.js
```diff
@@ -0,0 +1,320 @@
...
+          margin.right - margin.left;
+      x = d3.scale.linear().range([0, width]);
```
gberaudo: There is an issue if the profile is populated while the div is hidden:
in this case the width and height are 0 and the generated SVG is incorrect.
A workaround is to watch clientWidth / clientHeight for change and regenerate the SVG.
I have some working code here.

src/d3-ext/profile.js
```diff
@@ -0,0 +1,320 @@
...
+          margin.top - margin.bottom;
+      y = d3.scale.linear().range([height, 0]);
+
+      var xAxis = d3.svg.axis().scale(x).orient('bottom');
+      var yAxis = d3.svg.axis()
+          .scale(y)
+          .orient('left');
+
+      var area = d3.svg.area()
+          .x(function(d) { return x(d['dist']); })
```
gberaudo: A custom function could be used here to extract the dist value from the data row.
`extractorFn.getDistance(data[i]);`
The other values: `x, y and z` could be extracted in the same way.

src/d3-ext/profile.js
```diff
@@ -0,0 +1,320 @@
...
+   * @type {function(Object)}
+   * Hover callback function.
+   */
+  var hover = function() {};
+
+  /**
+   * @type {function()}
+   * Out callback function.
+   */
+  var out;
```
gberaudo: What about using `outCallback` and assigning a nope function like in `hover`?

src/d3-ext/profile.js
```diff
@@ -0,0 +1,320 @@
...
+  /**
+   * Factor to determine whether to use 'm' or 'km'.
+   */
+  var xFactor;
+
+  /**
+   * @type {function(Object)}
+   * Hover callback function.
+   */
+  var hover = function() {};
```
gberaudo: What about using a more clear name like `hoverCallback`?

src/d3-ext/profile.js
```diff
@@ -0,0 +1,320 @@
...
+      }
+    });
+  };
+
+  profile.hover = function(cb) {
+    if (!arguments.length) {
+      return hover;
+    }
+    hover = cb;
+    return profile;
```
gberaudo: It would be more clear to return `this`.

src/d3-ext/profile.js
```diff
@@ -0,0 +1,320 @@
...
+  };
+
+  profile.out = function(cb) {
+    if (!arguments.length) {
+      return out;
+    }
+    out = cb;
+    return profile;
```
gberaudo: Return `this`.

src/d3-ext/profile.js
```diff
@@ -0,0 +1,320 @@
...
+            .style('text-anchor', right ? 'end' : 'start')
+            .attr('transform', 'translate(' + xtranslate + ',' +
+                (y(elevation) - 10) + ')');
+        hover.call(null, point);
+      }
+
+      function mouseout() {
+        g.selectAll('.grid-hover')
+            .style('display', 'none');
+        if (out) {
```
gberaudo: The if could be removed by using a nope function like for `hover`.

src/d3-ext/profile.js
```diff
@@ -0,0 +1,320 @@
...
+ * @return {Object}
+ * @export
+ *
+ */
+ngeo.profile = function() {
+
+  /**
+   * The values for margins around the chart defined in pixels.
+   */
+  var margin = {top: 10, right: 10, bottom: 30, left: 40};
```
gberaudo: What about removing the margin logics and taking all the space?

gberaudo:
I am also integrating D3 profile. It would be nice to have it reasonably customizable now so that I can directly use the directive from ngeo. The main issue is the different data format and the generation of a wrong SVG when the profile div is hidden.

src/d3-ext/profile.js
```diff
@@ -0,0 +1,320 @@
+/**
+ * @fileoverview Provides a D3js component to be used to draw a elevation
```
gberaudo: Typo: s/a/an/

examples/profile.html
gberaudo: In this case, how do you update data?

src/d3-ext/profile.js
```diff
@@ -0,0 +1,320 @@
...
+      svg
+          .attr('width', width + margin.left + margin.right)
+          .attr('height', height + margin.top + margin.bottom);
+
+      // Update the inner dimensions.
+      var g = svg.select('g')
+          .attr('transform', 'translate(' + margin.left + ',' +
+              margin.top + ')');
+
+      var xDomain = d3.extent(data, function(d) { return d['dist']; });
```
gberaudo: If an extractor as defined is https://github.com/camptocamp/ngeo/pull/173/files#r25346757 is used, the extractor method can be passed directly here: `d3.extent(data, extractor.getDistance);`

src/d3-ext/profile.js
```diff
@@ -0,0 +1,320 @@
...
+          .attr('class', 'y axis');
+
+      gEnter.append('text')
+          .attr('class', 'y label')
+          .attr('text-anchor', 'end')
+          .attr('y', 6)
+          .attr('dy', '.75em')
+          .attr('transform', 'rotate(-90)')
+          .style('fill', 'grey')
+          .text('elevation (m)');
```
gberaudo: There are a few strings needing translation. How are we going to handle it?
Do we introduce a translator object with methods like `getElevationText()`?

src/d3-ext/profile.js
```diff
@@ -0,0 +1,320 @@
...
+      xAxis.tickFormat(function(d) {
+        return d / xFactor;
+      });
+
+      g.select('.x.axis')
+          .transition()
+          .call(xAxis);
+
+      g.select('.x.label')
+          .text('distance (' + units + ')')
```
gberaudo: Need translation, see https://github.com/camptocamp/ngeo/pull/173/files#r25350768
In this case we could pass a parameter to the translator.

src/d3-ext/profile.js
```diff
@@ -0,0 +1,320 @@
...
+            .attr('y1', y(elevation))
+            .attr('x2', width)
+            .attr('y2', y(elevation));
+
+        var right = x0 > xDomain[1] / 2;
+        var xtranslate = x(dist);
+        xtranslate += right ? -10 : 10;
+
+        g.select('.x.grid-hover text')
+            .text(parseFloat(dist.toPrecision(3) / xFactor) + ' ' + units)
```
gberaudo: Need translation, see https://github.com/camptocamp/ngeo/pull/173/files#r25350768

src/d3-ext/profile.js
```diff
@@ -0,0 +1,320 @@
...
+            .style('text-anchor', right ? 'end' : 'start')
+            .attr('transform', 'translate(' + xtranslate + ',' +
+                (height - 10) + ')');
+
+        g.select('.y.grid-hover text')
+            .text(Math.round(elevation) + ' m')
```
gberaudo: Need translation, see https://github.com/camptocamp/ngeo/pull/173/files#r25350768

src/d3-ext/profile.js
```diff
@@ -0,0 +1,320 @@
...
+
+      // Update the outer dimensions.
+      svg
+          .attr('width', width + margin.left + margin.right)
+          .attr('height', height + margin.top + margin.bottom);
+
+      // Update the inner dimensions.
+      var g = svg.select('g')
+          .attr('transform', 'translate(' + margin.left + ',' +
+              margin.top + ')');
```
gberaudo: Margin handling, related to https://github.com/camptocamp/ngeo/pull/173/files#r25347799

examples/profile.html
pgiraud: @gberaudo is right. The idea is to be able to update the data though the example doesn't show it. We should provide a way to load a new profile in the example.

src/d3-ext/profile.js
pgiraud: Indeed, my example doesn't take this case into account.
A similar issue may happen if the element is resized.

src/d3-ext/profile.js
pgiraud: I understand.
I wonder if it would be better to extract the data for each item on-the-fly. We have to take care of potential performance issues.

src/d3-ext/profile.js
pgiraud: profile != this

`profile` is a function here
`this` is the ngeo.profile instance

It can be changed of course.

pgiraud: Forget about my previous comment.

src/d3-ext/profile.js
pgiraud: This would be great indeed, I agree that it would simplify the code. But did you give it a try?
How can we ensure that the axis labels are visible?

src/d3-ext/profile.js
pgiraud: This makes sense.

src/d3-ext/profile.js
pgiraud: I'm not sure yet.
We'll need to be able to refresh the translations on the fly.

src/d3-ext/profile.js
gberaudo: Right, I missed the axis of the enlarged profile.

src/d3-ext/profile.js
```diff
@@ -0,0 +1,316 @@
...
+goog.provide('ngeo.profile');
+
+
+/**
+ *
+ * @return {Object}
+ * @export
+ *
+ */
+ngeo.profile = function() {
```
gberaudo: What about using a constructor instead of a factory?
We would then use `new` to instantiate it.

src/d3-ext/profile.js
gberaudo: Do you mean the texts are currently set once at creation time?
When is a text update needed?

src/d3-ext/profile.js
pgiraud: Indeed. Do you want me to give it a try? Or can you fork my branch and work on this yourself?

src/d3-ext/profile.js
pgiraud: In Luxembourg's application, the user may want to switch from one language to an other anytime. The page is not reloaded. The different texts needs to be translated in the new language though.

gberaudo: The profile could register itself to be notified on language changes.
Then we could refresh texts. This may be a feature for the translator
object.

gberaudo:
@pgiraud, I think your PR should be merged to master as it is now, or with just the use of a constructor. Then I will create PRs to make the code more generics. Is everyone OK with this?

buildtools/example.json
```diff
@@ -3,6 +3,7 @@
   "src": [
     "node_modules/openlayers/src/**/*.js",
     "node_modules/openlayers/build/ol.ext/*.js",
+    "node_modules/openlayers/build/d3.ext/*.js",
```
elemoine: Are you sure about this path?

examples/profile.html
elemoine: I don't think the examples should show bad practices. If the data does not change then one-time binding should be used.

examples/profile.html
```diff
@@ -0,0 +1,33 @@
...
+         ngeo-profile-on-hover="ctrl.onProfileHover(point)"
+         ngeo-profile-on-out="ctrl.onProfileOut()"></div>
```
elemoine: I'd use `onhover` and `onout` (no space) as we do anywhere else. In this way we always have one dash (two words) for the directive itself and two dashes (three words) for directive attributes.

examples/profile.js
```diff
@@ -0,0 +1,52 @@
...
+};
+
+
+/**
+ * @param {Object} point The point object with x, y, elevation, distance
+ * information.
+ */
+app.MainController.prototype.onProfileHover = function(point) {
+  this['point'] = point;
+  this.scope_.$apply();
```
elemoine: You should not need to call `$apply` here.

src/d3-ext/profile.js
```diff
@@ -0,0 +1,316 @@
...
+   * Out callback function.
+   */
+  var outCallback = goog.nullFunction;
+
+  /**
+   * The color to be used for filling the area.
+   * Can be overriden using the '.area' CSS selector.
+   * @type {string}
+   */
+  var FILL_COLOR = 'rgba(222, 222, 222, 0.5)';
```
elemoine: Why capital letters? Because it's a constant? You may want to use a `@const` annotation.

examples/profile.js
```diff
@@ -0,0 +1,52 @@
...
+
+  this['point'] = null;
+
+  this.scope_ = $scope;
+};
+
+
+/**
+ * @param {Object} point The point object with x, y, elevation, distance
+ * information.
```
elemoine: Are you sure `elevation` and `distance` are properties of the point? It does not look like so looking at the profile code.

elemoine: And I think this is a dangerous zone. The point object passed to the hover and out callbacks should be consistently typed, to avoid the risk of inconsistent renaming by the compiler. The easier option is to pass multiple args to the callbacks.

asaunier:
This looks like a nice feature!

That would be cool if it could be used in combination with the map directive to display for instance a track on the map with the matching profile in a profile area below the map (sharing the same geojson data). In addition being able to drag a cursor either on the map or on the profile and see the matching cursor in the other view would be useful as well. Is it already available? If not, that would be interesting to keep that in mind for the future.

examples/profile.js
pgiraud: This seems to be required. Should I do it in the directive?

src/d3-ext/profile.js
pgiraud: I'm not really sure whether I want constants here. I don't think this is worth.

examples/profile.js
pgiraud: The object sent to callbacks is supposed to be the point as it was given to the profile. It's the raw data.

src/directives/profile.js
```diff
@@ -0,0 +1,51 @@
...
+         * @param {angular.Attributes} attrs Attributes.
+         */
+        function(scope, element, attrs) {
+
+          var selection = d3.select(element[0]);
+          var profile = ngeo.profile();
+
+          scope.$watch(attrs['ngeoProfile'], function(newVal, oldVal) {
+            if (goog.isDef(newVal)) {
+              selection.datum(newVal['profile']).call(profile);
```
gberaudo: Would it be possible to get directly the profile data with `selection.datum(newVal).call(profile);`
and write the html like `<div class="profile" ngeo-profile="ctrl.profileData.profile"...`?

examples/profile.js
elemoine: Yes, for the controller the `onProfileHover` function is executed in the
context of Angular, so the controller should not have to call $apply. You
can try this in the directive:

```js
profile.onHover(function() {
  scope.$apply(function() {
    scope.$eval(attrs['ngeoProfileOnhover'], {'point': point});
  });
});
```




On Fri, Feb 27, 2015 at 2:10 PM, Pierre GIRAUD <notifications@github.com>
wrote:

> In examples/profile.js
> <https://github.com/camptocamp/ngeo/pull/173#discussion_r25504857>:
>
> > +    this['profileData'] = resp.data;
> > +  }));
> > +
> > +  this['point'] = null;
> > +
> > +  this.scope_ = $scope;
> > +};
> > +
> > +
> > +/**
> > + * @param {Object} point The point object with x, y, elevation, distance
> > + * information.
> > + */
> > +app.MainController.prototype.onProfileHover = function(point) {
> > +  this['point'] = point;
> > +  this.scope_.$apply();
>
> This seems to be required. Should I do it in the directive?
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/camptocamp/ngeo/pull/173/files#r25504857>.
>



-- 
Eric Lemoine

Camptocamp France SAS
Savoie Technolac, BP 352
73377 Le Bourget du Lac, Cedex

Tel : 00 33 4 79 44 44 94
Mail : eric.lemoine@camptocamp.com
http://www.camptocamp.com

examples/profile.js
elemoine: > The object sent to callbacks is supposed to be the point as it was given to the profile. It's the raw data.


Are you sure that is the case currently?

-- 
Eric Lemoine

Camptocamp France SAS
Savoie Technolac, BP 352
73377 Le Bourget du Lac, Cedex

Tel : 00 33 4 79 44 44 94
Mail : eric.lemoine@camptocamp.com
http://www.camptocamp.com

src/directives/profile.js
```diff
@@ -0,0 +1,51 @@
...
+            }
+          });
+
+          profile.onHover(function(point) {
+            scope.$eval(attrs['ngeoProfileOnhover'], {'point': point});
+          });
+
+          profile.onOut(function() {
+            scope.$eval(attrs['ngeoProfileOnout']);
+          });
```
gberaudo: I plan to introduce an `options` to pass to the `ngeo.profile` factory.
It would look like:
```js
/**
 * Options for the D3 Profile.
 *  @typedef {{
 *    light: (boolean|undefined),
 *    extractor: ngeox.profile.ProfileExtractor,
 *    ... other options including hover and out
 *  }}
 */
ngeox.profile.ProfileOptions;
```

@elemoine, @pgiraud, is the next step to add a `ngeoProfileOptions` attribute and remove the `ngeoProfileOnhover` and `ngeoProfileOnout` attributes? Then in the directive to call `scope.$watchCollection(attrs['ngeoProfileOptions'],` and recreate the profile on change?

src/directives/profile.js
pgiraud: Indeed. The directive doesn't have to know that the data is actually behin a "profile" key.

examples/profile.js
pgiraud: It's actually an item in the data array given to the profile.
What makes you think it's not the case?


On Fri, Feb 27, 2015 at 3:03 PM, Éric Lemoine <notifications@github.com>
wrote:

> In examples/profile.js
> <https://github.com/camptocamp/ngeo/pull/173#discussion_r25507676>:
>
> > + */
> > +app.MainController = function($http, $scope) {
> > +
> > +  $http.get('data/profile.json').then(angular.bind(this, function(resp) {
> > +    this['profileData'] = resp.data;
> > +  }));
> > +
> > +  this['point'] = null;
> > +
> > +  this.scope_ = $scope;
> > +};
> > +
> > +
> > +/**
> > + * @param {Object} point The point object with x, y, elevation, distance
> > + * information.
>
>  The object sent to callbacks is supposed to be the point as it was given
> to the profile. It's the raw data.
> Are you sure that is the case currently?
> … <#14bcb59076b834f7_>
> -- Eric Lemoine Camptocamp France SAS Savoie Technolac, BP 352 73377 Le
> Bourget du Lac, Cedex Tel : 00 33 4 79 44 44 94 Mail :
> eric.lemoine@camptocamp.com http://www.camptocamp.com
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/camptocamp/ngeo/pull/173/files#r25507676>.
>



-- 
Pierre GIRAUD
Géomaticien, Analyste

Camptocamp France SAS
Savoie Technolac, BP 352
73377 Le Bourget du Lac, Cedex

Tel : 00 33 4 79 44 44 93
Mail : pierre.giraud@camptocamp.com
http://www.camptocamp.com

src/directives/profile.js
pgiraud: What will the html code look like?
Why watching options? Are the options supposed to change?

On Fri, Feb 27, 2015 at 3:13 PM, gberaudo <notifications@github.com> wrote:

> In src/directives/profile.js
> <https://github.com/camptocamp/ngeo/pull/173#discussion_r25508251>:
>
> > +          var selection = d3.select(element[0]);
> > +          var profile = ngeo.profile();
> > +
> > +          scope.$watch(attrs['ngeoProfile'], function(newVal, oldVal) {
> > +            if (goog.isDef(newVal)) {
> > +              selection.datum(newVal['profile']).call(profile);
> > +            }
> > +          });
> > +
> > +          profile.onHover(function(point) {
> > +            scope.$eval(attrs['ngeoProfileOnhover'], {'point': point});
> > +          });
> > +
> > +          profile.onOut(function() {
> > +            scope.$eval(attrs['ngeoProfileOnout']);
> > +          });
>
> I plan to introduce an options to pass to the ngeo.profile factory.
> It would look like:
>
> /** * Options for the D3 Profile. *  @typedef {{ *    light: (boolean|undefined), *    extractor: ngeox.profile.ProfileExtractor, *    ... other options including hover and out *  }} */
> ngeox.profile.ProfileOptions;
>
> @elemoine <https://github.com/elemoine>, @pgiraud
> <https://github.com/pgiraud>, is the next step to add a ngeoProfileOptions
> attribute and remove the ngeoProfileOnhover and ngeoProfileOnout
> attributes? Then in the directive to call
> scope.$watchCollection(attrs['ngeoProfileOptions'], and recreate the
> profile on change?
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/camptocamp/ngeo/pull/173/files#r25508251>.
>



-- 
Pierre GIRAUD
Géomaticien, Analyste

Camptocamp France SAS
Savoie Technolac, BP 352
73377 Le Bourget du Lac, Cedex

Tel : 00 33 4 79 44 44 93
Mail : pierre.giraud@camptocamp.com
http://www.camptocamp.com

src/d3-ext/profile.js
gberaudo: There seems to be no way to mark comment thread as obsolete. I tried removing the topmost comment without effect.

src/d3-ext/profile.js
gberaudo: A constructor might not be the right choice with D3. We will possibly look at it later.

src/directives/profile.js
```diff
@@ -0,0 +1,55 @@
+/**
+ * @fileoverview Provides a directive used to insert an elevation profile chart
+ * in the DOM.
+ *
+ * Example:
+ *
+ * <div ngeo-profile="ctrl.profileData"
+ *      ngeo-profile-onhover="ctrl.onHover(point)"
+ *      ngeo-profile-onout="ctrl.onProfileOut()"></div>
```
elemoine: I think this is where we should document that `point` is a reference to an element of the input, user-provide data (`profileData` in the example).

elemoine:
> @pgiraud, I think your PR should be merged to master as it is now, or with just the use of a constructor. Then I will create PRs to make the code more generics. Is everyone OK with this?

+1. I agree with merging this and building on it.

gberaudo:
+1 for merging.

PR #174, opened 2015-02-26
Changes to measure interaction

elemoine:
This PR makes four changes to the measure interaction:

* Move the registrations of `drawstrat` and `drawend` listeners to the constructor to avoid registering new listeners each time the measure interaction is added to the map.
* Make sure the draw interaction is removed from the previous map when switching maps.
* Inherit from the base interaction class instead of the pointer interaction class. The measure interaction is indeed not a pointer interaction in the sense that it does not need to handle drag sequences.
* Rename the `interactions` directory to `interaction`. The directory is named after the namespace in ol3.


pgiraud:
Can you please do a rebase, please?

elemoine:
Done.

PR #176, opened 2015-02-27
Use goog.bind instead of angular.bind

elemoine:
This PR follows the guidelines to use `goog.bind` instead of `angular.bind` in the ngeo code.


pgiraud:
Looks good. Please merge.

PR #177, opened 2015-02-27
Introduce an ngeox.profile.ProfileOptions

gberaudo:
An `ngeox.profile.ProfileOptions` typedef has been introduced and must be passed to the `ngeo.profile` factory. Additionally, the profile now depends on an `ngeo.ProfileExtractor`  to support any tabular data.

The directive has been updated to require the options attribute.
Basic support for updating the profile was added.

### DIFF
--- a/examples/profile.html
+++ b/examples/profile.html
@@ -19,7 +19,7 @@
   </head>
   <body ng-controller="MainController as ctrl">
     <div class="profile" ngeo-profile="::ctrl.profileData"
-         ngeo-profile-options="::ctrl.profileOptions">
+         ngeo-profile-options="ctrl.profileOptions">
     </div>
 
     <ul ng-show="ctrl.point != null">
--- a/examples/profile.html
+++ b/examples/profile.html
@@ -19,8 +19,8 @@
   </head>
   <body ng-controller="MainController as ctrl">
     <div class="profile" ngeo-profile="::ctrl.profileData"
-         ngeo-profile-onhover="ctrl.onProfileHover(point)"
-         ngeo-profile-onout="ctrl.onProfileOut()"></div>
+         ngeo-profile-options="::ctrl.profileOptions">
+    </div>
 
     <ul ng-show="ctrl.point != null">
       <li>Longitude: {{ctrl.point.x}}</li>
--- a/examples/profile.js
+++ b/examples/profile.js
@@ -25,22 +25,21 @@ app.MainController = function($http, $scope) {
     this['profileData'] = resp.data['profile'];
   }));
 
-  this['point'] = null;
-};
+  var profileOptions = {
+    extractor: {
+      z: function(item) { return item['values']['mnt']; },
+      dist: function(item) { return item['dist']; }
+    },
+    hoverCallback: angular.bind(this, function(point) {
+      // An item in the list of points given to the profile.
+      this['point'] = point;
+    }),
+    outCallback: angular.bind(this, function() {
+      this['point'] = null;
+    })
+  };
 
-
-/**
- * @param {Object} point The point object. It correspond to an item in the list
- * of points given to the profile.
- */
-app.MainController.prototype.onProfileHover = function(point) {
-  this['point'] = point;
-};
-
-
-/**
- */
-app.MainController.prototype.onProfileOut = function() {
+  this['profileOptions'] = profileOptions;
   this['point'] = null;
 };
 
--- a/examples/profile.js
+++ b/examples/profile.js
@@ -25,21 +25,53 @@ app.MainController = function($http, $scope) {
     this['profileData'] = resp.data['profile'];
   }));
 
-  var profileOptions = {
-    extractor: {
-      z: function(item) { return item['values']['mnt']; },
-      dist: function(item) { return item['dist']; }
-    },
-    hoverCallback: angular.bind(this, function(point) {
-      // An item in the list of points given to the profile.
-      this['point'] = point;
-    }),
-    outCallback: angular.bind(this, function() {
-      this['point'] = null;
-    })
+  /**
+   * @param {Object} item
+   * @return {number}
+   */
+  var z = function(item) {
+    return item['values']['mnt'];
   };
 
-  this['profileOptions'] = profileOptions;
+  /**
+    * @param {Object} item
+    * @return {number}
+    */
+  var dist = function(item) {
+    return item['dist'];
+  };
+
+  /**
+   * @type {ngeox.profile.ProfileExtractor}
+   */
+  var extractor = {z: z, dist: dist};
+
+
+  // Using closures for hoverCallback and outCallback since
+  // wrapping in angular.bind leads to a closure error.
+  // See PR https://github.com/google/closure-compiler/pull/867
+  var that = this;
+
+  /**
+   * @param {Object} point
+   */
+  var hoverCallback = function(point) {
+    // An item in the list of points given to the profile.
+    that['point'] = point;
+  };
+
+  var outCallback = function() {
+    that['point'] = null;
+  };
+
+
+  this['profileOptions'] = {
+    extractor: extractor,
+    hoverCallback: hoverCallback,
+    outCallback: outCallback
+  };
+
+
   this['point'] = null;
 };
 
--- a/externs/ngeox.js
+++ b/externs/ngeox.js
@@ -49,3 +49,68 @@ ngeox.interaction.MeasureOptions.prototype.style;
  * @type {ol.style.Style|Array.<ol.style.Style>|ol.style.StyleFunction|undefined}
  */
 ngeox.interaction.MeasureOptions.prototype.sketchStyle;
+
+
+/**
+ * Namespace.
+ * @type {Object}
+ */
+ngeox.profile;
+
+
+/**
+ * Options for the profile.
+ *  @typedef {{
+ *    extractor: ngeox.profile.ProfileExtractor,
+ *    light: (boolean|undefined),
+ *    hoverCallback: (function(Object)|undefined),
+ *    outCallback: (function()|undefined)
+ *  }}
+ */
+ngeox.profile.ProfileOptions;
+
+
+/**
+ * @type {ngeox.profile.ProfileExtractor}
+ */
+ngeox.profile.ProfileOptions.prototype.extractor;
+
+
+/**
+ * @type {boolean|undefined}
+ */
+ngeox.profile.ProfileOptions.prototype.light;
+
+
+/**
+ * @type {function(Object)|undefined}
+ */
+ngeox.profile.ProfileOptions.prototype.hoverCallback;
+
+
+/**
+ * @type {function()|undefined}
+ */
+ngeox.profile.ProfileOptions.prototype.outCallback;
+
+
+/**
+ * Profile data extractor.
+ * @typedef {{
+ *   z: function(Object): number,
+ *   dist: function(Object): number
+ * }}
+ */
+ngeox.profile.ProfileExtractor;
+
+
+/**
+ * @type {function(Object, string=): number}
+ */
+ngeox.profile.ProfileExtractor.prototype.z;
+
+
+/**
+ * @type {function(Object): number}
+ */
+ngeox.profile.ProfileExtractor.prototype.dist;
--- a/src/d3-ext/profile.js
+++ b/src/d3-ext/profile.js
@@ -29,10 +29,12 @@ goog.provide('ngeo.profile');
 /**
  *
  * @return {Object}
+ * @param {ngeox.profile.ProfileOptions} options
  * @export
  *
  */
-ngeo.profile = function() {
+ngeo.profile = function(options) {
+
 
   /**
    * The values for margins around the chart defined in pixels.
@@ -42,7 +44,9 @@ ngeo.profile = function() {
   /**
    * Method to get the coordinate in pixels from a distance.
    */
-  var bisectDistance = d3.bisector(function(d) { return d['dist']; }).left;
+  var bisectDistance = d3.bisector(function(d) {
+    return extractor.dist(d);
+  }).left;
 
   /**
    * Distance units. Either 'm' or 'km'.
@@ -58,13 +62,15 @@ ngeo.profile = function() {
    * @type {function(Object)}
    * Hover callback function.
    */
-  var hoverCallback = goog.nullFunction;
+  var hoverCallback = goog.isDef(options.hoverCallback) ?
+      options.hoverCallback : goog.nullFunction;
 
   /**
    * @type {function()}
    * Out callback function.
    */
-  var outCallback = goog.nullFunction;
+  var outCallback = goog.isDef(options.outCallback) ?
+      options.outCallback : goog.nullFunction;
 
   /**
    * The color to be used for filling the area.
@@ -95,6 +101,12 @@ ngeo.profile = function() {
    */
   var y;
 
+  /**
+   * Extractor used to ge the dist and elevation from data.
+   */
+  var extractor = options.extractor;
+
+
   var profile = function(selection) {
     selection.each(function(data) {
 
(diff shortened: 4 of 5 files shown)


externs/ngeox.js
```diff
@@ -49,3 +49,33 @@ ngeox.interaction.MeasureOptions.prototype.style;
...
+ * Options for the profile.
+ *  @typedef {{
+ *    extractor: ngeox.profile.ProfileExtractor,
+ *    light: (boolean|undefined),
+ *    hoverCallback: (function(Object)|undefined),
+ *    outCallback: (function()|undefined),
+ *    zKeys: (Array.<string>|undefined)
+ *  }}
+ */
+ngeox.profile.ProfileOptions;
```
elemoine: It's not sufficient for the externs definitions. We need to define a prototype and prototype properties. See MeasureOptions for an example.

src/d3-ext/profile.js
```diff
@@ -58,13 +62,13 @@ ngeo.profile = function() {
    * @type {function(Object)}
    * Hover callback function.
    */
-  var hoverCallback = goog.nullFunction;
+  var hoverCallback = options.hoverCallback || goog.nullFunction;
```
elemoine: Please use `goog.isDef` for consistency reasons.

gberaudo: Using `goog.isDef` would be less readable, is it really an improvement?
`var hoverCallback = options.hoverCallback || goog.nullFunction;`
> var hoverCallback = goog.isDef(options.hoverCallback) ?  options.hoverCallback :
>   || goog.nullFunction;

Wouldn't it make sense to allow the short syntax when dealing with objects?

externs/ngeox.js
gberaudo: I thought it was only a requirement of the documentation generation tool used by Ol3.
Where is defined `ngeox.interaction.MeasureOptions.prototype`?

src/d3-ext/profile.js
elemoine: I think we want to be consistent and use `goog.isDef` everywhere. Much easier to explain to new developers.

externs/ngeox.js
elemoine: > I thought it was only a requirement of the documentation generation tool used by Ol3.

No.



> Where is defined ngeox.interaction.MeasureOptions.prototype?

Each measure option is defined in the prototype. E.g. https://github.com/camptocamp/ngeo/blob/master/externs/ngeox.js#L29. That is the way externs work.

src/d3-ext/profile.js
gberaudo: Consistency should be second after readability. There is no point in having a code base consistently hard to read.
Moreover, it is just plain and widely used javascript every developer should be familiar with.

elemoine: I disagree.

1. I don't think

 ```js
   var hoverCallback = goog.isDef(options.hoverCallback) ? options.hoverCallback : goog.nullFunction;
 ```

 is harder to read than

 ```js
   var hoverCallback = options.hoverCallback || goog.nullFunction;
 ```

 The first expression is even more explicit, because it makes it clear that the test is about whether `hoverCallback` is defined in the `options` object.

2. It's easier to explain to people that they should **always** use `goog.isDef(options.theOption)` to test if `theOption` is defined in the options.

 I don't want to have to tell people that they can use `||` when dealing with objects, and `goog.isDef` when dealing with number, strings and booleans. I prefer systematic approaches.

gberaudo: Honestly, having `options.hoverCallback` repeated makes it harder to read and easy to make typos.

While both expressions are not synonyms, they are both doing their job (I do not think we should use
goog.isDefAndNotNull).

I disagree with you on the second point: I expect all developers working on Ngeo to understand basics concepts of javascript. Then it is just a matter of having read the contributor guide. Don't you think preventing developers to use basic javascript is a bit tyrannical? ;-)

elemoine: > Honestly, having options.hoverCallback repeated makes it harder to read and easy to make typos.

Come on! You really can't say that `var foo = goog.isDef(options.foo) ? options.foo : "default"` is hard to read.

> I disagree with you on the second point: I expect all developers working on Ngeo to understand basics concepts of javascript.

People very often make mistakes because `undefined`, `null`, `0` and `''` are all falsy values. People who see `var callback = options.callback || goog.nullFunction` will do the same for their options of type `number`: `var myNum = options.myNum || -1`.

I think it's a good thing to teach people to be stricter and more explicit with types.

gberaudo:
@elemoine, I updated the code according to your two comments and it is now ready for merging.

src/d3-ext/profile.js
gberaudo: @elemoine, I think rules should be flexible enough so that they do not hinder developer creativity. Nevertheless, I updated the PR with your requirements.

examples/profile.js
```diff
@@ -25,22 +25,21 @@ app.MainController = function($http, $scope) {
...
- */
-app.MainController.prototype.onProfileHover = function(point) {
-  this['point'] = point;
-};
-
-
-/**
- */
-app.MainController.prototype.onProfileOut = function() {
+  var profileOptions = {
```
elemoine: This object does not have all the properties defined in the `ngeox.profile.ProfileOptions` type alias, so it may make sense to cast the right value.

externs/ngeox.js
```diff
@@ -49,3 +49,75 @@ ngeox.interaction.MeasureOptions.prototype.style;
...
+ * @type {Array.<string>|undefined}
+ */
+ngeox.profile.ProfileOptions.prototype.zKeys;
+
+
+/**
+ * Profile data extractor.
+ * @typedef {{
+ *   z: (function(Object, string=): number),
+ *   dist: (function(Object): number)
```
elemoine: Do you really need brackets here?

src/directives/profile.js
```diff
@@ -27,31 +27,37 @@ ngeo.profileDirective = function() {
...
+          }
+
           var selection = d3.select(element[0]);
-          var profile = ngeo.profile();
+          var profile, options, data;
 
-          scope.$watch(attrs['ngeoProfile'], function(newVal, oldVal) {
-            if (goog.isDef(newVal)) {
-              selection.datum(newVal).call(profile);
+          scope.$watch(attrs['ngeoProfileOptions'], function(newVal, oldVal) {
```
elemoine: Why do you watch the options?

src/directives/profile.js
```diff
@@ -27,31 +27,37 @@ ngeo.profileDirective = function() {
...
+              refreshData();
             }
           });
 
-          profile.onHover(function(point) {
-            scope.$apply(function() {
-              scope.$eval(attrs['ngeoProfileOnhover'], {'point': point});
-            });
-          });
+          function refreshData() {
```
elemoine: It would make more sense to define the function after the two watcher registrations.

src/directives/profile.js
```diff
@@ -27,31 +27,37 @@ ngeo.profileDirective = function() {
...
+       * @param {angular.Scope} scope Scope.
+       * @param {angular.JQLite} element Element.
+       * @param {angular.Attributes} attrs Attributes.
+       */
         function(scope, element, attrs) {
 
+          var optionsAttr = attrs['ngeoProfileOptions'];
+          if (!goog.isDef(optionsAttr)) {
+            throw new Error('The ngeoProfileOptions attribute is required');
+          }
```
elemoine: We just use `goog.asserts.assert` elsewhere. In this case:

```js
goog.asserts.assert(goog.isDef(optionsAttr);
```

elemoine:
The docs at the head of the `src/directives/profile.js` need to be updated.

examples/profile.html
```diff
@@ -19,8 +19,8 @@
   </head>
   <body ng-controller="MainController as ctrl">
     <div class="profile" ngeo-profile="::ctrl.profileData"
-         ngeo-profile-onhover="ctrl.onProfileHover(point)"
-         ngeo-profile-onout="ctrl.onProfileOut()"></div>
+         ngeo-profile-options="::ctrl.profileOptions">
```
elemoine: You can remove the one-time binding expression here if you remove the "options" watcher.

src/directives/profile.js
```diff
@@ -27,31 +27,37 @@ ngeo.profileDirective = function() {
...
     link:
         /**
-         * @param {angular.Scope} scope Scope.
-         * @param {angular.JQLite} element Element.
-         * @param {angular.Attributes} attrs Attributes.
-         */
+       * @param {angular.Scope} scope Scope.
+       * @param {angular.JQLite} element Element.
+       * @param {angular.Attributes} attrs Attributes.
+       */
```
elemoine: Why did you change the indentation? It looked correct to me.

elemoine:
I've added a few other comments. If you agree with my comments, please merge when they are addressed.

externs/ngeox.js
pgiraud: Why is the signature for `z` different?

examples/profile.js
```diff
@@ -25,22 +25,58 @@ app.MainController = function($http, $scope) {
...
+
+  /**
+   * @type {ngeox.profile.ProfileExtractor}
+   */
+  var extractor = {z: z, dist: dist};
+
+
+  // Using closures for hoverCallback and outCallback since
+  // wrapping in angular.bind leads to a closure error.
+  var that = this;
```
pgiraud: Can you tell us more about this closure error?

examples/profile.js
```diff
@@ -25,22 +25,58 @@ app.MainController = function($http, $scope) {
...
+
+
+  /**
+   * @type {ngeox.profile.ProfileOptions}
+   */
+  var profileOptions = {
+    extractor: extractor,
+    hoverCallback: hoverCallback,
+    light: false, // explicitely set to avoid closure error
+    zKeys: undefined, // explicitely set to avoid closure error
```
pgiraud: What is this `zKeys` option for?

examples/profile.js
gberaudo: ERR! compile .build/examples/all.js:1547: ERROR - initializing variable
ERR! compile found   : {extractor: {dist: function ((Object|null)): number, z: function ((Object|null), string=): number}, hoverCallback: (Function|null), light: boolean, outCallback: function (): undefined, ...}
ERR! compile required: {extractor: {dist: function ((Object|null)): number, z: function ((Object|null), string=): number}, hoverCallback: (function ((Object|null)): ?|undefined), light: (boolean|undefined), outCallback: (function (): ?|undefined), ...}
ERR! compile   var profileOptions = {

examples/profile.js
pgiraud: It's a bit unfortunate that we need to explicitely have to set values for options for which the default value is OK.

examples/profile.js
gberaudo: `angular.bind` is defined to get a `Function` and return a `Function` so it actually loses the type of the function. Where is angular.bind jsdocs coming from?

examples/profile.js
gberaudo: It is not used yet, I will remove it.

examples/profile.js
elemoine: Look at `.build/externs/angular-1.3.js` for the Angular-related types.

gberaudo: I see `Copyright 2012 The Closure Compiler Authors.`. Where was it downloaded from?

elemoine: https://github.com/google/closure-compiler/blob/master/contrib/externs/angular-1.3.js

gberaudo: OK, I found it in the Makefile.

gberaudo: I will try fixing the jsdoc.

examples/profile.js
```diff
@@ -25,22 +25,57 @@ app.MainController = function($http, $scope) {
...
+  };
+
+
+  /**
+   * @type {ngeox.profile.ProfileOptions}
+   */
+  var profileOptions = {
+    extractor: extractor,
+    hoverCallback: hoverCallback,
+    light: false, // explicitely set to avoid closure error
```
pgiraud: Once again, can't we use a default value? Does it really have to be explicitely set?

examples/profile.js
gberaudo: OK, I have a fix, I will send it upstream.

examples/profile.js
gberaudo: Indeed. @elemoine do you know of a better handling of this issue?
This case is similar to option typedefs we pass to constructors. Ex:
https://github.com/camptocamp/ngeo/blob/master/examples/geolocation.js#L36
```js
var map = new ol.Map({
  layers: [
    new ol.layer.Tile({
      source: new ol.source.OSM()
    })
  ],
  view: view
});
```

elemoine: On Tue, Mar 3, 2015 at 9:24 AM, gberaudo <notifications@github.com> wrote:
>
> In examples/profile.js:
>
> > +    that['point'] = point;
> > +  };
> > +
> > +  var outCallback = function() {
> > +    that['point'] = null;
> > +  };
> > +
> > +
> > +  /**
> > +   * @type {ngeox.profile.ProfileOptions}
> > +   */
> > +  var profileOptions = {
> > +    extractor: extractor,
> > +    hoverCallback: hoverCallback,
> > +    light: false, // explicitely set to avoid closure error
> > +    zKeys: undefined, // explicitely set to avoid closure error
>
> Indeed. @elemoine do you know of a better handling of this issue?
> This case is similar to option typedefs we pass to constructors. Ex:
> https://github.com/camptocamp/ngeo/blob/master/examples/geolocation.js#L36
>
> var map = new ol.Map({
>   layers: [
>     new ol.layer.Tile({
>       source: new ol.source.OSM()
>     })
>   ],
>   view: view
> });


You can use a typecast:


var profileOptions = /** @type {ngeox.profile.ProfileOptions} */ ({
  extractor: extractor,
  hoverCallback: hoverCallback
});


-- 
Eric Lemoine

Camptocamp France SAS
Savoie Technolac, BP 352
73377 Le Bourget du Lac, Cedex

Tel : 00 33 4 79 44 44 94
Mail : eric.lemoine@camptocamp.com
http://www.camptocamp.com

examples/profile.js
gberaudo: See https://github.com/google/closure-compiler/pull/867

elemoine: Our applications are compiled and use `goog.bind` instead of `angular.bind`, so this is not a major problem for us.

externs/ngeox.js
gberaudo: You are right, fixed.
I added this and zKeys in anticipation of multiple elevation values
handling. However, it is better if you add it when you actually implement
it.

examples/profile.js
gberaudo: Can we use `goog.bind` in examples?

elemoine: No. https://github.com/camptocamp/ngeo/blob/master/docs/developer-guide.md#authoring-examples.

gberaudo:
@elemoine, @pgiraud, does the current code is OK for you?
- use angular.bind with modified externs;
- directly set the ProfileOptions in this.

Makefile
```diff
@@ -233,7 +233,7 @@ node_modules/angular/angular.min.js node_modules/angular-animate/angular-animate
 
 .build/externs/angular-1.3.js:
 	mkdir -p $(dir $@)
-	wget -O $@ https://raw.githubusercontent.com/google/closure-compiler/master/contrib/externs/angular-1.3.js
+	wget -O $@ https://github.com/gberaudo/closure-compiler/raw/master/contrib/externs/angular-1.3.js
```
elemoine: I'd rather use a workaround in the example, because we don't know whether your patch will be merged.

pgiraud:
@gberaudo we should avoid using forks as much as possible. We should open a new issue to keep track of the pull request in google closure.

Otherwise, it looks good to me.

Makefile
gberaudo: In this case, let's use the initial workaround I proposed and we will change if my PR to closure is merged.

pgiraud: +1

elemoine:
+1 for using the initial workaround for `angular.bind`.

gberaudo:
I think I have addressed all comments and will now merge. Thank you @elemoine and @pgiraud for the review.

gberaudo:
@elemoine, I do not have write access to the ngeo repository. Could you please add me to the project collaborators or merge this PR? Thank you.

elemoine:
> @elemoine, I do not have write access to the ngeo repository. Could you please add me to the project collaborators or merge this PR? Thank you.

Done.

gberaudo:
Thanks.

PR #180, opened 2015-03-03
Make the measure tooltips translatable

elemoine:
This PR makes it possible to use Angular (angular-gettext) to translate measure tooltips. The `measure` example shows how it works.

Please review.


pgiraud:
Perfect. Please merge.

PR #181, opened 2015-03-04
Azimut distance

pgiraud:
The goal of this pull request it to enhance the azimut measure interaction. Distance is now shown. Also the user can add additional information when the measurement is finished.
Please review.

### DIFF
--- a/src/ol-ext/interaction/measure.js
+++ b/src/ol-ext/interaction/measure.js
@@ -1,3 +1,4 @@
+goog.provide('ngeo.MeasureEvent');
 goog.provide('ngeo.interaction');
 goog.provide('ngeo.interaction.Measure');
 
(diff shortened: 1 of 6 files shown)


elemoine:
Looks good, but build failed.

pgiraud:
My bad. I stupidly didn't run make check again after adding a simple comment.

src/ol-ext/interaction/measure.js
```diff
@@ -25,6 +26,45 @@ goog.require('ol.style.Style');
...
 
+/**
+ * @enum {string}
+ */
+ngeo.MeasureEventType = {
+  /**
+   * Triggered upon feature draw end
+   * @event ol.MeasureEvent#end
+   */
+  END: 'end'
```
elemoine: I'd have named this measureend for consistency with ol3 (which has drawstart and drawend), but it's up to you.

PR #183, opened 2015-03-05
Fix goog.provides in measure.js

elemoine:
This PR removes the `goog.provides('ngeo.interaction')` statement from `interaction/measure.js`. It also adds `goog.provides('ngeo.MeasureEventType')` to that file.


fredj:
+1

PR #185, opened 2015-03-09
Add an examples index page

elemoine:
This PR adds an index page for the examples on github.io.

See http://erilem.net/ngeo/examplesindex/ to see what the index looks like. It could probably be a lot nicer, but I didn't want to spend time on the styling.

The index is generated automatically, based on the examples' title and description. The title is read from the `<title>` tag, and the description is read from the element whose `id` is `"desc"`.

Please review.


asaunier:
Looks good to me, thanks.

The page layout is simple but clear. Perhaps we'll have to use another layout (blocks and filter field like for OL?) if the number of examples grows to big.

elemoine:
> Perhaps we'll have to use another layout (blocks and filter field like for OL?) if the number of examples grows to big.

Ctrl+f to search in the page works great for me. I don't feel the change for a search field.

elemoine:
Thanks for the review @asaunier. I may add more text to the README.md file before merging.

asaunier:
> Ctrl+f to search in the page works great for me. I don't feel the change for a search field.

It really depends how many examples you have. It's cool to have such a search field to show only the examples that match your search. I admit it is not required though.

PR #187, opened 2015-03-11
Add an ngeoSearch directive

elemoine:
This PR adds an `ngeoSearch` directive, an `ngeoCreateGeoJSONBloodhound` service, and a `search` example.

The `ngeoSearch` directive is very generic/low-level. It just provides a directive for decorating an element with `typeahead`.

The `ngeoCreateGeoJSONBloodhound` service is a function returning a Bloodhound engine that expects GeoJSON responses from the remote service, and create `ol.Feature` objects as the suggestions.

The `search` example shows how to use the `ngeoSearch` directive and the `ngeoCreateGeoJSONBloodhound` service.

Please review.

### DIFF
--- a/src/services/creategeojsonbloodhound.js
+++ b/src/services/creategeojsonbloodhound.js
@@ -1,0 +1,80 @@
+/**
+ * @fileoverview Provides a function that creates a Bloodhound engine
+ * expecting GeoJSON responses from the search web service, and creating
+ * `ol.Feature` objects as suggestions.
+ *
+ * Example #1:
+ *
+ * var bloodhound = ngeoCreateGeoJSONBloodhound(
+ *   'http://example.com/fulltextsearch?query=%QUERY',
+ *   ol.proj.get('EPSG:3857'));
+ * bloodhound.initialize();
+ *
+ * Example #2:
+ *
+ * var bloodhound = ngeoCreateGeoJSONBloodhound({
+ *   remote: {
+ *     limit: 10,
+ *     url: mySearchEngineUrl,
+ *     replace: function(url, query) {
+ *       return url +
+ *           '?qtext=' + encodeURIComponent(query) +
+ *           '&lang=' + gettextCatalog.currentLanguage;
+ *     }
+ *   }
+ * }, ol.proj.get('EPSG:3857'), ol.proj.get('EPSG:21781'));
+ * bloodhound.initialize
+ */
+
+goog.provide('ngeo.CreateGeoJSONBloodhound');
+
+goog.require('ngeo');
+goog.require('ol.format.GeoJSON');
+
+
+/**
+ * @typedef {function(string, ol.proj.Projection=,
+ *     ol.proj.Projection=):Bloodhound}
+ */
+ngeo.CreateGeoJSONBloodhound;
+
+
+/**
+ * @param {BloodhoundOptions|string} options Bloodhound options or a URL to the
+ *     search service. If a URL is provided then default Bloodhound options are
+ *     used.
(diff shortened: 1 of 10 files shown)


elemoine:
The `search` example is available at http://erilem.net/ngeo/search/search.html.

src/services/creategeojsonbloodhound.js
```diff
@@ -0,0 +1,59 @@
...
+ */
+ngeo.createGeoJSONBloodhound = function(url, opt_featureProjection,
+    opt_dataProjection) {
+  var geojsonFormat = new ol.format.GeoJSON();
+  return new Bloodhound(/** @type {BloodhoundOptions} */ ({
+    remote: {
+      url: url,
+      ajax: {
+        dataType: 'jsonp'
+      },
```
gberaudo: We are setting the `limit` option and using a `replace` function to handle localization.
Would it be possible to update the remote `url` of a `Bloodhound` afterwards?

gberaudo:
@elemoine, I looked at this PR and added a comment. I think @fredj will have more.

src/services/creategeojsonbloodhound.js
elemoine: > We are setting the limit option and using a replace function to handle localization.

Yes, I see.

And your `filter` function doesn't create `ol.Feature` objects either. So this Bloodhound factory is not useful to you…

This factory is really about creating a `Bloodhound` engine with a `filter` function returning `ol.Feature` created out of GeoJSON documents. The rest should be customizable through options.

What about the following?

```js
/**
 * @params {BloodhoundOptions|string} options Bloodhound options or a URL. If a string is passed the
 *     default options are used, with this string as the `url` in the `remote` object.
 * @param {ol.proj.Projection=} opt_featureProjection Feature projection.
 * @param {ol.proj.Projection=} opt_dataProjection Data projection.
 * @return {Bloodhound} The Bloodhound object.
 */
ngeo.createGeoJSONBloodhound = function(
    options, opt_featureProjection, opt_dataProjection) {
  var bloodhoundOptions = /** @type BloodhoundOptions */ ({
    remote: {
      url: goog.isString(options) ? options : '',
      ajax: {
        dataType: 'jsonp'
      },
      filter: function(parsedResponse) {
        // …
      }
    },
    datumTokenizer: goog.nullFunction,
    queryTokenizer: Bloodhound.tokenizers.whitespace
  };
  if (!goog.isString(options)) {
    goog.object.extend(bloodhoundOptions, options);
  }
  return new Bloodhound(bloodhoundOptions);
};
```

I don't know if it's worth it. Maybe the user should just create his Bloodhound engine himself if he needs something more specific than what this factory does.

Thoughts?


> Would it be possible to update the remote url of a Bloodhound afterwards?

Not that I know.

fredj: > I don't know if it's worth it. Maybe the user should just create his Bloodhound engine himself if he needs something more specific than what this factory does.

The `BloodhoundOptions|string` parameter is a good comprise.

elemoine: Thanks. I changed it, and added an example in the fileoverview showing how to pass a Bloodhound options object.

elemoine:
Note that the search example currently relies on the Luxembourg dev instance's search service (http://devv3.geoportail.lu/main/wsgi/fulltextsearch). Using a static file doesn't make sense in this case.

elemoine:
Any objection against merging this? Happy to make changes if necessary.

asaunier:
From my perspective, it is OK for merging.

gberaudo:
Looks good to me.

fredj:
+1

PR #188, opened 2015-03-11
Building the examples index requires beautifulsoup4

elemoine:


asaunier:
Makes sense!

elemoine:
It works this time. See http://camptocamp.github.io/ngeo/master/.

PR #195, opened 2015-03-16
Drag drop

pgiraud:
With this pull request I let developer style the dragged element.

Also I'm using DragListGroup::dispose to remove any listener on the drag list group. Without doing that the number of listeners was growing each time the sortable was reset. This can easily seen by checking the dom inspector and looking for the clone(s) of the currently dragged item.
https://github.com/camptocamp/ngeo/commit/ae0e60074983010ac70d89863b71b87a5652b272

Please review.

### DIFF
--- a/src/directives/sortable.js
+++ b/src/directives/sortable.js
@@ -89,7 +89,7 @@ ngeo.sortableDirective = function($timeout) {
             }
 
             if (!goog.isNull(dragListGroup)) {
-              goog.events.removeAll(dragListGroup);
+              dragListGroup.dispose();
             }
 
             dragListGroup = new goog.fx.DragListGroup();
--- a/src/directives/sortable.js
+++ b/src/directives/sortable.js
@@ -37,7 +37,10 @@ goog.require('ngeo');
 
 
 /**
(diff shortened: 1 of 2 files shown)


src/directives/sortable.js
```diff
@@ -86,7 +89,7 @@ ngeo.sortableDirective = function($timeout) {
             }
 
             if (!goog.isNull(dragListGroup)) {
-              goog.events.removeAll(dragListGroup);
+              dragListGroup.dispose();
```
elemoine: You could also use `goog.dispose`.

elemoine:
LGTM

src/directives/sortable.js
pgiraud: Is there any real advantage?

elemoine: No, and using `dragListGroup.dispose()` may be better, because `gooo.dispose`  is not strict with types, and [does](https://github.com/google/closure-library/blob/master/closure/goog/disposable/disposable.js#L279-L288) runtime checks on the object passed.

PR #196, opened 2015-03-18
Make sortable directive use getOptions

elemoine:
Make the `sortable` directive use the `getOptions` function. Calling that function is necessary to get the default handle class name when it's not specified by the user.


asaunier:
Looks good to me

PR #197, opened 2015-03-19
Fire a change event when new background layer is changed

pgiraud:


elemoine:
I realize that ngeo.js users won't be able to register listeners, because there's no exported register function they can use. One option is to inherit from ol.Observable instead of goog.events.EventTarget. With that ngeo.js users will be able to use the on and once functions. And users using Closure will still be able to use goog.events.listen.

pgiraud:
Pull request updated.
It now inherits from `ol.Observable` + a `previous` property is provisioned.

externs/ngeox.js
```diff
@@ -326,3 +326,14 @@ ngeox.SearchDirectiveListeners.prototype.selected;
  * @type {function(jQuery.event, Object, TypeaheadDataset)|undefined}
  */
 ngeox.SearchDirectiveListeners.prototype.autocompleted;
+
+
+/**
+ * @typedef {{previous: (ol.layer.Base|undefined)}}
```
elemoine: This should be interface.

src/services/backgroundlayermgr.js
```diff
@@ -21,19 +21,67 @@
...
+  CHANGE: 'change'
+};
+
+
+
+/**
+ * @constructor
+ * @extends {goog.events.Event}
+ * @param {ngeo.BackgroundEventType} type Type.
+ * @param {ol.layer.Base|undefined=} opt_previous
```
elemoine: `@implements` missing.

pgiraud:
Updated.

elemoine:
Please merge.

PR #199, opened 2015-03-20
Fix scale order in scale selector

elemoine:
With this commit scales are ordered by their zoom levels in the scale dropdown.


asaunier:
Looks good

elemoine:
Thanks for the review @asaunier!

PR #200, opened 2015-03-23
Add getPath and setPath methods to Location

elemoine:
Useful when you want to change the path of the current location instead of just updating query parameters.


pgiraud:
OK, looks good.

PR #204, opened 2015-03-25
Add MapFish Print v3 support

elemoine:
This PR is about adding support for printing with MapFish Print v3. This is very WIP for the moment.

### DIFF
--- a/buildtools/example.json
+++ b/buildtools/example.json
@@ -23,7 +23,8 @@
       ".build/externs/angular-1.3-http-promise.js",
       ".build/externs/jquery-1.9.js",
       "externs/d3.js",
-      "externs/typeahead.js"
+      "externs/typeahead.js",
+      "externs/mapfish-print-v3.js"
     ],
     "define": [
       "goog.dom.ASSUME_STANDARDS_MODE=true",
--- a/buildtools/examples-all.json
+++ b/buildtools/examples-all.json
@@ -22,7 +22,8 @@
       ".build/externs/angular-1.3-http-promise.js",
       ".build/externs/jquery-1.9.js",
       "externs/d3.js",
-      "externs/typeahead.js"
+      "externs/typeahead.js",
+      "externs/mapfish-print-v3.js"
     ],
     "define": [
       "goog.dom.ASSUME_STANDARDS_MODE=true",
--- a/buildtools/ngeo.json
+++ b/buildtools/ngeo.json
@@ -26,7 +26,8 @@
       ".build/externs/jquery-1.9.js",
       "externs/ngeox.js",
       "externs/d3.js",
-      "externs/typeahead.js"
+      "externs/typeahead.js",
+      "externs/mapfish-print-v3.js"
     ],
     "define": [
       "goog.DEBUG=false"
--- a/examples/mapfishprint.html
+++ b/examples/mapfishprint.html
@@ -1,0 +1,28 @@
+<!DOCTYPE>
+<html ng-app='app'>
+  <head>
+    <title>Print example</title>
+    <meta charset="utf-8">
+    <meta name="viewport"
+          content="initial-scale=1.0, user-scalable=no, width=device-width">
+    <meta name="apple-mobile-web-app-capable" content="yes">
+    <link rel="stylesheet" href="../node_modules/openlayers/css/ol.css" type="text/css">
+    <style>
+      #map {
+        width: 600px;
+        height: 400px;
+      }
+    </style>
+  </head>
+  <body ng-controller="MainController as ctrl">
+    <div id="map" ngeo-map="ctrl.map"></div>
+    <button ng-click="ctrl.print()">Print</button>
+    <span ng-bind="ctrl.printState"></span>
+    <p id="desc">This example shows how to use <code>ngeo.Print</code> to print an OpenLayers map with MapFish Print v3.</p>
+    <p>In this example the best print scale is chosen based on the current map resolution.</p>
+    <script src="http://cdnjs.cloudflare.com/ajax/libs/proj4js/2.2.1/proj4.js" type="text/javascript"></script>
+    <script src="http://epsg.io/21781.js" type="text/javascript"></script>
+    <script src="../node_modules/angular/angular.js"></script>
+    <script src="/@?main=mapfishprint.js"></script>
+  </body>
+</html>
--- a/examples/mapfishprint.js
+++ b/examples/mapfishprint.js
@@ -1,0 +1,253 @@
+goog.provide('mapfishprint');
+
+goog.require('ngeo.CreatePrint');
+goog.require('ngeo.Print');
+goog.require('ngeo.PrintUtils');
+goog.require('ngeo.mapDirective');
+goog.require('ol.Map');
+goog.require('ol.View');
+goog.require('ol.layer.Image');
+goog.require('ol.proj');
+goog.require('ol.source.ImageWMS');
+
+
+/** @const **/
+var app = {};
+
+
+/** @type {!angular.Module} **/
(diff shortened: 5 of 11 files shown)


examples/print.html
```diff
@@ -0,0 +1,27 @@
...
+        width: 600px;
+        height: 400px;
+      }
+    </style>
+  </head>
+  <body ng-controller="MainController as ctrl">
+    <div id="map" ngeo-map="ctrl.map"></div>
+    <button ng-click="ctrl.print()">Print</button>
+    <span ng-bind="ctrl.printState"></span>
+    <p id="desc">This example shows how to use the <code>ngeo-map</code> directive to insert an OpenLayers map in a page.</p>
```
elemoine: Change example description.

examples/print.js
```diff
@@ -0,0 +1,248 @@
...
+
+  var i, ii;
+  for (i = 0, ii = app.PRINT_SCALES_.length; i < ii; i++) {
+    if (pageScale > app.PRINT_SCALES_[i]) {
+      optimal = app.PRINT_SCALES_[i];
+    }
+  }
+
+  return optimal;
+};
```
elemoine: This function does not seem to work for my example.

examples/print.js
```diff
@@ -0,0 +1,248 @@
...
+
+
+/**
+ * @param {string} ref Ref.
+ * @param {Object} data Data.
+ * @private
+ */
+app.MainController.prototype.handleGetStatusSuccess_ = function(ref, data) {
+  var done = data['done'];
```
elemoine: We need new types in the mapfish-print-v3.js externs file for the print responses.

elemoine:
This PR is now ready for review.

There is still a lot that to be done, for example to support more layer/source types. But this PR provides the base, and I think it makes sense to review and merge this before adding more functionality.

So this PR adds two services: `ngeoCreatePrint` and `ngeoPrintUtils`.

`ngeoCreatePrint` is a factory function for creating `ngeo.Print` objects. `ngeo.Print` objects are used to interact with a MapFish Print service. `ngeo.Print` exposes functions like `createReport` and `getStatus` that send requests to the MapFish Print service.

`ngeoPrintUtils` is a object with two functions: `getOptimalScale` and `createPrintMaskPostcompose`. The first function can be used to get the optimal map print scale for a given size and resolution of the OpenLayers map. The second function creates a function that can be used as a map `postcompose` listener to add a "print mask" on the map.

See http://erilem.net/ngeo/print/print.html for a simple example showing how to use the `ngeoCreatePrint` and `ngeoPrintUtils` services in an application. This example relies on a MapFish Print instance on http://geomapfish.demo-camptocamp.com.

Note that the build fails on Travis because this PR depends on https://github.com/openlayers/ol3/pull/3612, which is not yet merged.

Please review.

marcjansen:
ol3#3612 is ready to be merged.

examples/print.js
```diff
@@ -0,0 +1,245 @@
...
+
+  var spec = this.print_.createSpec(map, scale, dpi, layout, {
+    'datasource': [],
+    'debug': 0,
+    'comments': 'My comments',
+    'title': 'My print'
+  });
+
+  this.print_.createReport(spec).then(
+      angular.bind(this, this.handleCreateReportSuccess_),
```
asaunier: I don't remember why we sometimes use goog.bind and sometimes angular.bind in ngeo?

elemoine: This is explained in the [developer guide](https://github.com/camptocamp/ngeo/blob/master/docs/developer-guide.md#authoring-examples):

« As described above the gh-pages make target can be used to publish the examples to github.io. The examples published on github.io are not compiled, they rely on the standalone ngeo.js build.

Because of that the examples cannot rely on non-exported symbols or properties. For example they cannot use goog objects and they cannot use ol objects that are not part of the OpenLayers API (that is marked with @api in the OpenLayers source code). »

asaunier:
Looks good to me.

Would it makes to add scale/orientation selectors in the example? Those are features that would most likely be very useful in real UIs.

examples/print.js
asaunier: Right! :)

elemoine:
> Would it makes to add scale/orientation selectors in the example? Those are features that would most likely be very useful in real UIs.

The example had that at some point, but I finally decided against it because it made the example too complex to my taste. I could try again if you feel strong about that.

I'll also need to make changes for https://github.com/camptocamp/ngeo/pull/216.

examples/print.js
elemoine: No problem. I know these notions are complex, and require quite some experience with Closure.

examples/print.js
```diff
@@ -0,0 +1,245 @@
...
+ */
+app.MainController = function($timeout, ngeoCreatePrint, ngeoPrintUtils) {
+
+  var projection = ol.proj.get('EPSG:21781');
+  projection.setExtent([485869.5728, 76443.1884, 837076.5648, 299941.7864]);
+
+  /**
+   * @type {ol.Map}
+   */
+  this['map'] = new ol.Map({
```
elemoine: Use `@export`. Here and elsewhere in the example.

gberaudo:
I have the impression that most of the code is about implementing a pure Mapfish print client in javascript, with no dependency on Angular.

What about extracting these files (externs/mapfish-print-v3.js, src/services/print.js, src/services/printutils.js, test/spec/services/print.spec.js) in an independant project and only have a very small glue in ngeo?

That project could be called `ol3-mapfish` and be used as an Ol3 plugin by the larger Ol3/Mapfish community.

elemoine:
I'd like to have more opinions about that idea…

elemoine:
I seek other opinions, because I do not agree :-) What about the extensions we have in d3-ext and ol3-ext? Are you also suggesting to create separate libraries/repositories for those as well? That is not the way I see things. I don't want to have to pull dozens of dependencies in our applications. This would make our work, both as project and library developers, a lot more complex. I'd like to avoid that. For example, ngeo provides all the infrastructure we need, and I would not like to repeat/duplicate that infrastructure. Let's ngeo be our library we use with OpenLayers 3, and let's use the Closure Compiler to select what we use in our applications.

fredj:
I agree that creating a new project implies a lot a work.

Do we want to provide and maintain a "generic" library to integrate ol3 and mapfish print v3?
If other people are interested it would be a good idea to move it outside of ngeo.

gberaudo:
> I seek other opinions, because I do not agree :-) What about the
> extensions we have in d3-ext and ol3-ext? Are you also suggesting to
> create separate libraries/repositories for those as well? That is not

If these extensions are tiny and may not be of interest for a broader
community it is of course fine to keep them in ngeo.

However Mapfish-print looks different as it may interest the whole Ol3
community. I remember some discussion at the OL3 code sprint where
someone used Mapfish-print v2 and was looking to port to v3 + Ol3.

It seems similar to Ol3-cesium where we have a standalone repository and
we integrate using closure.

elemoine:
Ok. The MapFish Print code is just Angular services right now, so it's very easy to extract if we need to. So what we can add is add it to ngeo for now and move it out to a separate project if people are interested. How does that sound to you?

fredj:
Sounds very good, thanks for your work on this.

gberaudo:
Sounds good to me and a nice addition to ngeo.

elemoine:
For some reason, I can't get Travis to run for this PR. Any idea?

fredj:
I already had this issue once: try to cancel the build (there's a button in https://travis-ci.org/camptocamp/ngeo/builds/62231749) and then restart it

elemoine:
I did that multiple times already, but to no available.

elemoine:
I'm going to try creating a new PR.

PR #205, opened 2015-03-31
Formatter can be configured partially

pgiraud:
I'm using `goog.object.extend` so that the user can configure only part of the `formatter`.
Please review.


gberaudo:
Though user code still need a cast, it is a nice improvement.
Please merge.

PR #206, opened 2015-03-31
Profile hovercallback

pgiraud:
#205 should be merged first.

The goal for this pull request is to send the distance and elevation to the hoverCallback so that we can format display on the chart and on the map the same way.

Please review.


gberaudo:
LGTM

PR #207, opened 2015-04-07
Fix compile errors

elemoine:
This PR fixes compile errors that we get because we now [use](https://github.com/openlayers/closure-util/pull/51) a more recent version of the compiler.

The new compiler is much more strict with respect to the presence/absence of `goog.provide` and `goog.require`.

And, unfortunately, I had to move `missingRequire` from `jscomp_error` to `jscomp_off` for the compilation of the examples. Otherwise we get compilation errors when `ngeox.js` is used as a source file rather than an externs file, which is what to do when compiling the application code. Indeed, `ngeox.js` cannot include `goog.provide`'s because it is used as an externs file for `ngeo.js`.

Please review.

### DIFF
--- a/buildtools/example.json
+++ b/buildtools/example.json
@@ -22,7 +22,6 @@
       ".build/externs/angular-1.3-q.js",
       ".build/externs/angular-1.3-http-promise.js",
       ".build/externs/jquery-1.9.js",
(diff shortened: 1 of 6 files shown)


pgiraud:
LGTM.

PR #210, opened 2015-04-08
Interaction map -> profile

pgiraud:
With this pull request I added a way to manage interaction between map and interaction.
When the user hovers the map and the cursor is close to the line, then the corresponding value in the chart is highlighted.

Changes are required in the HTML code:
 - `ngeo-profile` attribute is now called `ngeo-profile-data`,
 - `ngeo-profile` attribute can now be used to store the D3 profile component.

Please review.

### DIFF
--- a/examples/profile.html
+++ b/examples/profile.html
@@ -10,6 +10,10 @@
     <link rel="stylesheet" href="http://getbootstrap.com/dist/css/bootstrap.min.css" type="text/css">
     <style>
 
+        #map {
+          width: 600px;
+          height: 400px;
+        }
         .profile {
           width: 450px;
           height: 250px;
@@ -18,9 +22,11 @@
     </style>
   </head>
   <body ng-controller="MainController as ctrl">
+    <div id="map" ngeo-map="ctrl.map"></div>
     <div class="profile" ngeo-profile="::ctrl.profileData"
          ngeo-profile-options="::ctrl.profileOptions"
-         ngeo-profile-pois="::ctrl.profilePoisData">
+         ngeo-profile-pois="::ctrl.profilePoisData"
+         ngeo-profile-highlight="ctrl.profileHighlight">
     </div>
 
     <ul ng-show="ctrl.point != null">
--- a/examples/profile.js
+++ b/examples/profile.js
@@ -2,7 +2,20 @@ goog.provide('profile');
 
 
 goog.require('ngeo');
+goog.require('ngeo.mapDirective');
 goog.require('ngeo.profileDirective');
+goog.require('ol');
+goog.require('ol.Attribution');
+goog.require('ol.Feature');
+goog.require('ol.FeatureOverlay');
+goog.require('ol.Map');
+goog.require('ol.View');
+goog.require('ol.geom.GeometryLayout');
+goog.require('ol.geom.LineString');
+goog.require('ol.layer.Image');
+goog.require('ol.proj');
+goog.require('ol.proj.Projection');
+goog.require('ol.source.ImageWMS');
 
 
 /** @const **/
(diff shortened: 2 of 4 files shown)


examples/profile.html
```diff
@@ -18,7 +22,10 @@
     </style>
   </head>
   <body ng-controller="MainController as ctrl">
-    <div class="profile" ngeo-profile="::ctrl.profileData"
+    <div id="map" ngeo-map="ctrl.map"></div>
+    <div class="profile"
+         ngeo-profile="ctrl.profile"
+         ngeo-profile-data="::ctrl.profileData"
```
gberaudo: It would also be possible to add an `ngeo-profile-highlight` attribute and call `profile.highlight` in the directive.

examples/profile.js
```diff
@@ -21,16 +34,92 @@ app.module = angular.module('app', ['ngeo']);
  */
 app.MainController = function($http, $scope) {
 
+  this['profile'] = null;
```
gberaudo: It seems unusual to have the object set to null in the controller and have it instanciated by the directive.

examples/profile.js
```diff
@@ -21,16 +34,92 @@ app.module = angular.module('app', ['ngeo']);
...
 
   $http.get('data/profile.json').then(angular.bind(this, function(resp) {
-    this['profileData'] = resp.data['profile'];
+    var data = resp.data['profile'];
+    this['profileData'] = data;
+
+    var i;
+    var len = data.length;
+    var lineString = new ol.geom.LineString([],
+        /** @type {ol.geom.GeometryLayout} */ ('XYM'));
```
gberaudo: What about directly using `ol.geom.GeometryLayout.XYM`?

elemoine: No, because the examples use ngeo.js when hosted on github.io.

examples/profile.js
```diff
@@ -123,4 +210,24 @@ app.MainController = function($http, $scope) {
...
+  var dy = closestPoint[1] - coordinate[1];
+  var dist = Math.sqrt(dx * dx + dy * dy);
+  var pixelDist = dist / this['map'].getView().getResolution();
+
+  if (pixelDist < 8) {
+    this['profile'].highlight(closestPoint[2]);
+  } else {
+    this['profile'].highlight(-1);
+  }
+};
```
gberaudo: An alternative is to modify a `this['profileHighlight']` property which would be watched by the directive.

src/d3-ext/profile.js
```diff
@@ -368,56 +387,79 @@ ngeo.profile = function(options) {
...
+    outCallback.call(null);
+  };
+
+  /**
+   * Highlight the given distance and corresponding elevation on chart.
+   * Fire the hoverCallback callback with corresponding point.
+   * @param {number} distance Distance.
+   */
+  profile.highlight = function(distance) {
+    if (goog.isNull(distance) || distance == -1) {
```
gberaudo: May a number be `null`? Should not it be tested against `undefined` instead?

examples/profile.js
```diff
@@ -92,22 +181,20 @@ app.MainController = function($http, $scope) {
...
-  var that = this;
-
   /**
    * @param {Object} point
    */
   var hoverCallback = function(point) {
     // An item in the list of points given to the profile.
     that['point'] = point;
+    that.snappedPoint_.setGeometry(new ol.geom.Point([point.x, point.y]));
+    $scope.$apply();
```
gberaudo: Instead of calling `$apply` here, it would be more robust and clear to call it from the first function receiving the browser event.

pgiraud:
@gberaudo I actually tried with a `ngeo-profile-highlight` attribute. I had several problems. One of them was `$apply` being called several times.
I had a discussion with @elemoine on this topic yesterday and we concluded that the current implementation is the best one.
I'd be happy to give you a more detailed information if you're not convinced.

examples/profile.js
pgiraud: I don't get the point here.

gberaudo: IMO, `$apply` should only be called in event listeners. Otherwise you cannot know whether it was called or not so it may throw. I was suggesting to call the `hoverCallback`  and `outCallback` from an applied function like:
`$apply(function() {hoverCallback(point);});` from the profile code since it is the one listening for browser events.

pgiraud: But the code in `d3-ext/profile.js` doesn't know nothing about scope and it shouldn't.

elemoine: doesn't -> does, no?

gberaudo: Technically, the profile directive could proxy the callback.

For example for `hoverCallback` it could do:
```js
var newHoverCallback = function(point) {
  originalHoverCallback(point);
  scope.$apply();
  }
```

pgiraud: I just tried:
https://github.com/pgiraud/ngeo/commit/793dec6f94f0eaed0b75d228557b794525031132

But I then get a "too much recursion" error.
Did you have something similar in mind?

pgiraud:
Remarks taken into account. Please review again.

src/d3-ext/profile.js
```diff
@@ -368,56 +387,79 @@ ngeo.profile = function(options) {
...
+    outCallback.call(null);
+  };
+
+  /**
+   * Highlight the given distance and corresponding elevation on chart.
+   * Fire the hoverCallback callback with corresponding point.
+   * @param {number} distance Distance.
+   */
+  profile.highlight = function(distance) {
+    if (distance == -1) {
```
gberaudo: Testing for `< 0` would handle all negative values.

examples/profile.js
gberaudo: Do we still need to export the profile and call the highlight or should it now use the directive highlight attribute?
If yes, then you can use `clearHighlight` instead of passing `-1`.

pgiraud: Using an attribute in the directive for that isn't possible in my opinion. Or at least too complex.
You're right, I'll use `clearHighlight` directly.

src/directives/profile.js
```diff
@@ -65,6 +93,12 @@ ngeo.profileDirective = function() {
               profile.showPois(poiData);
             }
           }
+
+          scope.$watch(attrs['ngeoProfileHighlight'], function(newVal, oldVal) {
+            if (goog.isDef(profile) && goog.isDef(elevationData)) {
+              profile.highlight(newVal);
+            }
+          });
```
pgiraud: This piece should be removed. It's a remaining code from when I was experimenting with a `highlight` directive attribute.

examples/profile.js
gberaudo: I see there is still an issue when `highlight()` is called while we are in a digest cycle: the wrapped hoverCallback will call `$apply()` again. I think it would be better to remove the wrapper and instead have the profile know about `$apply()`. Then it would be able to call apply only when handling a browser events and in the other cases, it would be assumed that the user is in a digest cycle or will apply herself.
We can use a workaround to hide the scope from the profile function, but why would we want to hide it?

pgiraud:
After a good discussion with @elemoine on when it's safe to call `$apply` and after discovering `$applyAsync`, I reworked on this pull request. I'm coming back with a slightly different implementation where the application doesn't know nothing about the profile component and only talks to the directive.

src/directives/profile.js
```diff
@@ -39,11 +40,34 @@ ngeo.profileDirective = function() {
...
 
           scope.$watchCollection(optionsAttr, function(newVal) {
-            options = newVal;
+
+            var options = /** @type {ngeox.profile.ProfileOptions} */
+                (goog.object.clone(newVal));
+
             if (goog.isDef(options)) {
+
+              // proxy the hoverCallback in order to be able to call $apply
```
elemoine: the hoverCallback **and** the outCallback

src/directives/profile.js
```diff
@@ -39,11 +40,34 @@ ngeo.profileDirective = function() {
...
+              if (goog.isDef(options.hoverCallback)) {
+                var origHoverCallback = options.hoverCallback;
+                var hoverCallback = function(point) {
+                  origHoverCallback(point);
+                  scope.$applyAsync();
+                };
+                options.hoverCallback = hoverCallback;
```
elemoine: Minor, but you don't need the `hoverCallback` variable:

```js
if (goog.isDef(options.hoverCallback)) {
  var origHoverCallback = options.hoverCallback;
  options.hoverCallback = function(point) {
    scope.$applyAsync(function() {
      origHoverCallback(point);
    });
  };
}
```

elemoine: Also, reading the code, it is not clear why you need to use `$applySync` rather than `$apply` here. I do think this deserves a line of explanation.

gberaudo:
I agree this new version is easier to use. LGTM for merging when @elemoine comments are addressed.

PR #212, opened 2015-04-10
Use $applyAsync in scale selector directive

elemoine:
This PR changes the scale selector directive to use `$applyAsync` instead of `$apply` in the `change:resolution` listener. It is wrong to use `$apply` in a `change:resolution` listener, because it is not guaranteed that a `change:listener` will be executed outside Angular context.

The PR also adds tests for the scale selector directive. It also sets things up to be able to test directives with templates.

Please review.


elemoine:
Test simplified, as discussed with @pgiraud.

pgiraud:
Excellent. Please merge when travis is happy.

PR #213, opened 2015-04-13
Fix hoverCallback not being called with correct arguments

pgiraud:
Google closure didn't detect my mistake.
Please review.


elemoine:
> Google closure didn't detect my mistake.

Maybe because the code is not correctly typed. The Closure Compiler is lenient, it won't report any error if it does not have type information.

elemoine:
LGTM.

PR #214, opened 2015-04-20
Fix ol3 externs

fredj:


elemoine:
Thank you @fredj. Please merge.

PR #215, opened 2015-04-23
Remove svg tag when there's no data

pgiraud:
With this pull request the d3 component can receive `undefined` for data in which case the `svg` element is removed.
Please review.

### DIFF
--- a/src/d3-ext/profile.js
+++ b/src/d3-ext/profile.js
@@ -194,6 +194,10 @@ ngeo.profile = function(options) {
 
   var profile = function(selection) {
     selection.each(function(data) {
+      if (!goog.isDef(data)) {
+        d3.select(this).selectAll('svg').remove();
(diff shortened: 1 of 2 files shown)


gberaudo:
Out of curiosity, isn't the data cleared when passing an empty array? (see how it is done for the POIs).
That beeing said I am fine with this PR.

src/d3-ext/profile.js
```diff
@@ -194,6 +194,10 @@ ngeo.profile = function(options) {
 
   var profile = function(selection) {
     selection.each(function(data) {
+      if (!data) {
```
elemoine: Use `goog.isDef` for consistency and more strictness with types?

pgiraud:
@gberaudo I want to manage the case where data is undefined and remove the svg tag.

PR #216, opened 2015-05-05
Use export_local_property_definitions compile flag

elemoine:
Use the new export_local_property_definitions compiler option. Using this option properties can be marked with `@export` in constructors. This means we can now use the following in controller constructors instead of using the square bracket notation:

```js
app.MyController = function() {

  /**
   * @type {Object}
   * @export
   */
  this.aProperty = {};

  // …
};
```

### DIFF
--- a/examples/control.js
+++ b/examples/control.js
@@ -23,8 +23,11 @@ app.module = angular.module('app', ['ngeo']);
  */
 app.MainController = function() {
 
-  /** @type {ol.Map} */
-  this['map'] = new ol.Map({
+  /**
+   * @type {ol.Map}
+   * @export
+   */
+  this.map = new ol.Map({
     layers: [
       new ol.layer.Tile({
         source: new ol.source.OSM()
@@ -36,8 +39,11 @@ app.MainController = function() {
     })
   });
 
-  /** @type {ol.control.Control} */
-  this['control'] = new ol.control.MousePosition({
+  /**
+   * @type {ol.control.Control}
+   * @export
+   */
+  this.control = new ol.control.MousePosition({
     className: 'mouse-position'
   });
 };
--- a/examples/interactionbtngroup.js
+++ b/examples/interactionbtngroup.js
@@ -55,8 +55,11 @@ app.MainController = function(ngeoDecorateInteraction) {
     })
   });
 
-  /** @type {ol.Map} */
-  var map = new ol.Map({
+  /**
+   * @type {ol.Map}
+   * @export
+   */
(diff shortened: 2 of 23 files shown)


asaunier:
Makes sense :)

elemoine:
@asaunier thanks for the review!

@fredj could you please review this as well, given that you now have some experience with `export_local_property_definitions`? Thanks.

examples/control.js
```diff
@@ -36,8 +39,11 @@ app.MainController = function() {
     })
   });
 
-  /** @type {ol.control.Control} */
-  this['control'] = new ol.control.MousePosition({
+  /**
+   * @type {ol.control.Control}
```
fredj: Not related to this PR but shouldn't it be `@type {ol.control.MousePosition}` ?

elemoine: Yes, it could be. I'll make the change.

examples/interactionbtngroup.js
```diff
@@ -68,40 +71,54 @@ app.MainController = function(ngeoDecorateInteraction) {
...
+   * @type {ol.interaction.Draw}
+   * @export
+   */
+  this.drawLine = new ol.interaction.Draw(
       /** @type {olx.interaction.DrawOptions} */ ({
         type: 'LineString',
         source: source
       }));
+
+  var drawLine = this.drawLine;
```
fredj: minor: this local variable is not needed

examples/interactiontoggle.js
```diff
@@ -39,21 +42,26 @@ app.MainController = function(ngeoDecorateInteraction) {
...
+   * @type {ol.interaction.Draw}
+   * @export
+   */
+  this.interaction = new ol.interaction.Draw(
       /** @type {olx.interaction.DrawOptions} */ ({
         type: 'Point',
         features: featureOverlay.getFeatures()
       }));
+
+  var interaction = this.interaction;
```
fredj: minor: this local variable is not needed

examples/interactionbtngroup.js
elemoine: Well, it's used below. This is to avoid dereferencing `this` multiple times.

fredj:
LGTM, thanks

examples/interactionbtngroup.js
fredj: > This is to avoid dereferencing this multiple times.

For performance reasons?

elemoine: Possibly yes, and to avoid typing `this` each time. You never do that?

PR #217, opened 2015-05-06
Make the goog.Uri object available in ngeoLocation

asaunier:
This change implies to rename getUri() to getUriString() in the projects using ngeoLocation. Is there a changelog to advertise this?


elemoine:
Could you please add getUriString to https://github.com/camptocamp/ngeo/blob/master/exports/services/location.js?

asaunier:
> Could you please add getUriString

Thanks for the reminder. Done.

elemoine:
Please merge. Thanks.

PR #219, opened 2015-05-13
Add requestEncoding to mapfish print externs

elemoine:
Add missing `requestEncoding` property to `MapFishPrintWmtsLayer` in the `mapfish-print-v3.js` externs file.

Please review.


asaunier:
Makes sense!

PR #221, opened 2015-05-19
Support printing of vector layers

elemoine:
This PR adds support for printing vector layers. It basically adds new encoders to the  Print service. This is still work in progress as icon and text styles are not yet supported.


elemoine:
So this PR adds support for printing vector layers. With this PR styled polygons and lines, and points styled with `ol.style.Circle` and `ol.style.Icon` are printed. Labels (`ol.style.Text`) and regular shapes (`ol.style.RegularShape`) are not yet supported.

Please review.

The mapfishprint example is available at http://erilem.net/ngeo/print-vector/mapfishprint.html.

elemoine:
Thanks for the good review @tsauerwein! Taking a close look at your comments now…

elemoine:
> Or we could try to use the in operator in the style filter. That is create a separate mfp style for each ol style (like before) and make the style property on the feature a list which contains the ids of all styles. Then the filter expression would be something like this:
> 
>  '[' + styleId + ' in ' + ngeo.Print.FEAT_STYLE_PROPERTY_ + ']'
> 
> I am skeptical if this is valid, but it would be worth a try.

This does not seem to work. I get a GeoTools exception when I use that:

```
13:49:03.279 [PrintJobManager-1] DEBUG o.mapfish.print.servlet.job.PrintJob - Print Job d4fb6f53-a1a7-43c3-a0ab-f0ae4f6dccf9@91c37f58-5624-4937-ad03-d260afc542ae completed in 817ms
13:49:04.777 [http-8080-1] ERROR o.m.print.servlet.BaseMapServlet - Error while processing request: org.geotools.filter.text.cql2.CQLException: Encountered "\'28\' in" at line 1, column 2.
Was expecting one of:
    "not" ...
    "include" ...
    "exclude" ...
    "(" ...
    "[" ...
    "id" ...
    "in" ...
    <IDENTIFIER> ...
    <DATE_TIME> ...
    <DATE> ...
    "-" ...
    <INTEGER_LITERAL> ...
    <FLOATING_LITERAL> ...
    <STRING_LITERAL> ...
    <STRING_LITERAL> "*" ...
    <STRING_LITERAL> "/" ...
    <STRING_LITERAL> "+" ...
    <STRING_LITERAL> "-" ...
    <STRING_LITERAL> "]" ...
    <STRING_LITERAL> "not" ...
    <STRING_LITERAL> "like" ...
    <STRING_LITERAL> "ilike" ...
    <STRING_LITERAL> "exists" ...
    <STRING_LITERAL> "does-not-exist" ...
    <STRING_LITERAL> "is" ...
    <STRING_LITERAL> "between" ...
    <STRING_LITERAL> "=" ...
    <STRING_LITERAL> ">" ...
    <STRING_LITERAL> "<" ...
    <STRING_LITERAL> ">=" ...
    <STRING_LITERAL> "<=" ...
    <STRING_LITERAL> "<>" ...
     Parsing : ['28' in _ngeo_style].
```

elemoine:
Another option: use a feature style property for each style. E.g.

```
styles: {
  "[ _ngeo_style_0 = '28' ]": {
    "symbolizers": […]
  },
  "[_ngeo_style_1 = '29' ]": {
    "symbolizers": […]
  }
}
```

I think that should work…

tsauerwein:
> Another option: use a feature style property for each style. 

Good idea!

elemoine:
> Yes, right. Though I am not sure if this technique will still work when array of styles are converted correctly.

I think it works. Say I have two features, `feature0` and `feature1`, both styled with the same style objects, `style0` and `style1`. This is what the print spec will look like:

```json
styles: {
  "[ _ngeo_style_0 = '28' ]": {
    "symbolizers": [{
    }]
  },
  "[_ngeo_style_1 = '29' ]": {
    "symbolizers": [{
    }]
  }
}
```

The `encodeVectorStyle_` returns immediately if there already is a style object for this feature style property and style id.

The tests cover that case.

tsauerwein:
> I think it works. Say I have two features, feature0 and feature1, both styled with the same style objects, style0 and style1. 

Yes, with this technique it will work.

elemoine:
Radians now converted to degrees for the icon rotation. See https://github.com/camptocamp/ngeo/pull/221/files#diff-78a84f2934a358cd32ef75a64910711cR442. I've also verified that a positive value means clockwise both in SLD and OpenLayers 3.

elemoine:
@tsauerwein, I think I've addressed all of your comments. Please tell me if you have more.

tsauerwein:
No, this is looking good to me!

elemoine:
Travis finally ran! Thanks again for the excellent review @tsauerwein.

PR #222, opened 2015-05-21
Make it possible to pass $http options to print

elemoine:
This commit does two things: (1) it allows passing `$http` options to `createReport` and `getStatus`, making it possible to pass a request canceler to `$http`, and (b) change the promise returned by `createReport` and `getStatus` to be an http promise instead of a regular promise, providing more flexibility and making it possible to be notified in case of HTTP errors.

### DIFF
--- a/examples/mapfishprint.js
+++ b/examples/mapfishprint.js
@@ -192,11 +192,12 @@ app.MainController.prototype.print = function() {
 
 
 /**
- * @param {MapFishPrintReportResponse} resp Response.
+ * @param {!angular.$http.Response} resp Response.
  * @private
  */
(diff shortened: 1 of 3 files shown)


pgiraud:
Reading this http://stackoverflow.com/questions/13928057/how-to-cancel-an-http-request-in-angularjs helps understanding this pull request.

pgiraud:
After a good explanation this makes sense. Please merge.

elemoine:
Yes, the cancellation of an `$http` request occurs when the user resolves the promise that was to `$http` through the `timeout` config option. The user is responsible for creating and resolving that promise.

PR #223, opened 2015-05-21
Remove unused externs/angular-local-storage.js

### LINKED ISSUE #220: remove externs/angular-local-storage.js ?
We're not using this library anymore (in schweizmobil_re3) . Do we want to keep the extern?

fredj:
Fixes #220


elemoine:
It would not harm to keep this externs file but it  does not matter to me. Please merge if you think this is really useless.

elemoine:
@fredj, feel free to merge this.

PR #226, opened 2015-06-03
Allow tiles from a TileWMS to be printed

Jenselme:
- Use encodeImageWmsLayer to encode layer of type Tile and of source TileWMS
- Add a function to get the correct URL if the source is TileWMS and not
  ImageWMS
- Add a custom parameter in the object to get images with transparency

### DIFF
--- a/src/services/print.js
+++ b/src/services/print.js
@@ -193,18 +193,30 @@ ngeo.Print.prototype.encodeImageLayer_ = function(arr, layer) {
 
 /**
  * @param {Array.<MapFishPrintLayer>} arr Array.
- * @param {ol.layer.Image} layer Layer.
+ * @param {ol.layer.Tile|ol.layer.Image} layer Layer.
  * @private
  */
 ngeo.Print.prototype.encodeImageWmsLayer_ = function(arr, layer) {
-  goog.asserts.assertInstanceof(layer, ol.layer.Image);
+  var url;
   var source = layer.getSource();
-  goog.asserts.assertInstanceof(source, ol.source.ImageWMS);
-  var url = source.getUrl();
+
+  if (source instanceof ol.source.ImageWMS) {
+    goog.asserts.assertInstanceof(layer, ol.layer.Image);
+    goog.asserts.assertInstanceof(source, ol.source.ImageWMS);
+    url = source.getUrl();
+  } else {
+    goog.asserts.assertInstanceof(layer, ol.layer.Tile);
+    goog.asserts.assertInstanceof(source, ol.source.TileWMS);
+    url = source.getUrls()[0];
+  }
+
   var params = source.getParams();
   var object = /** @type {MapFishPrintWmsLayer} */ ({
     baseURL: url,
     imageFormat: 'FORMAT' in params ? params['FORMAT'] : 'image/png',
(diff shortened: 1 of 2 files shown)


src/services/print.js
```diff
@@ -222,6 +234,8 @@ ngeo.Print.prototype.encodeTileLayer_ = function(arr, layer) {
   var source = layer.getSource();
   if (source instanceof ol.source.WMTS) {
     this.encodeTileWmtsLayer_(arr, layer);
+  } else if (source instanceof ol.source.TileWMS) {
+    this.encodeImageWmsLayer_(arr, layer);
```
elemoine: My plan was to have an `encodeTileWmsLayer_` function for tile layers with an `ol.source.TileWMS` source. And we can have an `encodeWmsLayer_` function if `encodeImageWmsLayer_` and `encodeTileWmsLayer_` have common code.

src/services/print.js
```diff
@@ -193,18 +193,30 @@ ngeo.Print.prototype.encodeImageLayer_ = function(arr, layer) {
...
+    url = source.getUrls()[0];
+  }
+
   var params = source.getParams();
   var object = /** @type {MapFishPrintWmsLayer} */ ({
     baseURL: url,
     imageFormat: 'FORMAT' in params ? params['FORMAT'] : 'image/png',
+    customParams: {
+      transparent: true
+    },
```
elemoine: Not sure about this one. Why assuming "transparent" here?

elemoine:
Thanks a lot for your contribution. I've added two comments. I look forward to merging your branch.

src/services/print.js
Jenselme: It is compulsory to print multiple layers.

Say I want to print layer1 (let's assume it is the background) and layer2 (rivers for instance). Without the transparent option, I will only see the image of layer2 (so the rivers) and not the background since the image of layer2 is the rivers with blank pixel between the rivers.

Should I add a comment to make this clearer?

Jenselme:
> My plan was to have an encodeTileWmsLayer_ function for tile layers with an ol.source.TileWMS source. And we can have an encodeWmsLayer_ function if encodeImageWmsLayer_ and encodeTileWmsLayer_ have common code.

Proposed in 3ced295

src/services/print.js
elemoine: I think we should do this:

```js
  var customParams = goog.object.extend({
    'TRANSPARENT': true
  }, params);
  goog.object.remove(customParams, 'LAYERS');
  goog.object.remove(customParams, 'FORMAT');
  var object = /** @type {MapFishPrintWmsLayer} */ ({
    baseURL: url,
    imageFormat: 'FORMAT' in params ? params['FORMAT'] : 'image/png',
    layers: params['LAYERS'].split(','),
    customParams: customParams,
    type: 'wms'
  });
```

In this way we use what the user set in the OpenLayers layer instance. Does it make sense?

Jenselme: It does. Do you want me to update the PR?

elemoine: That would be great.

Jenselme:
PR updated

elemoine:
Awesome.

PR #228, opened 2015-06-04
Add a getCapabilities method to ngeo.Print

Jenselme:


src/services/print.js
```diff
@@ -550,6 +549,24 @@ ngeo.Print.prototype.getReportUrl = function(ref) {
 
 
 /**
+ * Get the capabilities of the report
+ * @param {function()} success Success callback
+ * @param {function()=} opt_error Error callback
```
elemoine: These callbacks are not needed. The `getCapabilities` returns the http promise so the user can just call `success`, `error `or `then` on the promise.

Jenselme:
> These callbacks are not needed. The getCapabilities returns the http promise so the user can just call success, erroror then on the promise. 

Indeed. Fixed (commits squashed).

src/services/print.js
```diff
@@ -550,6 +549,19 @@ ngeo.Print.prototype.getReportUrl = function(ref) {
 
 
 /**
+ * Get the capabilities of the report
```
elemoine: Can you please just rephrase this to "Get the print capabilities from MapFish Print."? Once done I'll merge your branch. Thanks a lot for your contribution.

Jenselme:
Done

src/services/print.js
```diff
@@ -550,6 +549,19 @@ ngeo.Print.prototype.getReportUrl = function(ref) {
 
 
 /**
+ * Get the print capabilities from MapFish Print.
+ * @param {angular.$http.Config=} opt_httpConfig $http config object
+ * @return {Object}
```
elemoine: I've just realized that the return type is not correct. It should be `angular.$http.HttpPromise`.

src/services/print.js
```diff
@@ -550,6 +549,19 @@ ngeo.Print.prototype.getReportUrl = function(ref) {
 
 
 /**
+ * Get the print capabilities from MapFish Print.
+ * @param {angular.$http.Config=} opt_httpConfig $http config object
```
elemoine: Please add a `.` at end of line.

Jenselme:
> Please add a . at end of line.

Done

> I've just realized that the return type is not correct. It should be angular.$http.HttpPromise.

Corrected

elemoine:
Cool, will merge when the build passed on Travis.

PR #229, opened 2015-06-05
Fix layertree example

### LINKED ISSUE #227: Layer tree example is broken
It seems to me.
I don't see any tree in the page :http://camptocamp.github.io/ngeo/master/layertree.html

elemoine:
This commit fixes #227, reported by @fgravin. The bug is related to Angular's one-time binding (correctly) considering `null` values as defined values. The fix involves using `undefined` instead of `null` as the initial value for the `tree` data model.

Fixes #227.


pgiraud:

PR #234, opened 2015-06-09
Display measurements even on touch devices

pgiraud:
With this pull request, I make it so the measurement can be displayed even on touch devices.
Now the help tooltip is the only one to be shown on mousemove. The measurement tooltip displays and updates (content and position) as soon as the geometry is changed.
Please review.

### DIFF
--- a/src/ol-ext/interaction/measure.js
+++ b/src/ol-ext/interaction/measure.js
@@ -2,6 +2,7 @@ goog.provide('ngeo.MeasureEvent');
 goog.provide('ngeo.MeasureEventType');
 goog.provide('ngeo.interaction.Measure');
 
+goog.require('goog.asserts');
 goog.require('goog.dom');
 goog.require('goog.dom.classlist');
 goog.require('goog.events');
@@ -123,11 +124,24 @@ ngeo.interaction.Measure = function(opt_options) {
   this.sketchFeature = null;
 
   /**
+   * Message to show after the first point is clicked.
+   * @type {?Element}
+   */
+  this.continueMsg = null;
+
+  /**
    * The message to show when user is about to start drawing.
    * @type {Element}
    */
   this.startMsg = goog.isDef(options.startMsg) ? options.startMsg :
       goog.dom.createDom(goog.dom.TagName.SPAN, {}, 'Click to start drawing.');
+
+  /**
+   * The key for geometry change event.
+   * @type {?goog.events.Key}
+   * @private
+   */
+  this.changeEventKey_ = null;
 
   var style = goog.isDef(options.style) ? options.style :
       [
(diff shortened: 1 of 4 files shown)


src/ol-ext/interaction/measure.js
```diff
@@ -123,12 +123,24 @@ ngeo.interaction.Measure = function(opt_options) {
...
    * @type {Element}
    */
   this.startMsg = goog.isDef(options.startMsg) ? options.startMsg :
       goog.dom.createDom(goog.dom.TagName.SPAN, {}, 'Click to start drawing.');
 
+  /**
+   * The key for geometry change event.
+   * @type {?goog.events.Key}
+   */
+  this.changeEventKey = null;
```
elemoine: This should be made `@private` no?

src/ol-ext/interaction/measure.js
```diff
@@ -250,6 +256,16 @@ ngeo.interaction.Measure.prototype.onDrawStart_ = function(evt) {
   this.sketchFeature = evt.feature;
   this.overlay_.getFeatures().clear();
   this.createMeasureTooltip_();
+
+  this.changeEventKey = this.sketchFeature.getGeometry().on('change',
```
elemoine: Please use `goog.events.listen`.

pgiraud: I tried to do the following
```
goog.events.listen(geometry, goog.events.EventType.CHANGE, ...
```
but I get the following error when compiling:
```
ERR! compile /home/pierre/public_html/ngeo/src/ol-ext/interaction/measure.js:263: ERROR - actual parameter 1 of goog.events.listen does not match formal parameter
ERR! compile found   : (null|ol.geom.Geometry|undefined)
ERR! compile required: (EventTarget|goog.events.Listenable|null)
ERR! compile   this.changeEventKey_ = goog.events.listen(geometry,
ERR! compile                                             ^
ERR! compile 
ERR! compile 
ERR! compile 1 error(s), 0 warning(s), 98.0% typed
ERR! compile 
```

Doesn't `ol.geom.Geometry` inherit from `goog.events.Listenable`?

pgiraud: Forget my comment. I had to check that geometry is defined.

elemoine: The problem may be related to `undefined`. What if you add `goog.asserts.assert(goog.isDef(geometry))` before registering the listener?

pgiraud: So, is it better to use `goog.asserts.assert` or `if (goog.isDef(...))`?

elemoine:
> So, is it better to use goog.asserts.assert or if (goog.isDef(...))?

It depends. You should use an assertion if you, as the developer, know that `geometry` should never be `undefined` at this place in the code. And you'd use `if goog.isDef(geometry)` if there are actual cases where `geometry` will be `undefined`.

pgiraud:
Pull request updated with @elemoine's comments.

elemoine:
+1

PR #235, opened 2015-06-11
Fix OL3 version to prevent error due to FeatureOverlay deprecation

pgiraud:


elemoine:
Thank you @pgiraud. I created an issue for the work we need to do: https://github.com/camptocamp/ngeo/issues/236.

PR #238, opened 2015-06-16
Remove ol.FeatureOverlay

### LINKED ISSUE #236: ol.FeatureOverlay removal
`ol.FeatureOverlay` has been [removed](https://github.com/openlayers/ol3/pull/3758). ngeo uses `ol.FeatureOverlay` in examples and in the ol-ext measure interactions. Examples are not a problem, as we can just replace the feature overlay by a vector layer in each example. For the ol-ext measure interactions we may want to use the immeditate API instead of using an "unmanaged" vector layer.

elemoine:
This PR removes the use of `ol.FeatureOverlay`. Feature overlays are replaced by "unmanaged" vector layers, that is vector layers that are always on top.

Fixes #236.

### DIFF
--- a/examples/geolocation.js
+++ b/examples/geolocation.js
@@ -3,12 +3,13 @@ goog.provide('geolocation');
 goog.require('ngeo.DecorateGeolocation');
 goog.require('ngeo.mapDirective');
 goog.require('ol.Feature');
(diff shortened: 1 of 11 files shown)


pgiraud:
LGTM. Please merge.

PR #239, opened 2015-06-19
Solve cyclic reference when printing imported KML

Jenselme:


elemoine:
I think this is a bug in OpenLayers. The GeoJSON parser should not add a Style property to the GeoJSON feature.

Jenselme:
For your information, I reported the issue in the ol3 bugtracker: https://github.com/openlayers/ol3/issues/3832

Jenselme:
This is indeed a problem with openlayers, the issue is being solve there, no need for this patch. Closing.

elemoine:
For reference, this problem was fixed in OpenLayers. https://github.com/openlayers/ol3/pull/3900

PR #240, opened 2015-06-22
Use new geomapfish server

elemoine:
This changes the mapfishprint example to use the new geomapfish server as the previous one is not in service anymore.

But there still is a problem. The WMS layer works but print requests fail because of the CORS issue. This is the error I get in the console:

```
XMLHttpRequest cannot load http://geomapfish-demo.camptocamp.net/1.6/wsgi/printproxy/report.pdf. No 'Access-Control-Allow-Origin' header is present on the requested resource. Origin 'http://localhost:3000' is therefore not allowed access. The response had HTTP status code 404.
```

@sbrunner, I think we need your help for this.

### DIFF
--- a/examples/mapfishprint.js
+++ b/examples/mapfishprint.js
@@ -27,7 +27,7 @@ app.module = angular.module('app', ['ngeo']);
  * @const
  * @private
  */
-app.WMS_URL_ = 'http://geomapfish.demo-camptocamp.com/1.6/wsgi/' +
+app.WMS_URL_ = 'http://geomapfish-demo.camptocamp.net/1.6/wsgi/' +
     'mapserv_proxy';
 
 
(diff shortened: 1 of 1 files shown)


sbrunner:
Strange I have this header ...
Where can I test ?

elemoine:
I can merge this PR and we will test on http://camptocamp.github.io/ngeo/master/mapfishprint.html.

elemoine:
I'll let you know when the example in updated on github.io.

elemoine:
Btw, it was working properly before the server update.

elemoine:
Can now be tested on http://camptocamp.github.io/ngeo/master/mapfishprint.html…

sbrunner:
I see a `OPTIONS` request do you relay need this ?

elemoine:
It is the browser that uses an OPTIONS request to determine if the server has CORS headers.

sbrunner:
OK :-(

elemoine:
So you don't know what the problem could be?

sbrunner:
Yes I know, than the solution should be: https://github.com/camptocamp/c2cgeoportal/pull/1422

PR #241, opened 2015-06-22
Add a print cancel method

elemoine:
This PR adds a cancel method to the print service. 

Please review.


elemoine:
I'll merge this today. So now is a good time to review :)

PR #243, opened 2015-06-22
Add the possibility to print ol.style.Text

Jenselme:


src/services/print.js
```diff
@@ -488,6 +492,49 @@ ngeo.Print.prototype.encodeVectorStyleStroke_ =
...
 /**
+ * @param {Array.<MapFishPrintSymbolizer>} symbolizers Array of MapFish Print
+ *     symbolizers.
+ * @param {!ol.style.Text} textStyle Text style.
+ * @private
+ */
+ngeo.Print.prototype.encodeTextStyle_ = function(symbolizers, textStyle) {
+  var symbolizer = {
+    type: 'Text'
+  };
```
elemoine: I think we need a new type for this in the mapfish-print-v3.js externs file.

src/services/print.js
```diff
@@ -488,6 +492,49 @@ ngeo.Print.prototype.encodeVectorStyleStroke_ =
...
+  var font = textStyle.getFont().split(' ');
+  var labelAlign = textStyle.getTextAlign();
+  if (labelAlign) {
+    symbolizer.labelAlign = labelAlign;
+  }
+  var labelRotation = textStyle.getRotation();
+  if (labelRotation) {
+    symbolizer.labelRotation = labelRotation;
```
elemoine: Do we need to convert from radians to degrees here?

src/services/print.js
Jenselme: I don't know. I have no particular experience with closure. If you give me a hit, I may be able to write it.

src/services/print.js
Jenselme: We need, thanks for pointing that out. PR updated!

src/services/print.js
elemoine: Look at the [MapFishPrintSymbolizerPolygon type](https://github.com/camptocamp/ngeo/blob/master/externs/mapfish-print-v3.js#L428-L462) as an example. You'd need something similar for Text.

Jenselme: Type added

externs/mapfish-print-v3.js
```diff
@@ -460,3 +460,70 @@ MapFishPrintSymbolizerPolygon.prototype.strokeOpacity;
...
+ */
+MapFishPrintSymbolizerText.prototype.labelRotation;
+
+/**
+ * @type {string}
+ */
+MapFishPrintSymbolizerText.prototype.fontWeight;
+
+/**
+ * @type {string}
```
elemoine: `fontSize` is expected to be a `string` too?

src/services/print.js
```diff
@@ -374,6 +374,10 @@ ngeo.Print.prototype.encodeVectorStyle_ =
   } else if (!goog.isNull(imageStyle)) {
     this.encodeVectorStylePoint_(styleObject.symbolizers, imageStyle);
   }
+
+  if (!goog.isNull(textStyle) && textStyle.getText()) {
```
elemoine: Please use `goog.isDef(textStyle.getText())`.

elemoine: Or maybe remove that test and test that within the `encodeTextStyle_` function. I think that would make more sense.

src/services/print.js
```diff
@@ -488,6 +492,63 @@ ngeo.Print.prototype.encodeVectorStyleStroke_ =
...
+ *     symbolizers.
+ * @param {!ol.style.Text} textStyle Text style.
+ * @private
+ */
+ngeo.Print.prototype.encodeTextStyle_ = function(symbolizers, textStyle) {
+  var symbolizer = /** @type {MapFishPrintSymbolizerText} */ ({
+    type: 'Text'
+  });
+  var label = textStyle.getText();
+  if (label) {
```
elemoine: Do you want to use `goog.isDef(label)` here?

src/services/print.js
```diff
@@ -488,6 +492,63 @@ ngeo.Print.prototype.encodeVectorStyleStroke_ =
...
+    symbolizer.label = label;
+  }
+
+  var labelAlign = textStyle.getTextAlign();
+  if (labelAlign) {
```
elemoine: Please consider using `goog.isDef(labelAlign)`?

src/services/print.js
```diff
@@ -488,6 +492,63 @@ ngeo.Print.prototype.encodeVectorStyleStroke_ =
...
+    symbolizer.labelAlign = labelAlign;
+  }
+
+  var labelRotation = textStyle.getRotation();
+  if (labelRotation) {
```
elemoine: `goog.isDef`

src/services/print.js
```diff
@@ -488,6 +492,63 @@ ngeo.Print.prototype.encodeVectorStyleStroke_ =
...
+    // Mapfish print expects a string, not a number to rotate text
+    symbolizer.labelRotation = (labelRotation * 180 / Math.PI).toString();
+  }
+
+  var font = textStyle.getFont().split(' ');
```
elemoine: `getFont` may return an `undefined` value.

src/services/print.js
```diff
@@ -488,6 +492,63 @@ ngeo.Print.prototype.encodeVectorStyleStroke_ =
...
+    symbolizer.fontFamily = font.splice(2).join(' ');
+  }
+
+  var strokeStyle = textStyle.getStroke();
+  if (!goog.isNull(strokeStyle)) {
+    var strokeColorRgba = ol.color.asArray(strokeStyle.getColor());
+    symbolizer.haloColor = goog.color.rgbArrayToHex(strokeColorRgba);
+    symbolizer.haloOpacity = strokeColorRgba[3];
+    var width = strokeStyle.getWidth();
+    if (width) {
```
elemoine: `goog.isDef`?

elemoine:
I've just added a few comments. But it looks good. I'll merge your branch when you've addressed the comments. Thanks!

src/services/print.js
```diff
@@ -488,6 +492,63 @@ ngeo.Print.prototype.encodeVectorStyleStroke_ =
 
 
 /**
+ * @param {Array.<MapFishPrintSymbolizer>} symbolizers Array of MapFish Print
+ *     symbolizers.
+ * @param {!ol.style.Text} textStyle Text style.
+ * @private
+ */
+ngeo.Print.prototype.encodeTextStyle_ = function(symbolizers, textStyle) {
+  var symbolizer = /** @type {MapFishPrintSymbolizerText} */ ({
```
gberaudo: `MapFishPrintSymbolizerText` is a constructor, is it fine to cast a hash to an instance?
Have you considered instanciating the class?

elemoine: It's not really a constructor. It is a constructor because it is the way types are defined in an externs file.

gberaudo: You are right.

externs/mapfish-print-v3.js
Jenselme: Yes. By instance, `fontSize` is `16px`.

Jenselme:
If have just updated the PR to address all your comments.

externs/mapfish-print-v3.js
elemoine: Obviously.

PR #244, opened 2015-06-22
Preserve user defined opacity when printing WMS layers

Jenselme:


src/services/print.js
```diff
@@ -206,18 +206,20 @@ ngeo.Print.prototype.encodeImageWmsLayer_ = function(arr, layer) {
 
   var url = source.getUrl();
   var params = source.getParams();
-  this.encodeWmsLayer_(arr, url, params);
+  this.encodeWmsLayer_(arr, layer, url, params);
```
elemoine: Please pass the opacity value directly here.

Jenselme: Done. Commits squashed.

PR #246, opened 2015-06-23
Add missing goog.requires

elemoine:
This PR adds missing `goog.require`'s. I'll merge it when Travis is green.

cc @Jenselme.


Jenselme:
I find it strange that tests pass without these requires. Shouldn't the closure compiler detect them?

elemoine:
We do have the `checkRequire` flag [enabled](https://github.com/camptocamp/ngeo/blob/master/buildtools/ngeo.json#L58-Lundefined) so I wonder too.

PR #247, opened 2015-06-23
Prevent LineStrings to be filled

Jenselme:
Lines imported from KML have a filled style which makes them look as polygon once printed. This PR solves this problem.

However, as for #239, it may be a bug in openlayer. @elemoine any opinion?

### DIFF
--- a/src/services/print.js
+++ b/src/services/print.js
@@ -40,6 +40,7 @@ goog.require('goog.math');
 goog.require('goog.object');
 goog.require('ngeo');
 goog.require('ol.format.GeoJSON');
+goog.require('ol.geom.GeometryType');
 goog.require('ol.layer.Image');
 goog.require('ol.layer.Tile');
 goog.require('ol.layer.Vector');
@@ -360,12 +361,19 @@ ngeo.Print.prototype.encodeVectorLayer_ = function(arr, layer, resolution) {
       if (goog.isNull(geojsonFeature.properties)) {
         geojsonFeature.properties = {};
       }
+      var geometry = feature.getGeometry();
+      var geometryType;
+      if (goog.isDefAndNotNull(geometry)) {
+        geometryType = geometry.getType();
+      } else {
+        continue;
+      }
       for (var j = 0, jj = styles.length; j < jj; ++j) {
         var style = styles[j];
         var styleId = goog.getUid(style).toString();
         var featureStyleProp = ngeo.Print.FEAT_STYLE_PROP_PREFIX_ + j;
         this.encodeVectorStyle_(
-            mapfishStyleObject, style, styleId, featureStyleProp);
+            mapfishStyleObject, geometryType, style, styleId, featureStyleProp);
         geojsonFeature.properties[featureStyleProp] = styleId;
       }
     }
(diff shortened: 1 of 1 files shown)


src/services/print.js
```diff
@@ -364,8 +364,15 @@ ngeo.Print.prototype.encodeVectorLayer_ = function(arr, layer, resolution) {
         var style = styles[j];
         var styleId = goog.getUid(style).toString();
         var featureStyleProp = ngeo.Print.FEAT_STYLE_PROP_PREFIX_ + j;
+        var geometryType;
+        if (goog.isDef(geojsonFeature.geometry) &&
+                goog.isDef(geojsonFeature.geometry.type)) {
+          geometryType = geojsonFeature.geometry.type;
+        } else {
+          geometryType = 'FeatureCollection';
+        }
```
elemoine: Can't you just use `feature.getGeometry().getType()`?

src/services/print.js
```diff
@@ -408,7 +416,7 @@ ngeo.Print.prototype.encodeVectorStyle_ =
   var imageStyle = style.getImage();
   var strokeStyle = style.getStroke();
   var textStyle = style.getText();
-  if (!goog.isNull(fillStyle)) {
+  if (!goog.isNull(fillStyle) && geometryType !== 'LineString') {
```
elemoine: What about MultiLineString?

elemoine:
Fixing/working around this problem in the ngeo print service makes sense to me.

src/services/print.js
Jenselme: I forgot them.

src/services/print.js
Jenselme: I can, thanks

Jenselme:
PR updated, commits squashed.

src/services/print.js
```diff
@@ -364,8 +364,9 @@ ngeo.Print.prototype.encodeVectorLayer_ = function(arr, layer, resolution) {
         var style = styles[j];
         var styleId = goog.getUid(style).toString();
         var featureStyleProp = ngeo.Print.FEAT_STYLE_PROP_PREFIX_ + j;
+        var geometryType = feature.getGeometry().getType();
```
elemoine: You just need to call `getGeometry().getType()` once for each feature. And you also want to deal with the case where the feature does not have a geometry (`getGeometry` returns `null` or `undefined`).

src/services/print.js
```diff
@@ -408,7 +410,9 @@ ngeo.Print.prototype.encodeVectorStyle_ =
   var imageStyle = style.getImage();
   var strokeStyle = style.getStroke();
   var textStyle = style.getText();
-  if (!goog.isNull(fillStyle)) {
+  var isLine = geometryType === 'LineString' ||
+          geometryType === 'MultiLineString';
```
elemoine: Please use `ol.geom.GeometryType.LINE_STRING` and `ol.geom.GeometryType.MULTI_LINE_STRING`. And add `goog.include('ol.geom.GeometryType')`.

elemoine:
Sorry, I still have comments.

Jenselme:
> Sorry, I still have comments.

No problem. PR updated.

src/services/print.js
```diff
@@ -360,12 +361,19 @@ ngeo.Print.prototype.encodeVectorLayer_ = function(arr, layer, resolution) {
       if (goog.isNull(geojsonFeature.properties)) {
         geojsonFeature.properties = {};
       }
+      var geometry = feature.getGeometry();
+      var geometryType;
+      if (goog.isDef(geometry) && !goog.isNull(geometry)) {
+        geometryType = geometry.getType();
+      } else {
+        geometryType = '';
+      }
```
elemoine: I think we can just skip the feature if it doesn't have a geometry. And note that you can use `goog.isDefAndNotNull`.

Jenselme: > And note that you can use goog.isDefAndNotNull.

Noted.

> I think we can just skip the feature if it doesn't have a geometry

What do you mean by that?

elemoine: In the loop over the features we can just skip the feature (`continue`) if the feature does not have a geometry.

Jenselme:
PR updated

elemoine:
Still not exactly to my taste :) But I'll merge this and slightly change the code with a separate commit/PR.

Thanks for your contribution. (It may be time to create a contributors file and add your name there.)

Jenselme:
> It may be time to create a contributors file and add your name there.

This up to you I guess

PR #249, opened 2015-06-24
Refactor code in the print service

elemoine:
This follows up on #247. We immediately `continue` the loop when we find a feature with no geometry.

@Jenselme, does this look ok to you?


Jenselme:
It looks better, yeah.

PR #250, opened 2015-07-01
Set stopEvent to true in measure tooltip overlay

elemoine:
This fixes a bug where the user cannot add a point under the measure tooltip.


pgiraud:
LGTM. Please merge.

PR #252, opened 2015-07-01
Make ngeo.Print#encodeLayer public

elemoine:
This commit makes `ngeo.Print#encodeLayer` public. This, for example, makes it possible to applications to print "unmanaged" vector layers.


pgiraud:
LGTM. Please merge.

PR #254, opened 2015-07-02
Remove Angular externs from closure-compiler.js

elemoine:
With https://github.com/google/closure-compiler/pull/1030 and https://github.com/google/closure-compiler/pull/1031 now closed we can remove these extern definitions.


elemoine:
Trivial. Travis happy. Merging.

PR #256, opened 2015-07-02
Strictly listen on container, not children

pgiraud:
When we listen on $animate events, callbacks may be called even for children of the element we listen to. This pull request prevents this by ensuring that the target of the event is actually the element itself.
Please review.


elemoine:
Please merge.

PR #258, opened 2015-07-06
Make getOptimalScale return -1 if no optimal scale

elemoine:
This commit makes `getOptimalScale` return `-1` if no optimal scale can be found.  This occurs when the optimal scale is lower than or equal to the smallest value of the `printMapScales` array. When this happens the user can take the action of her choice (e.g. use the smallest print scale).


pgiraud:
This looks to me. Please merge.

elemoine:
Compilation fails because of https://github.com/camptocamp/ngeo/issues/259.

PR #261, opened 2015-07-06
Select print style type based on geometry type

elemoine:
With this commint the vector style type used in the print spec is selected based on the geometry type, instead of being selected based on the properties found in the `ol.style.Style` object. Relying on the properties found in the `ol.style.Style` object is fragile, as a style object may be used for drawing different types of geometries.

### DIFF
--- a/src/services/print.js
+++ b/src/services/print.js
@@ -426,18 +426,23 @@ ngeo.Print.prototype.encodeVectorLayer_ = function(arr, layer, resolution) {
     }
   }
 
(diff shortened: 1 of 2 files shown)


elemoine:
This was tested in the Luxembourg project, and it works as expected there.

elemoine:
@Jenselme, feel free to review this and check that it works for you… Thanks.

elemoine:
Ok, I'm going to merge this without waiting for a review, as it fixes a critical bug in the Luxembourg geoportailv3 project.

PR #262, opened 2015-07-08
Add externs definitions on angular.JQLite

elemoine:
`angular.JQLite` was de-typedef'ed in closure-compiler. See https://github.com/google/closure-compiler/commit/d785704f64e95e8764db5a9fb02809aaabf42c7e. Because of that we need to add externs definitions to `angular.JQLite`.

### DIFF
--- a/externs/twbootstrap.js
+++ b/externs/twbootstrap.js
@@ -13,10 +13,24 @@ jQuery.prototype.alert = function(opt_action) {};
 
 
 /**
-* @param {(string|Object.<string,*>)=} opt_options
-* @return {jQuery}
-*/
+ * @param {string=} opt_action
+ * @return {!angular.jQLite}
+ */
+angular.JQLite.prototype.alert = function(opt_action) {};
+
+
+/**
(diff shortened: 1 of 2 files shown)


elemoine:
@gberaudo, @fredj, this change should be necessary for the Schweizmobil project as well.

gberaudo:
@elemoine, your PR extends JQLite extern with some bootstrap and typehead extern definitions.
I do not understand why it worked before the de-typedef change?
Was `angular.JQLite` type considered equal to `jQuery`? How?

elemoine:
It worked probably the compiler was less strict, and it didn't care that we were using an unknown function on `angular.JQLite`. Now that `angular.JQLite` is not a `typedef` anymore, but just defined as a regular extern definition, the compiler complains when it detects an unknown method. That's all I can say about this.

elemoine:
I am going to merge this, as it fixes the build of the Luxembourg project. Please tell me if you would have fixed this differently.

gberaudo:
Thank you for the explanations and +1 for merging.

PR #263, opened 2015-07-08
Fix typo in twbootstrap externs file

elemoine:
This one was not caught by the ngeo build because this extern definition is not used in the ngeo code or in one of the examples.


gberaudo:
Please merge when travis is OK.

PR #267, opened 2015-07-13
Always push symboliser for image style if defined

Jenselme:


src/services/print.js
```diff
@@ -561,6 +560,10 @@ ngeo.Print.prototype.encodeVectorStylePoint_ =
       }
     }
   }
+
```
elemoine: Please remove that empty line.

elemoine:
Interesting bug. I'll merge this when the empty line is removed. Thanks!

Jenselme:
Empty line removed, commits squashed.

PR #270, opened 2015-07-14
Make sure baseURL are absolute when printing

Jenselme:


src/services/print.js
```diff
@@ -271,7 +271,7 @@ ngeo.Print.prototype.encodeWmsLayer_ = function(arr, opacity, url, params) {
   goog.object.remove(customParams, 'FORMAT');
 
   var object = /** @type {MapFishPrintWmsLayer} */ ({
-    baseURL: url,
+    baseURL: url.match(/^https?:/) ? url : 'http:' + url,
```
elemoine: This assume that, if `url` doesn't start with `http` or `https` it will start with `//`. That doesn't sound correct to me. What if `url` looks like `path/to/wms` or `/path/to/wms`?

GeoExt's print provider [includes](https://github.com/geoext/geoext/blob/master/lib/GeoExt/data/PrintProviderBase.js#L633-L647) a `getAbsoluteUrl` function that we could copy to ngeo. The function would be simpler in ngeo as we don't need to support IE < 9.

Jenselme:
PR updated: I now use the getAbsuloteUrl from geoExt.

src/services/print.js
```diff
@@ -283,6 +283,17 @@ ngeo.Print.prototype.encodeWmsLayer_ = function(arr, opacity, url, params) {
 
 
 /**
+ * @param {string} url
+ * @return {string} Absolute URL.
+ */
+ngeo.Print.prototype.getAbsoluteUrl_ = function(url) {
```
elemoine: You can make this function static: `ngeo.Print.getAbsolute_`.

elemoine: And do not forget the `@private` annotation.

Jenselme:
PR update:
- `ngeo.Print.getAbsolute_` is static
- add `@private`

src/services/print.js
```diff
@@ -252,14 +252,16 @@ ngeo.Print.prototype.encodeImageWmsLayer_ = function(arr, layer) {
 
   var url = source.getUrl();
   var params = source.getParams();
-  this.encodeWmsLayer_(arr, layer.getOpacity(), url, params);
+  if (goog.isDefAndNotNull(url)) {
```
elemoine: You can just use `if (goog.isDef(url))` here no?

Jenselme: I think so yes. Is it problematic to use `goog.isDefAndNotNull(url)` instead?

elemoine: It's more processing. It's confusing for readers of the code, who may think that `source.getUrl()` may return `null`. I've become strict with typing, and I think it's right :)

src/services/print.js
```diff
@@ -283,6 +285,18 @@ ngeo.Print.prototype.encodeWmsLayer_ = function(arr, opacity, url, params) {
 
 
 /**
+ * @param {string} url
+ * @return {string} Absolute URL.
+ * @private
+ */
+ngeo.Print.getAbsoluteUrl_ = function(url) {
+  var a = document.createElement('a');
+  a.href = decodeURI(url);
```
elemoine: It should be `encodeURI` here, no?

elemoine:
This PR's branch needs to be rebased.

Jenselme:
- Use `encodeURI` where needed instead of `decodeURI`
- More strict about typing
- Rebased agains master.

elemoine:
Thanks.

PR #272, opened 2015-07-27
Set license to MIT in package.json

fredj:


fredj:
trivial, merging

asaunier:
Indeed:
https://github.com/camptocamp/ngeo/blob/master/LICENSE

PR #274, opened 2015-07-27
Initialize layers array to 'undefined'

### LINKED ISSUE #273: backgroundlayer example not working
http://camptocamp.github.io/ngeo/master/backgroundlayer.html

The select element is empty

fredj:
Otherwise the one time binding doesn't work.

fixes #273


elemoine:
Yep, the issue is related to the one-time binding expression used in the [backgroundlayer.html partial](https://github.com/camptocamp/ngeo/blob/master/examples/partials/backgroundlayer.html#L2). Angular's one-time binding code considers `null` as an actual value, so it immediately removes the binding.

Please merge.

PR #281, opened 2015-08-03
Use language:python in .travis.yml

elemoine:
This is an attempt to use `language: python` in the Travis config. Not sure it works and makes sense really.


fredj:
makes sense ...

elemoine:
It seems to work. So I'd say we can go with that.

PR #284, opened 2015-08-05
Add ngeo.gmf.mapDirective

asaunier:
Port of https://github.com/camptocamp/c2cgeoportal/pull/1471

To be discussed:
* ~~the new ngeo.gmf directive/controller are in a dedicated ``directives/gmf/`` to make GMF directives apart (clearer?).~~
* ~~on the other hand the ``gmf-simple.html`` example is at the root of ``examples/``. If files are placed in a ``gmf/`` dir, the page complains that ``Main script not in manager paths: ...``. I also suspect troubles with the hosted examples deployment.~~
* ~~the map expands on the whole viewport, the explanation test being pushed ouside of the window, despite the "#map" style provided in the example page. Any idea?~~

### DIFF
--- a/examples/gmf-simple.html
+++ b/examples/gmf-simple.html
@@ -1,0 +1,24 @@
+<!DOCTYPE>
+<html ng-app='app'>
+  <head>
+    <title>Simple GeoMapFish example</title>
+    <meta charset="utf-8">
+    <meta name="viewport"
+          content="initial-scale=1.0, user-scalable=no, width=device-width">
+    <meta name="apple-mobile-web-app-capable" content="yes">
+    <link rel="stylesheet" href="../node_modules/openlayers/css/ol.css" type="text/css">
+    <style>
+      ngeo-gmf-map > div {
+        width: 600px;
+        height: 400px;
+      }
+    </style>
+  </head>
+  <body ng-controller="MainController as ctrl">
+    <ngeo-gmf-map ngeo-gmf-map-map="::ctrl.map"></ngeo-gmf-map>
+    <p id="desc">This example shows how to use the <code>ngeo-gmf-map</code> directive to insert an OpenLayers map in a GeoMapFish page.</p>
+    <script src="../node_modules/angular/angular.js"></script>
+    <script src="/@?main=gmf-simple.js"></script>
+    <script src="../utils/watchwatchers.js"></script>
+  </body>
+</html>
--- a/src/directives/gmf-map.js
+++ b/src/directives/gmf-map.js
@@ -1,0 +1,74 @@
+/**
+ * @fileoverview This file provides the "map" directive
+ * for GeoMapFish applications.
+ *
+ * Example:
+ *
+ * <ngeo-gmf-map ngeo-gmf-map-map="::mainCtrl.map"></ngeo-gmf-map>
+ */
+goog.provide('ngeo.gmf.mapDirective');
+
+goog.require('goog.asserts');
+goog.require('ngeo.Debounce');
+goog.require('ngeo.Location');
+goog.require('ngeo.mapDirective');
+
+
+/**
+ * @return {angular.Directive} The Directive Definition Object.
+ * @ngInject
+ */
+ngeo.gmf.mapDirective = function() {
+  return {
+    scope: {},
+    bindToController: {
+      'map': '=ngeoGmfMapMap'
+    },
+    controller: 'NgeoGmfMapController',
+    controllerAs: 'ctrl',
+    template: '<div ngeo-map="ctrl.map"></div>'
+  };
+};
+
+ngeoModule.directive('ngeoGmfMap', ngeo.gmf.mapDirective);
+
+
+
+/**
+ * @param {ngeo.Location} ngeoLocation ngeo Location service.
+ * @param {ngeo.Debounce} ngeoDebounce ngeo Debounce service.
+ * @constructor
+ * @ngInject
+ */
+ngeo.gmf.MapController = function(ngeoLocation, ngeoDebounce) {
+
+  /**
+   * @type {!ol.Map}
+   */
+  var map = this['map'];
+  goog.asserts.assert(goog.isDef(map));
+  var view = map.getView();
+
+  var x = ngeoLocation.getParam('map_x');
+  var y = ngeoLocation.getParam('map_y');
+  var zoom = ngeoLocation.getParam('map_zoom');
+
+  if (goog.isDef(x) && goog.isDef(y)) {
+    view.setCenter([+x, +y]);
+  }
+  if (goog.isDef(zoom)) {
+    view.setZoom(+zoom);
+  }
+
+  view.on('propertychange', ngeoDebounce(function() {
+    var center = view.getCenter();
+    var zoom = view.getZoom();
+    ngeoLocation.updateParams({
+      'map_zoom': zoom,
+      'map_x': Math.round(center[0]),
+      'map_y': Math.round(center[1])
+    });
+  }, 300, /* invokeApply */ true));
+};
+
+ngeoModule.controller('NgeoGmfMapController', ngeo.gmf.MapController);
--- a/examples/gmf-simple.js
+++ b/examples/gmf-simple.js
@@ -1,0 +1,42 @@
+goog.provide('gmf-simple');
+
+goog.require('ngeo.gmf.mapDirective');
+goog.require('ol.Map');
+goog.require('ol.View');
+goog.require('ol.layer.Tile');
+goog.require('ol.source.OSM');
+
+
+/** @const **/
+var app = {};
+
+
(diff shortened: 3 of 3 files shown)


src/directives/gmf/map.js
```diff
@@ -0,0 +1,74 @@
...
+    var zoom = view.getZoom();
+    ngeoLocation.updateParams({
+      'map_zoom': zoom,
+      'map_x': Math.round(center[0]),
+      'map_y': Math.round(center[1])
+    });
+  }, 300, /* invokeApply */ true));
+};
+
+ngeoModule.controller('ngeoGmfMapController', ngeo.gmf.MapController);
```
elemoine: Controller names always start with a capital letter, because they reference a constructor. See https://github.com/camptocamp/ngeo/blob/master/src/directives/layertree.js#L140 for example.

examples/gmf-simple.html
```diff
@@ -0,0 +1,24 @@
...
+<html ng-app='app'>
+  <head>
+    <title>Simple GeoMapFish example</title>
+    <meta charset="utf-8">
+    <meta name="viewport"
+          content="initial-scale=1.0, user-scalable=no, width=device-width">
+    <meta name="apple-mobile-web-app-capable" content="yes">
+    <link rel="stylesheet" href="../node_modules/openlayers/css/ol.css" type="text/css">
+    <style>
+      #map {
```
elemoine: You may want to use this instead:

```css
ngeo-gmf-map > div {
  width: 600px;
  height: 400px;
}
```

elemoine:
> To be discussed:
> 
>  * the new ngeo.gmf directive/controller are in a dedicated directives/gmf/ to make GMF directives apart (clearer?).

I would not bother with a separate  directory personally. I'd just name the file `gmf-map.js`.


>  * on the other hand the gmf-simple.html example is at the root of examples/. If files are placed in a gmf/ dir, the page complains that Main script not in manager paths: .... I also suspect troubles with the hosted examples deployment.

If we don't add a `gmf` subdir in the `directives` dir we don't need to have `gmf` subdirs everywhere else.

elemoine:
> the map expands on the whole viewport, the explanation test being pushed ouside of the window, despite the "#map" style provided in the example page. Any idea?

I think it's because the map is actually included in the div that is the child of the div whose id is "map". I suggested a CSS rule that should fix the problem.

asaunier:
Thanks @elemoine for reviewing.
I have applied the changes you have suggested:
* About the ``gmf/`` I agree it's better to be consistent between the directives and examples dirs.
* The CSS trick indeed works.

src/directives/gmf-map.js
```diff
@@ -0,0 +1,74 @@
...
+/**
+ * @return {angular.Directive} The Directive Definition Object.
+ * @ngInject
+ */
+ngeo.gmf.mapDirective = function() {
+  return {
+    scope: {
+      'map': '=ngeoGmfMapMap'
+    },
+    bindToController: true,
```
fgravin: 1.4 seems to prefer to use 

    bindToController : {
      'map': '=ngeoGmfMapMap'
    }
instead.
FYI, the `bindToController` object can also be used in directive with no isolated scope.

fgravin:
Looks like a simple view decorator.
It generates useless DOM.
Also, I wonder if there is no other way to attach a custom controller to an existing directive.

src/directives/gmf-map.js
asaunier: @fgravin Thanks for the tip, but if I do so the page fails loading with error:

``
.
angular.js:12330 AssertionError: Assertion failed
    at new goog.asserts.AssertionError (http://localhost:3000/@/home/alex/Bureau/ngeo/node_modules/openlayers/node_…2b6477e735055b8f53284d5cf6a4af1b29da/closure/goog/asserts/asserts.js:62:20)
    at Object.goog.asserts.doAssertFailure_ (http://localhost:3000/@/home/alex/Bureau/ngeo/node_modules/openlayers/node_…b6477e735055b8f53284d5cf6a4af1b29da/closure/goog/asserts/asserts.js:119:11)
    at Object.goog.asserts.assert [as assert] (http://localhost:3000/@/home/alex/Bureau/ngeo/node_modules/openlayers/node_…b6477e735055b8f53284d5cf6a4af1b29da/closure/goog/asserts/asserts.js:149:18)
    at new ngeo.gmf.MapController (http://localhost:3000/@/home/alex/Bureau/ngeo/src/directives/gmf-map.js:48:16)
    at Object.invoke (http://localhost:3000/node_modules/angular/angular.js:4450:17)
    at $get.extend.instance (http://localhost:3000/node_modules/angular/angular.js:8999:34)
    at nodeLinkFn (http://localhost:3000/node_modules/angular/angular.js:8109:36)
    at compositeLinkFn (http://localhost:3000/node_modules/angular/angular.js:7541:13)
    at nodeLinkFn (http://localhost:3000/node_modules/angular/angular.js:8145:24)
    at compositeLinkFn (http://localhost:3000/node_modules/angular/angular.js:7541:13)
``

It seems that ``this['map']`` is then not available at
https://github.com/camptocamp/ngeo/blob/gmf-map/src/directives/gmf-map.js#L48
What's the correct way to access the map property in that case?

elemoine: Do you have `scope: {}`?

asaunier: Nope. It works better with it indeed. :)

fgravin: > Do you have scope: {}?

Yes, bindToController works on new scope. You either need `scope: true` or `scope: {}` to make it work.
In your case `scope: {}`.
Note that you can have both :

    scope: {toto:'='},
    bindToController: {tata: '='}

elemoine:
> Looks like a simple view decorator.

So what?

Our plan is too start adding higher-level, GMF-specific components. And `ngeo-gmf-map` would be the first component of that sort.

To be honest, I also feel that adding GMF-specific components pollutes ngeo a bit. Yet, I wouldn't like and support the idea of creating and maintaining a separate library for GMF.

> It generates useless DOM.

True. I'd also like to think about a way to wrap a directive (`ngeo-map` in that case), without adding useless DOM elements.

> Also, I wonder if there is no other way to attach a custom controller to an existing directive.

And also, in that case we probably don't need to use an isolate scope. I feel that it makes things unnecessarily complex. For example it forces the user of the directive to use a one-time binding expression for the map to prevent Angular from adding an unnecessary watcher.

I think I'd still merge this PR, but I'd also like to think a bit more about the aforementioned issues.

src/directives/gmf-map.js
asaunier: scope/bindToController updated and commits rebased

elemoine: @fgravin, if you have `bindToController: {tata: '='}` I don't see when you'd use `scope: true` rather than `scope: {}`.

src/directives/gmf-map.js
```diff
@@ -0,0 +1,74 @@
...
+
+
+/**
+ * @return {angular.Directive} The Directive Definition Object.
+ * @ngInject
+ */
+ngeo.gmf.mapDirective = function() {
+  return {
+    scope: {},
+    bindToController: {
```
fgravin: @elemoine maybe to put parent scope properties in your directive controller ? You can still want a `scope:true` but use the binding for you controller through dom attributes. I don't really know what could be the usecase.

elemoine:
> I'd also like to think a bit more about the aforementioned issues.

One thing we could do is add a controller to the `ngeo-map` directive. The `ngeo-map` directive would add a reference to the map to the controller.

The DDO of `ngeo-gmf-map` directive would then have `require: 'ngeoMap'`, so its `link` function would receive the controller of `ngeo-map` as a fourth parameter.

```js
ngeo.gmf.mapDirective = function() {
  return {
    require: 'ngeoMap',
    link: function(scope, element, attrs, ctrl) {
      var map = crtl.map;
      map.getView().on('propertychange', function() {
        // …
      });
    };
};
```

And this is how you'd use `ngeo-map` and `ngeo-gmf-map` together:

```html
<div ngeo-map ngeo-map-map="mainCtrl.map" ngeo-gmf-map>
```

Not saying I like this better, but I thought I could just throw the idea to the table…

elemoine:
Another alternative would be to just add a service that takes care of the map-related permalink parameters. The application would just call that service from its main controller.

My opinion is that the `ngeo-gmf-map` directive as developed by @asaunier provides for a better and easier API to users. It is true that the directive adds markup to the page, but that may not be that bad.

In the future, the `ngeo-gmf-map` directive (its controller really) may also create the `ol.Map` instance based on parameters set by the user.

I am +1 on merging this.

asaunier:
OK, thanks all for argumenting. I will merge soon if no one protest.

elemoine:
Thanks @asaunier.

@fgravin, still interested in your comments if you have any.

PR #285, opened 2015-08-07
Send layer opacity to MapFish Print

elemoine:
Currently, the opacity of WMTS and Vector layers are not set in the print specs sent to MapFish Print. This PR fixes that.

Please review.


fredj:
LGTM

elemoine:
Thanks for the review @fredj.

PR #288, opened 2015-08-07
Add a contribs/gmf directory

elemoine:
I'm opening this PR for discussion.

The PR adds a `contribs/gmf` directory to be used for GeoMapFish-related components and examples. In this way we make it clear to external users that ngeo is not tied to GeoMapFish.

The infrastructure is not yet complete, but I'd like to know whether people agree with this way of structuring things before continuing on.


fgravin:
I share the same ideas.
+1

tonio:
It’s much clear like this. :+1:

fredj:
+1

elemoine:
Things to do:

* [ ] Check GMF examples with PhantomJS
* [ ] Add GMF examples to GH pages

asaunier:
Makes sense indeed. But shouldn't the README advertise these new contribs examples as it does for the standard examples?

elemoine:
> Makes sense indeed. But shouldn't the README advertise these new contribs examples as it does for the standard examples?

Yes, indeed.

elemoine:
> Makes sense indeed. But shouldn't the README advertise these new contribs examples as it does for the standard examples?

I decided not to do that at this point, given that we currently just have the gmf simple example. We can advertise the gmf examples when we have more examples to show.

That being said the gmf examples will be available at https://camptocamp.github.io/ngeo/master/gmf.

elemoine:
I am done. Please review.

elemoine:
I changed paths, and the gmf examples should now be available at https://camptocamp.github.io/ngeo/master/contribs/gmf (when this is merged).

elemoine:
I am done again. I'll merge this, unless there are other comments.

asaunier:
For me it's OK.

PR #290, opened 2015-08-12
Remove binding in gmf-map directive

elemoine:
This removes the binding used for the gmf-map directive's "map" property. With this change the directive user no longer needs to use a one-time binding expression to prevent Angular from creating an unnecessary watcher.

### DIFF
--- a/contribs/gmf/src/directives/map.js
+++ b/contribs/gmf/src/directives/map.js
@@ -4,7 +4,7 @@
  *
  * Example:
  *
- * <gmf-map gmf-map-map="::mainCtrl.map"></gmf-map>
+ * <gmf-map gmf-map-map="mainCtrl.map"></gmf-map>
  */
 goog.provide('gmf.mapDirective');
 
@@ -13,6 +13,7 @@ goog.require('goog.asserts');
 goog.require('ngeo.Debounce');
 goog.require('ngeo.Location');
 goog.require('ngeo.mapDirective');
+goog.require('ol.Map');
 
 
 /**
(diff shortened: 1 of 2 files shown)


contribs/gmf/src/directives/map.js
```diff
@@ -21,9 +22,8 @@ goog.require('ngeo.mapDirective');
  */
 gmf.mapDirective = function() {
   return {
-    scope: {},
-    bindToController: {
-      'map': '=gmfMapMap'
+    scope: {
+      'getMapFn': '&gmfMapMap'
     },
```
asaunier: Just curious why you eventually used scope instead of bindToController (by the way don't we need a ``bindToController: true`` line?)?

elemoine: > Just curious why you have eventually used scope instead of bindToController?

Because I did not want to pollute the directive's controller with a property (`getMapFn`) that I don't use in the directive's template.

> (by the way don't we need a bindToController: true line?)

I'd pollute the controller with `getMapFn` if I used `bindToController: true`.

asaunier:
OK looks good to me.
Curious again: is it such a problem to use the one-time binding operator here or is this a way to make sure that the watcher will never be used even though the user forgets the :: ?

elemoine:
> Curious again: is it such a problem to use the one-time binding operator here or is this a way to make sure that the watcher will never be added even though the user forgets the :: ?

It is indeed a way to ensure that user won't create an unnecessary watcher. It is a bit silly to rely on the user using a one-time binding expression here.

PR #292, opened 2015-08-14
Merge layertree and layertreenode directives

elemoine:
This PR merges the `ngeoLayertree` and `ngeoLayertreenode` directives into one single directive, `ngeoLayertree`.

The main benefit of this merging is that we no longer have two partials with similar markup (`<ul><li>`). There now is one partial file, used both for the root node and non-root nodes.

This also makes things easier to applications that need to provide custom markup for the tree, as there's only one partial to write.

At last, I want to provide a new way for the application to specify a custom template (through a `ngeo-layertree-templateurl` attribute), and the merging of the two directives will make that much easier.


elemoine:
No need to review this for now. I want to test it in the Luxembourg project first.

elemoine:
I tested it in the Luxembourg project, and it works as expected. Please review.

elemoine:
> At last, I want to provide a new way for the application to specify a custom template (through a ngeo-layertree-templateurl attribute), and the merging of the two directives will make that much easier.

I finally decided to do that as part of this PR.

asaunier:
Thanks! No objection from my perspective.

Issue #293, opened 2015-08-14
The gmf examples are not on github.io


elemoine:
It is expected that the gmf examples are published on github.io at http://camptocamp.github.io/ngeo/master/contribs/gmf. But this URL gives a 404.

elemoine:
http://camptocamp.github.io/ngeo/master/contribs/gmf/ now returns the index page.

There is a new problem: the examples do not work on github.io anymore, because `ngeo.js`, `ngeo.css` and `gmf.js` are missing. New PR coming up for this.

elemoine:
> There is a new problem: the examples do not work on github.io anymore, because ngeo.js, ngeo.css and gmf.js are missing. New PR coming up for this.

Fixed with https://github.com/camptocamp/ngeo/pull/295.

PR #294, opened 2015-08-14
Fix gh-pages-from-travis Makefile target

elemoine:
12558b9 introduced a regression where `$(GIT_BRANCH)` is used instead of `master` in the `gh-pages-from-travis` target.


elemoine:
Merging this, hoping it'll fix https://github.com/camptocamp/ngeo/issues/293.

PR #296, opened 2015-08-14
Fix the gmf.js build

elemoine:
The gmf simple example currently does not work on github.io. See http://camptocamp.github.io/ngeo/master/contribs/gmf/simple.html. This is because the `gmf.js` build does not include the `gmfMap` directive because of a mis-configuration in `gmf.json`. This PR fixes it.


asaunier:
Looks good!

elemoine:
Thanks @asaunier. I hope that this will definitely fix the examples on github.io.

PR #291, opened 2015-08-12
[WIP] Add a layer tree component to contribs/gmf

elemoine:
This is very very early stage right now. I am just opening this PR for people who want to track progress on this development.

PR #283, opened 2015-08-04
Update typeahead  to v0.11.1

fredj:
This change is not backward compatible

Fixes #282.

PR #297, opened 2015-08-17
Bases for GeoMapFish FTS

ger-benjamin:
First draw of the GeoMapFish2's FTS. It is already able to group results and this, by accepting multiple sources (multiple urls with or without filters).
This example works and is in an "understandable state" but it's not really usable "as is".

Two important notes:
 - I've added a "filter" parameter in the src/service/creategeojsonbloodhound.js
 - I've used "geomapfish-demo.camptocamp.net/1.6" and, most important,"map.cartoriviera.ch"  as datasources for my queries. Is that acceptable ?

### DIFF
--- a/contribs/gmf/examples/search.html
+++ b/contribs/gmf/examples/search.html
@@ -1,0 +1,103 @@
+<!DOCTYPE>
+<html ng-app='app'>
+  <head>
+    <title>Search GeoMapFish example</title>
+    <meta charset="utf-8">
+    <meta name="viewport"
+          content="initial-scale=1.0, user-scalable=no, width=device-width">
+    <meta name="apple-mobile-web-app-capable" content="yes">
+    <link rel="stylesheet" href="../../../node_modules/openlayers/css/ol.css" type="text/css">
+    <link rel="stylesheet" href="../../../node_modules/bootstrap/dist/css/bootstrap.css" type="text/css">
+    <style>
+      gmf-map > div {
+        width: 600px;
+        height: 400px;
+      }
+      /* CSS stolen from https://github.com/bassjobsen/typeahead.js-bootstrap-css/ */
+      span.twitter-typeahead .tt-menu {
+        position: absolute;
+        top: 100%;
+        left: 0;
+        z-index: 1000;
+        display: none;
+        float: left;
+        min-width: 160px;
+        padding: 5px 0;
+        margin: 2px 0 0;
+        list-style: none;
+        font-size: 14px;
+        text-align: left;
+        background-color: #ffffff;
+        border: 1px solid #cccccc;
+        border: 1px solid rgba(0, 0, 0, 0.15);
+        border-radius: 4px;
+        -webkit-box-shadow: 0 6px 12px rgba(0, 0, 0, 0.175);
+        box-shadow: 0 6px 12px rgba(0, 0, 0, 0.175);
+        background-clip: padding-box;
+      }
+      span.twitter-typeahead .header {
+       background-color: #CCDDEE;
+       padding: 4px 0;
+      }
+      span.twitter-typeahead .tt-suggestion {
+        display: block;
+        padding: 3px 20px;
+        margin-bottom: 0px;
+        clear: both;
+        font-weight: normal;
+        line-height: 1.42857143;
+        color: #333333;
+        white-space: nowrap;
+      }
+      span.twitter-typeahead .tt-suggestion:hover,
+      span.twitter-typeahead .tt-suggestion:focus {
+        color: #ffffff;
+        text-decoration: none;
+        outline: 0;
+        background-color: #428bca;
+      }
+      span.twitter-typeahead .tt-suggestion.tt-cursor {
+        color: #ffffff;
+        background-color: #428bca;
+      }
+      span.twitter-typeahead .tt-suggestion button {
+        background: none;
+        border: 0;
+        float: right;
+      }
+      span.twitter-typeahead {
+        width: 100%;
+      }
+      .input-group span.twitter-typeahead {
+        display: block !important;
+      }
+      .input-group span.twitter-typeahead .tt-menu {
+        top: 32px !important;
+      }
+      .input-group.input-group-lg span.twitter-typeahead .tt-menu {
+        top: 44px !important;
+      }
+      .input-group.input-group-sm span.twitter-typeahead .tt-menu {
+        top: 28px !important;
+      }
+    </style>
+  </head>
+  <body ng-controller="MainController as ctrl">
+    <div class="container-fluid">
+      <gmf-map gmf-map-map="ctrl.map"></gmf-map>
+      <p>Search for « Laus » for example…<p>
+      <gmf-search 
+        gmf-search-map="ctrl.map"
+        gmf-search-datasources="ctrl.searchDatasources">
+      </gmf-search>
+      <p id="desc">This example shows how to use the <code>gmf-search</code> directive, which
+      is based on Twitter's <code>typeahead</code> component.</p>
+    </div>
+    <script src="../../../node_modules/jquery/dist/jquery.js"></script>
+    <script src="../../../node_modules/angular/angular.js"></script>
+    <script src="../../../node_modules/typeahead.js/dist/typeahead.bundle.js"></script>
+    <script src="http://cdnjs.cloudflare.com/ajax/libs/proj4js/2.2.1/proj4.js" type="text/javascript"></script>
+    <script src="/@?main=search.js"></script>
+    <script src="../../../utils/watchwatchers.js"></script>
+  </body>
+</html>
--- a/contribs/gmf/examples/search.js
+++ b/contribs/gmf/examples/search.js
@@ -1,0 +1,63 @@
+goog.provide('gmf-search');
+
+goog.require('gmf.mapDirective');
+goog.require('gmf.searchDirective');
+goog.require('ol.Map');
+goog.require('ol.View');
+goog.require('ol.layer.Tile');
+goog.require('ol.source.OSM');
+
+
+/** @const **/
+var app = {};
+
+
+/** @type {!angular.Module} **/
+app.module = angular.module('app', ['gmf']);
+
+
+
+/**
+ * @constructor
+ */
+app.MainController = function() {
+
+  proj4.defs('EPSG:21781',
+      '+proj=somerc +lat_0=46.95240555555556 +lon_0=7.439583333333333 +k_0=1 ' +
+      '+x_0=600000 +y_0=200000 +ellps=bessel ' +
+      '+towgs84=674.374,15.056,405.346,0,0,0,0 +units=m +no_defs');
+
+  /**
+   * @type {Array.<gmfx.SearchDirectiveDatasource>}
+   * @export
+   */
+  this.searchDatasources = [{
+    title: 'From demo 1.6',
+    url: 'http://geomapfish-demo.camptocamp.net/1.6/wsgi/fulltextsearch',
+    projection: 'EPSG:21781'
+  }, {
+    title: 'From Cartoriviera - ',
+    url: 'http://map.cartoriviera.ch/main/wsgi/fulltextsearch',
+    projection: 'EPSG:21781',
+    groupByProperties: ['04rue', '03adresse']
+  }];
+
+  /**
+   * @type {ol.Map}
+   * @export
+   */
+  this.map = new ol.Map({
+    layers: [
+      new ol.layer.Tile({
+        source: new ol.source.OSM()
+      })
+    ],
+    view: new ol.View({
+      center: [0, 0],
+      zoom: 4
+    })
+  });
+};
+
+
+app.module.controller('MainController', app.MainController);
--- a/contribs/gmf/src/directives/search.js
+++ b/contribs/gmf/src/directives/search.js
@@ -1,0 +1,280 @@
+/**
+ * @fileoverview this file provides the "search" directive and controller for
+ * a GeoMapFish application.
+ *
+ * FIXME: The 'placeholder' in the input field is hard-coded and can't currently
+ * be internationalised.
+ *
+ * Example:
+ *
+ * <gmf-search gmf-search-map="ctrl.map"
+ *             gmf-search-datasources="ctrl.searchDatasources">
+ * </gmf-search>
+ */
+goog.provide('gmf.searchDirective');
+
+
+goog.require('gmf');
+goog.require('gmf.mapDirective');
+goog.require('ngeo.CreateGeoJSONBloodhound');
+goog.require('ngeo.searchDirective');
+goog.require('ol.Map');
+goog.require('ol.View');
+goog.require('ol.layer.Tile');
+goog.require('ol.layer.Vector');
+goog.require('ol.proj');
+goog.require('ol.source.OSM');
+goog.require('ol.source.Vector');
+
+
+/**
+ * @return {angular.Directive} The Directive Definition Object.
+ * @ngInject
+ */
+gmf.searchDirective = function() {
+  return {
+    restrict: 'E',
+    scope: {
+      'getMapFn': '&gmfSearchMap',
+      'getDatasourcesFn': '&gmfSearchDatasources'
+    },
+    controller: 'GmfSearchController',
+    controllerAs: 'ctrl',
+    template:
+        '<input type="text" placeholder="search…" ' +
+        'ngeo-search="ctrl.options" ' +
+        'ngeo-search-datasets="ctrl.datasets" ' +
+        'ngeo-search-listeners="ctrl.listeners">',
+    link:
+        /**
+         * @param {angular.Scope} scope Scope.
+         * @param {angular.JQLite} element Element.
+         * @param {angular.Attributes} attrs Atttributes.
+         */
+        function(scope, element, attrs) {
+          // Empty the search field on focus and blur.
+          element.find('input').on('focus blur', function() {
+            $(this).val('');
+          });
+        }
+  };
+};
+
+gmfModule.directive('gmfSearch', gmf.searchDirective);
+
+
+
+/**
+ * @constructor
+ * @param {angular.Scope} $scope The directive's scope.
+ * @param {angular.$compile} $compile Angular compile service.
+ * @param {ngeo.CreateGeoJSONBloodhound} ngeoCreateGeoJSONBloodhound The ngeo
+ * create GeoJSON Bloodhound service.
+ * @ngInject
+ * @export
+ */
+gmf.SearchController = function($scope, $compile,
+    ngeoCreateGeoJSONBloodhound) {
+
+  /**
+   * @type {angular.Scope}
+   * @private
+   */
+  this.scope_ = $scope;
+
+  /**
+   * @type {angular.$compile}
+   * @private
+   */
+  this.compile_ = $compile;
+
+  /**
+   * @type {ngeo.CreateGeoJSONBloodhound}
+   * @private
+   */
+  this.ngeoCreateGeoJSONBloodhound_ = ngeoCreateGeoJSONBloodhound;
+
+  var map = this.scope_['getMapFn']();
+  goog.asserts.assertInstanceof(map, ol.Map);
+
+  /**
+   * @type {!ol.Map}
+   * @private
+   */
+  this.map_ = map;
+
+  /**
+   * @type {ol.layer.Vector}
+   * @private
+   */
+  this.vectorLayer_ = this.createVectorLayer_();
+
+  var datasources = this.scope_['getDatasourcesFn']();
+  goog.asserts.assertArray(datasources);
+
+  /**
+   * @type {Array.<gmfx.SearchDirectiveDatasource>}
+   * @private
+   */
+  this.datasources_ = datasources;
+
+  /**
+   * @type {TypeaheadOptions}
+   * @export
+   */
+  this.options = /** @type {TypeaheadOptions} */ ({
+    highlight: true,
+    hint: undefined,
+    minLength: undefined
+  });
+
+  /**
+   * @type {Array.<TypeaheadDataset>}
+   * @export
+   */
+  this.datasets = [];
+
+  for (var i = 0; i < this.datasources_.length; i++) {
+    var datasource = this.datasources_[i];
+    var groupByProperties = datasource.groupByProperties || [];
+    var filter;
+
+    do {
+      var title = datasource.title;
+
+      if (groupByProperties[0]) {
+        //Add an optional filter function to keep objects only from one
+        //"layername" from a GMF's fulltextsearch service.
+        filter = this.filterLayername_(groupByProperties[0]);
+        title = title + groupByProperties[0];
+        groupByProperties.shift();
+      } else {
+        filter = undefined;
+      }
+
+      this.datasets.push(this.createDataset_({
+        'title': title,
+        'filter': filter,
+        'projection': datasource.projection,
+        'url': datasource.url
+      }));
+
+    } while (groupByProperties.length > 0);
+  }
+
+
+
+  /**
+   * @type {ngeox.SearchDirectiveListeners}
+   * @export
+   */
+  this.listeners = /** @type {ngeox.SearchDirectiveListeners} */ ({
+    selected: goog.bind(gmf.SearchController.selected_, this)
+  });
+
+};
+
+
+/**
+ * @return {ol.layer.Vector} The vector layer.
+ * @private
+ */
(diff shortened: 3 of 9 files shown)


elemoine:
https://github.com/camptocamp/ngeo/pull/283 is closed, you can rebase this.

ger-benjamin:
@elemoine Done, thanks

ger-benjamin:
@elemoine I've removed my own import of the commit of fredj. Thanks

elemoine:
@ger-benjamin is this ready for review?

ger-benjamin:
@elemoine Yep

contribs/gmf/examples/search.js
```diff
@@ -0,0 +1,48 @@
...
+
+/**
+ * @constructor
+ */
+app.MainController = function() {
+
+  proj4.defs('EPSG:21781',
+      '+proj=somerc +lat_0=46.95240555555556 +lon_0=7.439583333333333 +k_0=1 ' +
+      '+x_0=600000 +y_0=200000 +ellps=bessel ' +
+      '+towgs84=674.374,15.056,405.346,0,0,0,0 +units=m +no_defs');
```
elemoine: Do you need proj4?

contribs/gmf/src/directives/search.js
```diff
@@ -0,0 +1,281 @@
...
+ * @ngInject
+ */
+gmf.searchDirective = function() {
+  return {
+    restrict: 'E',
+    scope: {
+      'getMapFn': '&gmfSearchMap'
+    },
+    controller: 'GmfSearchController',
+    bindToController: true,
```
elemoine: I wouldn't use `bindToController: true` in this case, because you don't need to have `getMapFn` in the controller.

contribs/gmf/src/directives/search.js
```diff
@@ -0,0 +1,281 @@
...
+ * @ngInject
+ */
+gmf.SearchController = function($rootScope, $compile,
+    ngeoCreateGeoJSONBloodhound) {
+
+  /**
+   * @type {angular.Scope}
+   * @private
+   */
+  this.rootScope_ = $rootScope;
```
elemoine: I don't think you need the `rootScope`. You can just use the directive's scope to create a new (isolate) scope in `createDataset_`.

contribs/gmf/src/directives/search.js
```diff
@@ -0,0 +1,281 @@
...
+  this.map = map;
+
+  /**
+   * @type {ol.layer.Vector}
+   * @private
+   */
+  this.vectorLayer_ = this.createVectorLayer_();
+
+  /**
+   * TODO must be definable by an integrator
```
elemoine: Yes, I'd suggest that you change this before merging. I also had to do that in https://github.com/camptocamp/ngeo/pull/291.

contribs/gmf/src/directives/search.js
```diff
@@ -0,0 +1,281 @@
...
+   * @type {Array<string>}
```
elemoine: It is not an array of strings.

contribs/gmf/src/directives/search.js
```diff
@@ -0,0 +1,281 @@
...
+        groupByProperties.shift();
+      } else {
+        filter = null;
+      }
+
+      this.datasets.push(this.createDataset_(datasource['url'],
+                                             filter,
+                                             dataTitle));
+    }
+    while (groupByProperties.length > 0);
```
elemoine: Please place the closing curly brace and the `while` on the same line:

```js
    } while (groupByProperties.length > 0);
```

contribs/gmf/src/directives/search.js
elemoine: And we currently use a `.` after `Array` everywhere else.

contribs/gmf/src/directives/search.js
```diff
@@ -0,0 +1,281 @@
...
+  }
+
+
+
+  /**
+   * @type {ngeox.SearchDirectiveListeners}
+   * @export
+   */
+  this.listeners = /** @type {ngeox.SearchDirectiveListeners} */ ({
+    selected: angular.bind(this, gmf.SearchController.selected_)
```
elemoine: Please use goog functions (`goog.bind` here) outside examples.

contribs/gmf/src/directives/search.js
```diff
@@ -0,0 +1,281 @@
...
+};
+
+
+/**
+ * @return {ol.layer.Vector} The vector layer.
+ * @private
+ */
+gmf.SearchController.prototype.createVectorLayer_ = function() {
+  var vectorLayer = new ol.layer.Vector({
+    source: new ol.source.Vector()
```
elemoine: See https://github.com/openlayers/ol3/blob/master/src/ol/interaction/drawinteraction.js#L276-L279 for options used for unmanaged layers.

contribs/gmf/src/directives/search.js
```diff
@@ -0,0 +1,281 @@
...
+
+/**
+ * @param {string} url A ws url that return geojson
+ * @param {Function?} filter A filter based on a featuresCollection's array
+ * @param {string} dataTitle the tile of this dataset's results
+ * @return {TypeaheadDataset} a typeahead dataset.
+ * @private
+ */
+gmf.SearchController.prototype.createDataset_ = function(url, filter,
+                                                             dataTitle) {
```
elemoine: Please use 4 spaces for the identation.

contribs/gmf/src/directives/search.js
```diff
@@ -0,0 +1,281 @@
...
+  var rootScope = this.rootScope_;
+  var compile = this.compile_;
+  var bloodhoundEngine = this.createAndInitBloodhound_(url, filter);
+  return /** @type {TypeaheadDataset} */ ({
+    source: bloodhoundEngine.ttAdapter(),
+    display: function(suggestion) {
+      var feature = /** @type {ol.Feature} */ (suggestion);
+      return feature.get('label');
+    },
+    templates: {
```
elemoine: You need to cast the object here, to make sure Closure recognizes it as a `TypeaheadTemplates`.

contribs/gmf/src/directives/search.js
```diff
@@ -0,0 +1,281 @@
...
+        return '<div class="header">' + dataTitle + '</div>';
+      },
+      suggestion: function(suggestion) {
+        var feature = /** @type {ol.Feature} */ (suggestion);
+
+        // A scope for the ng-click on the suggestion's « i » button.
+        var scope = rootScope.$new(true);
+        scope['feature'] = feature;
+        scope['click'] = function(event) {
+          window.alert(feature.get('label'));
```
elemoine: We do not want to leave an alert in a directive.

contribs/gmf/src/directives/search.js
```diff
@@ -0,0 +1,281 @@
...
+};
+
+
+/**
+ * @param {string} layername the dataset "type" to keep
+ * @return {Function} A filter function based on a featuresCollection's array
+ * @private
+ */
+gmf.SearchController.prototype.filterLayername_ = function(layername) {
+  return function(feature) {
```
elemoine: Please provide a signature for this function.

contribs/gmf/src/directives/search.js
```diff
@@ -0,0 +1,281 @@
...
+        return compile(html)(scope);
+      }
+    }
+  });
+};
+
+
+/**
+ * @param {string} layername the dataset "type" to keep
+ * @return {Function} A filter function based on a featuresCollection's array
```
elemoine: The return type should be more specific. And note that `Function` should be `function`.

contribs/gmf/src/directives/search.js
```diff
@@ -0,0 +1,281 @@
...
+ * @private
+ */
+gmf.SearchController.prototype.filterLayername_ = function(layername) {
+  return function(feature) {
+    return (feature['properties']['layer_name'] === layername);
```
elemoine: `feature.getProperties`?

elemoine: or `feature.get('layer_name')`?

contribs/gmf/src/directives/search.js
```diff
@@ -0,0 +1,281 @@
...
+ * @param {Function?} filter A filter based on a featuresCollection's array
+ * results
+ * @return {Bloodhound} The bloodhound engine.
+ * @private
+ */
+gmf.SearchController.prototype.createAndInitBloodhound_ = function(url,
+        filter) {
+  url = url + '?query=%QUERY&';
+  var bloodhound = this.ngeoCreateGeoJSONBloodhound_(url, filter,
+      ol.proj.get('EPSG:3857'), ol.proj.get('EPSG:21781'));
```
elemoine: We should not hard-code the projections.

src/services/creategeojsonbloodhound.js
```diff
@@ -58,7 +60,13 @@ ngeo.createGeoJSONBloodhound = function(options, opt_featureProjection,
       },
       filter: function(parsedResponse) {
         var featureCollection = /** @type {GeoJSONFeatureCollection} */
-            (parsedResponse);
+            (goog.object.clone(parsedResponse));
+
+        if (opt_filter) {
```
elemoine: Please use `goog.isDef(opt_filter)`.

elemoine:
First review round done. Thanks for the work @ger-benjamin.

ger-benjamin:
Tanks you @elemoine I'll clean that soon (probably monday)
Some things are just copy-pasted from the "ngeo search's example" (not the gmf-search). Do you want that I clean the ngeo search example ? (in another pull request, and in a few days)

contribs/gmf/examples/search.js
ger-benjamin: Yes, because I need to reproject the search's results from EPSG:21781 to EPSG:3857 (osm)

elemoine: Ok. It's probably obvious but we should not use specific projections in the search directive.

contribs/gmf/src/directives/search.js
ger-benjamin: feature is GeoJSONFeature. Not ol.Feature...

contribs/gmf/examples/search.js
ger-benjamin: I've removed the hard-coded call to this projection in the directive but not this declaration in the example file.

ger-benjamin:
Except two remarks, I've finished the first review round.
@elemoine  Can you check again ?

elemoine:
Will do…

contribs/gmf/src/directives/search.js
elemoine: Adding types to the returned function would help understand this.

contribs/gmf/examples/search.html
```diff
@@ -0,0 +1,112 @@
...
+        gmf-search-datasources='[{
+            dataTitle: "From demo 1.6",
+            url: "http://geomapfish-demo.camptocamp.net/1.6/wsgi/fulltextsearch",
+            projection: "EPSG:21781"
+          }, {
+            dataTitle: "From Cartoriviera - ",
+            url: "http://map.cartoriviera.ch/main/wsgi/fulltextsearch",
+            projection: "EPSG:21781",
+            groupByProperties: ["04rue", "03adresse"]
+          }]'>
```
elemoine: I don't think this is good practice to define such config objects in the HTML. Instead, this config object should be defined on the main controller's scope.

contribs/gmf/src/directives/search.js
```diff
@@ -0,0 +1,277 @@
+/**
+ * @fileoverview this file provides the "search" directive and controller for
+ * a GeoMapFish application.
+ *
+ * Example:
+ *
+ * <gmf-search gmf-search-map="ctrl.map"></gmf-search>
```
elemoine: `gmf-search-datasources` is missing.

contribs/gmf/src/directives/search.js
```diff
@@ -0,0 +1,277 @@
...
+      header: function() {
+        return '<div class="header">' + config['dataTitle'] + '</div>';
+      },
+      suggestion: function(suggestion) {
+        var feature = /** @type {ol.Feature} */ (suggestion);
+
+        var scope = directive_scope.$new(true);
+        scope['feature'] = feature;
+        scope['click'] = function(event) {
+          window.alert(feature.get('label'));
```
ger-benjamin: I've not removed that. I'm not sur if I must remove all the "button" part of this "suggestion" function or if I must use a "cleaner" fonction to make an alert. I don't know if we already have something to create an alert..?

contribs/gmf/src/directives/search.js
```diff
@@ -0,0 +1,277 @@
...
+  return {
+    restrict: 'E',
+    scope: {
+      'getMapFn': '&gmfSearchMap',
+      'getDatasourcesFn': '&gmfSearchDatasources'
+    },
+    controller: 'GmfSearchController',
+    controllerAs: 'ctrl',
+    template:
+        '<input type="text" placeholder="search…" ' +
```
elemoine: We have a translation issue here. And I'm not sure how to address that issue at this point. Maybe we'll add angular-gettext to contribs/gmf. In the mean time, we should just mention to the problem with a FIXME in the `@fileoverview` comment block.

contribs/gmf/src/directives/search.js
```diff
@@ -0,0 +1,277 @@
...
+
+
+
+/**
+ * @constructor
+ * @param {angular.Scope} $scope The directive's scope.
+ * @param {angular.$compile} $compile Angular compile service.
+ * @param {ngeo.CreateGeoJSONBloodhound} ngeoCreateGeoJSONBloodhound The ngeo
+ * create GeoJSON Bloodhound service.
+ * @ngInject
```
elemoine: You need an `@export` annotation here. See https://github.com/camptocamp/ngeo/blob/master/docs/guidelines.md#use-export-for-controllers if you want to understand why.

contribs/gmf/src/directives/search.js
```diff
@@ -0,0 +1,277 @@
...
+   * @type {ol.layer.Vector}
+   * @private
+   */
+  this.vectorLayer_ = this.createVectorLayer_();
+
+  var datasources = this.scope_['getDatasourcesFn']();
+  goog.asserts.assertArray(datasources);
+
+  /**
+   * @type {Array.<Object.<string,*>>}
```
elemoine: We should define an `ngeo.SearchDatasource` type (with a `@typedef`). Something like this:

```js
/**
 * @typedef {{dataTitle: string, groupByProperties: (Array.<string>|undefined)}}
 */
ngeo.SearchDatasource;
```

And this should actually go in an `gmfx.js` externs file (as `ngeox.js`) to prevent renaming for the gmf examples.

I can work on that later, if you don't feel comfortable doing it.

contribs/gmf/src/directives/search.js
```diff
@@ -0,0 +1,277 @@
...
+    highlight: true,
+    hint: undefined,
+    minLength: undefined
+  });
+
+  /**
+   * @type {Array.<TypeaheadDataset>}
+   * @export
+   */
+  this.datasets = [];
```
elemoine: I'd recommend adding an empty line after this line.

contribs/gmf/src/directives/search.js
```diff
@@ -0,0 +1,277 @@
...
+      }
+
+      this.datasets.push(this.createDataset_({
+        'dataTitle': dataTitle,
+        'filter': filter,
+        'projection': datasource['projection'],
+        'url': datasource['url']
+      }));
+
+    } while (groupByProperties.length > 0);
```
elemoine: You use a `do while` loop. But shouldn't you use a `while do` loop instead? Indeed, what about the case where the `groupByProperties` array is empty? Also, why not just use a `for` loop here? With a `for` loop you wouldn't need to mutate the user data.

contribs/gmf/src/directives/search.js
elemoine: Correction: the type would be `gmfx.SearchDirectiveDatasource`.

contribs/gmf/src/directives/search.js
elemoine: We should create a `gmf-search.js` externs file, with types for the objects returned in the response from the full text search. We can do that later.

contribs/gmf/src/directives/search.js
```diff
@@ -0,0 +1,277 @@
...
+  });
+  // Use vectorLayer.setMap(map) rather than map.addLayer(vectorLayer). This
+  // makes the vector layer "unmanaged", meaning that it is always on top.
+  vectorLayer.setMap(this.map_);
+  return vectorLayer;
+};
+
+
+/**
+ * @param {Object} config the config of te dataset
```
elemoine: "the config of te dataset" -> "The config of the dataset."

contribs/gmf/src/directives/search.js
```diff
@@ -0,0 +1,277 @@
...
+ */
+gmf.SearchController.prototype.filterLayername_ = function(layername) {
+  return function(feature) {
+    return (feature['properties']['layer_name'] === layername);
+  };
+};
+
+
+/**
+ * @param {Object} config the config of te dataset
```
elemoine: "the config of te dataset" -> "The config of the dataset."

contribs/gmf/src/directives/search.js
elemoine: Again, we should define a specific type for this. This time it wouldn't need to define in `gmfx.js` as it is an internal type only.

src/services/creategeojsonbloodhound.js
```diff
@@ -58,7 +61,14 @@ ngeo.createGeoJSONBloodhound = function(options, opt_featureProjection,
       },
       filter: function(parsedResponse) {
         var featureCollection = /** @type {GeoJSONFeatureCollection} */
-            (parsedResponse);
+            (goog.object.clone(parsedResponse));
+
+        if (goog.isDef(opt_filter) && !goog.isNull(opt_filter)) {
+          var features = /* Array.<GeoJSONFeature> */
+              (featureCollection['features']);
```
elemoine: We have a type defined in an externs file for `GeoJSONFeatureCollection`, so you don't need to use the square bracket notation here.

Use this:

```js
  if (goog.isDef(opt_filter)) {
    var features = featureCollection.features.filter(opt_filter);
    featureCollection.features = features;
  }
```

src/services/creategeojsonbloodhound.js
```diff
@@ -58,7 +61,14 @@ ngeo.createGeoJSONBloodhound = function(options, opt_featureProjection,
       },
       filter: function(parsedResponse) {
         var featureCollection = /** @type {GeoJSONFeatureCollection} */
-            (parsedResponse);
+            (goog.object.clone(parsedResponse));
```
elemoine: Any reason for cloning the object?

src/services/creategeojsonbloodhound.js
```diff
@@ -22,7 +23,7 @@
  *           '&lang=' + gettextCatalog.currentLanguage;
  *     }
  *   }
- * }, ol.proj.get('EPSG:3857'), ol.proj.get('EPSG:21781'));
+ * }, null, ol.proj.get('EPSG:3857'), ol.proj.get('EPSG:21781'));
```
elemoine: Use `undefined`.

src/services/creategeojsonbloodhound.js
```diff
@@ -33,8 +34,8 @@ goog.require('ol.format.GeoJSON');
 
 
 /**
- * @typedef {function(string, ol.proj.Projection=,
- *     ol.proj.Projection=):Bloodhound}
+ * @typedef {function(string, (function(Object): boolean)=,
```
elemoine: `Object` -> `GeoJSONFeature`

contribs/gmf/src/directives/search.js
elemoine: What would you want to display an alert? It does not make sense to me. At this point, I think we should just remove that functionality.

elemoine:
Thanks for the work @ger-benjamin. It looks good. I think it still requires work, but this is taking shape!

ger-benjamin:
@elemoine Thank you too, I'll update this soon.

contribs/gmf/src/directives/search.js
ger-benjamin: Just to keep the example, but I agree, that makes no sense right now.

elemoine: What do you mean by "keep the example"? We're not talking about the example code but the gmf search directive code which will be used in every GMF project.

contribs/gmf/src/directives/search.js
ger-benjamin: In GMF 1, we have a "tooltipTitle" parameter in the configuration of the FTS's widget. That can be a solution...

elemoine: The problem is the dynamic translation of that thing.

contribs/gmf/src/directives/search.js
ger-benjamin: All cases sems to work like this (unlike with a ``while do`` loop).
But I'll do a more understandable, logic, loop procedure.

elemoine: I see how now. I did not notice the `groupByProperties[0]` test.

contribs/gmf/src/directives/search.js
ger-benjamin: that's why I've wrote "I agree"

elemoine: Ok. I was just trying to understand "keep the example".

src/services/creategeojsonbloodhound.js
elemoine: We usually rely on the `goog.array` functions. So you may want to use `goog.array.filter` here. And we could use `goog.array.ASSUME_NATIVE_FUNCTIONS=true` in the build config files, to make the compiler actually use the native function in the compiled code.

contribs/gmf/src/directives/search.js
elemoine: I did for GMF's "themes" web service. See https://github.com/elemoine/ngeo/commit/2afea3fc8174484cf4d1e92e6d7724cc54e7fdac. You can do something similar for the "search" web service.

src/services/creategeojsonbloodhound.js
ger-benjamin: When you use a service (one url) multiple times with differents configuration of datasources, Typeahead will do only one request to this service (this is expected). But when you filter the results in one datasource, the results of this filter will be the new "parsedResponse" from each "same-service" datasource.

So I must work on a cloned object to avoid modifications on the service's response.

(Request a service multiple times is used to make groups in Typeahead.)

elemoine: In that case I'd use this:

```js
  /** @type {GeoJSONFeatureCollection} */
  var featureCollection = /** @type {GeoJSONFeatureCollection} */ (parsedResponse);
  if (goog.isDef(filter)) {
    featureCollection = {
      type: 'FeatureCollection',
      features: goog.array.filter(featureCollection.features, opt_filter);
    };
  }
```

Much more explicit IMO.

contribs/gmf/src/directives/search.js
ger-benjamin: Thanks -> wip...

ger-benjamin:
@elemoine I suggest that I close this PR and create a new one to clean all the comments.

contribs/gmf/src/directives/search.js
ger-benjamin: I've added a FIXME in the ``@fileoverview``

contribs/gmf/src/directives/search.js
ger-benjamin: Done (but under your comment, Github did not understand)

contribs/gmf/src/directives/search.js
ger-benjamin: Done (but under your comment, Github did not understand)

contribs/gmf/src/directives/search.js
```diff
@@ -0,0 +1,280 @@
...
+ * @fileoverview this file provides the "search" directive and controller for
+ * a GeoMapFish application.
+ *
+ * FIXME: The 'placeholder' in the input field is hard-coded and can't currently
+ * be internationalised.
+ *
+ * Example:
+ *
+ * <gmf-search gmf-search-map="ctrl.map"
+ *             gmf-search-datasources="ctrl.search_datasources">
```
elemoine: `search_datasource` -> `searchDatasources`

elemoine:
> @elemoine Updated. I suggest that I close this PR and create a new one to clean all the comments.

It's up to you. I don't mind.

PR #298, opened 2015-08-18
Use first scale if no optimal scale can be determined

Jenselme:


elemoine:
See https://github.com/Jenselme/ngeo/commit/402bcaa3bf6378945ca4eb7afdcc5fb2cf76733e, where I decided that the user is responsible for choosing what to use when there's no optimal scale. That doesn't work for you?

Jenselme:
Not really. I already had this behavior with `optimal = Infinity` and I find confusing not to have a true scale selected by default. I'd rather have a non optimum one, it feels better from a user point of view I think.

Feel free to reject this if you disagree.

elemoine:
I just want to give more control to the user. If `getOptimalScale` returns `-1` the user may decide to raise an error or just use the first scale.

Jenselme:
OK, good for me.

PR #300, opened 2015-08-26
Bases for gmf search

ger-benjamin:
First draw of the GeoMapFish2's search. It is already able to group results and this, by accepting multiple sources (multiple urls with or without filters).

I've added a "filter" parameter in the src/service/creategeojsonbloodhound.js

### DIFF
--- a/contribs/gmf/examples/search.js
+++ b/contribs/gmf/examples/search.js
@@ -1,0 +1,65 @@
+goog.provide('gmf-search');
+
+goog.require('gmf.mapDirective');
+goog.require('gmf.searchDirective');
+goog.require('ngeo.FeatureOverlayMgr');
+goog.require('ol.Map');
+goog.require('ol.View');
+goog.require('ol.layer.Tile');
+goog.require('ol.source.OSM');
+
+
+/** @const **/
+var app = {};
+
+
+/** @type {!angular.Module} **/
+app.module = angular.module('app', ['gmf']);
+
+
+
+/**
+ * @param {ngeo.FeatureOverlayMgr} ngeoFeatureOverlayMgr The ngeo feature
+ *     overlay manager service.
+ * @constructor
+ * @export
+ * @ngInject
+ */
+app.MainController = function(ngeoFeatureOverlayMgr) {
+
+  proj4.defs('EPSG:21781',
+      '+proj=somerc +lat_0=46.95240555555556 +lon_0=7.439583333333333 +k_0=1 ' +
+      '+x_0=600000 +y_0=200000 +ellps=bessel ' +
+      '+towgs84=674.374,15.056,405.346,0,0,0,0 +units=m +no_defs');
+
+  /**
+   * @type {Array.<gmfx.SearchDirectiveDatasource>}
+   * @export
+   */
+  this.searchDatasources = [{
+    title: 'From demo 1.6',
+    url: 'http://geomapfish-demo.camptocamp.net/1.6/wsgi/fulltextsearch',
+    projection: 'EPSG:21781'
+  }];
+
+  /**
+   * @type {ol.Map}
+   * @export
+   */
+  this.map = new ol.Map({
+    layers: [
+      new ol.layer.Tile({
+        source: new ol.source.OSM()
+      })
+    ],
+    view: new ol.View({
+      center: [0, 0],
+      zoom: 4
+    })
+  });
+
+  ngeoFeatureOverlayMgr.init(this.map);
+};
+
+
+app.module.controller('MainController', app.MainController);
--- a/contribs/gmf/externs/gmfx.js
+++ b/contribs/gmf/externs/gmfx.js
@@ -1,0 +1,39 @@
+/**
+ * @type {Object}
+ */
+var gmfx;
+
+/**
+ * Datasource configuration options for the search directive.
+ * @typedef {{
+ *    title: (string|undefined),
+ *    groupByProperties: (Array.<string|undefined>|undefined),
+ *    projection: (string|undefined),
+ *    url: (string|undefined)
+ * }}
+ */
+gmfx.SearchDirectiveDatasource;
+
+/**
+ * The title of this set of data.
+ * @type {string|undefined}
+ */
+gmfx.SearchDirectiveDatasource.prototype.title;
+
+/**
+ * An array of propertie's keys used to group data.
+ * @type {Array.<string>|undefined}
+ */
+gmfx.SearchDirectiveDatasource.prototype.groupByProperties;
+
+/**
+ * The geometry's projection for this set of data.
+ * @type {string|undefined}
+ */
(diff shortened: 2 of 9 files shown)


elemoine:
> I've used "geomapfish-demo.camptocamp.net/1.6" and, most important,"map.cartoriviera.ch" as datasources for my queries. Is that acceptable ?

We can rely on geomapfish-demo.camptocamp.net/1.6, but I don't think we should rely on map.cartoriviera.ch.

contribs/gmf/externs/gmfx.js
```diff
@@ -0,0 +1,39 @@
...
+
+/**
+ * The title of this set of data.
+ * @type {string|undefined}
+ */
+gmfx.SearchDirectiveDatasource.prototype.title;
+
+/**
+ * An array of propertie's keys used to group data.
+ * @type {Array.<string|undefined>|undefined}
```
elemoine: May the array include `undefined` values really?

contribs/gmf/src/directives/search.js
```diff
@@ -0,0 +1,280 @@
...
+
+  for (var i = 0; i < this.datasources_.length; i++) {
+    var datasource = this.datasources_[i];
+    var groupByProperties = datasource.groupByProperties || [];
+    var filter;
+
+    do {
+      var title = datasource.title;
+
+      if (groupByProperties[0]) {
```
elemoine: I'd use `groupByProperties.lengh > 0` here, to use the same test as in the `while` condition.

contribs/gmf/src/directives/search.js
```diff
@@ -0,0 +1,280 @@
...
+      } else {
+        filter = undefined;
+      }
+
+      this.datasets.push(this.createDataset_({
+        'title': title,
+        'filter': filter,
+        'projection': datasource.projection,
+        'url': datasource.url
+      }));
```
elemoine: You shouldn't use quotes for the object passed to `createDataset_`. I also realize that you use a `filter` property here that is not defined in the `gmfx.SearchDirectiveDatasource`. This requires more rigor in my opinion. Maybe you could just pass the filter with an extra argument to `createDataset_`.

contribs/gmf/src/directives/search.js
```diff
@@ -0,0 +1,280 @@
...
+   * @type {!ol.Map}
+   * @private
+   */
+  this.map_ = map;
+
+  /**
+   * @type {ol.layer.Vector}
+   * @private
+   */
+  this.vectorLayer_ = this.createVectorLayer_();
```
elemoine: We may want to use an `ngeo.FeatureOverlay` rather than a vector layer here, but we can change that later.

contribs/gmf/src/directives/search.js
```diff
@@ -0,0 +1,280 @@
...
+
+  /**
+   * @type {Array.<TypeaheadDataset>}
+   * @export
+   */
+  this.datasets = [];
+
+  for (var i = 0; i < this.datasources_.length; i++) {
+    var datasource = this.datasources_[i];
+    var groupByProperties = datasource.groupByProperties || [];
```
elemoine: Please use `goog.isDef`:

```js
/** @type {Array.<string>} */
var groupByProperties = goog.isDef(datasource.groupByProperties) ?
    datasource.groupByProperties : [];
```

contribs/gmf/src/directives/search.js
```diff
@@ -0,0 +1,280 @@
...
+};
+
+
+/**
+ * @param {gmfx.SearchDirectiveDatasource} config The config of the dataset.
+ * @return {TypeaheadDataset} A typeahead dataset.
+ * @private
+ */
+gmf.SearchController.prototype.createDataset_ = function(config) {
+  var directive_scope = this.scope_;
```
elemoine: Please use `directiveScope` as the variable name.

contribs/gmf/src/directives/search.js
```diff
@@ -0,0 +1,280 @@
...
+        return compile(html)(scope);
+      }
+    })
+  });
+};
+
+
+/**
+ * @param {string} layername The dataset "type" to keep.
+ * @return {(function(Object): boolean)} A filter function based on a
```
elemoine: `Object` -> `GeoJSONFeature`, no?

src/services/creategeojsonbloodhound.js
```diff
@@ -58,8 +61,16 @@ ngeo.createGeoJSONBloodhound = function(options, opt_featureProjection,
         return settings;
       },
       transform: function(parsedResponse) {
+        /** @type {GeoJSONFeatureCollection} */
         var featureCollection = /** @type {GeoJSONFeatureCollection} */
             (parsedResponse);
+        if (goog.isDef(opt_filter)) {
+          featureCollection = /** @type {GeoJSONFeatureCollection} */ ({
+            type: 'FeatureCollection',
+            features: goog.array.filter(featureCollection.features, opt_filter)
```
elemoine: Please add a `goog.require('goog.array')`.

elemoine:
I've added a few more comments. Please merge when addressed.

ger-benjamin:
All done. Thanks @elemoine I'll merge tomorow, when travis agree. 
(I've also replaced the vectorlayer by the ngeoFeatureOverlay service. It works fine !)

contribs/gmf/src/directives/search.js
```diff
@@ -0,0 +1,272 @@
...
+  var map = this.scope_['getMapFn']();
+  goog.asserts.assertInstanceof(map, ol.Map);
+
+  /**
+   * @type {!ol.Map}
+   * @private
+   */
+  this.map_ = map;
+
+  ngeoFeatureOverlayMgr.init(this.map_);
```
elemoine: This is not correct. This should be done at the application level, in the main controller.

elemoine:
I have one remaining comment. Please merge when addressed. 

Excellent work @ger-benjamin.

contribs/gmf/examples/search.js
```diff
@@ -0,0 +1,65 @@
...
+/** @type {!angular.Module} **/
+app.module = angular.module('app', ['gmf']);
+
+
+
+/**
+ * @param {ngeo.FeatureOverlayMgr} ngeoFeatureOverlayMgr The ngeo feature
+ *     overlay manager service.
+ * @constructor
+ * @export
```
elemoine: We don't need the `@export` tag here. Very minor.

elemoine: I'll go ahead and create a PR. There's also another minor change I'd like to make.

elemoine:
The gmf search example does not work on github.io. See http://camptocamp.github.io/ngeo/master/contribs/gmf/search.html. I'll have a look.

elemoine:
See https://github.com/camptocamp/ngeo/pull/301.

PR #302, opened 2015-08-27
Let MapFish Print choose WMS version

Jenselme:
For instance, we may pass VERSION=1.3.0, MapFish Print will add
VERSION=1.1.0 and the WMS server will use version 1.3.0 and print will
not work as required parameters for version 1.3.0 won't be passed.

### DIFF
--- a/src/services/print.js
+++ b/src/services/print.js
@@ -271,6 +271,7 @@ ngeo.Print.prototype.encodeWmsLayer_ = function(arr, opacity, url, params) {
 
   goog.object.remove(customParams, 'LAYERS');
   goog.object.remove(customParams, 'FORMAT');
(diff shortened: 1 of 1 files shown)


elemoine:
Shoudn't we add `version` to `MapFishPrintWmsLayer`?

elemoine:
See https://github.com/camptocamp/ngeo/compare/master...elemoine:print-wms-version.

What do you think? Does this work for you?

Jenselme:
It does work correctly this way. Do you want me to update this PR?

elemoine:
See https://github.com/camptocamp/ngeo/pull/305. I'll merge it when Travis is happy.

PR #303, opened 2015-08-27
Print only visible layers

Jenselme:


elemoine:
We'd need to check the resolution range (`minResolution` and `maxResolution`), but this can be done with a separate PR.

Jenselme:
I don't really understand what you mean by that nor what this has to do with this PR. Can you give me more explaination?

elemoine:
Let's merge this first :) I'll provide more information…

PR #304, opened 2015-08-27
Add ngeo.format.FeatureHash

elemoine:
This PR adds an OpenLayers format for encoding and decoding features in permalinks. The code is based on sbrunner's OpenLayers 2 URLCompressed plugin.

TODOs:

* [x] Encoding and decoding of styles
* [ ] Simplify geometries before encoding
* [x] Add a test to verify compatibility with URLCompressed
* [x] Support multi-geometries
* [ ] Custom attribute and style transformers
* [ ] Maximum number of vertices

### DIFF
--- a/examples/permalink.js
+++ b/examples/permalink.js
@@ -1,12 +1,20 @@
 goog.provide('permalink');
 
 goog.require('ngeo.Debounce');
+goog.require('ngeo.DecorateInteraction');
 goog.require('ngeo.Location');
+goog.require('ngeo.format.FeatureHash');
 goog.require('ngeo.mapDirective');
 goog.require('ol.Map');
 goog.require('ol.View');
+goog.require('ol.geom.GeometryType');
+goog.require('ol.interaction.Draw');
 goog.require('ol.layer.Tile');
+goog.require('ol.layer.Vector');
 goog.require('ol.source.OSM');
+goog.require('ol.source.Vector');
+goog.require('ol.style.Stroke');
+goog.require('ol.style.Style');
 
 
 /** @const **/
@@ -98,6 +106,133 @@ app.MapDirectiveController = function(ngeoLocation, ngeoDebounce) {
 app.module.controller('AppMapController', app.MapDirectiveController);
 
 
+/**
+ * A draw directive that adds a simple draw tool.
+ *
+ * @return {angular.Directive} The directive specs.
+ */
+app.drawDirective = function() {
+  return {
+    restrict: 'E',
+    scope: {
+      'map': '=appDrawMap',
+      'layer': '=appDrawLayer'
+    },
+    controller: 'AppDrawController',
+    controllerAs: 'ctrl',
+    bindToController: true,
+    template:
+        '<label>Enable drawing:' +
+        '<input type="checkbox" ng-model="ctrl.interaction.active" />' +
+        '</label><br>' +
+        '<button ng-click="ctrl.clearLayer()">Clear layer</button>'
+
+  };
+};
+
+
+app.module.directive('appDraw', app.drawDirective);
+
+
+
+/**
+ * @param {angular.Scope} $scope Scope.
+ * @param {ngeo.DecorateInteraction} ngeoDecorateInteraction Decorate
+ *     interaction service.
+ * @param {ngeo.Location} ngeoLocation ngeo Location service.
+ * @constructor
+ * @export
+ * @ngInject
+ */
+app.DrawDirectiveController =
+    function($scope, ngeoDecorateInteraction, ngeoLocation) {
+
+  /**
+   * @type {ol.Map}
+   * @export
+   */
+  this.map;
+
+  /**
+   * @type {ol.layer.Vector}
+   */
+  this.layer;
+
+  /**
+   * @type {ngeo.Location}
+   * @private
+   */
+  this.ngeoLocation_ = ngeoLocation;
+
+  /**
+   * @type {number}
+   * @private
+   */
+  this.featureSeq_ = 0;
+
+  var vectorSource = this.layer.getSource();
+
+  /**
+   * @type {ol.interaction.Draw}
+   * @export
+   */
+  this.interaction = new ol.interaction.Draw({
+    type: /** @type {ol.geom.GeometryType} */ ('LineString'),
+    source: vectorSource
+  });
+
+  var interaction = this.interaction;
+  interaction.setActive(false);
+  this.map.addInteraction(interaction);
+  ngeoDecorateInteraction(interaction);
+
+  this.interaction.on('drawend', function(e) {
+    e.feature.set('id', ++this.featureSeq_);
+  }, this);
+
+
+  // Deal with the encoding and decoding of features in the URL.
+
+  var fhFormat = new ngeo.format.FeatureHash();
+
+  vectorSource.on('addfeature', function(e) {
+    var feature = e.feature;
+    feature.setStyle(new ol.style.Style({
+      stroke: new ol.style.Stroke({
+        color: [255, 0, 0, 1],
+        width: 2
+      })
+    }));
+    var features = vectorSource.getFeatures();
+    var encodedFeatures = fhFormat.writeFeatures(features);
+    $scope.$applyAsync(function() {
+      ngeoLocation.updateParams({'features': encodedFeatures});
+    });
+  });
+
+  var encodedFeatures = ngeoLocation.getParam('features');
+  if (angular.isDefined(encodedFeatures)) {
+    var features = fhFormat.readFeatures(encodedFeatures);
+    this.featureSeq_ = features.length;
+    vectorSource.addFeatures(features);
+  }
+
+};
+
+
+/**
+ * Clear the vector layer.
+ * @export
+ */
+app.DrawDirectiveController.prototype.clearLayer = function() {
+  this.layer.getSource().clear(true);
+  this.featureSeq_ = 0;
+  this.ngeoLocation_.deleteParam('features');
+};
+
+app.module.controller('AppDrawController', app.DrawDirectiveController);
+
+
 
 /**
  * @constructor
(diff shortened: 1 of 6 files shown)


elemoine:
This is WIP and not ready for review yet.

elemoine:
There are missing features, such as the encoding and decoding of styles. But at this point I'd still like to get reviews and merge this work.
 
Here is a link that can be used for testing: http://erilem.net/ngeo/pla-format/permalink.html?z=4&x=0&y=0&features=Fl(6r5p-yvs*n95s-m5xw-snta-7hym-~id*1).

elemoine:
I've spent another day on this, adding support for multi geometries and for the encoding and decoding of styles. There are still missing bits when compared to `URLCompressed`, but the base is here, and extending the format with new features should be relatively easy.

Can someone please review this?

elemoine:
Quick question for @sbrunner: @pmauduit and I looked at the [character set](https://github.com/camptocamp/ngeo/pull/304/files#diff-55d1e76bb600bad24ba3a6abeb908f4eR104) used for the encoding of geometries. It is very close to the standard [base64](https://en.wikipedia.org/wiki/Base64) character set, where `i`, `o`, `l`, `+` and `/` are replaced by `.`, `-`, `_`, `!` and `*`.  We understand that `+` and `/` are replaced because they are URL-encoded as 3 characters each (`%2B` and `%2F`, respectively). But do you what is the reason for not using `i`, `o` and `l`?

examples/permalink.js
```diff
@@ -98,6 +103,127 @@ app.MapDirectiveController = function(ngeoLocation, ngeoDebounce) {
...
+
+  /**
+   * @type {ol.interaction.Draw}
+   * @export
+   */
+  this.interaction = new ol.interaction.Draw(
+      /** @type {olx.interaction.DrawOptions} */ ({
+        type: 'LineString',
+        source: vectorSource
+      }));
```
pgiraud: Is it better to cast the DrawOptions or the GeometryType like the following?

```
this.interaction = new ol.interaction.Draw({
    source: vectorSource,
    type: /** @type {ol.geom.GeometryType} */ ('LineString')
});
```

elemoine: The `DrawOptions` record type has many more properties than what's used here. See https://github.com/openlayers/ol3/blob/master/externs/olx.js#L2370-L2386. This is why a cast is used here. But the compiler may now be able to infer the type. I'll try.

elemoine: You're right. It compiles the way you wrote it.

src/ol-ext/format/featurehash.js
```diff
@@ -0,0 +1,940 @@
...
+ * @return {string} String.
+ * @private
+ */
+ngeo.format.FeatureHash.encodeSignedNumber_ = function(num) {
+  var sgnNum = num << 1;
+  if (num < 0) {
+    sgnNum = ~(sgnNum);
+  }
+  return ngeo.format.FeatureHash.encodeNumber_(sgnNum);
+};
```
pgiraud: Really minor but can `sgnNum` be renamed to `signedNumber`?

elemoine: I'll rename it to `signedNum` because the argument name is `num` and I don't want to rename `num` to `number` as `number` is reserved word.

src/ol-ext/format/featurehash.js
```diff
@@ -0,0 +1,940 @@
...
+  }
+  encodedNumber += ngeo.format.FeatureHash.CHAR64_.charAt(num);
+  return encodedNumber;
+};
+
+
+/**
+ * @param {Array.<ol.style.Style>} styles Styles.
+ * @param {ol.geom.GeometryType} geometryType Geometry type.
+ * @param {Array.<string>} encodedStyles Array of encoded strings.
```
pgiraud: Array of encoded styles.

src/ol-ext/format/featurehash.js
```diff
@@ -0,0 +1,940 @@
...
+ * @private
+ */
+ngeo.format.FeatureHash.encodeStyles_ =
+    function(styles, geometryType, encodedStyles) {
+  var styleType = ngeo.format.FeatureHashStyleTypes_[geometryType];
+  if (!goog.isDef(styleType)) {
```
pgiraud: Would it be useful to use `goog.assert` here?

src/ol-ext/format/featurehash.js
elemoine: Will change.

src/ol-ext/format/featurehash.js
elemoine: Yes, also more consistent with how it's done in `readGeometryFromText`.

src/ol-ext/format/featurehash.js
```diff
@@ -0,0 +1,940 @@
...
+    var fillStyle = imageStyle.getFill();
+    if (!goog.isNull(fillStyle)) {
+      ngeo.format.FeatureHash.encodeStyleFill_(fillStyle, encodedStyles);
+    }
+    var strokeStyle = imageStyle.getStroke();
+    if (!goog.isNull(strokeStyle)) {
+      ngeo.format.FeatureHash.encodeStyleStroke_(strokeStyle, encodedStyles);
+    }
+  }
+};
```
pgiraud: `ol.style.Icon` is not taken into account yet, is it?

elemoine: Indeed. I'll add a note in the header of the file.

src/ol-ext/format/featurehash.js
```diff
@@ -0,0 +1,940 @@
...
+
+
+/**
+ * @param {string} text Text.
+ * @param {olx.format.ReadOptions=} opt_options Read options.
+ * @return {ol.geom.LineString} Line string.
+ * @this {ngeo.format.FeatureHash}
+ * @private
+ */
+ngeo.format.FeatureHash.readLineStringGeometry_ = function(text, opt_options) {
```
pgiraud: opt_options doesn't seem to be used

pgiraud: Same for the other readXxxxxGeometry_ functions.

elemoine: True. It's related to a limitation, where the format doesn't support transforming the decoded geometries. I'll remove the unused `opt_options`.

src/ol-ext/format/featurehash.js
```diff
@@ -0,0 +1,940 @@
...
+
+/**
+ * @param {ol.geom.Geometry} geometry Geometry.
+ * @param {olx.format.WriteOptions=} opt_options Write options.
+ * @return {string} Encoded geometry.
+ * @this {ngeo.format.FeatureHash}
+ * @private
+ */
+ngeo.format.FeatureHash.writeLineStringGeometry_ =
+    function(geometry, opt_options) {
```
pgiraud: `opt_options` should be removed from the `writeXxxxxGeometry_` functions as well.

elemoine: Yes. There are places where it cannot be removed though, because it's part of the signature from the parent method.

src/ol-ext/format/featurehash.js
```diff
@@ -0,0 +1,940 @@
...
+    for (var j = 0; j < linearRingCount; ++j) {
+      // skip the "closing" point
+      var end = ends[j] - stride;
+      var text = this.encodeCoordinates_(flatCoordinates, stride, offset, end);
+      if (j !== 0) {
+        textArray.push('\'');
+      }
+      textArray.push(text);
+      offset = ends[j];
+    }
```
pgiraud: Cannot the linearRing loop be factorized? It is used similarly in `writePolygonGeometry_`.

src/ol-ext/format/featurehash.js
```diff
@@ -0,0 +1,940 @@
...
+   */
+  this.prevY_ = 0;
+
+};
+goog.inherits(ngeo.format.FeatureHash, ol.format.TextFeature);
+
+
+/**
+ * Characters used to encode the data.
+ * ~'() will be used as separator.
```
pgiraud: Reading this, I don't understand what the separator is.

src/ol-ext/format/featurehash.js
```diff
@@ -0,0 +1,940 @@
...
+  return features;
+};
+
+
+/**
+ * @inheritDoc
+ */
+ngeo.format.FeatureHash.prototype.readGeometryFromText =
+    function(text, opt_options) {
+  var firstChar = text[0];
```
pgiraud: Is it useful to use a variable here?

src/ol-ext/format/featurehash.js
```diff
@@ -0,0 +1,940 @@
...
+ * @type {Object.<string, function(string, olx.format.ReadOptions=):ol.geom.Geometry>}
+ */
+ngeo.format.FeatureHash.GEOMETRY_READERS_ = {
+  'P': ngeo.format.FeatureHash.readMultiPointGeometry_,
+  'L': ngeo.format.FeatureHash.readMultiLineStringGeometry_,
+  'A': ngeo.format.FeatureHash.readMultiPolygonGeometry_,
+  'l': ngeo.format.FeatureHash.readLineStringGeometry_,
+  'p': ngeo.format.FeatureHash.readPointGeometry_,
+  'a': ngeo.format.FeatureHash.readPolygonGeometry_
+};
```
pgiraud: Those prefixes could have been used when writing geometries as well.

src/ol-ext/format/featurehash.js
elemoine: No. I'll remove it.

src/ol-ext/format/featurehash.js
elemoine: Separators to separate coordinates and properties and key/value pairs within properties blocks.

pgiraud: Should we read the following?
~, ', ( and ) will be used as separators.

elemoine: Right. I changed the text.

src/ol-ext/format/featurehash.js
elemoine: I haven't taken this comment into account, as I don't quite see how I could address it in practice.

elemoine:
Do you have more comments @pgiraud?

pgiraud:
You can merge. Thanks.

elemoine:
Thanks for the review. I am conscious that the format is not fully complete, but I think we can easily extend it, and I've added TODOs in the header of the file.

PR #307, opened 2015-09-02
Add goog.provide's for constructors

elemoine:
Recent versions of the compiler require a goog.provide for every type defined as constructor.


fredj:
+1

fgravin:
why ?

elemoine:
Because constructors define types and it is a convention in Closure that types have associated `goog.provide`'s. And recent versions of the compiler now checks that the files have correct `goog.provide`'s and `goog.require`'s.

PR #308, opened 2015-09-04
Adding a modal directive

pgiraud:
Please review.

### DIFF
--- a/examples/modal.js
+++ b/examples/modal.js
@@ -1,0 +1,28 @@
+goog.provide('modal');
+
+goog.require('ngeo.modalDirective');
+
+
+/** @const **/
+var app = {};
+
+
+/** @type {!angular.Module} **/
+app.module = angular.module('app', ['ngeo']);
+
+
+
+/**
+ * @constructor
+ */
+app.MainController = function() {
+
+  /**
+   * @type {boolean}
+   * @export
+   */
+  this.modalShown = false;
+};
+
+
+app.module.controller('MainController', app.MainController);
--- a/src/directives/modal.js
+++ b/src/directives/modal.js
@@ -1,0 +1,74 @@
+/**
+ * @fileoverview Provides the "ngeoModal" directive. This directive allows
+ * displaying a modal window. The directive takes care or the show/hide
+ * actions.
+ *
+ *
+ * Example:
+ *
+ * <ngeo-modal ngeo-modal-shown="modalShown">
+ *   <div class="modal-header">
+ *     <button type="button" class="close" data-dismiss="modal"
+ *         aria-hidden="true">&times;</button>
+ *     <h4 class="modal-title">The Title</h4>
+ *   </div>
+ *   <div class="modal-body">Some content</div>
+ * </ngeo-modal>
+ *
+ * This directive requires Bootstrap's `modal` classes and associated jQuery
+ * plugin.
+ */
+
+goog.provide('ngeo.modalDirective');
+
+goog.require('ngeo');
+
+
+/**
+ * @param {angular.$parse} $parse Angular parse service.
+ * @return {angular.Directive} The directive specs.
+ * @ngInject
+ */
+ngeo.modalDirective = function($parse) {
+  return {
+    template: '<div class="modal fade" tabindex="-1" role="dialog">' +
+        '<div class="modal-dialog">' +
+        '<div class="modal-content">' +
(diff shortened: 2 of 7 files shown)


examples/modal.js
```diff
@@ -0,0 +1,23 @@
...
+/** @type {!angular.Module} **/
+app.module = angular.module('app', ['ngeo']);
+
+
+
+/**
+ * @constructor
+ */
+app.MainController = function() {
+  this['modalShown'] = false;
```
elemoine: We now use `@export`.

```js
    /**
     * @type {boolean}
     * @export
     */
    this.modalShown = false;
```

The `@export` annotation prevents renaming.

src/directives/modal.js
```diff
@@ -0,0 +1,67 @@
+/**
+ * @fileoverview Provides the "ngeoModal" directive. This directive allows
+ * displaying a modal window. The directive takes care or the show/hide
+ * actions.
+ *
+ *
+ * Example:
+ *
+ * <modal shown="modalShown">
```
elemoine: `<modal>` -> `<ngeo-modal>`

elemoine: Even:

`<modal shown="modalShown">`  -> `<ngeo-modal ngeo-modal-shown="modalShown">`. This is to match our conventions.

src/directives/modal.js
```diff
@@ -0,0 +1,67 @@
...
+    template: '<div class="modal fade" tabindex="-1" role="dialog">' +
+        '<div class="modal-dialog">' +
+        '<div class="modal-content">' +
+        '<ng-transclude></ng-transclude>' +
+        '</div>' +
+        '</div>' +
+        '</div>',
+    restrict: 'E',
+    transclude: true,
+    replace: true,
```
elemoine: `replace` is deprecated. We don't use it in ngeo.

src/directives/modal.js
```diff
@@ -0,0 +1,67 @@
...
+ */
+
+goog.provide('ngeo.modalDirective');
+
+goog.require('ngeo');
+
+
+/**
+ * @param {angular.$parse} $parse Angular parse service.
+ * @return {angular.Directive} The directive specs.
```
elemoine: You need to use `@ngInject` or it won't work when compiled.

src/directives/modal.js
```diff
@@ -0,0 +1,67 @@
...
+        '<div class="modal-dialog">' +
+        '<div class="modal-content">' +
+        '<ng-transclude></ng-transclude>' +
+        '</div>' +
+        '</div>' +
+        '</div>',
+    restrict: 'E',
+    transclude: true,
+    replace: true,
+    link: function(scope, element, attrs) {
```
elemoine: I think it's a good idea to type the function. See https://github.com/camptocamp/ngeo/blob/master/src/directives/btngroup.js#L107-L113. In this way, `scope`, `element` and `attrs` are typed, and you're then guaranteed to 
correctly use these variables.

src/directives/modal.js
```diff
@@ -0,0 +1,67 @@
+/**
+ * @fileoverview Provides the "ngeoModal" directive. This directive allows
+ * displaying a modal window. The directive takes care or the show/hide
+ * actions.
```
elemoine: Is the directive bootstrap-specific? If so it should be documented. Also, does the directive requires bootstrap.js and jQuery? If so it should be documented as well.

src/directives/modal.js
```diff
@@ -0,0 +1,67 @@
...
+ */
+ngeo.modalDirective = function($parse) {
+  return {
+    template: '<div class="modal fade" tabindex="-1" role="dialog">' +
+        '<div class="modal-dialog">' +
+        '<div class="modal-content">' +
+        '<ng-transclude></ng-transclude>' +
+        '</div>' +
+        '</div>' +
+        '</div>',
```
elemoine: I'd recommend adding a partial.

elemoine:
I added a couple comments, but it looks good to me. I think you can merge this when the comments are addressed. Thanks!

src/directives/modal.js
pgiraud: There's a note for this a few lines below.

src/directives/modal.js
pgiraud: Using a partial was a bit painful in this simple case.

PR #314, opened 2015-09-11
Add a setFeatures method to feature overlays

elemoine:
This PR adds the possibility to configure an ngeo feature overlay with a feature collection (`ol.Collection.<ol.Feature>`).

This makes it possible to bind a drawing tool to an feature overlay:

```js
var features = new ol.Collection();

var overlay = ngeoFeatureOverlayMgr.getFeatureOverlay();
overlay.setStyle(myStyle);
overlay.setFeatures(features);

var draw = new ol.interaction.Draw({
  type: 'point',
  features: features
});
```

With this features drawn with the draw interaction are added to the ngeo feature overlay.

### DIFF
--- a/src/services/featureoverlay.js
+++ b/src/services/featureoverlay.js
@@ -21,6 +21,9 @@ goog.provide('ngeo.FeatureOverlayMgr');
 
 goog.require('goog.object');
 goog.require('ngeo');
+goog.require('ol.Collection');
+goog.require('ol.CollectionEvent');
+goog.require('ol.CollectionEventType');
 goog.require('ol.Feature');
 goog.require('ol.layer.Vector');
 goog.require('ol.source.Vector');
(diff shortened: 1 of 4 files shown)


src/services/featureoverlay.js
```diff
@@ -239,4 +277,24 @@ ngeo.FeatureOverlay.prototype.setStyle = function(style) {
 };
 
 
+/**
+ * @param {ol.CollectionEvent} evt Feature collection event.
+ * @private
+ */
+ngeo.FeatureOverlay.prototype.handleFeaturesAdd_ = function(evt) {
```
pgiraud: It should be called `handleFeatureAdd_` since we add only one feature at a time.

pgiraud:
I added a minor comment. Otherwise it looks good to me.

elemoine:
Thanks for the comment. I will merge this when Travis is happy.

PR #317, opened 2015-09-16
Updated startpage with menu and added links for easy access to the ngeo files.

matjos:


matjos:
@elemoine This is kind of the begining to what i was requesting in the forum.

Merge if you think its a good idea or else discard.

elemoine:
@matjos The `gh-pages` branch is generated automatically, on each push to the master branch on GitHub. So pull requests with `camptocamp:gh-pages` as the target branch cannot be accepted.

matjos:
@elemoine okey

PR #318, opened 2015-09-17
Create EPSG21781 projection definition

fredj:
To avoid to copy / paste this file in (almost) all client project.

### DIFF
--- a/contribs/gmf/src/proj/epsg21781projection.js
+++ b/contribs/gmf/src/proj/epsg21781projection.js
@@ -1,0 +1,11 @@
+goog.provide('gmf.proj.EPSG21781');
+
+goog.require('ol.proj');
+
+if (typeof proj4 == 'function') {
+  proj4.defs('EPSG:21781',
+      '+proj=somerc +lat_0=46.95240555555556 +lon_0=7.439583333333333 ' +
+      '+k_0=1 +x_0=600000 +y_0=200000 +ellps=bessel' +
+      '+towgs84=674.374,15.056,405.346,0,0,0,0 +units=m +no_defs');
+  ol.proj.get('EPSG:21781').setExtent([420000, 30000, 900000, 350000]);
+}


src/ol-ext/proj/epsg21781projection.js
```diff
@@ -0,0 +1,17 @@
+goog.provide('ol.proj.EPSG21781');
+
+/**
+ * this ol.Map require is a workaround to be sure that ol.proj.common.add()
```
gberaudo: Missing capital letter.

gberaudo:
Makes sense to me.

src/ol-ext/proj/epsg21781projection.js
```diff
@@ -0,0 +1,17 @@
...
+ * is called before this projection is added.
+ *
+ * @suppress {extraRequire}
+ */
+goog.require('ol.Map');
```
elemoine: What is the reason for requiring `ol.Map`? (I am sure there's one.)

elemoine:
How about adding this to the `gmf` contrib instead?

src/ol-ext/proj/epsg21781projection.js
```diff
@@ -0,0 +1,17 @@
+goog.provide('ol.proj.EPSG21781');
+
+/**
+ * this ol.Map require is a workaround to be sure that ol.proj.common.add()
+ * is called before this projection is added.
```
elemoine: Could you please clarify why we need to call `ol.proj.common.add` before adding the EPSG:21781 projection.

fredj:
> How about adding this to the `gmf` contrib instead?

Yes, it's probably a better place ...

src/ol-ext/proj/epsg21781projection.js
fredj: It should not be needed, I need to check.

elemoine:
@fredj, can this be merged?

fredj:
Thanks for the reminder.
Yes, this be merged

contribs/gmf/src/proj/epsg21781projection.js
```diff
@@ -0,0 +1,11 @@
+goog.provide('ol.proj.EPSG21781');
```
elemoine: I don't think ngeo should provide namespaces prefixed with `ol`.

fredj: changed to `gmf.proj.EPSG21781`

elemoine:
LGTM

fredj:
Thanks for the review

PR #320, opened 2015-09-21
Copy ngeo-debug.js to examples/lib on github.io

elemoine:
Right now only ngeo.js is available on camptocamp.github.io. This adds ngeo-debug.js.


fredj:
+1

PR #321, opened 2015-09-21
Fix the print tests

elemoine:
With https://github.com/marcjansen/ol3/commit/80995159b1e71cc5c48b634fadc9d52f02830efb text styles now have a default color. This broke the ngeo print tests. This commits fixes the problem by adding a fill style to the text styles used in the tests.


elemoine:
Test-only change, so merging.

PR #322, opened 2015-09-21
Add a "properties" option to ngeo.format.FeatureHash

elemoine:
This commit adds an optional "properties" option to the ngeo.format.FeatureHash constructor. Using this new option the user can provide a function that returns serializable properties for a feature. This can be used to filter out feature properties. This can also be used to add properties that are not set in the feature.


pgiraud:
This looks very good to me.

PR #323, opened 2015-09-23
Add an encodeStyles option to ngeo.format.FeatureHash

elemoine:
This PR adds an `encodeStyles` option to `ngeo.format.FeatureHash`. When set to `false` the format does not encode styles.

Please review.


fredj:
LGTM

PR #324, opened 2015-09-24
Latest jquery-1.9.js externs file is broken

elemoine:
We get a compile error with the latest jquery-1.9.js externs file:

```
ERR! compile .build/externs/jquery-1.9.js:94: ERROR - accessing name jQuery in externs has no effect. Perhaps you forgot to add a var keyword?
ERR! compile var $ = jQuery;
ERR! compile         ^
ERR! compile
ERR! compile
ERR! compile 1 error(s), 0 warning(s)
ERR! compile
```

This commit fixes the problem by fetching the jquery-1.9.js file from a specific commit.

This is a temporary fix, until we (or someone) really fix the actual problem.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -333,7 +333,7 @@ $(EXTERNS_ANGULAR_HTTP_PROMISE):
 
 $(EXTERNS_JQUERY):
 	mkdir -p $(dir $@)
-	wget -O $@ https://raw.githubusercontent.com/google/closure-compiler/master/contrib/externs/jquery-1.9.js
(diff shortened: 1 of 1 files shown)


fredj:
According to https://github.com/google/closure-compiler/pull/1165 the closure-compiler v20150920 is needed

elemoine:
Yes, that's what I've been trying. But the new compiler also requires a new version of the Closure Library. May I merge this for now?

fredj:
please merge

PR #325, opened 2015-09-24
Use version 20150920 of Closure Compiler

elemoine:
OpenLayers now uses version 20150920 of Closure Compiler. This PR updates ngeo to

* use the `master` version of the jquery-1.9.js externs again
* remove the `checkStructDictInheritance` option which does no longer exist
* remove the definition of `Navigator.prototype.languages`


fredj:
LGTM

elemoine:
Thanks @fredj.

PR #331, opened 2015-10-01
Export gmf.source.AsitVD

fredj:


elemoine:
It is not the way we do it in ngeo. Indeed, we want  `gmf.source.AsitVD` to be exported in `gmf.js` (used by the examples on github.io), but we don't want `gmf.source.AsitVD` to be exported in our applications, which are compiled together with ngeo. If you use `@export` then `gmf.source.AsitVD` will unconditionally be exported. This is why we use separate exports files. Makes sense?

fredj:
makes sense.
we need to remove the `@export` annotation elsewhere in ngeo (https://github.com/camptocamp/ngeo/blob/master/src/directives/scaleselector.js#L92, https://github.com/camptocamp/ngeo/blob/master/src/directives/layertree.js#L135, ...) right ?

elemoine:
> makes sense.
> we need to remove the @export annotation elsewhere in ngeo (https://github.com/camptocamp/ngeo/blob/master/src/directives/scaleselector.js#L92, https://github.com/camptocamp/ngeo/blob/master/src/directives/layertree.js#L135, ...) right ?

No. Both controller classes include methods with `@export` (`getScale` for example). Contrary to `ngeo.source.AsitVD` these methods need to be unconditionally exported, because they're used in Angular templates. And for `@export` to work on these methods their constructors must have an `@export` as well. This is [explained](https://github.com/camptocamp/ngeo/blob/master/docs/guidelines.md#use-export-for-controllers) in our guidelines.

elemoine:
Please merge.

fredj:
And thanks for all your explanations

elemoine:
> And thanks for all your explanations

You can now be the one verifying that PRs/patches are correct with respect to exports/externs… :)

PR #332, opened 2015-10-01
Make check-example exit 1 on console errors

elemoine:
Currently exceptions thrown when checking the examples with the `check-example.js` PhantomJS script are not detected. This leads to PRs whose builds pass Travis but whose examples don't actually work on github.io. This PR aims to fix that by making `check-example` exit 1 when errors are detected on the console.

### DIFF
--- a/buildtools/check-example.js
+++ b/buildtools/check-example.js
@@ -30,4 +30,5 @@ page.open(examplePath, function(s) {
 });
 page.onConsoleMessage = function(msg) {
   console.log('console:', msg);
+  phantom.exit(msg.match(/error/i) ? 1 : 0);
 };


elemoine:
The Travis build failed, which is expected, and proves that this patch works :)

elemoine:
I've just rebased this. With https://github.com/Geoportail-Luxembourg/geoportailv3/pull/880 now merged the build should pass.

gberaudo:
Exceptions are supposed to be catched by `page.onError`, right?
Do you have an idea why it is not enough?

elemoine:
> Exceptions are supposed to be catched by page.onError, right?
> Do you have an idea why it is not enough?

Not really.

gberaudo:
I think I got it: angular catches all exceptions and only `prints` the error.

elemoine:
> I think I got it: angular catches all exceptions and only prints the error.

Sounds like a good explanation.

fredj:
Is it ready to be merged ?

elemoine:
Thanks for the reminder @fredj.

PR #339, opened 2015-10-15
Add a mobile app GMF example

elemoine:
This PR adds a mobile application to the GMF examples. The goal of this example is to develop a mobile layout for GMF. We'll start from a simple layout, with no functionality, and gradually functionality to the example.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -7,10 +7,13 @@ EXAMPLES_JS_FILES := $(shell find examples -maxdepth 1 -type f -name '*.js')
 EXAMPLES_HTML_FILES := $(shell find examples -maxdepth 1 -type f -name '*.html')
 
 GMF_SRC_JS_FILES := $(shell find contribs/gmf/src -type f -name '*.js')
-GMF_EXAMPLES_JS_FILES := $(shell find contribs/gmf/examples -maxdepth 1 -type f -name '*.js')
+GMF_EXAMPLES_JS_FILES := $(shell find contribs/gmf/examples -type f -name '*.js')
 GMF_EXAMPLES_HTML_FILES := $(shell find contribs/gmf/examples -maxdepth 1 -type f -name '*.html')
+GMF_EXAMPLES_LESS_FILES := $(shell find contribs/gmf/examples/style -maxdepth 1 -type f -name '*.less')
+GMF_EXAMPLES_MAIN_LESS_FILES := $(filter %-main.less, $(GMF_EXAMPLES_LESS_FILES))
+GMF_EXAMPLES_CSS_FILES := $(patsubst %-main.less,%.css,$(GMF_EXAMPLES_MAIN_LESS_FILES))
 
-BUILD_EXAMPLES_CHECK_TIMESTAMP_FILES := $(patsubst examples/%.html, .build/%.check.timestamp, $(EXAMPLES_HTML_FILES)) $(patsubst contribs/gmf/examples/%.html, .build/contribs/gmf/%.check.timestamp, $(GMF_EXAMPLES_HTML_FILES))
+BUILD_EXAMPLES_CHECK_TIMESTAMP_FILES := $(patsubst examples/%.html,.build/%.check.timestamp,$(EXAMPLES_HTML_FILES)) $(patsubst contribs/gmf/examples/%.html,.build/contribs/gmf/%.check.timestamp,$(GMF_EXAMPLES_HTML_FILES))
 
 EXTERNS_ANGULAR = .build/externs/angular-1.4.js
 EXTERNS_ANGULAR_Q = .build/externs/angular-1.4-q_templated.js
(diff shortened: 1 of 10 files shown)


contribs/gmf/examples/mobileapp.html
```diff
@@ -0,0 +1,90 @@
...
+           <li class="has-children">
+             <a href="#">Mobile Phones</a>
+             <ul class="is-hidden">
+               <li>
+                 <a href="#">Mobile Phone 1</a>
+               </li>
+             </ul>
+           </li>
+           <li>
+             <a href>Télévisions</a>
```
asaunier: Rather use the english word?

elemoine: Sure. That part is very temporary.

elemoine:
We may want to merge this if others want to start adding and testing functionality.

One thing that will may need to be changed. Currently the `mobileapp` example is not different from other GMF examples. This means that, for the example to work on github.io (with `gmf.js`), its Javascript code cannot use Closure Library (`goog` objects). In practice, in a real GMF project, the application will be compiled, and using Closure Library will be possible (and recommended).

@pgiraud, can you please review this?

Makefile
```diff
@@ -38,6 +41,7 @@ help:
 	@echo "- dist                    Compile the lib into an ngeo.js standalone build (in dist/)"
 	@echo "- check                   Perform a number of checks on the code"
 	@echo "- lint                    Check the code with the linter"
+	@echo "- less                    Compile less code to css"
```
sbrunner: `less code` can mean something different, I think that `less files` is better...

sbrunner:
I think that the `app.navDirective` and the `app.mainController` function should be in ngeo, not in the project part...

Makefile
elemoine: Yes. I actually planned to remove that PHONY target.

elemoine:
> I think that the app.navDirective and the app.mainController function should be in ngeo, not in the project part...

+1 to move the `app.navDirective` to the `gmf` library at some point. The `app.MainController` includes states and functions for `index.html`. So it's really application-specific.

contribs/gmf/examples/style/mobileapp-main.less
```diff
@@ -0,0 +1,23 @@
+@import (inline) '../../../../node_modules/openlayers/css/ol.css';
+@import (inline) 'reset.css';
+
+@import 'mobileapp-vars.less';
+
+@import '../../../../node_modules/more-or-less/less/css3/_transition';
+@import '../../../../node_modules/more-or-less/less/css3/_transform';
```
pgiraud: Instead of relying on `more-or-less`, we should use `bootstrap`'s mixins.
Same for `reset.css`.

I'll give it a try.

elemoine: > Instead of relying on more-or-less, we should use bootstrap's mixins.
> Same for reset.css.

Good call!

elemoine: I am happy to push the branch to camptocamp/ngeo for collaboration.

elemoine:
@pgiraud and I just decided to merge this, and continue the work with new pull requests. So if there's no objection I'll merge this when the build passes on Travis.

sbrunner:
> The app.MainController includes states and functions for index.html. So it's really application-specific.

Effectively the `app.MainController` constructor is really application-specific but the other methods looks relay generic, it it possible to extends a base controller ?

elemoine:
> Effectively the app.MainController constructor is really application-specific but the other methods looks relay generic, it it possible to extends a base controller ?

What methods? `rightNavIsVisible` and friends are all application-specific. They're just accessors for the application-specific states.

sbrunner:
```

+/**
+ * @export
+ */
+app.MainController.prototype.toggleLeftNavVisibility = function() {
+  this.leftNavVisible = !this.leftNavVisible;
+};
+
+
+/**
+ * @export
+ */
+app.MainController.prototype.toggleRightNavVisibility = function() {
+  this.rightNavVisible = !this.rightNavVisible;
+};
+
+
+/**
+ * Hide both navigation menus.
+ * @export
+ */
+app.MainController.prototype.hideNav = function() {
+  this.leftNavVisible = this.rightNavVisible = false;
+};
+
+
+/**
+ * @return {boolean} Return true if one of the navigation menus is visible,
+ * otherwise false.
+ * @export
+ */
+app.MainController.prototype.navIsVisible = function() {
+  return this.leftNavVisible || this.rightNavVisible;
+};
+
+
+/**
+ * @return {boolean} Return true if the left navigation menus is visible,
+ * otherwise false.
+ * @export
+ */
+app.MainController.prototype.leftNavIsVisible = function() {
+  return this.leftNavVisible;
+};
+
+
+/**
+ * @return {boolean} Return true if the right navigation menus is visible,
+ * otherwise false.
+ * @export
+ */
+app.MainController.prototype.rightNavIsVisible = function() {
+  return this.rightNavVisible;
+};
```

elemoine:
Yep, these are just accessors to the application-specific states. We could even remove these variables and access the states from `index.html`.

sbrunner:
OK thanks for your response :-)

elemoine:
> OK thanks for your response :-)

It is good to that you challenge us on having as much as code as possible in the library :)

elemoine:
Ok merging.

elemoine:
The merge of this branch broke the build. The problem is related to the new mobileapp example missing a description. I'll look into it.

PR #340, opened 2015-10-15
Generate API documentation

sbrunner:
View result: http://camptocamp.github.io/ngeo/jsdoc/apidoc/


asaunier:
Nice!
Is it possible to modify a bit the styles? It would be better if more space could be added in the right column after the "Classes" sections. Else it is a bit confusing to figure out what thoses classes are related to.

sbrunner:
@asaunier done :-)

elemoine:
There are still many problems to solve but this looks like a very good start!

asaunier:
Thanks, that looks much better :)

sbrunner:
> There are still many problems to solve but this looks like a very good start!

effectively :+1:

elemoine:
The merge of this branch broke the build. The problem is related to the new mobileapp example missing a description or something. I'll look into it.

elemoine:
https://camptocamp.github.io/ngeo/master/apidoc returns a 404. Is this expected?

sbrunner:
I don't have a 404 your are probably faster than pages :-)

elemoine:
Indeed.

PR #342, opened 2015-10-16
Add description to mobileapp example

elemoine:
The Travis build generated an index page for the examples. The index page generation script currently fails because the `mobileapp` gmf example does not have a description. This PR adds a description to the example.


elemoine:
Merging, and hoping that the build will pass this time.

PR #346, opened 2015-10-19
Get rid of more-or-less

pgiraud:
With this pull request, we use bootstrap instead of other third-party less mixins.


elemoine:
You should not have the "Add a GMF mobileapp example" commit in this PR.

elemoine:
Please merge when my comment is addressed.

PR #365, opened 2015-10-19
Add 'selectedLocation' scope property to gmf.locationchooserDirective

fredj:
A need to take a special action when a location is selected.

Comments welcome

### DIFF
--- a/contribs/gmf/examples/locationchooser.js
+++ b/contribs/gmf/examples/locationchooser.js
@@ -41,6 +41,13 @@ app.MainController = function() {
     extent: [12044030, -4921322, 17914393, -1007746]
   }];
 
+
+  /**
+   * @type {gmfx.LocationchooserLocation}
+   * @export
+   */
+  this.selectedLocation = this.locations[2];
+
   /**
    * @type {ol.Map}
    * @export
--- a/contribs/gmf/src/directives/locationchooser.js
+++ b/contribs/gmf/src/directives/locationchooser.js
@@ -12,7 +12,8 @@ goog.require('ol.Map');
  * @example
  * <gmf-locationchooser
  *   gmf-locationchooser-map="ctrl.map"
- *   gmf-locationchooser-locations="ctrl.locations">
+ *   gmf-locationchooser-locations="ctrl.locations"
+ *   gmf-locationchooser-selected="ctrl.selectedLocation">
  * </gmf-locationchooser>
  *
  * @return {angular.Directive} The directive specs.
@@ -24,8 +25,10 @@ gmf.locationchooserDirective = function() {
     restrict: 'E',
     scope: {
       'getMapFn': '&gmfLocationchooserMap',
-      'getLocationsFn': '&gmfLocationchooserLocations'
+      'getLocationsFn': '&gmfLocationchooserLocations',
+      'selectedLocation': '=gmfLocationchooserSelected'
     },
+    bindToController: true,
     controller: 'gmfLocationchooserController',
     controllerAs: 'ctrl',
     template: '<select ' +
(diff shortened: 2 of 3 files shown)


elemoine:
It would make more sense to add a scope property to me:

```js
    scope: {
      selectedLocation: '=gmfLocationchooserSelectedlocation'
    }
```

Through this scope property the directive can publish the currently selected location to its users. And users would be able to programmatically change the selected location.

Would that work for you?

fredj:
> Would that work for you?

Yes, very good idea. Thanks for the suggestion.

contribs/gmf/examples/locationchooser.js
```diff
@@ -41,6 +41,13 @@ app.MainController = function() {
     extent: [12044030, -4921322, 17914393, -1007746]
   }];
 
+
+  /**
+   * @type {gmfx.LocationchooserLocation}
```
elemoine: It looks like you have to use `{gmfx.LocationchooserLocation|undefined}` here. Or you could provide an initial location, as an illustration of using `gmf-locationchooser-selected` as an input mechanism.

contribs/gmf/src/directives/locationchooser.js
```diff
@@ -24,12 +25,13 @@ gmf.locationchooserDirective = function() {
...
     scope: {
       'getMapFn': '&gmfLocationchooserMap',
-      'getLocationsFn': '&gmfLocationchooserLocations'
+      'getLocationsFn': '&gmfLocationchooserLocations',
+      'selectedLocation': '=gmfLocationchooserSelected'
     },
     controller: 'gmfLocationchooserController',
     controllerAs: 'ctrl',
     template: '<select ' +
-        'ng-model="ctrl.selectedLocation" ' +
```
elemoine: You may want to use `bindToController`. In this way you can still use `ctrl.selectedLocation` here.

contribs/gmf/src/directives/locationchooser.js
```diff
@@ -72,11 +74,6 @@ gmf.LocationchooserController = function($scope) {
    */
   this.locations = locations;
 
-  /**
-   * @type {gmfx.LocationchooserLocation}
-   * @export
-   */
-  this.selectedLocation = this.locations[0];
```
elemoine: Don't you want to use something like this here (assuming you use `bindToController`)?

```js
/**
 * @type {gmfx.LocationchooserLocation}
 * @export
 */
this.selectedLocation;

if (!goog.isDef(this.selectedLocation)) {
  this.selectedLocation = this.locations[0];
}
```

elemoine:
@fredj I just made more suggestions. I hope they make sense.

fredj:
> I hope they make sense.

A lot of sense, thanks again

contribs/gmf/src/directives/locationchooser.js
```diff
@@ -76,7 +73,14 @@ gmf.LocationchooserController = function($scope) {
...
    * @export
    */
-  this.selectedLocation = this.locations[0];
+  this.selectedLocation;
+
+  if (!goog.isDef(this.selectedLocation)) {
+    this.selectedLocation = this.locations[0];
+  }
+
+  $scope.$watch('ctrl.selectedLocation', angular.bind(this, this.setLocation));
```
elemoine: We use `goog.bind` outside examples.

fredj: fixed

elemoine: Also, I'd recommend using:

```js
$scope.$watch(goog.bind(function() {
  return this.selectedLocation;
}, this), goog.bind(this.setLocation, this));
```

To avoid referring to `ctrl` from the js code.

contribs/gmf/src/directives/locationchooser.js
```diff
@@ -76,7 +73,14 @@ gmf.LocationchooserController = function($scope) {
...
    * @export
    */
-  this.selectedLocation = this.locations[0];
+  this.selectedLocation;
+
+  if (!goog.isDef(this.selectedLocation)) {
+    this.selectedLocation = this.locations[0];
+  }
+
+  $scope.$watch('ctrl.selectedLocation', goog.bind(this.setLocation, this));
```
elemoine: Did you see my comment suggesting to use a function instead of an expression as the first arg to `$watch`?

elemoine:
Please merge when my remaining comment is addressed. Thanks!

PR #368, opened 2015-10-23
Ensure that modal element is in body (ie. on top)

pgiraud:


elemoine:
Try this:

```js
angular.element(document.body).append(modal);
```

`angular.element` returns an `angular.JQLite`, so this should make the compiler happy.

src/directives/modal.js
```diff
@@ -47,6 +47,9 @@ ngeo.modalDirective = function($parse) {
          */
         function(scope, element, attrs) {
           var modal = element.children();
+          // move the modal to document body to ensure that it is on top of
+          // other elements even if in a positionned element initialy.
```
elemoine: positionned -> positioned and initialy -> initially

elemoine:
I think the docs should mention that the directive places the modal to the document body.

elemoine:
Please merge when you think this is ready.

elemoine:
Thanks for updating your branch. Please merge.

PR #371, opened 2015-10-28
Directive factories are not @constructor's

elemoine:
The primary consumer of annotations is the Closure Compiler. Directive factories are not constructors. And I don't think we should make the Closure Compiler believe they are constructors.

I know this is going to cause problems on the apidoc side, but I think we should address that problem in a different way.


fredj:
Do we want to keep the `goog.provide` for these controllers?

elemoine:
Yes, we need to provide the directives, for users to be able the associated code.

fredj:
LGTM

PR #372, opened 2015-10-28
Change ngeo-modal to use ngModelController

elemoine:
This changes the `ngeo-modal` directive to use `ngModelController`. With this `ngeo-modal` will be used together with `ng-model`. For example:

```html
<ngeo-modal ng-model="ctrl.modalShownHidden" ng-model-options="{getterSetter: true}">
  <!-- … -->
</ngeo-modal>
```

This makes the directive more flexible. For example it can be used with a getter/setter as the assignable expression (as shown in the example above).

Please review.


pgiraud:
This looks very good to me.
Please merge.

PR #375, opened 2015-11-02
Use htmlmin in templatecache.mako.js

elemoine:
`templatecache.mako.js` currently removes spaces and line breaks between HTML elements. This is not correct, and it may break the user's HTML code. This commit fixes the problem by using the `htmlmin` tool instead of minifying the HTML ourselves.

This is WIP in progress at the moment (being tested in https://github.com/Geoportail-Luxembourg/geoportailv3/pull/929).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -373,7 +373,7 @@ $(EXTERNS_JQUERY):
 # is done so ngeo.js works for the examples on github.io. If another key
 # pattern is needed this should be changed.
 .build/templatecache.js: buildtools/templatecache.mako.js .build/python-venv/bin/mako-render
-	.build/python-venv/bin/mako-render --var "partials=$(addprefix ../,$(SRC_DIRECTIVES_PARTIALS_FILES))" --var "basedir=src" $< > $@
+	PYTHONIOENCODING=UTF-8 .build/python-venv/bin/mako-render --var "partials=$(addprefix ../,$(SRC_DIRECTIVES_PARTIALS_FILES))" --var "basedir=src" $< > $@
 
 .build/apidoc-%: .build/node_modules.timestamp jsdoc.json $(SRC_JS_FILES)
 	rm -rf $@
--- a/Makefile
+++ b/Makefile
@@ -348,7 +348,7 @@ $(EXTERNS_JQUERY):
 	touch $@
 
 .build/python-venv/bin/mako-render: .build/python-venv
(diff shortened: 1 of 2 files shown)


sbrunner:
In general it looks good, I just don't completely like the quote replacement at (new) line 17...
Shouldn't be removed and use `repr` at line 36:
`$templateCache.put('${partial}', ${repr(_partials[partial])});`?

elemoine:
Why do you suggest to use `repr`? And why do you suggest to `repr` for the value and not for the key (`'${partial}'`)?

sbrunner:
Good  question just because I concentrate on the value :-)
+1 to do it on booth :-)

elemoine:
@sbrunner, `htmlmin.minify` returns a `unicode` object. So when we `repr` that object we get something like this: `u'unicode'`. Which is not what we want.

sbrunner:
Why it's not what we want?
I found one thing is not what we want is the accent encoding that we should use the evil eval to read it.
A other possibility is to use the `json.dumps` and `json.loads`...
what do you think?

elemoine:
I'd like to understand the problem first :)

elemoine:
> I found one thing is not what we want is the accent encoding that we should use the evil eval to read it.

I don't understand. Would you mind elaborating a bit? Thanks.

sbrunner:
Is it clearer with this?:
```
In [2]: "éàè"
Out[2]: '\xc3\xa9\xc3\xa0\xc3\xa8'

In [3]: repr("éàè")
Out[3]: "'\\xc3\\xa9\\xc3\\xa0\\xc3\\xa8'"

In [4]: '\\xc3\\xa9\\xc3\\xa0\\xc3\\xa8'
Out[4]: '\\xc3\\xa9\\xc3\\xa0\\xc3\\xa8'

In [5]: "éàè" == '\\xc3\\xa9\\xc3\\xa0\\xc3\\xa8'
Out[5]: False

In [7]: "éàè" == eval("'\\xc3\\xa9\\xc3\\xa0\\xc3\\xa8'")
Out[7]: True
```

elemoine:
I've just added a commit that set the `PYTHONIOENCODING` envvar to `UTF-8` when running `mako-render`. This prevents `mako-render` from failing when partials include non-ASCII characters.

elemoine:
@sbrunner, I am going to merge this when the Travis build passes. Please tell me if there are things I am still missing from your comments. Thanks.

PR #377, opened 2015-11-04
Add a gmfExportFeatures service

elemoine:
The `gmfExportFeatures` service is a function to be used for exporting a GPX or KML document. The service relies on GeoMapFish's `exportgpxkml` web service.

Providing an example would be good, but the `exportgpxkml` is not yet available on the GeoMapFish demo instance.


gberaudo:
LGTM

PR #378, opened 2015-11-04
Add an ngeoFilereader directive

elemoine:
This PR adds an `ngeoFilereader`  directive and an example showing how to use that directive to import features from a KML document read from the user's file system.

The directive is based on the [FileReader](https://developer.mozilla.org/en-US/docs/Web/API/FileReader) object, itself being part of the [File API](https://w3c.github.io/FileAPI/#FileReader-interface).

Works in Chrome, Firefox, Safari and IE > 9.

### DIFF
--- a/src/directives/filereader.js
+++ b/src/directives/filereader.js
@@ -1,0 +1,57 @@
+goog.provide('ngeo.filereaderDirective');
+
+goog.require('ngeo');
+
+
+/**
+ * This directive is to used on an input file element. When a file is selected
+ * the directive uses the browser `FileReader` API to read the file. The file
+ * content is provided to the directive user through the assignable expression.
+ * Only works for text file (`readAsText` used for reading the file).
+ *
+ * @example
+ * <input type="file" ngeo-filereader="ctrl.fileContent" />
+ *
+ * @param {angular.$window} $window The Angular $window service.
+ * @return {angular.Directive} Directive Definition Object.
+ * @ngInject
+ */
+ngeo.filereaderDirective = function($window) {
+  return {
+    restrict: 'A',
+    scope: {
+      'fileContent': '=ngeoFilereader',
+      'supported': '=ngeoFilereaderSupported'
+    },
+    link:
+        /**
+         * @param {angular.Scope} scope Scope.
+         * @param {angular.JQLite} element Element.
+         * @param {angular.Attributes} attrs Attributes.
+         */
+        function(scope, element, attrs) {
+          var supported = 'FileReader' in $window;
+          scope['supported'] = supported;
+          if (!supported) {
+            return;
+          }
+          element.bind('change', function(changeEvent) {
+            /** @type {!FileReader} */
+            var fileReader = new $window.FileReader();
+            fileReader.onload = (
+                /**
+                 * @param {!ProgressEvent} evt Event.
+                 */
+                function(evt) {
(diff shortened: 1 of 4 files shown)


src/directives/filereader.js
```diff
@@ -0,0 +1,54 @@
...
+ *
+ * @example
+ * <input type="file" ngeo-filereader="ctrl.fileContent" />
+ *
+ * @param {angular.$window} $window The Angular $window service.
+ * @return {angular.Directive} Directive Definition Object.
+ * @ngInject
+ */
+ngeo.filereaderDirective = function($window) {
+  goog.asserts.assert('FileReader' in $window);
```
elemoine: I need to change this, as the directive should not throw an exception when the browser doesn't support `FileReader`. Instead it should somehow report to the application code that the browser does not support the `FileReader` API required for the directive to do its job.

src/directives/filereader.js
```diff
@@ -0,0 +1,59 @@
...
+         */
+        function(scope, element, attrs) {
+          var supported = 'FileReader' in $window;
+          scope['supported'] = supported;
+          if (!supported) {
+            return;
+          }
+          element.bind('change', function(changeEvent) {
+            /** @type {!FileReader} */
+            var fileReader = new $window.FileReader();
```
fgravin: is filereader alway supported ?

fgravin:
Not sure about the fact that the directive set the fileReaderSupported property to its parent scope.
But it looks globally ok.

elemoine:
> Not sure about the fact that the directive set the fileReaderSupported property to its parent scope.

It does, through two-way binding (the `=` sign on the scope definition).

fgravin:
Yes i saw, but not sure about the purpose/design.
You're not even using it again if you have 2 times the `ngeoFilereader` directive in your app.

It's more oneway binding from directive to parent, as you don't use it in your directive.

elemoine:
> Yes i saw, but not sure about the purpose/design.

The goal is to be able to tell the application (parent scope) that the FileReader API is not supported and the directive cannot do its job.



> You're not even using it again if you have 2 times the ngeoFilereader directive in your app.

It is used in the example. See https://github.com/camptocamp/ngeo/blob/master/examples/importfeatures.html#L19-L20.

> It's more oneway binding from directive to parent, as you don't use it in your directive.

Yep. I said two-way binding because that's what the `=` sign is about in the directive definition.

elemoine:
Please tell me if I merged this too fast. I am happy to make changes if you have suggestions.

fgravin:
It's ok.

Other solutions would have been to set the `fileReaderSupported` in the parent controller and send it as read only to the directive.
If you use this directive multiple times in larger apps, what's would be the best design ? i think it is to set the `fileReaderSupported` value just once in a main controller, but it could be boring to pass it through all controllers and directive to reach your `ngeoFilereader` directive.
Maybe, as it is a global setting, put it in a `angular.value`, but it also have cons.

Anyway, you know all of these options, i don't really have an opinion about wich one is the best.

Note: In your case, you  must declare the variable in the parent controller even though you don't need it to avoid exception from angular.

elemoine:
> Note: In your case, you must declare the variable in the parent controller even though you don't need it to avoid exception from angular.

I thought so too, but I haven't been able to see exceptions.

PR #381, opened 2015-11-05
Add public getFormattedArea and getFormattedLength functions

elemoine:
This PR refactors the measure interaction code to add `getFormattedArea` and `getFormattedLength` static function and make them public for use from outside.

Ready for review.


fgravin:
Looks good.
Is there a way to override those 2 methods ?

elemoine:
> Is there a way to override those 2 methods ?

No, the goal here si just to make them public for other uses.

PR #382, opened 2015-11-09
Translate dropdown and placeholder

ger-benjamin:
This is a WIP, open for discussions.

 1. First, I had to add the ``gettext`` provider. Else, It throws an error because it doesn't find a provider for the ``translate`` keyword.

 ~~2. For the **locationchooser**, I had to change how I create the dropdown. ng-options doesn't work with "on-the-flight translation". So I've used a ``ng-repeat`` instead the ``ng-options``. The consequence is that I can't use an object in the "value" of the options. So now, I must use the extent as a String.~~

 3. I've tried my changes in a GeoMapFish. For the **search**, I've saw that the project doesn't add the translation in the "pot" file. I had to add the same string as in the placeholder in a hidden markup in the html file.

### DIFF
--- a/contribs/gmf/examples/locationchooser.js
+++ b/contribs/gmf/examples/locationchooser.js
@@ -63,7 +63,6 @@ app.MainController = function() {
       zoom: 4
     })
   });
-
 };
 
 
--- a/contribs/gmf/src/directives/locationchooser.js
+++ b/contribs/gmf/src/directives/locationchooser.js
@@ -31,9 +31,10 @@ gmf.locationchooserDirective = function() {
     bindToController: true,
     controller: 'gmfLocationchooserController',
     controllerAs: 'ctrl',
-    template: '<select ' +
-        'ng-model="ctrl.selectedLocation" ' +
-        'ng-options="location.label for location in ::ctrl.locations" ' +
+    template:
+        '<select ng-model="ctrl.selectedLocation" ' +
+        'ng-options="location as ctrl.translate(location.label) ' +
+        'for location in ctrl.locations" ' +
         'ng-change="ctrl.setLocation()">' +
         '</select>'
   };
@@ -44,6 +45,7 @@ gmfModule.directive('gmfLocationchooser', gmf.locationchooserDirective);
 
 
 /**
+ * @param {angularGettext.Catalog} gettextCatalog Gettext catalog.
  * @param {angular.Scope} $scope The directive's scope.
  * @constructor
  * @export
@@ -51,7 +53,13 @@ gmfModule.directive('gmfLocationchooser', gmf.locationchooserDirective);
  * @ngdoc controller
  * @ngname gmfLocationchooserController
  */
-gmf.LocationchooserController = function($scope) {
+gmf.LocationchooserController = function(gettextCatalog, $scope) {
+
+  /**
+   * @type {angularGettext.Catalog}
+   * @private
+   */
+  this.gettextCatalog_ = gettextCatalog;
 
   var map = this['getMapFn']();
   goog.asserts.assertInstanceof(map, ol.Map);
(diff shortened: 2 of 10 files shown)


contribs/gmf/src/directives/locationchooser.js
```diff
@@ -29,10 +29,12 @@ gmf.locationchooserDirective = function() {
...
     controllerAs: 'ctrl',
-    template: '<select ' +
-        'ng-model="ctrl.selectedLocation" ' +
-        'ng-options="location.label for location in ::ctrl.locations" ' +
+    template:
+        '<select ng-model="ctrl.selectedLocation" ' +
         'ng-change="ctrl.setLocation()">' +
+        '<option value="{{location.extent}}"' +
+        'ng-repeat="location in ctrl.locations">' +
+        '{{location.label | translate}}</option>' +
```
elemoine: If we decide to use the `translate` directive in GMF directive partials then the `gmf` Angular module should depend on the `gettext` module.

contribs/gmf/examples/locationchooser.js
```diff
@@ -43,10 +43,10 @@ app.MainController = function() {
 
 
   /**
-   * @type {gmfx.LocationchooserLocation}
+   * @type {string}
    * @export
    */
-  this.selectedLocation = this.locations[2];
+  this.selectedLocation = '[' + this.locations[2].extent + ']';
```
elemoine: What are these square brackets for?

contribs/gmf/examples/search.js
```diff
@@ -14,7 +14,7 @@ var app = {};
 
 
 /** @type {!angular.Module} **/
-app.module = angular.module('app', ['gmf']);
+app.module = angular.module('app', ['gmf', 'gettext']);
```
elemoine: See my other comment regarding `gettext` being a dependency of the `gmf` module.

contribs/gmf/examples/locationchooser.js
elemoine: Ok I think I understand. This is related to your comment in the issue description and the first "yellow" note in the Angular documentation for "select": https://docs.angularjs.org/api/ng/directive/select.

elemoine: But there must be a better way that using this (and `JSON.parse`).

elemoine:
> I've tried my changes in a GeoMapFish. For the search, I've saw that the project doesn't add the translation in the "pot" file. I had to add the same string as in the placeholder in a hidden markup in the html file.

It's related to the Makefile not considering the ngeo partials. If we decide that ngeo directives (or gmf directives) include translations that the extraction and translation should probably be done by ngeo itself. I hope you see what I mean.

ger-benjamin:
Yep, I understand. Thanks @elemoine

ger-benjamin:
I don't understand the Travis error. Someone has any clue ?

elemoine:
The build fails because the `check-example.js` PhantomJS script gives errors on the `locationchooser` example. This means that your example won't work when using `gmf.js` on github.io. You can reproduce the problem in your browser by opening http://localhost:3000/.build/examples-hosted/contribs/gmf/locationchooser.html.

elemoine:
The problem may be related to `angular-gettext.js` not being available. Look at what's done for other external libraries in the `Makefile` (d3 for example). You'll need to mimic that for `angular-gettext.js`.

ger-benjamin:
Thanks @elemoine

asaunier:
See also https://github.com/rubenv/angular-gettext/issues/58

ger-benjamin:
Thanks @asaunier but watch out: I've tried with Angular v1.4.7 and that doesn't work. Then I tried with your (older) v1.3.5 and that works... I don't currently know why...

ger-benjamin:
@asaunier Now it works with your solution. With Angular v1.4+ I've should remove the one-time binding on my 'for'.
 - Before: for location in ctrl.locationsfor location in **::ctrl.locations**
 - Now: for location in ctrl.locationsfor location in ctrl.locations

asaunier:
Good news! Thanks :)

sbrunner:
Can this pullrequest be merged ?

ger-benjamin:
For me, yes.

sbrunner:
=> #415 

Looks good, merge

PR #383, opened 2015-11-10
Gmf-mobileapp - layout with basics controls

ger-benjamin:
As the layout was defined. Even f.i. for the geolocation symbol (we can change it later)


contribs/gmf/examples/style/mobileapp-map.less
```diff
@@ -7,3 +18,162 @@ gmf-map {
...
+
+.nav-right-trigger {
+  margin-left: 77px;
+  &:after {
+    content: 'd';
+  }
+}
+
+//For tablet only
+@media (min-width: @screen-xs-max) {
```
tonio: You shouldn't mix min/max when using media queries & variables (use min-width with min values).
@screen-xs-max: 767px
@screen-sm-min: 768px
See https://github.com/twbs/bootstrap/blob/master/less/variables.less#L314-L317

tonio:
After a quick look, it seems styles could be less duplicated, some rules are repeated multiple times.
Hardcoded repeated values (ex: 13px) should be set in variables.

ger-benjamin:
Yep, I was thinking on that. Thanks @tonio I'll do some changes

contribs/gmf/examples/style/mobileapp-vars.less
```diff
@@ -1,8 +1,13 @@
+@app-margin: 13px;
```
pgiraud: It looks like this value is empirical and was determined to make it so the geolocation and + / - buttons are aligned, isn't it?

contribs/gmf/examples/style/mobileapp-map.less
```diff
@@ -7,3 +42,144 @@ gmf-map {
...
+    margin: 0 10px;
+    padding: 0 5px;
+    border-width: 0 1px;
+    border-radius: 0;
+  }
+}
+
+.nav-trigger {
+  top: @app-margin;
+  left: 50vw;
```
pgiraud: Is it really useful? It looks like it has no influence.

contribs/gmf/examples/style/mobileapp-map.less
```diff
@@ -7,3 +42,144 @@ gmf-map {
...
+  margin-right: auto;
+  padding: 6px 10px;
+  background-color: @map-tools-bg-color;
+  width: 162px + @map-tools-size + @map-tools-size;
+  border-radius: 25px;
+  border: solid 1px @map-tools-color;
+  z-index: @content-index;
+  input {
+    width: 150px;
+    margin: 0 10px;
```
pgiraud: Doesn't seem to have any influence.

contribs/gmf/examples/style/mobileapp-map.less
```diff
@@ -7,3 +42,144 @@ gmf-map {
...
+    padding: 0 5px;
+    border-width: 0 1px;
```
pgiraud: Using the input border to display the separators is a bit hackish to me, but I can leave with it for now.

contribs/gmf/examples/style/mobileapp-map.less
```diff
@@ -7,3 +42,144 @@ gmf-map {
...
+@media (min-width: @screen-sm-min) {
+  .ol-zoom {
+    top: auto;
+    bottom: 36px;
+  }
+
+  .geolocation {
+    top: auto;
+    bottom: 0px;
+    margin-bottom: 10px;
```
pgiraud: Can't we use `bottom` only? I don't think that `margin-bottom` is useful. Also `top` doesn't need to be set if you set `bottom`.

contribs/gmf/examples/style/mobileapp-map.less
```diff
@@ -7,3 +42,144 @@ gmf-map {
...
+  .nav-trigger {
+    top: @app-margin;
+    margin: 0;
+    background-color: @map-tools-bg-color;
+    border: solid 1px @map-tools-color;
+  }
+
+  .nav-left-trigger {
+    left: @app-margin;
+    right: auto;
```
pgiraud: `right` is not required here.

contribs/gmf/examples/style/mobileapp-map.less
```diff
@@ -7,3 +42,144 @@ gmf-map {
...
+  }
+
+  .nav-right-trigger {
+    left: auto;
```
pgiraud: `left` is not required here.

contribs/gmf/examples/style/mobileapp-vars.less
ger-benjamin: Yep. Based on the "+" and "-" button of Openlayers.

contribs/gmf/examples/style/mobileapp-map.less
ger-benjamin: ``top: auto`` is required by chrome at least

contribs/gmf/examples/style/mobileapp-map.less
ger-benjamin: ``left: auto`` is required by chrome at least

contribs/gmf/examples/style/mobileapp-vars.less
pgiraud: Let's remove the geolocation button for now, and re-add it later when functional. I'd really prefer the zoom and geolocation buttons to be displayed using the same classes.

contribs/gmf/examples/style/mobileapp-map.less
pgiraud: This is because you already set "top: 137px" for all medias.
I think it would be better to add a new media query for media with `max-width: @screen-sm-max`.

contribs/gmf/examples/style/mobileapp-vars.less
ger-benjamin: > Let's remove the geolocation button for now,  and re-add it later when functional.

That make sens. I'll remove the html "geolocation" button.

>I 'd really prefer the zoom and geolocation buttons to be displayed using the same classes.

I don't think that this is a good idea to link somehow two completely different functionalities :-/

pgiraud: What I mean is that the geolocation button should be treated as a map control. Maybe an extension of `ol.control`.
This would prevent errors if the OpenLayers CSS classes change.
An other option is to get rid of [+] and [-] buttons and manage them directly in our application.

ger-benjamin: Okay, I understand what you mean.

ger-benjamin:
Thanks for the review @pgiraud 
Can I merge this or someone sees something that must be changed now ?

pgiraud:
Please merge. We'll do ajustments later.

PR #384, opened 2015-11-11
Can pass options to configure Typeahead

ger-benjamin:
My initial aim was to set easily the fetch limit of Typeahead but I've also changed the manner to configure Bloodhound.

### DIFF
--- a/contribs/gmf/externs/gmfx.js
+++ b/contribs/gmf/externs/gmfx.js
@@ -15,15 +15,25 @@ var gmfx;
 /**
  * Datasource configuration options for the search directive.
  * @typedef {{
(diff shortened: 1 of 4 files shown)


contribs/gmf/externs/gmfx.js
```diff
@@ -27,6 +29,13 @@ gmfx.SearchDirectiveDatasource;
 
 
 /**
+ * The optional configuration for bloodhound for this set of data.
```
elemoine: You can change this to "Optional Bloodhound options for this data set." or "The optional Bloodhound configuration for this data set.". A link to the Bloodhound documentation would be good.

fredj:
LGTM

PR #394, opened 2015-11-18
Don't use the example jsdoc tag but indent the code

fredj:
"The @example tag is not intended to be used to generate 'inline' examples"
From: https://code.google.com/p/jsdoc-toolkit/wiki/TagExample


sbrunner:
Looks good, merge

Issue #345, opened 2015-10-19
[apidoc] Use tag to split or mark the angular components types?


sbrunner:
* Use the tag `@ngdoc` defied by `angular-jsdoc` https://github.com/allenhwkim/angular-jsdoc#tags-available

* Takes car about it in `buildNav`: https://github.com/camptocamp/ngeo/blob/master/jsdoc-template/publish.js#L341
  To group by angular types or add a mark.

elemoine:
Could you please be more specific?

sbrunner:
Description updated :-)

PR #400, opened 2015-11-19
Ngdoc - name default template

sbrunner:
Add the @ngdoc tag
Set the @name to the <package>.<angular name>
Use the standard template

PR #402, opened 2015-11-20
Add ngdoc and ngname tags

sbrunner:
Display angular types and names

alternative of #400


sbrunner:
Result: http://camptocamp.github.io/ngeo/Angular-JSDoc2/apidoc/

asaunier:
Same question than for #400 
I prefer this white background layout.

sbrunner:
And on this one the (more) links will works +1 for this one

ger-benjamin:
I prefer this layout too.

sbrunner:
fix #345

fredj:
+1

sbrunner:
The most important difference of those pull requests is not the layout but the generated doc.

In the #400 we are based on the angular-template witch is made for angular object but don't support wheel standard object... see: http://camptocamp.github.io/ngeo/Angular-JSDoc/apidoc/ngeo.interaction.Measure.html vs http://camptocamp.github.io/ngeo/Angular-JSDoc2/apidoc/ngeo.interaction.Measure.html

And in #402 the link generation is much better :-)

PR #406, opened 2015-11-23
Set search's rateLimitWait to 50ms

ger-benjamin:
Wait 50ms instead of the default 300ms on each request (~speedup the search).
https://github.com/twitter/typeahead.js/blob/master/doc/bloodhound.md#remote


fredj:
LGTM

PR #410, opened 2015-11-24
Refactor the demo mobile apps

sbrunner:
to be able to include it in c2cgeoportal

* Move the css, less, fonts outside the app (we needs them in the client application)
* Rename main to mobile (it should be different as in the desktop application)
* Move the nav to gmf
* Merge main.js and app.js in mobile.js for simplification
* Move search's urls in a ServerVars constant


ger-benjamin:
Not tested but seems alright

fredj:
LGTM too

sbrunner:
Thanks :-)

PR #411, opened 2015-11-25
Some backport from c2cgeoportal

sbrunner:


ger-benjamin:
Not sur for the Makefile, else it seems ready to merge.

PR #412, opened 2015-11-25
Add a Management for tools to activate

ger-benjamin:


sbrunner:
Except that the event methods signature should be more precise, that looks good, fix it and merge :-)

ger-benjamin:
Done. Thanks @sbrunner

PR #413, opened 2015-11-26
[WIP] Fix the major minor version

sbrunner:
This pull request is to open a discussion about the version.

As usual the version will be organize as: `major.minor.patch` actually we mostly use the `^version` (who fix the major version and minimum version), but sometimes the `^version` (who fix the major and minor version and minimum version) and `version` who fix everything.

Do we already have a reflection about this? I think that we should have a general politic about that ...


fredj:
For information, in ol3 we always use a strict version `"foo": "x.y.z"`

fredj:
IMHO we should use the strict mode and set exactly every version

PR #418, opened 2015-11-30
Add ngdoc and ngname in extra_annotation_name

fredj:


sbrunner:
Looks good, merge :-)

PR #428, opened 2015-12-03
[WIP] gmf-mobile-background-layer-selector

adube:
This is an early PR of the gmf-mobile-background-layer-selector directive, which is still under development.

I'm bringing this right now to the front in order to raise and issue I'm currently having with the `ngeo.BackgroundLayerMgr`.  The way it currently work is this: only one background layer gets added at a time to the map, i.e. the other layers are removed from the map.  This works fine.  The problem I'm having is that I need to detect which one is currently the base layer.  At first, I thought of using the `layer.getVisible()` property as a way to know that, but all background layers, regardless of if they are in the map or not, are always visible.  I'm doing this inside an Angular expression to add/remove some CSS for the GMF MobileBackgroundLayerSelector, which that looks like this:

      '    <div ng-class="layer.getVisible() ? ',
       '        \'gmf-check gmf-check-active\' : \'gmf-check\'"></div>',

Is there an other property I can use to know, for each background layer, if it is the current one being active ?  To me, using the visible property of the layer seems like a simple solution.  Thoughts ?

Oh, and feel free to comment/review what's currently in progress too, if you want.

fix: https://github.com/camptocamp/c2cgeoportal/issues/1654


adube:
On a related subject, this PR includes pieces of code that come from the Luxembourg Geoportal, such as  the theme, layerfactory and blanklayer services.  Those were 99% kept 'as-is'.

In the example, I use:

http://geomapfish-demo.camptocamp.net/2.0/wsgi/themes?'version=2&background=background

The layers returned are correctly added to the map, but they don't work.  I receive `400 (Bad Request)` errors.  Just raising this...

fgravin:
> Is there an other property I can use to know, for each background layer, if it is the current one being active ?

Have you tried https://github.com/camptocamp/ngeo/blob/master/src/services/backgroundlayermgr.js#L104

adube:
>Have you tried https://github.com/camptocamp/ngeo/blob/master/src/services/backgroundlayermgr.js#L104

No.  What would be the syntax of the Angular expression ?

fgravin:
something like `<div ng-class="{active: ctrl.ngeoBackgroundLayerMgr.get(ctrl.map) == layer}"`
Be sure of the name of your control, and that the bglayer service is in it.

adube:
I'll will try that.  Thanks a bunch.

adube:
@fgravin Your suggestion worked greatly.  I updated the PR and made some trivial changes following up our verbal discussions.  Thanks!

adube:
Here's a summary of the work currently in progress, i.e. what's remaining to do:

 * the WMTS layers are not showing any tile for me. The widget itself seems to work.  Can this be related to what the service replies ?
 * use a "checkmark" icon instead of a green box for the selected layer
 * add the widget in the "mobile" app
 * validate the components that were imported from Luxembourg Geoportal and how they are used in GMF.
 * fix Travis issue

@fgravin You can proceed with your "early" review, as lots of what's left to do may benefit from your input.

adube:
The last changes the inner `template` definition into a `templateUrl` instead.  I'm not entirely sure if everything's okay, here.  If I don't define the url in the example, then the default one isn't working because the path is incorrect.

adube:
@fgravin What is the way to pre-load templates in the template cache ?  That seems to be the reason why Travis is failing.

sbrunner:
if you do `make examples-hosted` does the example build in 
`.build/examples-hosted/contribs/gmf/mobilebackgroundlayerselector.html`
is working?

adube:
`make dist` currently fails:

    ERR! compile /opt/ngeo/adube/src/services/print.js:351: ERROR - actual parameter 1 of Object.keys does not match formal parameter
    ERR! compile found   : (Object|null)
    ERR! compile required: Object
    ERR! compile   var dimensionKeys = Object.keys(dimensions);
    ERR! compile                                   ^
    ERR! compile 
    ERR! compile 
    ERR! compile 1 error(s), 0 warning(s), 
    ERR! compile 97.4% typed
    ERR! compile 

adube:
The error comes from 9314ddf5.  @fredj ?

sbrunner:
`npm install --force` should fix it :-)

adube:
>npm install --force should fix it :-)

It did.  Thanks.

contribs/gmf/src/directives/mobilebackgroundlayerselector.js
```diff
@@ -0,0 +1,131 @@
...
+   * @private
+   */
+  this.backgroundLayerMgr_ = ngeoBackgroundLayerMgr;
+  gmfThemes.getBgLayers().then(goog.bind(
+      /**
+       * @param {Array.<ol.layer.Base>} bgLayers Array of background
+       *     layer objects.
+       */
+      function(bgLayers) {
+        this['bgLayers'] = bgLayers;
```
fgravin: Seems to can get rid of the `[]` annotation for scope values with the `@export` annotation ?
See https://github.com/fgravin/ngeo/blob/guidelines/docs/guidelines.md#property-renaming

Can you give it a try ?

fgravin:
I think this PR need to focus on the gmf background selector.
The input data of your directive has to be an array of layers, that's it.

Everything related to theme service and layer factory should be moved to another PR.
Also, i don't think you can design all of this alone by yourself, those are key components of the application and we need to think of how this gonna work.

Luxembourg work on this is a great start indeed.

adube:
>I think this PR need to focus on the gmf background selector.
The input data of your directive has to be an array of layers, that's it.

I agree.

>Everything related to theme service and layer factory should be moved to another PR.
Also, i don't think you can design all of this alone by yourself, those are key components of the application and we need to think of how this gonna work.
>
>Luxembourg work on this is a great start indeed.

How do you suggest we proceed next ?

adube:
@fgravin now using properties with `@export` to avoid using `this['myProperty']`, as suggested.

fgravin:
I've made a separated PR for the themeservice.
I have removed the logic in getBgLayer method, it is just to provide you the service and the interface.

Rebase your work on https://github.com/camptocamp/ngeo/pull/465/files when it's merged.
Also, remove blankLayer service and layerfactories from your code (for now).
Another layerHelper service is being made (maybe we put back the blanklayer later).

Thanks

adube:
Understood.  I'll stay on watch.

adube:
Here's a summary of the work in progress.  I 'rebased' and master, then tried to chain one request per background layer received and create a layer for each one, from there.  I currently have a

    No 'Access-Control-Allow-Origin' header is present on the requested resource.

error.  I'm currently blocked by this and require help.  Letting me know if the approach I took is good would also be nice, @fgravin ?

Thanks.

sbrunner:
@adube from where do you do the request (what's the referrer)?

adube:
@sbrunner, please tell me if this is what you're looking for:

    Accept:application/json, text/plain, */*
    Accept-Encoding:gzip, deflate, sdch
    Accept-Language:fr-FR,fr;q=0.8,en-US;q=0.6,en;q=0.4
    Cache-Control:no-cache
    Connection:keep-alive
    Host:ows.asitvd.ch
    Origin:http://localhost:3000
    Pragma:no-cache
    Referer:http://localhost:3000/contribs/gmf/examples/mobilebackgroundlayerselector.html
    User-Agent:Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/47.0.2526.80 Safari/537.36

sbrunner:
> Origin:http://localhost:3000

Then for `curl -I "http://geomapfish-demo.camptocamp.net/2.0/wsgi/themes?version=2&background=background" -H Origin:http://localhost:3000` I have to following headers:
```
HTTP/1.1 200 OK
Date: Mon, 14 Dec 2015 13:30:24 GMT
Server: Apache/2.4.10 (Debian)
Cache-Control: max-age=864000, public
Access-Control-Allow-Origin: *
Access-Control-Allow-Methods: HEAD, GET, POST, PUT, DELETE
Content-Length: 15403
Connection: close
Content-Type: application/json; charset=UTF-8
```
For me it should be OK ...

fgravin:
`Access-Control-Allow-Origin: *`
Is the important one here

adube:
The problem doesn't come from the request to get the themes/background layers.

The background layers return an WMTS GetCapabilities urls.  Those are the one that fail.  Example:

>XMLHttpRequest cannot load http://ows.asitvd.ch/wmts/1.0.0/WMTSCapabilities.xml. No 'Access-Control-Allow-Origin' header is present on the requested resource. Origin 'http://localhost:3000' is therefore not allowed access.

sbrunner:
Effectively I just see that..., in Openlayers 2 we can configure a proxy does it steal be the case in ol3?

fgravin:
Do we need to do a getCapabilities for background layers ?

contribs/gmf/src/services/themesservice.js
```diff
@@ -51,6 +56,12 @@ gmf.Themes = function($http, treeUrl, isThemePrivateUrl) {
   this.$http_ = $http;
 
   /**
+   * @type {angular.$q}
+   * @private
+   */
+  this.q_ = $q;
```
fgravin: Keep the `$` for Angular services

sbrunner:
@adube for the layer on `ows.asitvd.ch` they can't actually work, tor the layers on `geomapfish-demo.camptocamp.net` I just change the URL and it shouldn't have any cross origin issue.

contribs/gmf/src/services/themesservice.js
adube: Please, let me know if I'm mistaken.  See:

https://google.github.io/styleguide/angularjs-google-style.html

Look under "2 Angular Style Rules", Compare the "yes" and "no" there.

>Do not use $ to prepend your own object properties and service identifiers. Consider this style of naming reserved by AngularJS and jQuery.

>Why? It's useful to distinguish between Angular / jQuery builtins and things you add yourself. In addition, $ is not an acceptable character for variables names in the JS style guide.

Do I understand correctly that we shouldn't call our inner variables with a `$` for angular objects ?

sbrunner:
@fgravin Yes in `WMTS/REST` we needs to know many things that com from the capabilities do do the query, and it's the client who do that...

adube:
>I just change the URL and it shouldn't have any cross origin issue.

I still get the error.

contribs/gmf/src/services/themesservice.js
fgravin: > Do I understand correctly that we shouldn't call our inner variables with a $ for angular objects

Nope. It says that when we create your own service, we should'nt use `$` like for example `$ngeoBgLayerManager`, our prefix is `ngeo`, `$` is the angular prefix.
But here, the variable is a reference to the angular Service, so for a better understanding, we should keep the `$` prefix on the variable name.
Is it clear ?

contribs/gmf/src/services/themesservice.js
```diff
@@ -103,24 +120,51 @@ gmf.Themes.findTheme_ = function(themes, themeName) {
...
-       * @return {Array.<Object>} Array of background layer objects.
+       * @return {angular.$q.Promise}
        */
       function(data) {
-        var bgLayers = data['background_layers'].map(goog.bind(function(item) {
+        var configs = data['background_layers'];
+        this.backgroundLayerConfigs_ = configs;
+        var promises = configs.map(goog.bind(function(item) {
+          return this.$http_.get(item.url, {
+            cache: false
```
fgravin: why `false` ?

contribs/gmf/src/services/themesservice.js
adube: Yes.  Thanks.

adube:
@sbrunner Would you please explain again what I need to do to avoid the "same origin policy" error ?  Thanks.

contribs/gmf/src/services/themesservice.js
adube: I suppose we want this kind of request to always receive a "fresh" response.  After being logged out, if a user refreshes the page, we want this request to always be made and not use the cache that would return "the user is connected" when he's not.

Does that make sense ?

sbrunner:
1. Does your request be on http://geomapfish-demo.camptocamp.net/2.0/tiles/1.0.0/WMTSCapabilities.xml?
2. Do you have the header `access-control-allow-origin:"*"`?

If yes you shouldn't have a "same origin policy" on this layer your issue should com from somewhere else ...

contribs/gmf/src/services/themesservice.js
fgravin: > Does that make sense ?

Yes but, 
- cache is reset on page reload
- if you have 10 bg layers from same service, we don't want 10 request on init

You are right because the theme tree is loaded if you connect and disconnect, at this time, we want a fresh response.
Leave `false`, i will do some work on top of it afterwards too optimizes regarding the cases.

contribs/gmf/src/services/themesservice.js
```diff
@@ -103,24 +120,51 @@ gmf.Themes.findTheme_ = function(themes, themeName) {
...
+          });
+        }, this));
+        return this.q_.all(promises);
```
fgravin: `$q.all` could be a good approach, but i'm not sure that the callback function of your promise will have the array of background layers as its argument.

adube: That's why I keep a reference, i.e.:

    this.backgroundLayerConfigs_ = configs;

Does that make sense ?

adube:
@sbrunner 

I use this url: http://geomapfish-demo.camptocamp.net/2.0/wsgi/themes?version=2&background=background

Please, look at its response.  You'll see that there are 2 GetCapabilities urls that still point to http://ows.asitvd.ch/, thus I still get errors.

Could this please removed or replaced ?

sbrunner:
Why should I remove them?
One capabilities error shouldn't block the others layers...

contribs/gmf/src/services/themesservice.js
fgravin:
```js

var deferred = $q.defer();
this.$http_.get(item.url.then(function() { deferred.resolve(bgLayer);}, function() {defer.resolve();})
return deferred.promise;
```

adube:
FMI - https://github.com/ger-benjamin/ngeo/blob/gmf-layertree/src/services/layerHelper.js#L86

adube:
@sbrunner I got the background layer selector partially working.  I get the following errors when attempting to fetch tiles such as:

 * https://tiles8.geomapfish-demo.camptocamp.net/1.0.0/map/default/swissgrid_05/1/1/0.png

error:

    GET https://tiles8.geomapfish-demo.camptocamp.net/1.0.0/map/default/swissgrid_05/1/1/0.png net::ERR_INSECURE_RESPONSE

Probably because the url uses an `https` url where I'm on a simple `http` one.

@sbrunner Also, please note that this WMTS server is in `EPSG:21781` projection.  I had to include it in my example to make it work, and the tiles are currently reprojected in `EPSG:3857`.  The actual projection of the map (of its view) itself should be discussed in an other task, I think.  For the demo, it's okay.

adube:
@fgravin It worked well with `deferred`.  If you want to check at the code and review again, it would be nice.

It's quite a long `chain` of callbacks and I want to make sure it's easy enough to follow its logic, at least for you guys.

sbrunner:
> Also, please note that this WMTS server is in `EPSG:21781` projection.  I had to include it in my example to make it work, and the tiles are currently reprojected in `EPSG:3857`.

The Map projection should be changed to `EPSG:21781`, we actually use  `EPSG:3857` just to have the OSM background that should also be removed.

sbrunner:
I just change the capabilities to use the http protocol -> no security issue on localhost, but I don't know for later use ...

adube:
>I just change the capabilities to use the http protocol -> no security issue on localhost, but I don't know for later use ...

It's working now.  Thanks.

adube:
Here's a summary of the work in progress:

 * The theme service now supports WMTS
 * The theme service doesn't support **WMS** yet.  Please, **update the demo theme service response** to return one to be able to develop on top of it.
 * I don't know what the theme service `isThemePrivateUrl` property does.  Should I worry about this as part of this task ?
 * I need to include the tool in the **mobile app**.  To do so, I'll change its projection to **epsg:21781** to fit what the demo theme service currently returns as background layer.
 * Travis errors need to be fixed.

sbrunner:
> I don't know what the theme service `isThemePrivateUrl` property does.  Should I worry about this as part of this task ?

where do you see that ?

adube:
>> I  don't know what the theme service isThemePrivateUrl property does. Should I worry about this as part of this task ?

>where do you see that ?

Here: https://github.com/camptocamp/ngeo/pull/428/files#diff-b6bb5b6f78464fc5d679b1b3126e9449R70

sbrunner:
@fgravin do you know what this `isThemePrivateUrl`is?
we don't have something like it in c2cgeoportal ...

adube:
@sbrunner The GMF mobile application no longer works for me, on `master` branch.  Would you please look into that ?

sbrunner:
Same for me, I should revert this to make it working again:
https://github.com/camptocamp/ngeo/pull/468
@fgravin Do you have an idea about this?

fgravin:
I'll check that tomorow morning.
I've made a fix lately that fixed it, but maybe something is still broken.

> @fgravin do you know what this isThemePrivateUrlis?
> we don't have something like it in c2cgeoportal ...

Yes, but I'll remove it if we don't have private urls in gmf.
I will also remove some code if we cannot set the theme service url in the geoportal page request parameters.

contribs/gmf/src/services/themesservice.js
```diff
@@ -103,24 +114,76 @@ gmf.Themes.findTheme_ = function(themes, themeName) {
...
       /**
        * @param {gmf.ThemesResponse} data The "themes" web service response.
-       * @return {Array.<Object>} Array of background layer objects.
+       * @return {angular.$q.Promise}
        */
       function(data) {
-        var bgLayers = data['background_layers'].map(goog.bind(function(item) {
-          goog.asserts.assert('name' in item);
-          goog.asserts.assert('imageType' in item);
+        var promises = data['background_layers'].map(goog.bind(function(item) {
```
fgravin: `Array.protoype.map` accepts a `thisArg` as second argument, maybe we can get rid of the use of `goog.bind` ?

adube: Done.

adube:
Here's a new summary of the work in progress.  Please, disregard any previous one.

Here's a summary of the work in progress:

 * The GMF abstract mobile controller now uses the theme service to get its layer.  The map is also, from now on, in **epsg:21781** to fit the only available layer returned by the service.
 * The theme service doesn't support WMS yet. Please, update the demo theme service response to return one to be able to develop on top of it.
 * I need a **checkmark** icon or symbol.
 * There are Travis errors that need to be fixed.

sbrunner:
> There are Travis errors that need to be fixed. I would like help about this. @fgravin ?

Shouldn't you also do this change?
https://github.com/camptocamp/ngeo/pull/478/files#diff-b67911656ef5d18c4ae36cb6741b7965R444

adube:
>Shouldn't you also do this change?
>https://github.com/camptocamp/ngeo/pull/478/files#diff-b67911656ef5d18c4ae36cb6741b7965R444

Why not fix this right away to have the master branch clean ?

sbrunner:
=>  #479

adube:
New summary of the work in progress.  Please, disregard any previous one.

Here's a summary of the work in progress:

 * The GMF abstract mobile controller now uses the theme service to get its layer.  The map is also, from now on, in **epsg:21781** to fit the only available layer returned by the service.
 * The theme service doesn't support WMS yet. Please, update the demo theme service response to return one to be able to develop on top of it.
 * I need a **checkmark** icon or symbol.
 * The Travis error is gone.

contribs/gmf/examples/mobilebackgroundlayerselector.js
```diff
@@ -0,0 +1,58 @@
...
+/** @type {!angular.Module} **/
+app.module = angular.module('app', ['gmf']);
+
+
+app.module.constant(
+    'treeUrl',
+    'https://geomapfish-demo.camptocamp.net/2.0/wsgi/themes?' +
+        'version=2&background=background');
+
+app.module.constant(
```
fgravin: To remove after https://github.com/camptocamp/ngeo/pull/478/files is merged

contribs/gmf/src/controllers/abstractmobile.js
```diff
@@ -31,16 +33,31 @@ goog.require('ol.style.Style');
 gmfModule.constant('isMobile', true);
 
 
+gmfModule.constant(
+    'treeUrl',
+    'https://geomapfish-demo.camptocamp.net/2.0/wsgi/themes?' +
+        'version=2&background=background');
```
fgravin: We need to think about that @sbrunner. Those constants will be generated on load in the mako template (within c2cgeoportal).
Should we rather put those definitions in the `index.html` file of gmf contrib as well ?

contribs/gmf/src/controllers/abstractmobile.js
```diff
@@ -95,14 +112,12 @@ gmf.AbstractMobileController = function(ngeoFeatureOverlayMgr, serverVars) {
...
-      new ol.layer.Tile({
-        source: new ol.source.OSM()
-      })
-    ],
+    layers: [],
     view: new ol.View({
-      center: [0, 0],
-      zoom: 2
+      center: [632464, 185457],
+      projection: ol.proj.get('epsg:21781'),
```
fgravin: How did you choose that ? I'm not sure about this as well, this should be overwritten by config some how.

contribs/gmf/src/directives/mobilebackgroundlayerselector.js
```diff
@@ -0,0 +1,133 @@
...
+};
+
+
+/**
+ * @param {ol.layer.Base} layer Layer.
+ * @export
+ */
+gmf.MobileBackgroundLayerSelectorController.prototype.setLayer = function(
+    layer) {
+  this['bgLayer'] = layer;
```
fgravin: remove brakets

contribs/gmf/src/services/themesservice.js
```diff
@@ -103,24 +114,76 @@ gmf.Themes.findTheme_ = function(themes, themeName) {
...
-        var bgLayers = data['background_layers'].map(goog.bind(function(item) {
-          goog.asserts.assert('name' in item);
-          goog.asserts.assert('imageType' in item);
+        var promises = data['background_layers'].map(function(item) {
+          var deferred = this.$q_.defer();
+          this.$http_.get(item.url, {
+            cache: false
+          }).then(
+              goog.bind(
+                  function(deferred, item, response) {
```
fgravin: We gonna use https://github.com/camptocamp/ngeo/pull/444/files#diff-4e83f2b133c079468a24cd359e39a6e5R96 in the end.

contribs/gmf/src/services/themesservice.js
```diff
@@ -103,24 +114,76 @@ gmf.Themes.findTheme_ = function(themes, themeName) {
...
-          //return layer;
-        }, this));
+                    deferred.resolve(layer);
+                  },
+                  this,
+                  deferred,
+                  item
+              ),
+              goog.bind(function(deferred) {
+                deferred.resolve(null);
```
fgravin: `null` is useless

contribs/gmf/src/controllers/abstractmobile.js
```diff
@@ -95,14 +112,12 @@ gmf.AbstractMobileController = function(ngeoFeatureOverlayMgr, serverVars) {
...
+      minZoom: 3,
+      zoom: 3
```
fgravin: @sbrunner we will need to use contant for view settings as well (projection, center, zoom)

sbrunner: Right, I think that I can have a look later ...

adube:
Github IO pages:

 * the [example page](http://adube.github.io/ngeo/gmf-background-selector-mobile/examples/contribs/gmf/mobilebackgroundlayerselector.html?map_zoom=3&map_x=640078&map_y=198984)
 * the [mobile app](http://adube.github.io/ngeo/gmf-background-selector-mobile/examples/contribs/gmf/apps/mobile/index.html)

adube:
In the example page, the img thumbnail works.  In the mobile app, it doesn't.  I don't know why.  Any clue ?

Otherwise, this is ready for merge.

fgravin:
Try instead to debug the img url, i think they are wrong in the mobile app.

```html
src="{{layer.get('metadata')['thumbnail']}}"
```

fgravin:
@adube we need a blank layer as well (it's done in Lux but not sure we want a `angular.value` just for that).

adube:
Should the server return an entry for the blank background layer, or do we always create it ?

sbrunner:
I think that simpler to do it client side only ...

adube:
This task is ready for review.  Here's a wrap-up.  You can ignore everything noted before.

**Examples on GitHub.io**

 * [mobile bgLayer sel. in examples](http://adube.github.io/ngeo/gmf-background-selector-mobile/examples/contribs/gmf/mobilebackgroundlayerselector.html)
 * [mobile bgLayer sel. in app](http://adube.github.io/ngeo/gmf-background-selector-mobile/examples/contribs/gmf/apps/mobile/index.html)

The only thing remaining to do is use a "checkmark" icon instead of the green box for the selected background layer.  This is pretty trivial, so we could merge without this being fixed.

adube:
Ready for review.  I updated the GitHub.io pages.

fgravin:
You need to rebase on master.
You will have conflict cause of the `themeservice`. We need to fix that cause i prefer the old way with the injection, than the `init` method.
So maybe wait we finalize the theme service sorry.

adube:
@fgravin Okay, I'll wait for the themeservice to be completed, then.

fgravin:
@adube you can rebase onto master.

adube:
@fgravin I rebased onto master.  The tool stopped working after that and I indeed had conflict issues.  I quickly make the tool work again, but I noticed that there are work in progress regarding the issues I had, i.e. the `init` method and `treeUrl` constant.  I haven't looked at that, but I suppose it will affect my work so please keep me in touch.

fgravin:
oh yes sorry, i was waiting travis to pass before merging, and i though i had merged when i told you to rebase.

adube:
Okay, please let me know once that's merged.

fgravin:
it's merged

fgravin:
Go ahead

adube:
>Go ahead

I want to finish the modifications to the theme selector first.

adube:
@fgravin Travis fails.  See: https://travis-ci.org/camptocamp/ngeo/builds/101836412#L1181  Any hint ?

sbrunner:
I just try to restart the job ...

fgravin:
Probably an issue with the v`erifyNoOutstandingExpectation.` With your changes, the tests expectation are not fullfilled.
Digging

contribs/gmf/src/services/themesservice.js
```diff
@@ -96,26 +106,75 @@ gmf.Themes.findTheme_ = function(themes, themeName) {
...
+            cache: false
+          }).then(
+              goog.bind(
+                  function(deferred, item, response) {
+                    // TODO - we could use the layer helper here to create
+                    //        the layers...
+
+                    var layer = null;
+                    // WMTS
+                    if (item['type'] === 'WMTS') {
```
fgravin: Try to use the `layerHelper`.
Refer to the layerTree if needed.

adube:
@fgravin Done.  Here's also an updated summary of the work in progress. Please, disregard any previous one.

 * Theme service uses layer helper to create its background layers.
 * The GMF abstract mobile controller uses the theme service to get its layer. The map is also, from now on, in epsg:21781 to fit the only available layer returned by the service.
 * The theme service doesn't support WMS for its background layers yet. Please, update the demo theme service response to return one to be able to develop on top of it.
 * I need a checkmark icon or symbol.
 * Travis error is still present

adube:
@fgravin If you think we can merge without what's missing (see above), the let's go for it.

sbrunner:
> The theme service doesn't support WMS for its background layers yet. Please, update the demo theme service response to return one to be able to develop on top of it.

I just add a WMS layer (bus_stop) in the background, then you can test if it's can easily working, it's not a major requirement than you shouldn't lost much time on it...

adube:
@sbrunner FYI, I just had a discussion with @fgravin about the support of WMS layers.  In the scope of this task, we won't support WMS just yet.  Let me explain.

We currently have 2 components that can create layers: the layertree and the theme service (background layers).  Both should share the same method that create the layers.  They currently don't.  It would not have been in the scope of this task to merge the 2 of them, and the WMS layers came in very late so we'll go and merge this one.  The task of having only one method that creates the layers will be done in a separate task.

adube:
@fgravin You can merge.

Issue #437, opened 2015-12-04
Add isMobile constant


sbrunner:
Or whatever mechanism to be able in a directive to know on witch context we are

PR #441, opened 2015-12-04
Add isMobile value

sbrunner:
fix: #437

### DIFF
--- a/contribs/gmf/apps/mobile/js/mobile.js
+++ b/contribs/gmf/apps/mobile/js/mobile.js
@@ -24,6 +24,9 @@ goog.require('ol.layer.Tile');
 goog.require('ol.source.OSM');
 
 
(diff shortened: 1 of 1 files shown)


contribs/gmf/apps/mobile/js/mobile.js
```diff
@@ -24,6 +24,9 @@ goog.require('ol.layer.Tile');
 goog.require('ol.source.OSM');
 
 
+gmfModule.value('isMobile', true);
```
gberaudo: Is the value constant? In this case you may call `gmfModule.constant()` instead.

sbrunner: Thanks, fixed :-)

gberaudo:
LGTM

sbrunner:
Thanks :-)

PR #448, opened 2015-12-08
Standardise template cache

### LINKED ISSUE #425: Template cache URL...
The template cache URL building should be better

Actually in the template cache we have url like:
`../src/directives/partials/*.html`
`../contribs/gmf/src/directives/partials/*.html`

That work for example in `src/examples` but for for example in `contribs/gmf/examples` and in `contribs/gmf/apps/mobile`.
And in all example we need to add something like:
```js
// Use the default "scale selector" template.
app.module.value('ngeoScaleselectorTemplateUrl',
    '../src/directives/partials/scaleselector.html');
```

I think that should be auto-generated.

For the first issue we can build tree different template cache and for the second we should have something that will build automatically the `*TemplateUrl`.

sbrunner:
fix #425

### DIFF
--- a/contribs/gmf/apps/mobile/js/mobile.js
+++ b/contribs/gmf/apps/mobile/js/mobile.js
@@ -140,4 +140,4 @@ app.MobileController.prototype.rightNavIsVisible = function() {
 };
 
 
-app.module.controller('MobileController', app.MobileController);
+appModule.controller('MobileController', app.MobileController);
--- a/examples/layertree.js
+++ b/examples/layertree.js
@@ -25,11 +25,6 @@ var app = {};
 app.module = angular.module('app', ['ngeo']);
 
 
-// Use the default "popup" template.
-app.module.value('ngeoPopupTemplateUrl',
-    '../src/directives/partials/popup.html');
-
-
 /**
  * An application-specific directive wrapping the ngeo tree layer directive.
  * The directive includes a controller defining the tree tree.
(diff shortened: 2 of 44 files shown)


contribs/gmf/apps/mobile/js/mobile.js
```diff
@@ -140,4 +140,4 @@ app.MobileController.prototype.rightNavIsVisible = function() {
 };
 
 
-app.module.controller('MobileController', app.MobileController);
+appModule.controller('MobileController', app.MobileController);
```
fgravin: why ?

sbrunner: To be the same as in ngeo and gmf :-)
see: https://github.com/camptocamp/ngeo/pull/448/files#diff-c72fa8a7df6c1c2a1cf6b46a1eaca607L14

fgravin:
Looks globally good, add the same principle for gmf examples, as discussed.
Thanks for the work.

sbrunner:
Done :-)

examples/layertree.js
```diff
@@ -25,11 +25,6 @@ var app = {};
 app.module = angular.module('app', ['ngeo']);
 
 
-// Use the default "popup" template.
-app.module.value('ngeoPopupTemplateUrl',
```
fgravin: Is the https://github.com/camptocamp/ngeo/blob/master/examples/partials/layertree.html not used anymore then ?

fgravin: forget it, it was just that before the override of this value was useless cause it's passed trhough attributes

PR #452, opened 2015-12-08
Fix less imports

sbrunner:
I have to split the `mobile.less` in two parts because `less.js` creates the links from the first included less file.


sbrunner:
Visible at: http://camptocamp.github.io/ngeo/fix-less/examples/contribs/gmf/apps/mobile/index.html

fgravin:
Maybe `build.less` is not very relevant for name, what do you think ?

sbrunner:
I don't know, do you have a better purpose?
`for_build.less`?

fgravin:
Ok nevermind i don't think we can get any better.
It is good for me, you can merge.

sbrunner:
Thanks @fgravin

PR #460, opened 2015-12-09
Add link to AngularJS documentation on AngularJS objects

ger-benjamin:
Example: http://ger-benjamin.github.io/ngeo/master/apidoc/gmf.MapController.html

(link to the ``angular.Scope`` is not really clean. But it's like an exception...)


sbrunner:
Looks good, thanks :-)

fgravin:
Scope doc should go there https://docs.angularjs.org/api/ng/type/$rootScope.Scope no ?
It depends on what is your target for those doc links.

PR #462, opened 2015-12-09
[WIP] gmf-authentication

adube:
This is a very early state of the `gmf-authentication` directive.  The idea is to collect comments from the core developers of ngeo/gmf.

The base idea is this: an `authentication` service, configured with a base url, will attempt to ask to the server upon creation if the user is connected or not.  The main `authentication` directive listens to the `load` event of the service and checks if the user is connected or not.  It will then decide which inner directive to include.

Is this a good start ?


adube:
**Update**

For the purpose of developing, I created dummy files to act as the request responses.  You can try to modify the `isloggedin` file to see the template show the proper content depending on the response.

This is still a pretty early state, but I would very much like to hear if I'm on the right track.

**Question**  In the template, I use a `ng-switch` on the controller `isLoggedIn` property.  It works upon loading, but if I try to log in, the inner property gets updated but the template doesn't do its magic.  What am I missing ?

adube:
@fgravin Would you please review this work in progress ?  I would like to know if I'm on the right track.

contribs/gmf/src/directives/authenticationdirective.js
```diff
@@ -0,0 +1,137 @@
...
+ * @return {angular.Directive} The Directive Definition Object.
+ * @ngInject
+ * @ngdoc directive
+ * @ngname gmfAuthentication
+ */
+gmf.authenticationDirective = function(gmfAuthenticationTemplateUrl) {
+  return {
+    scope: {},
+    controller: 'GmfAuthenticationController',
+    controllerAs: 'ctrl',
```
fgravin: more explicit name

contribs/gmf/src/directives/authenticationdirective.js
```diff
@@ -0,0 +1,137 @@
...
+   */
+  this.gmfAuthentication_ = gmfAuthentication;
+
+  /**
+   * @type {angular.Scope}
+   * @private
+   */
+  this.scope_ = $scope;
+
+  this.scope_.formData = {};
```
fgravin: No.
Put your data into the controller itself, and access them from the view with the `ctrl` that comes from `controllerAs: 'ctrl'`

adube: Do you have an example ?

contribs/gmf/src/directives/authenticationdirective.js
adube: how about `authenticationCtrl` ?

fgravin: `authCtrl` looks good

contribs/gmf/src/directives/authenticationdirective.js
fgravin: Yes, all ngeo directives and your own PR :)
https://github.com/camptocamp/ngeo/pull/428/files

adube: Still learning here... My own PR have a lot of copy/pasted code I have no perfect understanding of yet :)  Which line ?

fgravin: i've updated the link sorry i missed the copy/paste

fgravin: Just understand that you have a `controllerAs: 'authCtrl'`that will make your controller reachable in the view.

```js
this.formData = 'toto'
```

then in your view 
```html
<div>{{::authCtrl.formData}}</div>
```

adube: Oooh, I think I get it.  I'll try it soon and report back.

fgravin: It's the same principle as https://github.com/fgravin/ngeo/blob/guidelines/docs/guidelines.md#use-the-controller-as-syntax

But i will update the guideline for this case.

adube: Thanks for you precious help.

contribs/gmf/src/directives/partials/authentication.html
```diff
@@ -0,0 +1,24 @@
+<div ng-switch on="ctrl.isLoggedIn">
+    <div ng-switch-when="true">
```
fgravin: I don't really like `ngSwitch`.
You should use `ngIf` instead in that case.

contribs/gmf/src/services/authenticationservice.js
```diff
@@ -0,0 +1,174 @@
...
+ * @export
+ */
+gmf.Authentication.prototype.login = function(login, pwd, callback) {
+
+  var url = goog.uri.utils.appendPath(this.baseUrl_, this.loginRouteSuffix_);
+
+  // FIXME - use post
+  this.http_.get(url, {
+    login: login,
+    password: pwd
```
fgravin: for POST content, you might need to add some headers

```js
method: 'POST',
data: $.param({login: login, password: pwd}),
headers: {'Content-Type': 'application/x-www-form-urlencoded'}
```

contribs/gmf/src/services/authenticationservice.js
```diff
@@ -0,0 +1,174 @@
...
+  })
+    .then(goog.bind(function(response) {
+        var responseData = /** @type {gmf.AuthenticationLoginResponse} */ (
+            response.data);
+        if (responseData.status) {
+          this.isLoggedIn = true;
+        }
+        callback(responseData);
```
fgravin: No this is not the way it should be done.
Look at the themservice, it should return $promises, then you call your callback from where you call the method (don't pass a callback to the method).
https://github.com/fgravin/ngeo/blob/9a95af08e373d1928394bf8cbec03ae0c3499739/contribs/gmf/src/services/themesservice.js#L103

Watch doc https://docs.angularjs.org/api/ng/service/$q

adube:
>Question In the template, I use a ng-switch on the controller isLoggedIn property. It works upon loading, but if I try to log in, the inner property gets updated but the template doesn't do its magic. What am I missing ?

@fgravin Do you know what I'm missing ?

contribs/gmf/src/services/authenticationservice.js
adube: `$` requires jQuery.  Does GMF includes jQuery ?

adube: What I mean is: is it okay to have a dependency on jQuery within the GMF source code ?

fgravin: hmm it was more focused on the `headers`.
not sure for jQuery, it's there in the mobile application but i need to confirm to put some jQuery code in core sources.

fgravin:
> @fgravin Do you know what I'm missing ?

not yet. Try to use {{ctrl.isLoggedIn}} in your view to debug the view value, to see if it's not ngSwitch that just sucks.

adube:
>not yet. Try to use {{ctrl.isLoggedIn}} in your view to debug the view value, to see if it's not ngSwitch that just sucks.

It stays `false`, even with the new `ng-if`.  Still investigating...

adube:
I noticed that printing `{{ formData }}` outside the `<form>` doesn't update its content.  It only does if it's inside the form.

adube:
@fgravin, here's a summary of the work in progress with some questions along the way.

 * Angular `value` -> see how I did it.  I added a single `authenticationCache` value to hold everything related to the authentication.  I'm not sure if this follows the Angular philosophy, but I thought it would be better to only need one parameter to get everything related to the authentication.  Let me know what you think about this.

 * `goog.bind(f, this)` VS `f.bind(this)`, as it turns out I was wrong.  I still causes the build to fail unless I use `goog.bind`.  I reverted that change and still use `goog.bind` for now, if that's okay with you.

 * I still use a `then` inside the service because I want to return the request data instead of the request itself to the directive.  Do you agree ?

 * Travis still fails anyway.  When I run a `make check` locally, I don't see any error.  I haven't found it in the travis report either.

@sbrunner I also have a couple questions / things to show you:

 * when I try to use `http://geomapfish-demo.camptocamp.net/2.0/wsgi/` in my example, I get `'Access-Control-Allow-Origin'` errors.  Would you please look into that ?

 * if you could take a look at the approach I'm currently taking here to see if I'm on the right track, it would be appreciated.  I'm currently assuming that the API routes of the auth. service will return JSON responses.

contribs/gmf/src/services/authenticationservice.js
```diff
@@ -0,0 +1,225 @@
...
+      /**
+       * @param {angular.$http.Response} resp Ajax response.
+       * @return {gmf.AuthenticationLoginResponse}
+       */
+      function(resp) {
+        var respData = /** @type {gmf.AuthenticationLoginResponse} */ (
+            resp.data);
+        if (respData.status) {
+          if (respData.login !== undefined) {
+            this.isLoggedIn = true;
```
fgravin: `isLoggedIn` is useless, `cache.login` should be enough ?

contribs/gmf/src/services/authenticationservice.js
```diff
@@ -0,0 +1,225 @@
...
+/**
+ * @typedef {{
+ *     fullname: (string|undefined),
+ *     login: (string|undefined)
+ * }}
+ */
+gmf.AuthentificationCache;
+
+
+gmfModule.value('authentificationCache', {});
```
fgravin: This maps the user, maybe just call it `gmfUser`.

contribs/gmf/src/directives/partials/authentication.html
```diff
@@ -0,0 +1,35 @@
+<div ng-if="authCtrl.gmfAuthentication.isLoggedIn">
+  Logged in as <strong>{{ authCtrl.gmfAuthentication.cache.fullname }}</strong>.
```
fgravin: same, access through `authCache` directly.

contribs/gmf/src/directives/authenticationdirective.js
```diff
@@ -0,0 +1,140 @@
...
+/**
+ * @export
+ */
+gmf.AuthenticationController.prototype.login = function() {
+  this.gmfAuthentication.login(this.loginVal, this.pwdVal).then(goog.bind(
+      /**
+       * @param {gmf.AuthenticationLoginResponse} respData
+       */
+      function(respData) {
+        if (!respData.status) {
```
fgravin: I'm not sure what's you're doing here.
Can you get a 200 but not being logged ?
It seems the server return a 401.

Plug it to the failure callback then.

fgravin:
Do you know the server side auth service behavior ?
For all requests :
- object returned when success
- what happen on error

You seem you manage a `status` `true` or `false` value, but i guess the service returns a 401 on loggin if fails.
Please be sure you know all scpecs before going ahead.

adube:
About the specs of the server, that's the reason I was asking @sbrunner.  I don't know the spec exactly yet.  That is also why I asked him to check what I did and tell me what to adjust.  I know the server is still under development, thus my questions.

I'll do the other required changes (remove isLoggedIn, rename authenticationCache -> gmfUser, use gmfUser directly in view).

fgravin:
@adube ok.
You can work on the error handling. Just use a bad url for your request and see how you can catch the error from the failure callback of the promise.

sbrunner:
doc: https://camptocamp.github.io/c2cgeoportal/master/developer/webservices.html#authentication

adube:
FYI, there's no documentation for the responses, yet.  That's what I'm looking for.

adube:
Here's an update of this task:

**Done**

 * The client-side forms are all done
 * All possible actions are supported on client-side: login, logout, isloggedin, resetpassword, changepassword
 * Bootstrap style used
 * Validation and errors managed
 * Use of `angularGettext.Catalog` for the translation strings

**To do**

 * @fgravin: review the code
 * validate the current translation text (English)
 * translate in French (should this be part of this task ?)
 * @sbrunner: give me the response information (documentation) from the server, please **Urgent**
 * connect to the demo server in the example instead of using the dummy local files
 * add in **mobile app**: it is **currently broken**

**Question**: I ran `make gh-pages` to be able to show you what's done so far, but the link isn't working.  I expected the following to work, but it doesn't: http://adube.github.io/ngeo/gmf-authentication/contribs/gmf/examples/authentication.html.  What's odd is that the examples from `ngeo` work, where all examples from `gmf` don't.  See: http://adube.github.io/ngeo/gmf-authentication/examples/desktopgeolocation.html. Any clue ?

Let me know if I forgot anything.

sbrunner:
> translate in French (should this be part of this task ?)

No we will use Transifex for that :-)

sbrunner:
The right URL is: http://adube.github.io/ngeo/gmf-authentication/examples/contribs/gmf/authentication.html :-)

fgravin:
> add in mobile app: it is currently broken

Should be fixed now

Makefile
```diff
@@ -1,6 +1,6 @@
 SRC_JS_FILES := $(shell find src -type f -name '*.js')
 NGEO_DIRECTIVES_PARTIALS_FILES := $(shell ls -1 src/directives/partials/*.html)
-GMF_DIRECTIVES_PARTIALS_FILES := $(shell ls -1 contribs/gmf/src/directives/partials/*.html | "")
+GMF_DIRECTIVES_PARTIALS_FILES := $(shell ls -1 contribs/gmf/src/directives/partials/*.html)
```
fgravin: This fixes the gmf templateCache issue.

contribs/gmf/src/directives/authenticationdirective.js
```diff
@@ -0,0 +1,247 @@
...
+    function(element, attrs) {
+      var templateUrl = attrs['gmfAuthenticationTemplateurl'];
+      return templateUrl !== undefined ? templateUrl :
+          gmf.baseTemplateUrl + '/authentication.html';
+    });
+
+
+/**
+ * An "authentication" directive for a GeoMapFish application.
+ *
```
fgravin: Could you add some explanation about the content of the directive.
Just that it is a form with options blabla

contribs/gmf/src/directives/authenticationdirective.js
```diff
@@ -0,0 +1,247 @@
...
+
+  this.gmfAuthentication_.resetPassword(this.loginVal).then(
+      goog.bind(
+          /**
+           * @param {gmf.AuthenticationDefaultResponse} respData
+           */
+          function(respData) {
+            this.resetPasswordModalShown = true;
+            this.resetError_();
+            this.error = '';
```
fgravin: useless ?

contribs/gmf/src/services/authenticationservice.js
```diff
@@ -0,0 +1,247 @@
...
+
+  var url = goog.uri.utils.appendPath(this.baseUrl_, this.loginRouteSuffix_);
+
+  // FIXME - use post, get is currently used for dev purpose
+  /*
+  return this.$http_.post(url, $.param({login: login, password: pwd}), {
+    headers: {'Content-Type': 'application/x-www-form-urlencoded'}
+  }).then(goog.bind(
+  */
+  return this.$http_.get(url).then(
```
fgravin: use post instead.
It's ok tu use the `$.param({})` syntax as discussed previously.

contribs/gmf/src/services/authenticationservice.js
```diff
@@ -0,0 +1,247 @@
...
+/**
+ * Load the authentication service, which sends an asynch request to
+ * determine whether the user is currently connected or not.
+ * @private
+ */
+gmf.Authentication.prototype.load_ = function() {
+  var isLoggedInUrl = goog.uri.utils.appendPath(
+      this.baseUrl_, this.isLoggedInRouteSuffix_);
+  this.$http_.get(isLoggedInUrl, {
+    cache: false
```
fgravin: `false` is the default value i think, so it's useless.

contribs/gmf/src/services/authenticationservice.js
```diff
@@ -0,0 +1,247 @@
...
+};
+
+
+/**
+ * @param {angular.$http.Response} resp Ajax response.
+ * @return  {angular.$http.Response}
+ * @private
+ */
+gmf.Authentication.prototype.handleLogin_ = function(resp) {
+  var respData = /** @type {gmf.AuthenticationLoginResponse} */ (resp.data);
```
fgravin: I think that once we are logged in, we need to retrieve more user info from the `getuserinfoUrl`, doen't it @sbrunner ?

sbrunner: Yes :-)

adube:
Thanks for the review and answers.  Here's some more questions to follow them:

 * @fgravin I can't use `post` yet, because I'm still using the 'dummy' local server.  I did not receive any info about the server yet, so I'll leave them to `get` for now.  Please, note that I'll change them as soon as I'm able to connect to the server.

 * @sbrunner What is the `getuserinfoUrl` ?  Should the information be returned directly by the initial call, i.e. either `isloggedin` or `login` without needing to do an additional request ?

 * the example on github.io isn't currently working properly because some files were not copied, i.e. the dummy server.  This results in all requests to fail.  This should be fixed once we're using the server.

I'll make the other required fixes right away.

fgravin:
@adube ok no worries for th post, you'll do that in time.

fgravin:
@adube 

What i understand from the doc is that a success on `/login` should trigger a `/loginuser` request to fill the `gmfUser` object.

fgravin:
@adube 

For memo
- the doc https://camptocamp.github.io/c2cgeoportal/master/developer/webservices.html#authentication
- the luxembourg code https://github.com/Geoportail-Luxembourg/geoportailv3/blob/master/geoportailv3/static/js/usermanagerservice.js

adube:
Here's an update of this task:

To do:

 * validate the current translation text (English)
 * @sbrunner: I'm still **not able to use the auth. demo service**.  I still get the `No 'Access-Control-Allow-Origin'` error I talked you about 2 days ago.  Please, fix this.  I can't use the service otherwise.
 * Update the code to use the official service.
 * Update the mobile app.

sbrunner:
> I'm still **not able to use the auth. demo service**.  I still get the `No 'Access-Control-Allow-Origin'` error I talked you about 2 days ago.  Please, fix this.  I can't use the service otherwise.

Can you give us the curl?

adube:
>Can you give us the curl?

I don't understand what you mean.

sbrunner:
I most of the debugger you can do on the request: «Copy as cURL», can you past the result?

adube:
There you go.

    curl 'http://localhost:3000/contribs/gmf/examples/wsgi-local/loginuser' -H 'Pragma: no-cache' -H 'Accept-Encoding: gzip, deflate, sdch' -H 'Accept-Language: fr-FR,fr;q=0.8,en-US;q=0.6,en;q=0.4' -H 'User-Agent: Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/47.0.2526.106 Safari/537.36' -H 'Accept: application/json, text/plain, */*' -H 'Referer: http://localhost:3000/contribs/gmf/examples/authentication.html' -H 'Cookie: evsession=7dc5fce8fdcbc752056aa94ef9b3661b1f604455gAJVQ2V2XzIxZWRhYTEyNTgxYWFjMDU2Mjc2NmQxYmVmMjA1ODA4NTcwZjZmOGMyN2NkMzgzZDVhMzc2ZWFlMzExZDU3MzdxAS4=; auth_tkt="9899c1bf61683266654d87e42a8cbf035672c0f2YWRtaW4%3D!userid_type:b64unicode"' -H 'Connection: keep-alive' -H 'Cache-Control: no-cache' --compressed

adube:
There.  I previously had the wrong link.  This is the right one.

    curl 'http://geomapfish-demo.camptocamp.net/2.0/wsgi/loginuser' -H 'Pragma: no-cache' -H 'Origin: http://localhost:3000' -H 'Accept-Encoding: gzip, deflate, sdch' -H 'Accept-Language: fr-FR,fr;q=0.8,en-US;q=0.6,en;q=0.4' -H 'User-Agent: Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/47.0.2526.106 Safari/537.36' -H 'Accept: application/json, text/plain, */*' -H 'Referer: http://localhost:3000/contribs/gmf/examples/authentication.html' -H 'Connection: keep-alive' -H 'Cache-Control: no-cache' --compressed

sbrunner:
@adube I just make some tests on the server (who is definitively bak in https :-) ) and it should work, can you try?

adube:
@sbrunner sure, just a min...

adube:
@sbrunner Nope, it's not working.  I have the same error.  I didn't check if it changed, but here's the curl content again, just in case:

    curl 'http://geomapfish-demo.camptocamp.net/2.0/wsgi/loginuser' -H 'Pragma: no-cache' -H 'Origin: http://localhost:3000' -H 'Accept-Encoding: gzip, deflate, sdch' -H 'Accept-Language: fr-FR,fr;q=0.8,en-US;q=0.6,en;q=0.4' -H 'User-Agent: Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/47.0.2526.106 Safari/537.36' -H 'Accept: application/json, text/plain, */*' -H 'Referer: http://localhost:3000/contribs/gmf/examples/authentication.html' -H 'Connection: keep-alive' -H 'Cache-Control: no-cache' --compressed

sbrunner:
on https I get:
`curl 'https://geomapfish-demo.camptocamp.net/2.0/wsgi/loginuser' -H 'Pragma: no-cache' -H 'Origin: http://localhost:3000' -H 'Accept-Encoding: gzip, deflate, sdch' -H 'Accept-Language: fr-FR,fr;q=0.8,en-US;q=0.6,en;q=0.4' -H 'User-Agent: Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/47.0.2526.106 Safari/537.36' -H 'Accept: application/json, text/plain, */*' -H 'Referer: http://localhost:3000/contribs/gmf/examples/authentication.html' -H 'Connection: keep-alive' -H 'Cache-Control: no-cache' --compressed -I`
```
HTTP/1.1 200 OK
Date: Thu, 17 Dec 2015 16:44:28 GMT
Server: Apache/2.4.10 (Debian)
Cache-Control: no-cache
Access-Control-Allow-Origin: http://localhost:3000
Access-Control-Allow-Methods: HEAD, GET, POST, PUT, DELETE
Access-Control-Allow-Credentials: true
Content-Length: 65
Connection: close
Content-Type: application/json; charset=UTF-8
```
Then what about if you replace http by https ?

adube:
>Then what about if you replace http by https?

I works.

Okay, so for the next steps: you told me 2 days ago that there were still work to do in the services for them to work all properly.  I haven't got any news about that.  Are they all ready, now ?  I.E. am I ready to proceed ?

sbrunner:
Yes, actually, for me, it's working correctly :-)

adube:
The ball is in my camp, now.  Thanks.

adube:
@sbrunner I just tried to login for the first time.

    GET https://geomapfish-demo.camptocamp.net/2.0/wsgi/login 
    authentication.html:1 XMLHttpRequest cannot load https://geomapfish-demo.camptocamp.net/2.0/wsgi/login.
    No 'Access-Control-Allow-Origin' header is present on the requested resource.
    Origin 'http://localhost:3000' is therefore not allowed access. The response had HTTP status code 400.

Bottom-line: I'm still unable to use the service.  Please, test them **all** - login, loginuser, logout, loginchange and loginresetpassword and make sure they all work properly.

I'll be waiting for your positive answer before moving on.  Thanks.

sbrunner:
I just try your example in `https` and I get an error 404 because it do a GET not a POST ...

sbrunner:
And if I activate the view a `Bad request` error because it don't pass any username and password...

adube:
Okay, thanks for the clarifications.  I guess I got stuck so many times in the past few days that this hasn't crossed my mind.  I'll make the changes and continue to work on this.

adube:
@sbrunner Even in POST, even if I correctly send the `login` and `password` parameters, I still get the `'Access-Control-Allow-Origin'` error.

adube:
Curl:

    curl 'https://geomapfish-demo.camptocamp.net/2.0/wsgi/login' -H 'Pragma: no-cache' -H 'Origin: http://localhost:3000' -H 'Accept-Encoding: gzip, deflate' -H 'Accept-Language: fr-FR,fr;q=0.8,en-US;q=0.6,en;q=0.4' -H 'User-Agent: Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/47.0.2526.106 Safari/537.36' -H 'Content-Type: application/x-www-form-urlencoded' -H 'Accept: application/json, text/plain, */*' -H 'Cache-Control: no-cache' -H 'Referer: http://localhost:3000/contribs/gmf/examples/authentication.html' -H 'Connection: keep-alive' --data 'login=a&password=a' --compressed

adube:
>XMLHttpRequest cannot load https://geomapfish-demo.camptocamp.net/2.0/wsgi/login. No 'Access-Control-Allow-Origin' header is present on the requested resource. Origin 'http://localhost:3000' is therefore not allowed access. The response had HTTP status code 401.

adube:
FYI, with a correct username/password, we get an error 500.

adube:
@sbrunner Here's a new summary:

 * `login` works, but the cookie is not stored because the request is not sent from the same domain
 * this causes `logout`, `loginuser` and `loginchange` to not work properly, because when sending those request it's as if I'm never connected
 * the `loginresetpassword` doesn't work, but I'm not sure why.  Look at the code here: https://github.com/camptocamp/c2cgeoportal/blob/master/c2cgeoportal/views/entry.py#L1448-L1458.  It would seem that if those responses are returned, then it is not what the `loginresetpassword` is supposed to expect as return.  I suppose that's why I always get `404` errors.

You may end up needing a proxy after all, at least for the purpose of making the examples work.  Please, talk to me about this.

I'll finish what I started and leave it "as-is", in the state where only the login works.  I'll keep you in touch.

sbrunner:
It's sure that the errors will never work ...

For the cookie story we should discus about it internally ..., it's sure that's not evident ...

adube:
Understood.  I'll put this on ice until you come back to me with a solution, then.  Thanks.

sbrunner:
You should add `withCredential` in your requests :-)
```diff
diff --git a/contribs/gmf/src/services/authenticationservice.js b/contribs/gmf/src/services/authenticationservice.js
index cbc292b..6833624 100644
--- a/contribs/gmf/src/services/authenticationservice.js
+++ b/contribs/gmf/src/services/authenticationservice.js
@@ -141,7 +141,7 @@ goog.inherits(gmf.Authentication, goog.events.EventTarget);
 gmf.Authentication.prototype.load_ = function() {
   var isLoggedInUrl = goog.uri.utils.appendPath(
       this.baseUrl_, this.isLoggedInRouteSuffix_);
-  this.$http_.get(isLoggedInUrl).then(
+  this.$http_.get(isLoggedInUrl, {withCredentials: true}).then(
       goog.bind(this.handleLogin_, this));
 };
 
@@ -181,7 +181,8 @@ gmf.Authentication.prototype.login = function(login, pwd) {
   // FIXME - use post, get is currently used for dev purpose
   //return this.$http_.get(url).then(
   return this.$http_.post(url, $.param({login: login, password: pwd}), {
-    headers: {'Content-Type': 'application/x-www-form-urlencoded'}
+    headers: {'Content-Type': 'application/x-www-form-urlencoded'},
+    withCredentials: true
   }).then(
       goog.bind(this.handleLogin_, this));
 };
```

adube:
I added `withCredentials: true` to the login, isLoggedIn and changePassword methods of the authentication service.

FYI, this morning, I'm no longer able to login.  It used to work, now it doesn't anymore.

Error:

    XMLHttpRequest cannot load https://geomapfish-demo.camptocamp.net/2.0/wsgi/login. No 'Access-Control-Allow-Origin' header is present on the requested resource. Origin 'http://127.0.0.1:3000' is therefore not allowed access.

Curl:

    curl 'https://geomapfish-demo.camptocamp.net/2.0/wsgi/login' -H 'Pragma: no-cache' -H 'Origin: http://127.0.0.1:3000' -H 'Accept-Encoding: gzip, deflate' -H 'Accept-Language: fr-FR,fr;q=0.8,en-US;q=0.6,en;q=0.4' -H 'User-Agent: Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/47.0.2526.106 Safari/537.36' -H 'Content-Type: application/x-www-form-urlencoded' -H 'Accept: application/json, text/plain, */*' -H 'Cache-Control: no-cache' -H 'Referer: http://127.0.0.1:3000/contribs/gmf/examples/authentication.html' -H 'Cookie: auth_tkt_2.0="a335a0ddad7db1508f26fad4768302ce4ef64855186ea00147c9961b51988ed2da46d3c2a0fbdb2c5f59fe1542f58467e36bfafc788a642045bbe144cfb02b295677f8d2YWR1YmU%3D!userid_type:b64unicode"; auth_tkt_2.0="a335a0ddad7db1508f26fad4768302ce4ef64855186ea00147c9961b51988ed2da46d3c2a0fbdb2c5f59fe1542f58467e36bfafc788a642045bbe144cfb02b295677f8d2YWR1YmU%3D!userid_type:b64unicode"' -H 'Connection: keep-alive' --data 'login=adube&password=adube' --compressed

sbrunner:
For me it's working ...

adube:
>For me it's working ...

For me, it's not.  Please, help me understand why.

adube:
FYI, the `curl` seems to work for me too, but from the example page it's not working.

adube:
I update the GitHub IO example page:

http://adube.github.io/ngeo/gmf-authentication/examples/contribs/gmf/authentication.html

In it, it is not working either.  It has the same kind of errors I'm experiencing locally.

adube:
Github IO pages:

 * the [example page](http://adube.github.io/ngeo/gmf-background-selector-mobile/examples/contribs/gmf/mobilebackgroundlayerselector.html?map_zoom=3&map_x=640078&map_y=198984)
 * the [mobile app](http://adube.github.io/ngeo/gmf-background-selector-mobile/examples/contribs/gmf/apps/mobile/index.html)

fgravin:
@sbrunner do we want a blank layer as well in the bg layers combo ?

fgravin:
Great, but there is an issue with the img urls in the application.

sbrunner:
We miss to specify that, but yes we need it...

adube:
This task is ready for review.  Here's a wrap-up.  You can ignore everything noted before.

**Examples on GitHub.io**

 * [auth. in examples](http://adube.github.io/ngeo/gmf-authentication/examples/contribs/gmf/authentication.html)
 * [auth. in mobile app](http://adube.github.io/ngeo/gmf-authentication/examples/contribs/gmf/apps/mobile/)

The only thing remaining to do is this:

 * in the GMF2 design interface, we have an `old password` field in the "change password".  In the actual API route, no such parameter is required.  I currently added the field in the gmf-authentication view, but it doesn't do anything. @sbrunner Please, tell me what to do with this.

sbrunner:
I should add it in the server interface...

adube:
>I should add it in the server interface...

I'll add it right away, then.  Please, let me know once that's added on your side for me to test it.

adube:
@sbrunner I'll use `oldPassword` as parameter.  Please, let me know if I should use something else.

sbrunner:
Without having a look at you do I use  the same name :-)
https://github.com/camptocamp/c2cgeoportal/pull/1892

contribs/gmf/apps/mobile/index.html
```diff
@@ -66,6 +62,9 @@
       <div id="themes" class="slide" data-header-title="Themes">
         <span>Themes</span>
       </div>
+      <div id="login" class="slide" data-header-title="Login">
```
fgravin: Do you really need a container div ?

`<gmf-authentication id="login" class="slide" data-header-title="Login"></gmf-authentication>` is not enough ?

contribs/gmf/src/directives/partials/authentication.html
```diff
@@ -0,0 +1,165 @@
+<div ng-if="gmfUser.username">
+  <div class="form-group">
+    <span>{{ authCtrl.translate('Logged in as') }}</span>
+    <strong>{{ gmfUser.username }}</strong>.
```
fgravin: one time binding ?

contribs/gmf/src/directives/partials/authentication.html
```diff
@@ -0,0 +1,165 @@
+<div ng-if="gmfUser.username">
+  <div class="form-group">
+    <span>{{ authCtrl.translate('Logged in as') }}</span>
```
fgravin: one time binding ?

contribs/gmf/src/directives/partials/authentication.html
```diff
@@ -0,0 +1,165 @@
...
+      role="form"
+      ng-submit="authCtrl.logout()"
+      ng-controller="GmfAuthenticationController"
+      ng-class="{'has-error': authCtrl.error}"
+      ng-show="!authCtrl.changingPassword">
+    <div class="form-group">
+      <input
+          type="submit"
+          class="form-control btn btn-primary"
+          value="{{ authCtrl.translate('Logout') }}" />
```
fgravin: one time binding ?

contribs/gmf/src/directives/partials/authentication.html
adube: Unless I misunderstand what a one time binding does, shouldn't this not be a one time binding ?  What if we log in a second time using an other username ?  Would a one time binding show the second username ?

adube: Okay, so with `::`, i.e. one time binding, the username is properly set.  That means I don't fully understand what the one time binding truly does...

I'll change this to one time binding, as requested.

adube:
@fgravin the required changes have been fixed.

adube:
@fgravin Ready for review.

adube:
I updated the GitHub.io example pages.

contribs/gmf/src/directives/authenticationdirective.js
```diff
@@ -0,0 +1,281 @@
...
+ * @param {gmf.User} gmfUser
+ * @constructor
+ * @ngInject
+ * @ngdoc controller
+ * @ngname GmfAuthenticationController
+ */
+gmf.AuthenticationController = function(gettextCatalog, $scope,
+    gmfAuthentication, gmfUser) {
+
+  $scope.gmfUser = gmfUser;
```
fgravin: why in the `$scope`and no in the controller ?

fgravin:
Example and mobile app still don't work cause of cross origin policy.

contribs/gmf/src/services/authenticationservice.js
fgravin: It seems you didn't do that.
You need to call `isLoggedInRouteSuffix_` after a login, to retrieve more information.
Do you confirm that it returns more info than the login service ?

contribs/gmf/src/services/authenticationservice.js
```diff
@@ -0,0 +1,263 @@
...
+   * @type {gmf.User}
+   * @private
+   */
+  this.user_ = gmfUser;
+
+  /**
+   * @type {string}
+   * @private
+   */
+  this.changePasswordRouteSuffix_ = 'loginchange';
```
fgravin: Not sure we should declare CONST this way.
Will see what other think about that, what to you think ?

adube: How about we move them in a

gmf.AuthenticationRouteSuffix = {
  CHANGE_PASSWORD: 'loginchange',
  // etc...
}

contribs/gmf/src/services/authenticationservice.js
adube: No need to do that.  The same response is returned when calling `userlogin` and `login`.

contribs/gmf/src/directives/authenticationdirective.js
adube: I don't remember why, but I'll have to check.

adube:
>Example and mobile app still don't work cause of cross origin policy.

@sbrunner ?

sbrunner:
I just change the URL to use the https protocol, is it working now?

contribs/gmf/src/directives/authenticationdirective.js
adube: I remember now.  I set it in the `$scope` to be able to do the following in the template:

    <div ng-if="gmfUser.username">

I haven't found a way to use an angular value inside a template otherwise, so far.  Is there a better way to accomplish this without having to set it in the scope ?

adube:
>I just change the URL to use the https protocol, is it working now?

No

On an other subject, I just re-based on top of master without changing anything and Travis fails.  It used to work before.  What am I missing ?

contribs/gmf/src/services/authenticationservice.js
adube: @fgravin ?

contribs/gmf/src/directives/authenticationdirective.js
fgravin: `this.gmfUser = gmfUser;`
in the template `authCtrl.gmfUser`
and remove all the` $scope` ref.

contribs/gmf/src/services/authenticationservice.js
fgravin: i thinks it's more ol3 like :)

sbrunner:
You have an error on witch URL?
I can't reproduce the error...

adube:
@sbrunner http://adube.github.io/ngeo/gmf-authentication/examples/contribs/gmf/authentication.html

sbrunner:
We will not add the http protocol as allowed origin, than the URL https://adube.github.io/ngeo/gmf-authentication/examples/contribs/gmf/authentication.html seam working ...

adube:
That famous extra **s**... okay, so the http**s** version IS working:

https://adube.github.io/ngeo/gmf-authentication/examples/contribs/gmf/authentication.html

 * Travis tests all pass
 * I made the modifications required by @fgravin 

Can we merge this, please ?

PR #469, opened 2015-12-10
Add more closure-compiler checks

### LINKED ISSUE #440: No check for unused 'goog.require's
When building there is no check for unused `goog.require`s like in ol3:
https://github.com/openlayers/ol3/blob/master/bin/check-requires.py#L120-L123

fredj:
Fixes #440

### DIFF
--- a/buildtools/example.json
+++ b/buildtools/example.json
@@ -40,42 +40,14 @@
       "externs/ngeox.js"
     ],
     "jscomp_error": [
-      "accessControls",
-      "ambiguousFunctionDecl",
-      "checkEventfulObjectDisposal",
-      "checkRegExp",
-      "checkTypes",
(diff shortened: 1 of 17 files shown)


sbrunner:
Looks good, thanks :-)

fgravin:
What the new comments with `@suppress` for ?

fredj:
`@suppress {extraRequire}` tells the compiler to not check the next `goog.require`.

Example:
In `src/services/popup.js` we require `ngeo.popupDirective`. The checker expect that `ngeo.popupDirective.*` is used in the file but that not the case.

fgravin:
Oh ok, so it's mainly for directives requires.

sbrunner:
This should be rebased and merged :-)

PR #471, opened 2015-12-10
Fix NgeoMobileGeolocation controller

adube:
This PR fixes the gmf mobile geolocation controller behaviour with the following rule: when receiving an update of the location while tracking, only recenter the map to this location if the map hadn't been panned (or zoomed).

### DIFF
--- a/src/directives/mobilegeolocation.js
+++ b/src/directives/mobilegeolocation.js
@@ -113,6 +113,21 @@ ngeo.MobileGeolocationController = function($scope, $element,
    */
   this.zoom_ = options.zoom;
 
+  /**
+   * Whether to recenter the map at the position it gets updated
+   * @type {boolean}
+   * @private
+   */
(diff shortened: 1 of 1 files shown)


adube:
@sbrunner This is ready for review, once Travis passes.

sbrunner:
Looks good for me :-)

adube:
Ready for merge, then.

src/directives/mobilegeolocation.js
```diff
@@ -132,6 +147,29 @@ ngeo.MobileGeolocationController = function($scope, $element,
       false,
       this);
 
+  var view = map.getView();
+
+  goog.events.listen(
```
fgravin: can't all of this could be simplified with

```js
ol.events.listen(
       view,
       'propertychange',

```

adube: Yes.

PR #473, opened 2015-12-11
Use stadard metadata

sbrunner:
see: 
https://developer.chrome.com/multidevice/android/installtohomescreen#comparison-to-ios


ger-benjamin:
+1

PR #481, opened 2015-12-23
Tests for gmf + tests for gmf's themesservice

ger-benjamin:
- Manage tests in ``contribs/gmf/test/...`` (commit 	*Add test in gmf*)
 - Make an example with ``contribs/gmf/src/services/themesservice.js`` (commit 	*Tests for gmf themesservice*)

I don't know if I should make two distinct "karma" files ..?


contribs/gmf/src/services/themesservice.js
```diff
@@ -54,13 +58,13 @@ gmf.Themes = function($http, treeUrl, isThemePrivateUrl) {
...
    */
-  this.treeUrl_ = treeUrl;
+  this.treeUrl_ = '';
 
   /**
    * @type {string}
    * @private
    */
-  this.isThemePrivateUrl_ = isThemePrivateUrl;
+  this.isThemePrivateUrl_ = '';
```
fgravin: `isThemePrivateUrl ` is useless, you can remove it

sbrunner: +1 to remove it :-)

contribs/gmf/src/services/themesservice.js
```diff
@@ -72,6 +76,21 @@ goog.inherits(gmf.Themes, goog.events.EventTarget);
 
 
 /**
+ * Init the service.
+ * @param {string} treeUrl URL to "themes" web service.
+ * @param {string} isThemePrivateUrl URL to check if theme is public.
+ * @export
+ */
+gmf.Themes.prototype.init = function(treeUrl, isThemePrivateUrl) {
```
ger-benjamin: I've changed how to initialize the service (I use an init function to be coherent with others services) but that causes trouble with the Luxembourg project... Must I revert this change ?

contribs/gmf/src/services/themesservice.js
ger-benjamin: Okay... then I pass the URL as parameter on function 'isThemePrivate' ? Luxembourg project will must be adapted.

fgravin: Luxembourg doesn't use this service yet.
If gmf2 doesn't support private theme url, then we need to remove it.
If a project like luxembourg want to use it, they need to add custom code in their own sources to handle it.

fgravin:
Ok for the change of the `ThemeService`, we need to go ahead with this cause it's used by many other components though.

Thanks for the tests it's really helpfull.

contribs/gmf/src/services/themesservice.js
fgravin: Luxembourg has its own service for the moment, when they want to migrate to use the ngeo one, then we'll see what are the impacts. They probably need to update their application code anyway.

ger-benjamin:
Thanks @fgravin 
So, I've removed the "isThemePrivate" function (and the corresponding test) and "isThemePrivateURL" string.

ger-benjamin:
Can I merge ?

fgravin:
Yes go ahead, thanks

PR #483, opened 2016-01-05
Add layerHelper service

ger-benjamin:
Will be used by the layertree and the backgroundmanager


src/services/layerHelper.js
```diff
@@ -0,0 +1,316 @@
...
+goog.require('ol.layer.Group');
+goog.require('ol.layer.Image');
+goog.require('ol.layer.Tile');
+goog.require('ol.source.ImageWMS');
+goog.require('ol.source.WMTS');
+
+
+
+/**
+ * Provides helpers fonctions that helps you to create and manage layers.
```
fgravin: `help`

src/services/layerHelper.js
```diff
@@ -0,0 +1,316 @@
...
+   * @type {angular.$http}
+   * @private
+   */
+  this.$http_ = $http;
+
+  /**
+   * @type {string}
+   * @export
+   */
+  this.helperID = 'helperID';
```
fgravin: It's a const, should be declared differently no ?
MAJ and maybe directly in the prototype

src/services/layerHelper.js
```diff
@@ -0,0 +1,316 @@
...
+ */
+ngeo.LayerHelper.prototype.makeHelperID = function(layerURL, layerName) {
+  return layerURL + '_' + layerName;
+};
+
+
+/**
+ * Create and return a basic WMS layer with only a source URL and a dot
+ * separated layers names (see {@link ol.source.ImageWMS}).
+ * This layer will be taged by a a helperID.
```
fgravin: tagged

src/services/layerHelper.js
```diff
@@ -0,0 +1,316 @@
...
+
+
+/**
+ * Create and return a promise that provides a WMTS layer with source on
+ * success, no layer else.
+ * The WMTS layer source will be configured by the capabilities that are
+ * loaded from the given capabilitiesUrl.
+ * This layer will be taged by a a helperID.
+ * @param {string} capabilitiesURL The getCapabilities url.
+ * @param {string} layerName The name of te layer.
```
fgravin: the layer

src/services/layerHelper.js
```diff
@@ -0,0 +1,316 @@
...
+  return layer;
+};
+
+
+/**
+ * Create and return a promise that provides a WMTS layer with source on
+ * success, no layer else.
+ * The WMTS layer source will be configured by the capabilities that are
+ * loaded from the given capabilitiesUrl.
+ * This layer will be taged by a a helperID.
```
fgravin: tagged

src/services/layerHelper.js
```diff
@@ -0,0 +1,316 @@
...
+  var deferred = this.$q_.defer();
+  this.setHelperID(layer, capabilitiesURL, layerName);
+
+  this.$http_.get(capabilitiesURL).then(function(response) {
+    var result;
+    if (response.data) {
+      result = parser.read(response.data);
+    }
+    if (goog.isDef(result)) {
+      var options = ol.source.WMTS.optionsFromCapabilities(result,
```
fgravin: we might need to add a `try catch` here, but i don't know yet.

src/services/layerHelper.js
```diff
@@ -0,0 +1,316 @@
...
+    group.setLayers(opt_layers);
+  }
+  return group;
+};
+
+
+/**
+ * Get the position of the layer in the map's layers array or -1 if the layer
+ * can't be found.
+ * The layer must be taged by a helperID (return -1 else).
```
fgravin: tagged

src/services/layerHelper.js
```diff
@@ -0,0 +1,316 @@
...
+    }
+  }
+  return null;
+};
+
+
+/**
+ * Add the given layer on the map if it doesn't already exists and if the layer
+ * has a source.
+ * The layer must be taged by a helperID.
```
fgravin: tagged

src/services/layerHelper.js
```diff
@@ -0,0 +1,316 @@
...
+ * @param {ol.layer.Layer} layer The concerned layer.
+ * @return {boolean} true if added, false else.
+ * @export
+ */
+ngeo.LayerHelper.prototype.addLayerToMap = function(layer) {
+  var added = false;
+  if (goog.isDefAndNotNull(layer.getSource())) {
```
fgravin: `if (layer.getSource())` does the same no ?
I think we want to remove `goog` usage if not needed.

src/services/layerHelper.js
```diff
@@ -0,0 +1,316 @@
...
+      added = true;
+    }
+  }
+  return added;
+};
+
+
+/**
+ * Remove the given layer from the map if it exists.
+ * The layer must be taged by a helperID.
```
fgravin: tagged

src/services/layerHelper.js
```diff
@@ -0,0 +1,316 @@
...
+    this.map_.removeLayer(layerToRemove);
+    removed = true;
+  }
+  return removed;
+};
+
+
+/**
+ * Add or remove some layers from the map.
+ * The layer must be taged by a helperID.
```
fgravin: tagged

src/services/layerHelper.js
```diff
@@ -0,0 +1,316 @@
...
+      this.addLayerToMap(layer);
+    } else {
+      this.removeLayerFromMap(layer);
+    }
+  }
+};
+
+
+/**
+ * Get an array of all layer in a group. The group can contain multiple levels
```
fgravin: layers

src/services/layerHelper.js
```diff
@@ -0,0 +1,316 @@
...
+ * @return {Array.<ol.layer.Layer>}
+ * @export
+ */
+ngeo.LayerHelper.prototype.getFlatLayers = function(layer) {
+  return this.getFlatLayers_(layer, []);
+};
+
+
+/**
+ * Get an array of all layer in a group. The group can contain multiple levels
```
fgravin: layers

fgravin:
Some minor comments but looks good.
A little question: why do you want to use a `helperID` ?

ger-benjamin:
Thanks @fgravin 
I want to use a ``helperID`` because a layer doesn't seem to have an unique ID and I must have a system to find a specific layer in the layers array of ol.Map. 

Before that I use a custom ID, I've tried with the name and the source URL of the layer but that doesn't work for WMTS layers (source URL is not contained in tiles layers).

src/services/layerHelper.js
```diff
@@ -0,0 +1,316 @@
...
+  this.helperID = 'helperID';
+};
+
+
+/**
+ * Init the service.
+ * @param {!ol.Map} map
+ * @export
+ */
+ngeo.LayerHelper.prototype.init = function(map) {
```
fgravin: This makes the service bound to only one map.
As services are Singleton, we can use just one service for one map. 
If you look at https://github.com/camptocamp/ngeo/blob/master/src/services/backgroundlayermgr.js#L117 for example, one service can manage multiple maps.

But, we can leave this like that for now, it may not be necessary...

src/services/layerHelper.js
```diff
@@ -0,0 +1,316 @@
...
+goog.require('ol.layer.Group');
+goog.require('ol.layer.Image');
+goog.require('ol.layer.Tile');
+goog.require('ol.source.ImageWMS');
+goog.require('ol.source.WMTS');
+
+
+
+/**
+ * Provides help fonctions that helps you to create and manage layers.
```
gberaudo: functions

src/services/layerHelper.js
ger-benjamin: If I add a "map" parameters at two fonctions (``addLayerToMap`` and ``removeLayerFromMap``, that seems not too bad to do that) It can manage multiple maps. Wait...

ger-benjamin:
Done.
No more need to init this service and it now doesn't depend to a map.

PR #482, opened 2016-01-05
Build data folder in exemples and update 'themes' json

ger-benjamin:

PR #444, opened 2015-12-07
Gmf Layertree

ger-benjamin:
Revival of the PR #291 
Work for https://github.com/camptocamp/c2cgeoportal/issues/1707

Example: http://ger-benjamin.github.io/ngeo/master/examples/contribs/gmf/layertree.html

Todo:
 - [x] [Question] Elemoine created externes for 'GmfThemesNode' in file 'gmf-layertree.js'  but perhaps it needs to be moved in file "gmfx" ? --> Let's keep one externs file by WS.
 - [x] [Question] Ev. rename the ``LayerManager`` in something more representative. --> Renamed to ``LayerHelper``
 - [x] [Question] Move the gmf's directive partial folder ?
 - [x] [Question] Move the gmf Layertree's data folder ?
 - [x] Debug make check (doesn't find 'partial' on examples-hosted builds)

For the 2nd point, I means that the 'LayerManager' doesn't manage layer. It only regroups some helpers fonctions related to layers and layergroups.
Some functions in the LayerManager is really... 'trivial'. But I won't put it in the gmf-layertree.

Adding gmf Layertree in the mobile folder will be in another pull request.

PR #484, opened 2016-01-05
Gmf layertree example

ger-benjamin:
Revival of the PR #444
Work for camptocamp/c2cgeoportal#1707

~~Waiting on PR #482 and #483 (please, review them).~~ (Done)


src/directives/layertree.js
```diff
@@ -220,6 +214,9 @@ ngeo.LayertreeController = function($scope, $element, $attrs) {
 
   $scope['layertreeCtrl'] = this;
 
+  $scope.$on('$destroy', angular.bind(this, function() {
+    this.map.removeLayer(this.layer);
```
ger-benjamin: I don't like to do this in ngeo but that's the cleaner (only cleaner) manner to remove layer when this tree 'on destroy'.

The only other way to do this is to proceed like the "ngeo-layertree-nodelayer". this attribute refers to a fonction (that help to link a layer to a node (a layertree).
I can do a similar system with a "ngeo-layertree-onevents" attributes that allow to bind custom function to layertree events. In that function, we can probably link a 'on destroy function'.

(On writing that, I just convince myself that's a better idea. I'll try this. Wait...)

ger-benjamin:
Ready to review

contribs/gmf/src/directives/layertree.js
```diff
@@ -0,0 +1,301 @@
...
+gmf.LayertreeController = function($scope, ngeoLayerHelper) {
+
+  var wmsUrl = this['getWmsUrlFn']();
+  goog.asserts.assert(goog.isString(wmsUrl));
+
+  /**
+   * @type {string}
+   * @private
+   */
+  this.wmsUrl_ = wmsUrl;
```
ger-benjamin: Currently passed as attribute but perhaps must be a constant ?

fgravin: Yes, like the theme url.
A constant is appropriate.

contribs/gmf/src/directives/layertree.js
```diff
@@ -0,0 +1,301 @@
...
+ * @return {angular.Directive} The directive specs.
+ * @ngInject
+ * @ngdoc directive
+ * @ngname gmfLayertreeDirective
+ */
+gmf.layertreeDirective = function(gmfLayertreeTemplate) {
+  return {
+    scope: {
+      'map': '=gmfLayertreeMap',
+      'getWmsUrlFn': '&gmfLayertreeWmsurl',
```
fgravin: You will remove it to use a constant, but anyway think to use `'='` instead.
Here you have both `'&'` and `::` in your example template, what is redondant.

contribs/gmf/src/directives/layertree.js
```diff
@@ -0,0 +1,301 @@
...
+
+
+/**
+ * Return a class name that match with the current node activation state.
+ * @param {ngeo.LayertreeController} treeCtrl ngeo layertree controller, from
+ *     the current node.
+ * @return {string} 'on' or 'off' or 'indeterminate'.
+ * @export
+ */
+gmf.LayertreeController.prototype.getSetActive = function(treeCtrl) {
```
fgravin: `getSetActive` name is not appropriate.

fgravin: This method seems a bit too complex to be called on each node on each `$digest loop`.

contribs/gmf/src/directives/partials/layertree.html
```diff
@@ -0,0 +1,17 @@
+<span ng-if="::!layertreeCtrl.isRoot"
+    ng-class="['depth-' + layertreeCtrl.depth, gmfLayertreeCtrl.getResolutionStyle(layertreeCtrl.node), gmfLayertreeCtrl.getNoSourceStyle(layertreeCtrl)]">
+  <a href ng-click="::gmfLayertreeCtrl.toggleActive(layertreeCtrl)">
+    <span class="state" ng-class="[gmfLayertreeCtrl.getSetActive(layertreeCtrl)]"></span>
```
fgravin: you could do a `ng-if` here with one time binding to know if there are children.
Only use the `ng-class` when children will avoid you a lot of useless watch functions executions.

contribs/gmf/src/directives/layertree.js
fgravin: Seems better to have a value that store the group node state.
This value is watched in the template (more simple watch function).
This value is updated each time a node is changed (actually your function is not called on digest, but manually when you need).

fgravin:
Looks globally very good, thanks for your wok.

I'm a bit worried though about performance optimisation in your whole `ng-class` watch functions. I think one could be optimized (for the group state). Please read my comments, if you think they are appropriate and that you could optimize your code, go ahead, else keep as it is.

ger-benjamin:
@fgravin Thanks a lot for the review. I'll read/correct/think about that

contribs/gmf/src/directives/layertree.js
ger-benjamin: > `getSetActive` name is not appropriate.

Done

> This method seems a bit too complex to be called on each node on each $digest loop.

True but...

> Seems better to have a value that store the group node state.
This value is watched in the template (more simple watch function).
This value is updated each time a node is changed (actually your function is not called on digest, but manually when you need).

The last point is just hard to do. Because a node doesn't know its parent(s). It just knows its layer and its node. So I must keep somewhere and manage all ol.layer.Group objects created in the gmf-layertree.
(And obviously, all node are not updated when one has its layer added on the map)

contribs/gmf/src/directives/partials/layertree.html
ger-benjamin: But in case of that's a leaf, I must get the 'state' style too... I don't undertsand :-/

ger-benjamin:
> `getSetActive` name is not appropriate.

Done

> This method seems a bit too complex to be called on each node on each $digest loop.

True but...

> Seems better to have a value that store the group node state.
This value is watched in the template (more simple watch function).
This value is updated each time a node is changed (actually your function is not called on digest, but manually when you need).

The last point is just hard to do. Because a node doesn't know its parent(s). It just knows its layer and its node. So I must keep somewhere and manage all ol.layer.Group objects created in the gmf-layertree.
(And obviously, all node are not updated when one has its layer added on the map)

fgravin:
> But in case of that's a leaf, I must get the 'state' style too... I don't undertsand :-/

```html
  <span ng-if="::layertreeCtrl.node.children" class="state" ng-class="[gmfLayertreeCtrl.getNodeState(layertreeCtrl)]"></span>
  <span ng-if="::!layertreeCtrl.node.children" class="state"></span>

```

fgravin:
> Can I merge that then, asap, propose a change to improve performances/logic ?

Ok, if it's too hard to optimize, forget it for now.
Thanks

fgravin:
You can merge

PR #491, opened 2016-01-06
Set all package version to strict

fredj:
Replace #413


sbrunner:
Looks good for me :-)

fredj:
Thanks for the review

PR #496, opened 2016-01-07
Fix GMF theme json example - human readable

adube:
This simple PR makes the theme json in the GMF example human readable.  In addition, it allows modifications to be more easily merged.  I got conflicts while rebasing and it took a short while to understand what was different.  Having the file like this would have avoided the conflict.


adube:
@fgravin What do you think ?

fgravin:
+1 thanks

adube:
Thanks.  Please, merge.

PR #497, opened 2016-01-07
[WIP] Add Theme Selector directive for GMF

adube:
Replaces #478.

Still in progress.  I just want to see the Travis report.  Please, wait for further details before reviewing.

### DIFF
--- a/contribs/gmf/apps/mobile/index.html
+++ b/contribs/gmf/apps/mobile/index.html
@@ -58,9 +58,13 @@
           </li>
         </ul>
       </div>
-      <div id="themes" class="slide" data-header-title="Themes">
-        <span>Themes</span>
-      </div>
+      <gmf-themeselector
+          id="themes"
+          class="slide"
+          data-header-title="Themes"
+          gmf-themeselector-currenttheme="mainCtrl.theme"
+          gmf-themeselector-filter="::mainCtrl.filter">
+      </gmf-themeselector>
       <gmf-authentication id="login" class="slide" data-header-title="Login">
       </gmf-authentication>
     </nav>
--- a/contribs/gmf/apps/mobile/js/mobile.js
+++ b/contribs/gmf/apps/mobile/js/mobile.js
@@ -36,6 +36,7 @@ appModule.constant(
  * @param {ngeo.StateManager} ngeoStateManager the state manager.
  * @param {angular.Scope} $scope Scope.
  * @param {ngeo.GetBrowserLanguage} ngeoGetBrowserLanguage
+ * @param {gmf.Themes} gmfThemes Themes service.
  * @constructor
  * @extends {gmf.AbstractMobileController}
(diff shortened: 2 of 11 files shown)


adube:
@fgravin:

 * template updated
 * filter option (function) now possible
 * added in mobile app

But in the mobile app it currently does nothing.  Please, help me understand how to make it actually do anything.

adube:
One more important point.  There are themes that have the same name: "Enseignement".  This may result in making the selector not working properly.  Are they supposed to be unique ?  Please, let me know.

fgravin:
> But in the mobile app it currently does nothing

It will update the layer tree, but the layertree is not yet in the modible application.
You must have a reference to the tree in the main controller, and this reference is updated when you select a theme.

Cf
```js
'currentTheme': '=gmfThemeselectorCurrenttheme'
```
and 

```html
 gmf-themeselector-currenttheme="ctrl.theme"
```

Main controller `theme` property will be the source of the layertree.
You may need to declare it in the main controller.

fgravin:
> This may result in making the selector not working properly. Are they supposed to be unique ? Please, let me know

id should be unique, name i don't know.
Is it an error in the theme example ? @sbrunner 
https://github.com/camptocamp/ngeo/blob/master/contribs/gmf/examples/data/themes.json#L479

adube:
>id should be unique, name i don't know.
Is it an error in the theme example ? @sbrunner 
https://github.com/camptocamp/ngeo/blob/master/contribs/gmf/examples/data/themes.json#L479

FYI, the name is used in the url.  If we switch to the id, it would work.  Otherwise, the name would need to be unique too.

sbrunner:
No the id isn't unique but to element with the same id should be the same.

This is because in the database we have a "graph" and the theme webservice return a graph.

theme

adube:
@sbrunner What's the solution, then ?

sbrunner:
... I probably miss something... what's the issue?

adube:
@sbrunner We use the name of the theme for selection purpose, and it's also added in the url.  If themes share the same name, then it causes the issue of both being selected when one is.

sbrunner:
@adube can you call me on hangout ?

fgravin:
merge all your commits in one before we merge your PR

adube:
@fgravin Okay, this is ready for review.

Please, note that I did not use the theme service `find*` methods as I needed to find in the inner themes with the potential filter applied (if any).

adube:
@sbrunner FYI, we're going to ignore the issue of the names that are not currently unique.

fgravin:
Looks good.
Can I merge ?

sbrunner:
Looks good to me to :-)

sbrunner:
We potentially already have the issue with cgxp :-)

adube:
>We potentially already have the issue with cgxp :-)

What do you mean ?

PR #507, opened 2016-01-11
Add legend icon and legend in gmf-layertree

ger-benjamin:
Example : http://ger-benjamin.github.io/ngeo/master/examples/contribs/gmf/layertree.html
Relative to c2c issue: https://github.com/camptocamp/c2cgeoportal/issues/1707

### DIFF
--- a/contribs/gmf/src/directives/layertree.js
+++ b/contribs/gmf/src/directives/layertree.js
@@ -77,7 +77,6 @@ gmfModule.directive('gmfLayertree', gmf.layertreeDirective);
 
 
 /**
- * @param {angular.Scope} $scope The directive's scope.
  * @param {ngeo.LayerHelper} ngeoLayerHelper Ngeo Layer Helper.
  * @param {string} gmfWmsUrl URL to the wms service to use by default.
  * @constructor
@@ -86,7 +85,7 @@ gmfModule.directive('gmfLayertree', gmf.layertreeDirective);
  * @ngdoc controller
  * @ngname gmfLayertreeController
  */
-gmf.LayertreeController = function($scope, ngeoLayerHelper, gmfWmsUrl) {
+gmf.LayertreeController = function(ngeoLayerHelper, gmfWmsUrl) {
 
   /**
    * @type {string}
@@ -146,6 +145,7 @@ gmf.LayertreeController.prototype.getLayer = function(node) {
     promise.then(function(layer) {
       if (goog.isDef(layer)) {
         newLayer.setSource(layer.getSource());
+        newLayer.set('capabilitiesStyles', layer.get('capabilitiesStyles'));
       }
     });
     layer = newLayer;
@@ -155,7 +155,7 @@ gmf.LayertreeController.prototype.getLayer = function(node) {
   this.existingLayers_.push(layer);
 
   // If layer is 'checked', add it on the map.
-  var metadata = node['metadata'];
+  var metadata = node.metadata;
   if (goog.isDefAndNotNull(metadata)) {
     if (metadata['isChecked'] == 'true') {
       this.layerHelper_.addLayerToMap(this.map, layer);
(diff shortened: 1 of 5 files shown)


contribs/gmf/src/directives/layertree.js
```diff
@@ -268,6 +269,120 @@ gmf.LayertreeController.prototype.getNodeState = function(treeCtrl) {
...
+
+/**
+ * Get the WMTS legend URL for the given layer.
+ * @param {ol.layer.Tile} layer Tile layer as returned by the
+ * gmf layerHelper service.
+ * @return {?string} The legend URL or null.
+ * @private
+ */
+gmf.LayertreeController.prototype.getWMTSLegendUrl_ = function(layer) {
+  // FIXME case of multiple styles ?  case of multiple legendUrl ?
```
ger-benjamin: I don't have any example of WMTS layers with multiple styles and, or, legend URL. So I don't know what I must do in this case...

ger-benjamin:
Perhaps, we can move methods `gmf.LayertreeController.prototype.getWMTSLegendUrl_` and `gmf.LayertreeController.prototype.getWMSLegendUrl_` in another file ?

contribs/gmf/src/directives/partials/layertree.html
```diff
@@ -1,10 +1,19 @@
 <span ng-if="::!layertreeCtrl.isRoot"
     ng-class="['depth-' + layertreeCtrl.depth, gmfLayertreeCtrl.getResolutionStyle(layertreeCtrl.node), gmfLayertreeCtrl.getNoSourceStyle(layertreeCtrl)]">
   <a href ng-click="::gmfLayertreeCtrl.toggleActive(layertreeCtrl)">
+    <span class="layerIcon">
```
fgravin: if there is no image ?

contribs/gmf/src/directives/partials/layertree.html
```diff
@@ -1,10 +1,19 @@
...
+      <img ng-src="{{gmfLayertreeCtrl.getLegendIconUrl(layertreeCtrl)}}"></img>
```
fgravin: single time binding ?

contribs/gmf/src/directives/partials/layertree.html
ger-benjamin: I use this class to align the line with others (this `span` element has always a width of 20px).

contribs/gmf/src/directives/partials/layertree.html
ger-benjamin: The icon change with the map scale no ?

contribs/gmf/src/directives/partials/layertree.html
fgravin: You are right

contribs/gmf/src/directives/partials/layertree.html
fgravin: You are right

fgravin: So it does a request per layer each time the resolution change.
It is a lot of requests.

fgravin:
Looks good.
Just a comment on the number of requests made, but i don't think we can avoid that ...
Maybe we can avoid the request if the legend image is collapsed ?

ger-benjamin:
@fgravin The good news is that I saw that in CGXP, we don't watch on current scale to set the little icon's layer image (what I call a legendIcon). We use a url without scale. So I can request my image with a one-time binding.

The bad news is that I don't know how to display legend only when its visible. Display it (or not) initially is easy. But I need to have a `ng-click` function to set the state of the node's image. And whatever I do, request an image as result of the `ng-click` function cause troubles with the collapse animation...

A `ng-if` (and no `ng-click`) on the `img` (or at a different place) doesn't work because... I need a `ng-click` to revalue the `ng-if`...

fgravin:
In the geomapfish demo, when the legend panel is open, then it request legend image on scale change.

A ng-if (and no ng-click) on the img (or at a different place) doesn't work because... I need a ng-click to revalue the ng-if...



> A ng-if (and no ng-click) on the img (or at a different place) doesn't work because... I need a ng-click to revalue the ng-if...

Forget it.
The only way is to observe the `collapsed` class value of the `data-toggle`.
You have the $attrs for a reference on the html node. But you would need to add something in ngeo layer tree, but i don't see how to make something generic right of it.

fgravin:
Be sure about the legend image request on scale change, see from spec or with @sbrunner.
When it's fixed, you can merge.

sbrunner:
- The request on the layer icon (using a rule) don't depends on the scale,
- The request for the legend depends on the scale,

PR #514, opened 2016-01-13
Update data/theme.json - Metadata keys use camelcasing now

ger-benjamin:
TODO

* [x] Debug obscur error from Angular (can't parse *something* in the new json) 
* [x] Adapt layertree


fredj:
+1

PR #518, opened 2016-01-14
mobile-measure

adube:
This PR adds a new mobile draw interaction and a mobile measure length interaction in ngeo, and a mobile measure length directive in gmf.

When the mobile draw is active, the directive appropriately displays start, add, clear, terminate or close buttons, depending on the state of the mobile draw interaction. 

 - [x] Mobile draw interaction
 - [x] Mobile measure length interaction
 - [x] Mobile measure length directive in gmf
 - [x] Example
 - [x] Build
 - [x] Mobile measure point interaction
 - [x] Mobile measure point directive in gmf
 - [x] Manage activation
 - [x] Add configuration (option) to determine the number of decimals
 - [x] Point measurement in meters, not lon/lat
 - [x] Use custom style in example
 - [x] Create a gmf-altitude service and use it in point measure directive
 - [x] Use the altitude service configured on demo server with the example
 - [x] Travis build - What is Coverall.io ?
 - [x] Review
 - [x] Altitude service error (same origin policy)

Live example:

 * http://adube.github.io/ngeo/mobile-measure/examples/contribs/gmf/mobilemeasure.html


src/services/decoratemobiledrawinteraction.js
```diff
@@ -0,0 +1,93 @@
...
+        // while angular is out of scope, which means that it doesn't detect
+        // it being changed. To work around this, the change event is listened
+        // and the property is manually changed
+        //
+        // A better approach to avoid doing this would be great, as this is
+        // not pretty at all...
+        Object.defineProperty(interaction, 'dirty', {
+          writable: true,
+          value: false
+        });
```
adube: See here.

src/services/decoratemobiledrawinteraction.js
```diff
@@ -0,0 +1,93 @@
...
+              Object.defineProperty(interaction, 'dirty', {
+                value: dirty
+              });
+
+              // this is where the angular scope is forced to be applied. We
+              // only need to do this when dirty, as going to "no being dirty"
+              // is made by a click on a button where Angular is within scope
+              if (dirty) {
+                $rootScope.$apply();
+              }
```
adube: And here.

src/services/decoratemobiledrawinteraction.js
```diff
@@ -0,0 +1,93 @@
+goog.provide('ngeo.DecorateMobileDrawInteraction');
```
fgravin: I would not do a specific decorator for this.
ngeo decorators are here to make generic ol3 components interfacable with angular. `decorateMobileDrawInteraction` is specific for a specific interaction `ngeo.interaction.MobileDraw`. I don't know what you think about this, but i would put the define properties inside the interaction itself, during creation.

src/services/decoratemobiledrawinteraction.js
```diff
@@ -0,0 +1,93 @@
...
+        });
+
+
+        goog.events.listen(
+            interaction,
+            ol.Object.getChangeEventType(
+                ngeo.interaction.MobileDrawProperty.DIRTY),
+            function() {
+              var dirty = interaction.getDirty();
+              Object.defineProperty(interaction, 'dirty', {
```
fgravin: Why do you define the property on event ? You could define it on object creation, just once.

src/services/decoratemobiledrawinteraction.js
```diff
@@ -0,0 +1,93 @@
...
+                value: dirty
+              });
+
+              // this is where the angular scope is forced to be applied. We
+              // only need to do this when dirty, as going to "no being dirty"
+              // is made by a click on a button where Angular is within scope
+              if (dirty) {
+                $rootScope.$apply();
```
fgravin: For all ol3 event, controller code is responsible of calling the `scope.$apply() method`.
Here, the directive that will manage the state of the button, depending on dirty value, will do the apply.

jlap:
Here we go, my latest commit in this PR makes the mobile measure length interaction & directive ready to be reviewed. I still have a few issues with the build, but one of them @adube has fixed in #540 (using partial in gmf examples) so that will help a lot. Because of that the tool can only be tested in dev mode for now.

The `mobilemeasurelength` example displays how the tool works.

 - [x] Mobile draw interaction
 - [x] Mobile measure length interaction
 - [x] Mobile measure length directive in gmf
 - [x] Example
 - [ ] Build
 - [ ] Review

Should I add tests to this list ? 

Please try it out and review. I'll get on fixing the build once #540 is merged!

contribs/gmf/examples/mobilemeasurelength.html
```diff
@@ -0,0 +1,70 @@
...
+<html ng-app='app'>
+  <head>
+    <title>Mobile Measure Length example</title>
+    <meta charset="utf-8">
+    <meta name="viewport"
+          content="initial-scale=1.0, user-scalable=no, width=device-width">
+    <meta name="mobile-web-app-capable" content="yes">
+    <link rel="stylesheet" href="../../../node_modules/openlayers/css/ol.css" type="text/css">
+    <link rel="stylesheet" href="../../../node_modules/font-awesome/css/font-awesome.css" type="text/css">
+    <link rel="stylesheet" href="http://getbootstrap.com/dist/css/bootstrap.min.css" type="text/css">
```
fgravin: get from local (node_modules)

fgravin:
Please provide an url to the example from your gh-pages

adube:
@fgravin We can't provide urls.  It won't work unless we're able to build.  The query needs to be merged first.

fgravin:
Ok it's merged

jlap:
Thanks a lot for the merged, helped a lot. 

Here's the example link: https://adube.github.io/ngeo/mobile-measure/examples/contribs/gmf/mobilemeasurelength.html

kalbermattenm:
Thanks a lot for the great work !

Just a little comment... I find it quite disturbing that everything disappears from the map when you click on "Set as starting point". Was that done on purpose ?

I think the client looses his "starting point context" by doing so... no ?

rbovard:
+1 (also after adding a new point)

I think the tooltip is not useful in mobile context.

contribs/gmf/examples/mobilemeasurelength.js
```diff
@@ -0,0 +1,112 @@
...
+
+
+/**
+ * App-specific directive wrapping the line measure tools. The directive's
+ * controller has a property "map" including a reference to the OpenLayers
+ * map.
+ *
+ * @return {angular.Directive} The Directive Definition Object.
+ */
+app.mobileMeasureLengthDirective = function() {
```
fgravin: You should avoid to introduce complexity with a directive in gmf examples.
Usually, in ngeo example we provide directives to show a way on how to use ngeo core components.
Gmf directives are another way to use ngeo core components, the example should just show the directive.

In other terms, ngeo examples directive = gmf directive, so don't add directive in gmf examples.

contribs/gmf/src/directives/mobilemeasurelength.js
```diff
@@ -0,0 +1,216 @@
...
+        this.valid = mobileDrawLine.getValid();
+      },
+      this
+  );
+
+  ol.events.listen(
+      mobileDrawLine,
+      ol.Object.getChangeEventType(ol.interaction.InteractionProperty.ACTIVE),
+      function() {
+        this.active = mobileDrawLine.getActive();
```
fgravin: I would prefer to use the `ngeoInteractionDecorator`, that set active property on the interaction.
`this.active`would be reachable through `this.mobileDrawLine.active`.

Could be the same for valid and drawing (add a new decorator).

contribs/gmf/src/directives/mobilemeasurelength.js
```diff
@@ -0,0 +1,216 @@
...
+gmf.mobileMeasureLengthDirective =
+    function(gmfMobileMeasureLengthTemplateUrl) {
+  return {
+    restrict: 'A',
+    scope: {
+      'map': '=gmfMobileMeasureLengthMap',
+      'mobileDrawLine': '=gmfMobileMeasureLengthDrawLine'
+    },
+    controller: 'GmfMobileMeasureLengthController',
+    controllerAs: 'measureLengthCtrl',
```
fgravin: i think it's a bit too long for nothing, when in a directive with isolate scope, you can use `ctrl`, or `mlCtrl`.
It's not dramatic but it's maybe better =)

contribs/gmf/src/directives/mobilemeasurelength.js
adube: We discussed about this already, remember ?  I had a decorator, but I had issues with the fact that the mobileDrawLine is an ol3 control and it doesn't natively contain an `active` property.  Well, it shouldn't because it already has an inner property set with the `set` method.

We decided together that we were to remove the decorator and put the logic in the directive instead.  That point is still valid, in my opinion.  Please, validate.

jlap: Do you mean like this?
https://github.com/adube/ngeo/blob/754ee10b4b825ed7bd7f79b9140dfa5a533bce4b/src/services/decoratemobiledrawinteraction.js 

Because I actually removed that decorator after this comment:
https://github.com/camptocamp/ngeo/pull/518#discussion-diff-49983744R1 

Or maybe you mean another decorator?

contribs/gmf/examples/mobilemeasurelength.js
adube: Neither @jlap nor I understand what you mean.  Please, contact @jlap about this.

jlap:
> Just a little comment... I find it quite disturbing that everything disappears from the map when you click on "Set as starting point". Was that done on purpose ?

> I think the client looses his "starting point context" by doing so... no ?

Yeah you're right I'll take a look, for the starting point it's definitely weird, for subsequent points it isn't so bad since you can see the line. But I think I'll make a point appear every time you hit "Add new point", since that's what the button says, plus in [1.6](https://geomapfish-demo.camptocamp.net/1.6/wsgi/mobile/?theme=Paysage&baselayer_ref=asitvd.fond_couleur&map_x=650196.29473857&map_y=47619.969846589&map_zoom=8&tree_layers=zoo%2Calpine_hut) the points are displayed on the map along with the line so it'd make sense to do the same thing. 

> I think the tooltip is not useful in mobile context.

You're right, it's not. I didn't give it a lot of attention though, because the tooltip doesn't appear at all when you're using a phone (it triggers on hover).

jlap:
I just updated the example with some UI changes: http://adube.github.io/ngeo/mobile-measure/examples/contribs/gmf/mobilemeasurelength.html 

There are two changes to notice:
- No more disappearing points! All points are kept on the line.
- No more tooltip when viewing the mobile example from a desktop.

Still have some requested changes to do in the code.

Also, I am not familiar at all with coveralls, so I'm a bit lost as to why it fails, who should I poke for help about that ? 

Thanks for everyone who left comments, please go check out the example to see if the change is good!

rbovard:
Thanks for this update, it's much better now!

jlap:
Just updated the mobile length directive so it uses an interaction decoractor instead of manually listening to activation of the mobile draw interaction.

adube:
The point measure has been added as well (see the updated live demo).  It's also ready to receive comments/reviews.

adube:
I made some additions.  You can see them in action in the live example:

 * point measure in meters instead of lon/lat
 * add 'decimal' options (used in the example with 2 as option set)
 * custom style to have a better look and feel of the tool
 * altitude added to the point measure (currently returning dummy data, still in progress)

contribs/gmf/src/directives/partials/mobilemeasurelength.html
```diff
@@ -0,0 +1,30 @@
+<a class="btn btn-default"
+   ng-if="ctrl.drawing && (!ctrl.valid)"
```
fgravin: does `ctrl.valid` still exists ? Or get the valid from the interaction ?

contribs/gmf/src/directives/mobilemeasurelength.js
```diff
@@ -0,0 +1,193 @@
...
+ * @ngInject
+ * @ngdoc directive
+ * @ngname gmfMobileMeasureLength
+ */
+gmf.mobileMeasureLengthDirective =
+    function(gmfMobileMeasureLengthTemplateUrl) {
+  return {
+    restrict: 'A',
+    scope: {
+      'drawInteraction': '=gmfMobileMeasureLengthDrawInteraction'
```
fgravin: Why the directive is not responsible for creating the interaction ?

contribs/gmf/src/directives/partials/mobilemeasurelength.html
fgravin: sorry it was for `active` not valid
but the example http://adube.github.io/ngeo/mobile-measure/examples/contribs/gmf/mobilemeasurelength.html never shows the buttons.

contribs/gmf/src/directives/mobilemeasurepoint.js
```diff
@@ -0,0 +1,202 @@
...
+ * @ngdoc directive
+ * @ngname gmfMobileMeasurePoint
+ */
+gmf.mobileMeasurePointDirective =
+    function(gmfMobileMeasurePointTemplateUrl) {
+  return {
+    restrict: 'A',
+    scope: {
+      'map': '=gmfMobileMeasurePointMap',
+      'measure': '=gmfMobileMeasurePointMeasureInteraction',
```
fgravin: don't understand this either
I don't think the main controller should pass those 2 interactions.

contribs/gmf/src/directives/partials/mobilemeasurelength.html
adube: This is not the correct example link.  See in the PR description.

contribs/gmf/src/directives/mobilemeasurepoint.js
adube: The map is needed for the altitude service.
The measure is used to be listened to on 'active' change.

contribs/gmf/src/directives/mobilemeasurelength.js
adube: The draw interaction is automatically created by the measure interaction.

adube:
@fgravin This is ready for merge, however I'm not able to use the altitude service from geomapfish-demo due to the 'same-origin' policy error.  We could wait for this to be fixed by @sbrunner before merging.

sbrunner:
I'm having a look on it :-)

sbrunner:
@adube I fixed the CORS issue, but I can't see the height anywhere...

contribs/gmf/src/directives/mobilemeasurelength.js
```diff
@@ -0,0 +1,270 @@
...
+ * @ngdoc directive
+ * @ngname gmfMobileMeasureLength
+ */
+gmf.mobileMeasureLengthDirective =
+    function(gmfMobileMeasureLengthTemplateUrl) {
+  return {
+    restrict: 'A',
+    scope: {
+      'active': '=gmfMobileMeasureLengthActive',
+      'decimals': '=?gmfMobileMeasureLengthDecimals',
```
fgravin: don't do a 2 way binding for decimals
either use `'@'` or get the value from `$attrs`

fgravin:
Looks good @adube , just one minor comment.

adube:
@fgravin You can merge, now.

fgravin:
@adube thanks for your work !

PR #524, opened 2016-01-18
Use polyfill to shim ES5 'bind' (not supported by PhantomJS)

pgiraud:
As stated by @tsauerwein:
> Because ngeo does not use a specifc version of ol3, it gets the current master of ol3. A PR related to .bind was merged 3 days ago. PhantomJS does not support .bind, you will have to use a shim, see:
https://github.com/openlayers/ol3/blob/db3ad70a3d7598254cbf11b7a6540b9fba3773ab/config/examples/example.html#L14

### DIFF
--- a/karma-conf.js
+++ b/karma-conf.js
@@ -29,6 +29,7 @@ module.exports = function(config) {
       'node_modules/angular/angular.js',
       'node_modules/angular-gettext/dist/angular-gettext.js',
       'node_modules/angular-mocks/angular-mocks.js',
+      'node_modules/polyfill/es5/Function.prototype.bind.js',
       'test/spec/beforeeach.js',
       'test/spec/data/*.js',
       'test/spec/**/*.spec.js',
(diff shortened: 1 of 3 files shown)


pgiraud:
Please review.

gberaudo:
Do we really need to depend on an external CDN for this?
Couldn't we use a local file instead?

pgiraud:
Now using a local version of polyfill.

karma-conf.js
```diff
@@ -66,6 +66,7 @@ module.exports = function(config) {
         watched: true,
         served: true
       },
+      'https://cdn.polyfill.io/v2/polyfill.min.js?features=Function.prototype.bind|gated',
```
gberaudo: Is this line still necessary?

pgiraud: It has to be changed, indeed.

gberaudo:
LGTM, thanks @pgiraud.

PR #526, opened 2016-01-19
Use local bootstrap

ger-benjamin:
Ref to Issue https://github.com/camptocamp/c2cgeoportal/issues/1937
...And local Font-Awesome


sbrunner:
Looks good, merge :-)

PR #527, opened 2016-01-19
Replace gjslint and jshint with eslint

### LINKED ISSUE #510: Use eslint instead of jshint and closure linter

fredj:
Fixes #510


Makefile
```diff
@@ -117,7 +117,7 @@ build-gmf-mobile-app: $(addprefix contribs/gmf/build/mobile,.js .css) \
 check-examples: $(BUILD_EXAMPLES_CHECK_TIMESTAMP_FILES)
 
 .PHONY: lint
-lint: .build/python-venv/bin/gjslint .build/node_modules.timestamp .build/gjslint.timestamp .build/jshint.timestamp
+lint: .build/node_modules.timestamp .build/eslint.timestamp
```
sbrunner: Why do we have a dependency on `.build/node_modules.timestamp`?

fredj: To make sure eslint is installed

Makefile
```diff
@@ -158,12 +158,8 @@ gh-pages: .build/ngeo-$(GITHUB_USERNAME)-gh-pages \
 .build/ngeo-$(GITHUB_USERNAME)-gh-pages:
 	git clone --branch gh-pages $(GIT_REMOTE_URL) $@
 
-.build/gjslint.timestamp: $(SRC_JS_FILES) $(EXPORTS_JS_FILES) $(EXAMPLES_JS_FILES) $(GMF_SRC_JS_FILES) $(GMF_EXAMPLES_JS_FILES) $(GMF_APPS_MOBILE_JS_FILES)
-	.build/python-venv/bin/gjslint --jslint_error=all --strict --custom_jsdoc_tags=event,fires,function,classdesc,api,observable,example,module,ngdoc,ngname $?
-	touch $@
-
-.build/jshint.timestamp: $(SRC_JS_FILES) $(EXPORTS_JS_FILES) $(EXAMPLES_JS_FILES) $(GMF_SRC_JS_FILES) $(GMF_EXAMPLES_JS_FILES) $(GMF_APPS_MOBILE_JS_FILES)
-	./node_modules/.bin/jshint --verbose $?
+.build/eslint.timestamp: $(SRC_JS_FILES) $(EXPORTS_JS_FILES) $(EXAMPLES_JS_FILES) $(GMF_SRC_JS_FILES) $(GMF_EXAMPLES_JS_FILES) $(GMF_APPS_MOBILE_JS_FILES)
```
sbrunner: missing dependency on `.build/node_modules.timestamp` (or on `node_modules/.bin/eslint`)

sbrunner:
I think that we should add a variable like `ALL_JS_FILES` and place the dependency at the right place...

Makefile
sbrunner: This should be in the `.build/eslint.timestamp` rule ...

sbrunner:
We should also be able use `$(filter-out $?)`?

fredj:
> We should also be able use `$(filter-out .build/node_modules.timestamp $?)`?

done

sbrunner:
Thanks :-)

fredj:
Will wait for the eslint 2.0 release

fredj:
Ready to be reviewed

sbrunner:
Looks good :-), but it should be rebased ..., Many thanks @fredj

Issue #399, opened 2015-11-19
mobile app scrolling issues


fredj:
from @pgiraud:  I figured that we have scrolling issues on mobile phones. Ideally, only the content of the sidebars should scroll, not the sidebar nor the map.

Problems:

* [x] The sidebar containers are not scrollable, for example the one containing the themes is cut off (see https://github.com/camptocamp/ngeo/pull/522)
* [x] On iOS: The bottom bar hides a part of the map (when in portrait mode).
* [x] On iOS: The page can be "scrolled" when a sidebar is shown (this [library](http://www.jaridmargolin.com/bouncefix.js/) might be interesting)

pgiraud:
I think I have a fix for the height of the map and scaleline being hidden.

PR #538, opened 2016-01-21
Improve mobile application for iOS

tsauerwein:
* Make sure the map container is fully visible (e.g. the scalebar is not hidden)
* Avoid that the sidebars are visible when they are not active and the map container is dragged.

Related to https://github.com/camptocamp/ngeo/issues/399


pgiraud:
Thanks. Please merge.

PR #548, opened 2016-01-25
Redirect gmf example to https

sbrunner:
related to https://github.com/camptocamp/c2cgeoportal/issues/1667

### DIFF
--- a/contribs/gmf/examples/https.js
+++ b/contribs/gmf/examples/https.js
@@ -1,0 +1,9 @@
+/*
+ * Redirect to https
+ */
+if (window.location.protocol == 'http:' &&
+    window.location.port != '3000') {
+  var restOfUrl = window.location.href.substr(5);
+  /** @type {Location} */
+  window.location = 'https:' + restOfUrl;
+}
--- a/Makefile
+++ b/Makefile
@@ -15,6 +15,7 @@ GMF_APPS_DESKTOP_JS_FILES := $(shell find contribs/gmf/apps/desktop/js -type f -
 GMF_APPS_MOBILE_LESS_FILES := $(shell find contribs/gmf/less -type f -name '*.less')
 GMF_APPS_DESKTOP_LESS_FILES := $(shell find contribs/gmf/less -type f -name '*.less')
 GMF_APPS_LIBS_JS_FILES += \
+	contribs/gmf/examples/https.js \
 	node_modules/jquery/dist/jquery.min.js \
 	node_modules/angular/angular.min.js \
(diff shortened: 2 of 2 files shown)


sbrunner:
Can be tested here:
http://camptocamp.github.io/ngeo/https/examples/contribs/gmf/apps/mobile/index.html

externs/closure-compiler.js
```diff
@@ -16,3 +16,11 @@ Navigator.prototype.systemLanguage;
  * @type {string}
  */
 Navigator.prototype.userLanguage;
+
+/**
```
pgiraud: 2 blank lines?

contribs/gmf/examples/https.js
```diff
@@ -0,0 +1,9 @@
+/*
+ * Redirect to https
+ */
+if (window.location.protocol == 'http:' &&
+    window.location.hostname != 'localhost') {
+  var restOfUrl = window.location.href.substr(5);
+  /** @type {string|WindowLocation} */
```
pgiraud: Is this really required?
Did you have any issue?

pgiraud:
With this pull request I can't test the hosted examples served with `make serve` anymore. Well I can if I use `localhost`. But this is not possible if I want to try on my device. In this case, I usually use a url like the following.
http://wrk31.wrk.cby.camptocamp.com:3000/.build/examples-hosted/contribs/gmf/apps/mobile/

In this case, I'm redirected to `https` which fails.

contribs/gmf/examples/https.js
sbrunner: If I build it I will have an error... (in the current case I don't build it!)

pgiraud: When is this code supposed to be built?
What is the error you get?

externs/closure-compiler.js
sbrunner: fixed

sbrunner:
@pgiraud I change to use the port to the the "serve" test :-)

contribs/gmf/examples/https.js
pgiraud: Oh, I just got the error when running `make check`.
However, we can use the `Location` type. It's provided by the compiler.
```javascript
/** @type {Location} */
```
See
https://github.com/google/closure-compiler/blob/master/externs/browser/window.js#L56

pgiraud:
Did you see my comment about the `Location` type?

pgiraud:
Good for me. Please merge.

sbrunner:
Thanks @pgiraud :-)

PR #550, opened 2016-01-25
Closing side navigation when dragging the map in

### LINKED ISSUE #549: [mobile] Allow user to close nav on swipe
See https://codyhouse.co/demo/mega-site-navigation/index.html

pgiraud:
Fixes #549

### DIFF
--- a/contribs/gmf/src/controllers/abstractmobile.js
+++ b/contribs/gmf/src/controllers/abstractmobile.js
@@ -12,6 +12,8 @@ goog.require('gmf.mobileMeasureLengthDirective');
 goog.require('gmf.mobileMeasurePointDirective');
 /** @suppress {extraRequire} */
 goog.require('gmf.mobileNavDirective');
+goog.require('goog.fx.Dragger');
+goog.require('goog.math.Rect');
 /** @suppress {extraRequire} */
 goog.require('ngeo.btnDirective');
 /** @suppress {extraRequire} */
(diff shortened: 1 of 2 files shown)


contribs/gmf/apps/appmodule.js
```diff
@@ -15,4 +15,4 @@ goog.require('gmf');
 /**
  * @type {!angular.Module}
  */
-var appModule = angular.module('app', [gmfModule.name]);
+var appModule = angular.module('app', [gmfModule.name, 'ngTouch']);
```
sbrunner: Shouldn't be in the gmf module ?

pgiraud:
With this pull request, the user can close side navs by swipping left or right on the map overlay.

This is not perfect since the swipe is taken into account only on release (touch end). It means that `ngTouch`'s `ngSwipeLeft` doesn't behave like `jQuery Mobile`'s 'swipeleft` event.

An other possible improvement would be to be able to pan (drag) the map and overlay. ie. that the map would follow the finger. We should be able to use `HammerJS` for that.

Please review.

contribs/gmf/apps/appmodule.js
pgiraud: You're probably right. But this means that we will have to load `ngTouch` in every single example or app in `gmf`.
In my opinion, the `ngTouch` module should only be required for the mobile app.

By the way, the desktop application is broken since dependency on `ngTouch` is not satisfied.

sbrunner: Why not but:
* This file is shared with the desktop app
* Done like this (As I understand) it don't add any code in the build file
* This will be copy in the client application => more difficult to maintain...

sbrunner:
Steal in WIP ?

ger-benjamin:
~same discution on: https://github.com/camptocamp/ngeo/pull/713

pgiraud:
I worked again on this pull request and instead of relying on swipe, I preferred to play with dragging.
Now the user can drag the map (and it's overlay) back to hide the nav and reveal the map again.
Demo: https://pgiraud.github.io/ngeo/close_nav_on_swipe/examples/contribs/gmf/apps/mobile
Please review.

pgiraud:
The performance decreases a bit here since we don't use `translateX` anymore.

pgiraud:
Still not perfect since if you drag to the opposite way a sufficient amount you may be able to hide the navigation.

pgiraud:
This is now ready for tests and reviews.
Demo: https://pgiraud.github.io/ngeo/close_nav_on_swipe/examples/contribs/gmf/apps/mobile

contribs/gmf/src/controllers/abstractmobile.js
```diff
@@ -145,6 +147,36 @@ gmf.AbstractMobileController = function(config, $scope, $injector) {
 
   goog.base(
       this, config, $scope, $injector);
+
+
+  var dragEl = document.querySelector('main');
+  var handleEl = document.querySelector('main .overlay');
+  this.dragger = new goog.fx.Dragger(dragEl, handleEl);
```
pgiraud: Indeed. Same to be done for `this.navWidth`.

fredj:
LGTM

PR #551, opened 2016-01-25
Rename ngeoModule and gmfModule

fredj:
to be discussed

### DIFF
--- a/buildtools/templatecache.mako.js
+++ b/buildtools/templatecache.mako.js
@@ -24,10 +24,6 @@
  * GENERATED FILE. DO NOT EDIT.
  */
 
-/**
- * This goog.require is needed because it provides ngeoModule.
- * @suppress {extraRequire}
- */
 goog.require('${app}');
 
 (function() {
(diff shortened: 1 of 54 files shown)


sbrunner:
And what about the provide/requires?

gberaudo:
As a second thought, I think it will be more consistent with `goog` not to provide/require the module explicitely. So +1 for @fredj suggestion.

fgravin:
> with goog not to provide/require the module explicitely

What do you mean ?
I don't really see the issue about the module variable.

sbrunner:
Same as @gberaudo :-)

gberaudo:
I mean it is fine to only require the namespace and get the module like this:
```
goog.require('ngeo');
angular.module('myapp', [ngeo.module.name]);
```

sbrunner:
Steal in WIP ?

fredj:
If everyone agrees, this can be merged.

fgravin:
Go ahead thanks

PR #553, opened 2016-01-26
Use visibility to 'add/remove' layers from layertree

ger-benjamin:
Fix: https://github.com/camptocamp/c2cgeoportal/issues/1969


sbrunner:
Looks good :-)

PR #554, opened 2016-01-26
Reduced zoom value when recentering. cf c2cgeoportal #1954.

ochriste:
fix https://github.com/camptocamp/c2cgeoportal/issues/1954


sbrunner:
Looks good, merge :-)

PR #568, opened 2016-02-01
Use phantomjs v2.1.3

fredj:
And remove the now unneeded Function.prototype.bind polyfill.

### DIFF
--- a/contribs/gmf/examples/mobilebackgroundlayerselector.html
+++ b/contribs/gmf/examples/mobilebackgroundlayerselector.html
@@ -63,7 +63,7 @@
     <gmf-map gmf-map-map="ctrl.map"></gmf-map>
     <script src="../../../node_modules/angular/angular.js"></script>
     <script src="../../../node_modules/angular-gettext/dist/angular-gettext.js"></script>
-    <script src="http://cdnjs.cloudflare.com/ajax/libs/proj4js/2.2.1/proj4.js" type="text/javascript"></script>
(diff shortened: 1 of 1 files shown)


fredj:
The `npm prune` in `travis.yml` is here to remove old packages from the travis cache

fgravin:
+1 looks good thanks

Makefile
```diff
@@ -44,7 +44,6 @@ EXAMPLE_HOSTED_REQUIREMENTS = .build/examples-hosted/lib/ngeo.js \
 	.build/examples-hosted/lib/watchwatchers.js \
 	.build/examples-hosted/lib/typeahead.bundle.min.js \
 	.build/examples-hosted/lib/proj4.js \
-	.build/examples-hosted/lib/Function.prototype.bind.js \
```
sbrunner: Should be removed :-)

sbrunner: allready don, :-)

.travis.yml
```diff
@@ -6,6 +6,9 @@ cache:
   directories:
   - node_modules
 
+before_script:
+- npm prune
```
sbrunner: Why this ?

fredj: To cleanup the node_modules directory (from the travis cache)

fredj: See https://docs.npmjs.com/cli/prune

fredj:
Travis fails because of #572

.travis.yml
sbrunner: OK, thanks :-)

PR #521, opened 2016-01-15
Use node modules version of proj4js

pgiraud:

Issue #570, opened 2016-02-01
GMF Layertree hosted example is broken


pgiraud:
https://camptocamp.github.io/ngeo/master/examples/contribs/gmf/layertree.html
GET https://camptocamp.github.io/ngeo/master/examples/contribs/gmf/data/themes.json 404 (Not Found)

pgiraud:
Same error for https://camptocamp.github.io/ngeo/master/examples/contribs/gmf/themeselector.html

sbrunner:
It's ma fault ..., I will fix it :-)

fredj:
@sbrunner see #573

PR #572, opened 2016-02-01
Use node modules version of proj4js

### LINKED ISSUE #571: GMF mobilebackgroundlayerselector example is broken
Mixed content issue; `proj4.js` is loaded over HTTP instead of HTTPS

fredj:
Fixes #571 
See #521


pgiraud:
Please merge.

fredj:
Travis fails because of #570

sbrunner:
Shouldn't we do something like this to have a referer?
https://gist.github.com/papoms/3453137
with a referer like https://camptocamp.github.io/ngeo/ ?

PR #574, opened 2016-02-01
Use phantomjs v2.1.3

fredj:
And remove the now unneeded Function.prototype.bind polyfill.

(was #568)


sbrunner:
I think that you can directly merge it if Travis is happy :-)

PR #575, opened 2016-02-01
Make the default theme configurable

### LINKED ISSUE #557: The default theme should be configurable
This shouldn't be configurable in the html?
https://github.com/camptocamp/ngeo/blob/master/contribs/gmf/src/directives/themeselector.js#L84

fredj:
Fixes #557

### DIFF
--- a/contribs/gmf/src/directives/themeselector.js
+++ b/contribs/gmf/src/directives/themeselector.js
@@ -7,7 +7,9 @@ goog.require('ngeo.Location');
 
 
 /**
- *
+ * @htmlAttribute {string} gmf-themeselector-defaulttheme The default theme.
+ * @htmlAttribute {string} gmf-themeselector-currenttheme The selected theme.
+ * @htmlAttribute {Function} gmf-themeselector-filter The themes filter.
  * @return {angular.Directive} The directive specs.
  * @ngInject
  * @ngdoc directive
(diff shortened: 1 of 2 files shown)


fgravin:
I think we were more talking about an `angular.value` in the index.html file, generated on page loading.

See https://github.com/Geoportail-Luxembourg/geoportailv3/blob/master/geoportailv3/templates/index.html#L275

sbrunner:
Personally I think about this ...

fgravin:
What this ? the PR or my comment ?
Where the default theme will come from ? DB ?

fredj:
What about adding a `defaultTheme` in the response from the 'themes' webservice?

sbrunner:
this => the Pull request :-)

sbrunner:
In standard case it's a configuration from the viewer.js.

fgravin:
Ok i don't mind.

contribs/gmf/src/directives/themeselector.js
```diff
@@ -18,6 +18,7 @@ gmf.themeselectorDirective = function() {
     restrict: 'E',
     controller: 'gmfThemeselectorController',
     scope: {
+      'defaultTheme': '@gmfThemeselectorDefaulttheme',
```
sbrunner: The `@htmlAttribute` should be added, see: https://github.com/camptocamp/ngeo/blob/master/contribs/gmf/src/directives/search.js#L47-L50

sbrunner: Then for me you can merge :-)

sbrunner:
Thanks :-)

Issue #558, opened 2016-01-26
The controlleurs should be more extandable


sbrunner:
https://github.com/camptocamp/ngeo/tree/master/contribs/gmf/src/controllers
e.-g.:
positionFeatureStyle, accuracyFeatureStyle, map.view, map.controls,.

PR #581, opened 2016-02-02
Be able to change everything what we needs in the application

sbrunner:
fix [#558]

### DIFF
--- a/contribs/gmf/src/controllers/abstractdesktop.js
+++ b/contribs/gmf/src/controllers/abstractdesktop.js
@@ -16,9 +16,13 @@ goog.require('ol.Map');
 goog.require('ol.View');
 goog.require('ol.control.ScaleLine');
 goog.require('ol.control.Zoom');
+// TODO: Remove when we add the base layer switcher
+goog.require('ol.format.WMTSCapabilities');
 goog.require('ol.interaction');
+// TODO: Remove when we add the base layer switcher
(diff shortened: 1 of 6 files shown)


contribs/gmf/src/controllers/abstractdesktop.js
```diff
@@ -54,34 +59,46 @@ gmf.AbstractDesktopController = function(
...
     ],
-    interactions: ol.interaction.defaults({
+    interactions: config.mapInteractions || ol.interaction.defaults({
       pinchRotate: false,
       altShiftDragRotate: false
     })
   });
 
+  // TODO: Remove when we add the base layer switcher
+  var parser = new ol.format.WMTSCapabilities();
```
ger-benjamin: Just for information (because that's a "remove-me code block"), you can do the same with https://github.com/camptocamp/ngeo/blob/master/src/services/layerHelper.js#L107

ger-benjamin:
Good, but I don't really like to have a ~"global value" that is named `Config`. Don't you think that something like `GmfConfig` is better ?

contribs/gmf/src/controllers/abstractdesktop.js
sbrunner: I will not change it because for that we needs a temporary injection ...

sbrunner:
It's not a global think that can be injected, it's just a type ...

PR #585, opened 2016-02-03
Fix the ol3 version to fix the master

sbrunner:


pgiraud:
I expect this fixes the recent build issues.
If so, please merge when travis is happy.

adube:
Hi guys.  I have 2 commits that were recently merged in ol3 master that are required for the query tool to work.   This change will break the query tool.

What should we do ?

gberaudo:
It is possible to depend on a specific git commit.
For example: `git://github.com/openlayers/ol3#473cbba26e005cf200acc779e2d69f0ba9524d13`.

sbrunner:
temporary +1 @gberaudo

PR #593, opened 2016-02-04
Use ol.events instead of goog.events

fredj:
Needs tests


sbrunner:
You should also update openlayers ...

fredj:
> You should also update openlayers ...

but do we want to go back to ol3 master ?

sbrunner:
(or an hash) temporary yes because @adube as needs an unreleased version...

adube:
Can we merge this soon ?

fredj:
Let's go

petzlux:
Is this breaking our build ? https://travis-ci.org/Geoportail-Luxembourg/geoportailv3/builds/107210332

Issue #576, opened 2016-02-01
Layers with same Id should be duplicated in GMF layer tree


fgravin:
In https://camptocamp.github.io/ngeo/master/examples/contribs/gmf/apps/mobile/index.html select the OSM theme and watch the layer tree.
The `osm_time` layer is used in 2 groups, but all interactions are bound to the same layer (visibility).

In 1.6, the layers are 2 disctincts layers.

This will may be fixed when we fix the layergroup issue.

sbrunner:
This one: https://github.com/camptocamp/c2cgeoportal/issues/1970 ?

fgravin:
Yes, we need to check #576 after https://github.com/camptocamp/c2cgeoportal/issues/1970 is done

sbrunner:
OK, thanks :-)

ger-benjamin:
Close with #601

PR #601, opened 2016-02-04
Layertree one wms

ger-benjamin:
Rel to: #576 and https://github.com/camptocamp/c2cgeoportal/issues/1970

Example: https://ger-benjamin.github.io/ngeo/master/examples/contribs/gmf/layertree.html
And: https://ger-benjamin.github.io/ngeo/master/examples/contribs/gmf/apps/mobile/index.html

Please review.
(Ignore coveralls ? I think it is not not happy because I've deleted some tests (but relatives methods too))

### DIFF
--- a/contribs/gmf/src/directives/layertree.js
+++ b/contribs/gmf/src/directives/layertree.js
@@ -18,7 +18,7 @@ gmf.module.value('gmfLayertreeTemplate',
       var subTemplateUrl = gmf.baseTemplateUrl + '/layertree.html';
       return '<div ngeo-layertree="gmfLayertreeCtrl.tree" ' +
           'ngeo-layertree-map="gmfLayertreeCtrl.map" ' +
-          'ngeo-layertree-nodelayer="gmfLayertreeCtrl.getLayer(node)" ' +
+          'ngeo-layertree-nodelayer="gmfLayertreeCtrl.getLayer(node, depth)" ' +
           'ngeo-layertree-listeners="gmfLayertreeCtrl.listeners(treeScope, ' +
           'treeCtrl)" ' +
           'ngeo-layertree-templateurl="' + subTemplateUrl + '"' +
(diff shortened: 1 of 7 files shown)


fgravin:
i'll review it later in the evening

contribs/gmf/src/directives/layertree.js
```diff
@@ -149,87 +149,272 @@ gmf.LayertreeController = function($http, $sce, ngeoCreatePopup,
...
-  layer = this.layerHelper_.findLayer(this.existingLayers_,
-      this.layerHelper_.makeHelperID(layerURL, layerName));
   if (goog.isDefAndNotNull(layer)) {
-    return layer;
+    layer.set('querySourceId', node.id);
+    layer.set('layerName', node.name);
+    // The layer must be upper than the background
+    layer.setZIndex(1);
+    // Add the new layer on the map but behind (before) others layers
+    this.map.getLayerGroup().getLayers().insertAt(0, layer);
```
fgravin: `this.map.getLayers()` is more understandable.

fgravin:
We should add some tests on gmf layertree directive methods.
The logic becomes a bit complex

contribs/gmf/src/directives/layertree.js
ger-benjamin: I'm note sur that that's completely the same things... (I'm not sur I've the `insertAt` method with your idea). But I'll try it.

ger-benjamin:
> We should add some tests on gmf layertree directive methods.

Ok, I'll do that.

> The logic becomes a bit complex

I know. That's hard to understand. But there is a lot of possible case to handle and I've also must take the performance in count the performance and future possibilities like activate layers by a directlink or by the search.... I've don't found a better solution... (And, at the end, I like this one).

fgravin:
Ok you can merge.
Thanks for the work

PR #603, opened 2016-02-05
Add a cleanallcache make target

sbrunner:
[https://github.com/camptocamp/c2cgeoportal/issues/1937]


pgiraud:
Looks good.
Can you please tell us in which case we should use this target? Any use case?

fredj:
Looks good

pgiraud:
Merging on behalf or @sbrunner.

Issue #577, opened 2016-02-01
Use a boolean property instead of a couple of functions for (de)activate


sbrunner:
https://github.com/camptocamp/ngeo/blob/master/src/services/toolActivateMgr.js

sbrunner:
@fgravin you can complete if it's required :-)

PR #615, opened 2016-02-05
Use `watch` in toolActivateMgr

tsauerwein:
This PR aims to simplify the usage of the `ngeo.ToolActivateMgr`. Now, the active state of a tool can be directly bound to the active property of an interaction (or any other object).

Example:

    this.drawPoint = new ol.interaction.Draw(...);
    ngeoDecorateInteraction(this.drawPoint);

    var drawPointTool = new ngeo.ToolActivate(this.drawPoint, 'active');
    ngeoToolActivateMgr.registerTool('mapTools', drawPointTool);

Closes https://github.com/camptocamp/ngeo/issues/577

### DIFF
--- a/examples/toolActivate.html
+++ b/examples/toolActivate.html
@@ -13,16 +13,14 @@
           width: 800px;
           height: 400px;
         }
-        .light {
-          opacity: 0.4;
-        }
     </style>
   </head>
   <body ng-controller="MainController as ctrl">
     <div id="map" ngeo-map="ctrl.map"></div>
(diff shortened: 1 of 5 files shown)


pgiraud:
In my opinion, we should remove the `light` className. It adds complexity to the example. The ngeo-btn combined with ng-model already toggles the `active` className.

tsauerwein:
> In my opinion, we should remove the light className.

I agree, I removed it.

fgravin:
Looks good thanks

PR #617, opened 2016-02-08
Handle add_theme results from search

### LINKED ISSUE #555: Error when searching theme elements
`Error: featureGeometry is null`

fredj:
- [x] theme is changed in the layer tree
 - [x]  theme is changed in the theme selector
 - [ ]  theme is changed in the url

fixes #555


fgravin:
Looks good.
I guess you checked the theme is also changed in the layerTree and in the theme Selector ?

fredj:
@fgravin the theme is not updated in the url but the layer tree and theme selector are ok

fredj:
I will merge this and fix the theme name in the url in an other pull request

PR #619, opened 2016-02-08
Don't load less js, not required anymore

### LINKED ISSUE #618: [desktop app] interface turns to white while loading
For un unknown (for now) reason, when loading the desktop app in the browser, the page becomes completely white after being loaded.
The following code seem to be added to the head:
```
<style type="text/css">body { display: none !important }</style>
```
After a quick bisect, it looks like it happens after the f7d7199910ac8b5c45d2d2639d4f57248312a123 commit.
Currently digging into the issue.

pgiraud:
Fixes #618 

After upgrading less to 2.6.0, the javascript compiler of less seem to hide the body when no less file is found.
We don't need to compile less in the browser so we don't need the less javascript file anymore.


fredj:
please merge

PR #620, opened 2016-02-09
Prevent default action on click on anchor link.

blattmann:
Fix disables displaying the href (anchor) values in the browser's address bar.

### DIFF
--- a/contribs/gmf/apps/mobile/index.html
+++ b/contribs/gmf/apps/mobile/index.html
@@ -35,16 +35,16 @@
     </main>
     <nav class="nav-left" gmf-mobile-nav>
       <header>
-        <a class="go-back" href="#"></a>
+        <a class="go-back" href="#" ng-click="$event.preventDefault()"></a>
       </header>
       <!-- main menu -->
       <div class="active slide">
         <ul>
           <li>
-            <a href="#background" data-toggle="slide-in">Background</a>
+            <a href="#background" data-toggle="slide-in" ng-click="$event.preventDefault()">Background</a>
           </li>
           <li>
-            <a href="#themes" data-toggle="slide-in">Themes</a>
+            <a href="#themes" data-toggle="slide-in" ng-click="$event.preventDefault()">Themes</a>
           </li>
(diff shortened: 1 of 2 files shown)


blattmann:
Issue URL: https://github.com/camptocamp/c2cgeoportal/issues/1979

fgravin:
```html
<a href="#background" data-toggle="slide-in" ng-click="$event.preventDefault()">
```

Can you try instead and see if it works or not.

```html
<a href="" target="#background" data-toggle="slide-in">
```

blattmann:
<a href="" target="#background" data-toggle="slide-in"> does not work.

fgravin:
http://getbootstrap.com/javascript/#via-data-attributes-1

`data-target` maybe ?

blattmann:
nope. sorry.

sbrunner:
Why not using buttons?

fgravin:
I've tried

```js
            // element to slide in
            var slideIn = $($(this).attr('target'));
```

```html
            <a href="" target="#background" data-toggle="slide-in">Background</a>
```

Works well.
See https://github.com/camptocamp/ngeo/blob/master/contribs/gmf/src/directives/mobilenav.js#L91

Do you agree @pgiraud ?

pgiraud:
@fgravin looks like a nice solution. To make it more compatible, I would even prefix the attribute "data-".

fgravin:
@blattmann please do the change as mentionned. use `data-target`.

blattmann:
If I do the changes as you proposed then it works fine in the address bar of the browser (does not show the anchor links) but the panel also does not show any more info regarding the backgrounds or the themes.

Here the code:
`<a href="" data-target="#background" data-toggle="slide-in">Background</a>`
`<a href="" data-target="#themes" data-toggle="slide-in">Themes</a>`

See my short screencast:
[https://drive.google.com/file/d/0B2m6lBH9E2lgbzVUMHNBTDI3b3M/view?usp=sharing](url)

pgiraud:
Replaced by #672.

PR #626, opened 2016-02-09
Desktop styling

pgiraud:
Here's a quick change to make the desktop look more coherent and with less "agressive", more neutral colors.

screenshot from 2016-02-09 14 10 22

Mobile also takes advantage of the modifications.
screenshot from 2016-02-09 14 16 25

Some work may still need to be done, but it's a good start with only a bit of CSS.

### DIFF
--- a/contribs/gmf/less/desktop.less
+++ b/contribs/gmf/less/desktop.less
@@ -13,8 +13,6 @@
 /* width for the vertical toolbar */
 @toolbar-width: 40px;
 
-@brand-secondary: lighten(@brand-primary, 20%);
-
 html, body {
   position: relative;
   height: 100%;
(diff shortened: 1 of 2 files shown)


contribs/gmf/less/desktop.less
```diff
@@ -229,13 +230,16 @@ gmf-layertree ul .treenode {
...
+      margin-top: -1px;
     }
 
     .btn {
       background-color: @brand-primary;
-      border: 0;
+      margin-left: 0;
+      border-right-width: 0;
+      border-left-width: 0;
       border-radius: 0 !important;
```
fredj: Is `border-radius: 0 !important;` still needed now that `@border-radius-base`is set to 0 ?

fredj:
LGTM

contribs/gmf/less/desktop.less
pgiraud: I'd prefer to keep this. This will prevent weird behaviour if a user chooses to force the radiuses back to a value greater than 0.

PR #629, opened 2016-02-09
[wip] Handle add_layer results from search

fredj:
add_layer:
 - if the layer exists in the current theme, set visibility to true
 - if not, add the group to the layer tree


fredj:
#749

PR #633, opened 2016-02-09
Add query to gmf mobile app

adube:
This PR adds the mobile query directive inside the GMF mobile application.

When any click is made inside the application, one or more request occurs.  At this stage, there's nowhere the result is put, yet.

The ToolActiveManager of ngeo is also added in there as well, with only the query tool being bound to it.  The query tool is also the "default" tool that has to remain active if no other tool is active.

Please, review.


adube:
@fgravin Would you please review this ?

fgravin:
Looks good thanks.

PR #630, opened 2016-02-09
Don't touch the webfonts in fontawesome node module

pgiraud:
Fixes issue with webfont files being removed when stopping (Ctrl-C) `make serve`.

```
^Cmake: *** [serve] Error 130
make: *** Deleting intermediate file `node_modules/font-awesome/fonts/fontawesome-webfont.woff'
make: *** Deleting intermediate file `node_modules/font-awesome/fonts/fontawesome-webfont.woff2'
make: *** Deleting intermediate file `node_modules/font-awesome/fonts/fontawesome-webfont.eot'
make: *** Deleting intermediate file `node_modules/font-awesome/fonts/fontawesome-webfont.ttf'
```
To be tested.

PR #638, opened 2016-02-10
Makefile osx compatible

pgiraud:
With this pull request I'm addressing compatibility issue that happen when running make targets on OSX due mainly to differences between the options available for `sed`, `touch`, and `stat` commands.

For the `sed` command, it took me a while and here's a very nice resource:
http://stackoverflow.com/questions/14846304/sed-command-error-on-macos-x

This hasn't been tested on linux yet since I am having issues with gflags python package not installing.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,6 +2,8 @@ SRC_JS_FILES := $(shell find src -type f -name '*.js')
 NGEO_DIRECTIVES_PARTIALS_FILES := $(shell ls -1 src/directives/partials/*.html)
 GMF_DIRECTIVES_PARTIALS_FILES := $(shell ls -1 contribs/gmf/src/directives/partials/*.html)
 
+OS := $(shell uname)
+
 EXPORTS_JS_FILES := $(shell find exports -type f -name '*.js')
 
 EXAMPLES_JS_FILES := $(shell find examples -maxdepth 1 -type f -name '*.js')
--- a/Makefile
+++ b/Makefile
@@ -381,7 +381,7 @@ node_modules/angular/angular.min.js: .build/node_modules.timestamp
 		-e 's|/@?main=$*.js|$*.js|' \
 		-e '/default\.js/d' \
 		-e 's|\.\./utils/watchwatchers.js|lib/watchwatchers.js|' \
-		-e '/$*.js/i\    <script src="lib/ngeo.js"></script>' $< > $@
+		-e '/$*.js/i\'$$'\n    <script src="lib/ngeo.js"></script>'$$'\n' $< > $@
 
 .PRECIOUS: .build/examples-hosted/contribs/gmf/%.html
 .build/examples-hosted/contribs/gmf/%.html: contribs/gmf/examples/%.html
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -76,12 +76,18 @@ EXTERNS_ANGULAR_HTTP_PROMISE = .build/externs/angular-1.4-http-promise_templated
 EXTERNS_JQUERY = .build/externs/jquery-1.9.js
 EXTERNS_FILES = $(EXTERNS_ANGULAR) $(EXTERNS_ANGULAR_Q) $(EXTERNS_ANGULAR_HTTP_PROMISE) $(EXTERNS_JQUERY)
 
+OS := $(shell uname)
+
```
sbrunner: For me all this OS specific block should be at the top of the file (actually we use the variable before we set them, it's working but it's not natural ...)

Makefile
```diff
@@ -379,7 +385,7 @@ node_modules/angular/angular.min.js: .build/node_modules.timestamp
 		-e 's|/@?main=$*.js|$*.js|' \
 		-e '/default\.js/d' \
 		-e 's|\.\./utils/watchwatchers.js|lib/watchwatchers.js|' \
-		-e '/$*.js/i\    <script src="lib/ngeo.js"></script>' $< > $@
+		-e '/$*.js/i\'$$'\n''    <script src="lib/ngeo.js"></script>'$$'\n' $< > $@
```
sbrunner: What does it will do?
isn't it the same as just `$$\n` (without any `'`)?

Makefile
```diff
@@ -451,7 +457,6 @@ node_modules/angular/angular.min.js: .build/node_modules.timestamp
 	touch $@
 
 node_modules/font-awesome/fonts/fontawesome-webfont.%: .build/node_modules.timestamp
-	touch --no-create $@
```
sbrunner: This ill create an issue, see:
https://github.com/camptocamp/ngeo/pull/630#issuecomment-182773312

Makefile
pgiraud: No, unfortunately. I know that it makes the syntax complex, but it's required.

sbrunner: And what 't for ?

sbrunner: As I test effectively the `$$` should be out of the quotes but the next `''` don't do anything...

pgiraud: What should the line look like in your opinion?
Did you test on OSX?

sbrunner: `-e '/$*.js/i\'$$'\n    <script src="lib/ngeo.js"></script>'$$'\n' $< > $@`
no, not actually...

sbrunner: And what those `$$` for?

pgiraud: You're right, it's possible to remove the quotes.
However, since `$'\n` is the new line character, I need to double the `$` to make it work in the Makefile.

pgiraud:
Please review again.

sbrunner:
The question about what will the `$$` do steal be open...

pgiraud:
@sbrunner you probably missed my previous answer:
> However, since $'\n is the new line character, I need to double the $ to make it work in the Makefile.

sbrunner:
Effectively I miss it, sorry ...

Than it's cosmetic change, not directly requires by OSX?

pgiraud:
As far as I understand, the new line character is required by OSX. Thus it's not "cosmetics". ;-)

sbrunner:
OK, thanks for your response, merge :-)

PR #643, opened 2016-02-12
Fix menu animation on Safari (mobile application)

tsauerwein:
In a sidebar, when clicking on a link in the main menu (e.g. "Background"), the title bar turned white during the animation on Safari.

The problem is that the newly inserted element did not have the right position after insertion. The "fix" forces Safari to calculate the position.

Closes https://github.com/camptocamp/c2cgeoportal/issues/1987


pgiraud:
Nice. It works well. Please merge.

tsauerwein:
Thanks for the review and help.

PR #649, opened 2016-02-15
Don't remove the fonts on the build

sbrunner:
fix https://github.com/camptocamp/ngeo/issues/640

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -450,6 +450,7 @@ node_modules/angular/angular.min.js: .build/node_modules.timestamp
 	mkdir -p $(dir $@)
 	touch $@
 
+.PRECIOUS: node_modules/font-awesome/fonts/fontawesome-webfont.%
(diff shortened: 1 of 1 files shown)


ger-benjamin:
+100 (but you'll must rebase this PR on master...)

Makefile
```diff
@@ -450,6 +450,7 @@ node_modules/angular/angular.min.js: .build/node_modules.timestamp
 	mkdir -p $(dir $@)
 	touch $@
 
+.PRECIOUS: node_modules/font-awesome/fonts/fontawesome-webfont.%
 node_modules/font-awesome/fonts/fontawesome-webfont.%: .build/node_modules.timestamp
 	touch --no-create $@
```
pgiraud: Can you please use the `-c` to make it work on OSX at the same time?

sbrunner: We will mix completely deferent things in the same pull request?

pgiraud: No you're right. Sorry.

PR #582, opened 2016-02-02
Gmf display queries

ger-benjamin:
Strange issue (reason of the [wip] tag on this PR): 
- The dropdown is empty on smartphone (at least on IOS)... (see #663)
- I must add a ng-swipe-left/right to do actions (but I can't test that because of the first point)

Link to, and current state of work, on: camptocamp/c2cgeoportal#1800

~~Example: https://ger-benjamin.github.io/ngeo/master/examples/contribs/gmf/mobilequery.html (try it on "edit" theme layers)~~

PR #651, opened 2016-02-15
Update to AngularJS 1.5.0

ger-benjamin:
Rel to: camptocamp/c2cgeoportal#1800
Needed by: #582 
Example on: https://ger-benjamin.github.io/ngeo/master/examples

To implement the required animations, the best way I found was to use the new Angluar directive "ngAnimateSwap" see (https://docs.angularjs.org/api/ngAnimate/directive/ngAnimateSwap)
It works really fine but to use it, I had to update angular to 1.5.0.

Angular changelog: https://github.com/angular/angular.js/blob/master/CHANGELOG.md#150-ennoblement-facilitation-2016-02-05 and http://angularjs.blogspot.ch/2016/02/angular-150-ennoblement-facilitation.html

The two small modifications in the filereader.js is also for AngularJS 1.5.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -20,6 +20,7 @@ GMF_APPS_LIBS_JS_FILES += \
 	contribs/gmf/examples/https.js \
 	node_modules/jquery/dist/jquery.min.js \
 	node_modules/angular/angular.min.js \
+	node_modules/angular-animate/angular-animate.min.js \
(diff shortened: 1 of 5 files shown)


sbrunner:
Looks good :-)

fredj:
LGTM

PR #665, opened 2016-02-16
Sed commands compatible with Mac (round 2)

### LINKED ISSUE #662: weird $ characters in hosted examples
@sbrunner you were right.
The following commit doesn't work correctly on Linux.
https://github.com/pgiraud/ngeo/commit/a08050f777a41bd889e4fbc5bf9dbd16fe8c5ed8

Extra "$" characters are added at the bottom of the page.

pgiraud:
Fixes #662.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -85,11 +85,13 @@ ifeq ($(OS),Darwin)
 	STAT_UNCOMPRESSED = stat -f 'uncompressed: %z bytes'
 	STAT_LAST_MODIFIED = stat -f '%m'
 	TOUCH_DATE = touch -t
+	SED_NEW_LINE = '$$'\n
 else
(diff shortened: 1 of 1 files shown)


sbrunner:
Looks good :-)

pgiraud:
The build fails because the coverage increased (+0.0%). What am I supposed to do in this case?

tsauerwein:
> The build fails because the coverage increased (+0.0%). What am I supposed to do in this case?

Remove tests!

tsauerwein:
No, seriously. Currently the Coverage threshold is configured not to fall below 35%. I guess another PR pulled the coverage down. I guess we have to readjust.

sbrunner:
Coveralls configurations adjusted to 30% of minimum coverage for the next runs...

PR #560, opened 2016-01-27
Handle error when the user has disabled the gelocation in his device.

ochriste:
TODO: disable button, notify user, waiting for people to answer my email.

PR #666, opened 2016-02-16
Emit event in case of geolocation error

### LINKED ISSUE #489: Handle errors in geolocation
Errors from ol.Geolocation should be handled

tsauerwein:
Supersedes #560
Closes #489
Closes #560

### DIFF
--- a/src/directives/desktopgeolocation.js
+++ b/src/directives/desktopgeolocation.js
@@ -12,6 +12,18 @@ goog.require('ol.geom.Point');
 
 
 /**
+ * @enum {string}
+ * @export
+ */
+ngeo.DesktopGeolocationEventType = {
+  /**
+   * Triggered when an error occures.
+   */
+  ERROR: 'desktop-geolocation-error'
+};
+
+
+/**
  * Provide a "desktop geolocation" directive.
  *
  * Example:
@@ -77,6 +89,12 @@ ngeo.DesktopGeolocationController = function($scope, $element,
   goog.asserts.assertObject(options);
 
   /**
+   * @type {!angular.Scope}
+   * @private
+   */
+  this.$scope_ = $scope;
+
+  /**
    * @type {ngeo.FeatureOverlay}
    * @private
    */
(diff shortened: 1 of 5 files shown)


fgravin:
Looks good.
But notice that only controllers that are parents (at any level) will receive the event.
For example, a same level directive controller won't catch this event, but this case may never be found.

pgiraud:
We should be able to translate the error message. This can be addressed later though.

src/directives/mobilegeolocation.js
```diff
@@ -12,6 +12,17 @@ goog.require('ol.geom.Point');
 
 
 /**
+ * @enum {string}
+ */
+ngeo.MobileGeolocationEventType = {
+  /**
+   * Triggered when an error occures.
+   *
```
pgiraud: extra line

src/directives/desktopgeolocation.js
```diff
@@ -12,6 +12,18 @@ goog.require('ol.geom.Point');
 
 
 /**
+ * @enum {string}
+ */
+ngeo.DesktopGeolocationEventType = {
+  /**
+   * Triggered when an error occures.
+   *
```
fredj: Extra line

fgravin:
Scope `$on` and `$emit` function are exported angular way to deal with event.
As, internally, we use closure library, i wonder if we should'nt use closure event system ?
It would got over scope tree view issues.

```js
goog.inherits(AbstractController, ol.events.EventTarget);

...

this.dispatchEvent(
```

What do you think ?

tsauerwein:
> As, internally, we use closure library, i wonder if we shouldn't use closure event system ?

This would mean, if you would want to listen to events emitted by the geo-location directive, you would need a reference to the geo-location controller to set up the listener. How practical is that?

As you can install the geo-location directive on any element, I don't think it is a problem that the event is only sent to parents.

fgravin:
Ok you are right

fgravin:
LGTM

PR #668, opened 2016-02-16
Load x,y,z when registering map in permalink

adube:
This PR simplifies the initial loading and reading of the X, Y and Z values from the permalink.  It now occurs when registering the map in the permalink, which is done in the GMF map directive.

Live demos:

 * http://adube.github.io/ngeo/app-mobile-permalink/examples/contribs/gmf/permalink.html
 * https://adube.github.io/ngeo/app-mobile-permalink/examples/contribs/gmf/apps/mobile/index.html

### DIFF
--- a/contribs/gmf/examples/permalink.html
+++ b/contribs/gmf/examples/permalink.html
@@ -21,8 +21,8 @@
     <gmf-map gmf-map-map="ctrl.map"></gmf-map>
 
     <p id="desc">
-      This example demonstrates the use of the <code>ngeo-permalink</code>
-      service.
+      This example demonstrates the use of the <code>gmf-permalink</code>
+      service, which is injected inside the <code>gmf-map</code> directive.
     </p>
 
     <script src="../../../node_modules/jquery/dist/jquery.js"></script>
--- a/contribs/gmf/examples/permalink.js
+++ b/contribs/gmf/examples/permalink.js
@@ -1,6 +1,5 @@
 goog.provide('gmf-permalink');
 
-goog.require('gmf.Permalink');
 goog.require('gmf.mapDirective');
 goog.require('gmf.proj.EPSG21781');
 goog.require('ngeo');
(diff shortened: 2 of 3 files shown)


adube:
@fgravin please, review.

contribs/gmf/src/controllers/abstract.js
```diff
@@ -180,4 +181,31 @@ gmf.AbstractController.prototype.initLanguage = function() {
   }
 };
 
+/**
+ * @param {angular.$injector} $injector Main injector.
+ * @return {olx.ViewOptions} The view options from the permalink.
+ * @protected
+ */
+gmf.AbstractController.prototype.getPermalinkViewConfig = function($injector) {
```
fgravin: Don't send `$injector` as parameter.
Either put it in `this` in the constructor, or put the `gmfPermalink` service in `this` in the constructor.

fgravin:
What about moving it in the gmf map directive ?
I know we will change the view afterwards, but maybe more consistent with what gmf map directive is used for.

contribs/gmf/src/controllers/abstract.js
adube: It would make sense to put the `$injector` in `this` in the abstract class, but the map is created before calling the parent constructor.  I suppose it's okay if it's created after.

adube:
>What about moving it in the gmf map directive ?
I know we will change the view afterwards, but maybe more consistent with what gmf map directive is used for.

I personally think the map object should be initialized with the proper values from the start, but that's my opinion.  Please, let me know which direction I must go:

a) do this at the controller level, before the map is created
b) do this at the map directive level, after the map is created and overriding the original view options.

Thanks.

adube:
@fgravin I'll leave this in your hands for now.  Please, come back to me once you've come up with a decision.

fgravin:
I prefer to move it in gmf map directive, or in the permalink service itself, for example on registering.
I think it's better than in the main controllers. It's not a big deal if we change the zoom and center afterwards.

adube:
@fgravin or @pgiraud, can this be merged ?

fgravin:
Nope, please see my previous comment.

adube:
I'm on it.

adube:
BTW, I like the idea `in the permalink service itself, for example on registering.`  I'll do that.

adube:
I made the changes.  This is now ready for merge, once Travis says it's okay.

I added github.io demos as well.

fgravin:
LG

PR #676, opened 2016-02-17
Hide the search-group for individual search result

### LINKED ISSUE #660: desktop: remove the search-group div in search result
The results are already grouped by search-group, no need to display the value
gmf-search

fredj:
Fixes #660


sbrunner:
Looks good, merge :-)

Issue #671, opened 2016-02-17
[Mobile/Tablet] Layout details - Search


blattmann:
* [x] [Mobile] Small shadows left and right of search box haven't the same size.
* [x] [Mobile] Buttons left and right of search box should be square just like other buttons.

Search results have a 1-px border all around:
* [x] [Mobile] Remove right/left/bottom borders so results take the whole space.
* [x] [Tablet] Remove bottom and top extra border.

* [x] [Mobile] For mobile with a wide screen (or mobile in landscape mode), align search box to the left and cross to hide search results to the rig.

* [ ] --low priority-- [Mobile] When results are present, map should be hidden.
Done here: https://github.com/camptocamp/ngeo/pull/801

PR #680, opened 2016-02-17
Layout details - Search [Mobile / Tablet]

blattmann:
Mobile: Small shadows left and right of search box have now the same size

Mobile: Buttons left and right of search box are now square just like other buttons

Mobile: Right / left / bottom borders removed so results take the whole space.

Tablet: Bottom and top extra border removed.

Mobile: For mobile with a wide screen or mobile in landscape mode: Search input field is aligned left and clear button (cross) is aligned right.

see: #671


blattmann:
This PR is no more used as it is split into several smaller PR's.

PR #685, opened 2016-02-17
Background / Themes thumbnails adjusted to be displayed as rectangle …

pgiraud:
…in mobile view

Initial work done by @blattmann.

### DIFF
--- a/contribs/gmf/less/mobile.less
+++ b/contribs/gmf/less/mobile.less
@@ -253,3 +253,20 @@ button[ngeo-mobile-geolocation] {
     }
   }
 }
+
+//mobile styles for the theme and background selector
+.gmf-theme-selector,
+.gmf-mobile-background-layer-selector {
+  .gmf-icon-check:after {
+    font-size: 1.2rem;
+    position: relative;
+    bottom: 0.8rem;
+  }
+  .gmf-text {
+    margin: 3% 10px 0 0;
+  }
+  .gmf-thumb {
+    height: auto;
+    width: 25%;
+  }
+}


contribs/gmf/less/mobile.less
```diff
@@ -253,3 +253,20 @@ button[ngeo-mobile-geolocation] {
     }
   }
 }
+
+//mobile styles for the theme and background selector
```
ger-benjamin: Please change this comment to: `// Mobile styles for the theme and background selector.`
(Space first, then an uppercase and a point at the end).

contribs/gmf/less/mobile.less
```diff
@@ -253,3 +253,20 @@ button[ngeo-mobile-geolocation] {
...
+    font-size: 1.2rem;
+    position: relative;
+    bottom: 0.8rem;
+  }
+  .gmf-text {
+    margin: 3% 10px 0 0;
+  }
+  .gmf-thumb {
+    height: auto;
+    width: 25%;
```
ger-benjamin: I hope that's not too small / compressed (I've not tried).

sbrunner: Why not just respect the image size (no width and height)?

blattmann: @sbrunner: Because the images are too big and would look like this if we do not adjust the CSS:
screen shot 2016-02-18 at 17 44 49

pgiraud:
Ok. LGTM. Please merge.

contribs/gmf/less/mobile.less
sbrunner: Effectively I just see it on the phone/tablet size, and it's wheal, previously I test it on desktop and the icons will already be on the new line ...

PR #691, opened 2016-02-17
(wip) Fix query integration in GMF

adube:
This PR focuses on fixing the `ngeo.Query` integration in GMF.  The main idea is, for now, deprecate the `gmf.QueryManager` and move the configuration of the query sources directly inside the layer tree directive.

Work in progress:

 * [x] define the format object directly using all layer names
 * [ ] separate the results per layer name instead of per source, since we now have a single source for multiple layer names
 * [ ] manage the `identifierAttributeField` properly
 * [ ] review


adube:
I'm second-guessing myself...  I'll create an other PR with an alternate solution that, in my opinion, will work better.

adube:
Yeah, I think that #691 is better.  I'm closing this one.

Issue #673, opened 2016-02-17
The background selector mix the layers


sbrunner:
* Open https://camptocamp.github.io/ngeo/master/examples/contribs/gmf/apps/mobile/index.html
* Add a theme -> a layer should be display (overlay)
* Set the backgound from map to blank
* The overlay isn't any more visible and the background map steal be visible ...

rbovard:
Just to be sure, the functionality to configure the basemap for each theme will still work in 2.0? http://docs.camptocamp.net/c2cgeoportal/1.6/integrator/functionality.html#using-functionalities-to-configure-the-basemap-to-use-for-each-theme

sbrunner:
No...

sbrunner:
In the version 2.0 we don't do what's was (in 1.6) specific to the desktop.

rbovard:
Seriously, we will lose all desktop related features?!

Ping @kalbermattenm @vvmalbre

sbrunner:
In the 2.0 yes, because the desktop steal be in CGXP, the desktop related features will be in the version 2.1!

ybolognini:
@rbovard @kalbermattenm @vvmalbre

Let me clarify:

When @sbrunner talks about the "2.0", he means the mobile version that will be presented on 2/29. This includes all 1.6 **mobile** features and also the same features applied to Desktop layout. 

The complete Desktop version is called "2.1". Il will include features that are currently **not** in 1.6 mobile. Configuration of basemap for a theme is a Desktop feature.

rbovard:
Ok, thanks a lot for the clarification @ybolognini. I've little mixed up v2 vs v2., sorry for that...

And for my information, is this functionality will be available in 2.1 mobile or not?

ybolognini:
As background layer components are different on Desktop and Mobile, it isn't planned to have this feature on the Mobile app (Mobile features are basically the same as 1.6). But this could be added as an enhancement.

PR #692, opened 2016-02-17
(wip) Fix query service integration in GMF

adube:
This PR is an alternate solution to fix the issue in the Query tool.  I think it's much better than #691.

The main idea behind this approach is this: "A source for query needs to be bound to a single layer, i.e. a single server-side layer".  A source and its result NEED to be bound together, as the source has its own label (used for title) and also requires a way to define a sub-title.  The latter is directly bound the the feature properties themselves, which cannot be controlled at the level of the source if it was responsible of managing multiple layers.

The approach of this PR is to keep the `gmf.QueryManager` service while allowing a layer to define multiple query sources.  This is defined using the `querySourceIds` property of the layer, which has a list of source ids instead of just one.  When looping in the layers, that property is also checked to get the sources associated.  The rest stays "as-is".

 - [x] Restore the previous behaviour, fixing the query tool in GMF
 - [x] Support layer groups, i.e. the "osm" layer returns nothing
 - [x] Fix the bug that occurs with layers that combine multiple layer names
 - [x] Address reviewer's comments / required fixes
 - [x] Review

Live demo:

 * https://adube.github.io/ngeo/fix-query-service-2nd/examples/contribs/gmf/mobilequery.html

### DIFF
--- a/contribs/gmf/src/directives/layertree.js
+++ b/contribs/gmf/src/directives/layertree.js
@@ -259,7 +259,8 @@ gmf.LayertreeController.prototype.getLayer = function(node, opt_depth,
   }
 
   if (goog.isDefAndNotNull(layer)) {
-    layer.set('querySourceId', node.id);
+    var ids = this.getNodeIds_(node);
+    layer.set('querySourceIds', ids);
     layer.set('layerName', node.name);
     // The layer must be upper than the background
     layer.setZIndex(1);
@@ -836,6 +837,26 @@ gmf.LayertreeController.prototype.zoomToResolution = function(node) {
   if (goog.isDef(resolution)) {
     view.setResolution(resolution);
   }
+};
+
+
+/**
+ * Collect and return all ids this node and all child nodes as well.
+ * @param {GmfThemesNode} node Layer tree node.
+ * @return {Array.<number|string>} Layer names.
+ * @private
+ */
+gmf.LayertreeController.prototype.getNodeIds_ = function(node) {
+  var ids = [];
+  var children = node.children || node;
+  if (children && children.length) {
+    children.forEach(function(childNode) {
+      ids = ids.concat(this.getNodeIds_(childNode));
+    }, this);
+  } else if (node.id !== undefined) {
+    ids.push(node.id);
+  }
+  return ids;
 };
 
 
--- a/contribs/gmf/src/directives/layertree.js
+++ b/contribs/gmf/src/directives/layertree.js
@@ -841,7 +841,7 @@ gmf.LayertreeController.prototype.zoomToResolution = function(node) {
 
 
 /**
- * Collect and return all ids this node and all child nodes as well.
+ * Collect and return all ids of this node and all child nodes as well.
  * @param {GmfThemesNode} node Layer tree node.
  * @return {Array.<number|string>} Layer names.
  * @private
--- a/contribs/gmf/src/services/querymanager.js
+++ b/contribs/gmf/src/services/querymanager.js
@@ -94,6 +94,21 @@ gmf.QueryManager.prototype.createSources_ = function(node) {
     }
   } else {
     if (!this.cache_[id]) {
+
+      // Some nodes have child layers, i.e. a list of layer names that are
+      // part of a group.  The name of the group itself can't be used 'as-is'
+      // as an identifier of the layers for this source.  For example, a
+      // group named 'osm' might result in returning 'restaurant' features.
+      // This override makes sure that those layer names are used instead of
+      // the original one.
+      if (node.childLayers) {
+        var childLayerNames = [];
+        node.childLayers.forEach(function(childLayer) {
+          childLayerNames.push(childLayer.name);
+        }, this);
+        layers = childLayerNames.join(',');
+      }
+
       var source = {
         'id': id,
         'identifierAttributeField': identifierAttributeField,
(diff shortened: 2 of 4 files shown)


adube:
@ger-benjamin This is ready for review.  You'll see that it's not the solution that we discussed together earlier, but you'll understand why that other solution couldn't more easily work.

Don't merge yet.  @fgravin will want to look at this on Monday.

contribs/gmf/src/services/querymanager.js
```diff
@@ -96,6 +96,21 @@ gmf.QueryManager.prototype.createSources_ = function(node) {
     }
   } else {
     if (!this.cache_[id]) {
+
+      // Some nodes have child layers, i.e. a list of layer names that are
+      // part of a group.  The name of the group itself can't be used 'as-is'
```
ger-benjamin: > ...of a group.  The name...

Too much space between "group." and "The"

src/services/query.js
```diff
@@ -308,30 +323,36 @@ ngeo.Query.prototype.issueWMSGetFeatureInfoRequests_ = function(
       return;
     }
 
-    // skip layers that don't have a source configured
+    // skip layers that don't have one or more sources configured
```
ger-benjamin: (I like comments starting with a capital letter and endings with a point :-) )

ger-benjamin:
Great work, I like this solution. I just wonder why that doesn't work when I try your fork on my own localhost... Are you experiment the same trouble ?

Just by curiosity, there is nothing right now to query WMTS layers, right ?

ger-benjamin:
The modified block codes are well isolated with this solution. I wonder if we can merge without fgravin...
If somethings is wrong, we can modify it later.

What do you think @ybolognini ?

ybolognini:
It would be great if I could test the query before my holidays. 

@tsauerwein can have a look and if everything looks fine we can merge.

contribs/gmf/src/services/querymanager.js
```diff
@@ -96,6 +96,21 @@ gmf.QueryManager.prototype.createSources_ = function(node) {
...
+      // as an identifier of the layers for this source.  For example, a
+      // group named 'osm' might result in returning 'restaurant' features.
+      // This override makes sure that those layer names are used instead of
+      // the original one.
+      if (node.childLayers) {
+        var childLayerNames = [];
+        node.childLayers.forEach(function(childLayer) {
+          childLayerNames.push(childLayer.name);
```
tsauerwein: I don't see the whole picture, but how unique are the names as identifiers?

tsauerwein: Sorry, ignore the above comment.

contribs/gmf/src/directives/layertree.js
```diff
@@ -838,4 +839,24 @@ gmf.LayertreeController.prototype.zoomToResolution = function(node) {
 };
 
 
+/**
+ * Collect and return all ids this node and all child nodes as well.
```
tsauerwein: "... **of** this node ..."?

src/services/query.js
```diff
@@ -308,30 +323,36 @@ ngeo.Query.prototype.issueWMSGetFeatureInfoRequests_ = function(
...
-      goog.asserts.assertString(url);
-      if (!itemsByUrl[url]) {
-        itemsByUrl[url] = [];
-      }
-      itemsByUrl[url].push(item);
-    } else {
-      // TODO - support other kinds of infoFormats
-      item['resultSource'].pending = false;
+    if (id) {
+      ids.push(id);
```
tsauerwein: `ids` is the array stored on the layer? If `issueWMSGetFeatureInfoRequests` is called multiple times, you also add `id` multiple times as you are not working on a clone of the array?

tsauerwein:
I don't have a complete understanding on how the layer tree/query manager work, but your changes look reasonable to me.

adube:
Thank you everyone for the review and comments.  I'll make the required changes shortly.

Hold the merge for a moment.  I thought of something yesterday (before I went to sleep... best place to think!)  I currently check if the layer is visible and if it is then all sources are queried.  That isn't always the case.  If, let's say, a layer has 5 inner layer names and only one out of the five are checked in the layertree, then only it is in the LAYERS param.  Currently, all five are added to the request.

I thought of a fix for this and I'll work on it in a few minutes.  I'll keep you in touch.

@tsauerwein I agree that from an external point of view, that tool isn't that simple to understand without looking at the code.  @ybolognini: I think it would make sense to document it some more.  I suggest in the examples themselves.  I could open an other PR for this if you agree.

ybolognini:
@adube More doc sounds good, after dev itself is finished

adube:
@ger-benjamin WMTS layers are supported already.

adube:
@ger-benjamin The last issue has been fixed, with the addition of all of the reviewer's comments.  I left them in separated commits for you to more easily check one last time.  If everything's okay, I'll combine them into a single commit for you to merge.

Thanks to everyone!

ger-benjamin:
Thanks to you, very good job.

Just one more problem. To fix #673 I've separate "data" layers from "background" layers by putting "data" layers in a group (see PR #703 for more details).

Layers is no more found in the map by the query service. Please use this fix : https://github.com/ger-benjamin/ngeo/commit/51a790188ca7c5f3177c960d4496e622dae9523d (perhaps you can "cherry-pick" it) to resolve the bug (I can't add it in this PR because the source is on your own fork).

Then please, rebase on master, merge commits in one and merge this pull request.

ger-benjamin:
I close this PR and open a new with my commit, so I can merge this sooner. See: https://github.com/camptocamp/ngeo/pull/706

adube:
@ger-benjamin Excellent.  Thank you very much for taking care of this.

ger-benjamin:
@adube Thanks to you :smiley:

PR #695, opened 2016-02-18
Mobile: Small shadows left and right of search box

blattmann:
shadows have now the same size


pgiraud:
LGTM. Will merge once travis is green.

PR #697, opened 2016-02-18
Mobile: Buttons left and right of search box

blattmann:
Buttons are now square just like other buttons

### DIFF
--- a/contribs/gmf/less/mobile-nav.less
+++ b/contribs/gmf/less/mobile-nav.less
@@ -221,12 +221,12 @@ nav.nav-right {
 }
 
 .nav-left-trigger {
-  left: ~"calc("@app-margin ~" + 2px)";
+  left: calc(@app-margin ~" + 1px");
   box-shadow: 3px 0 5px -2px #bbb;
 }
 
 .nav-right-trigger {
-  right: ~"calc("@app-margin ~" + 2px)";
+  right: calc(@app-margin ~" + 1px");
   box-shadow: -3px 0 5px -2px #bbb;
 }
 


sbrunner:
Not tested but that's looks good :-)

pgiraud:
It looks a bit hackish to me.
Why removing `0.1em`? Is it empirical?

pgiraud:
I just tried the following and it works well:
```CSS
.nav-trigger {
   ...
   /*
we remove 2px in order not to hide the border of the element behind
we were removing 3px which was wrong.
   */
   height: ~"calc("@map-tools-size ~" - 2px)";
   margin-top: 1px;
   ...
  }
  .nav-left-trigger {
   left: ~"calc("@app-margin ~" + 1px)";
  }
  .nav-right-trigger {
   right: ~"calc("@app-margin ~" + 1px)";
  }
```

sbrunner:
I'm ok with @pgiraud and in general is it a good idea to use the 'em' unit? for me we should never use this unit ...

blattmann:
@pgiraud I will use your code.

blattmann:
@sbrunner: I think em is ok. I like to use "em" as well as "rem" and as we have a flexible layout we should avoid "px" if possible. Here's a quite good article that explains when it's better to use "em" and when it's better to use "rem".
http://webdesign.tutsplus.com/tutorials/comprehensive-guide-when-to-use-em-vs-rem--cms-23984

sbrunner:
First, when I say that we should avoid using `em` it's not applicable to `rem` , that I think that the unit we should prefer.

For `em` it say in the article: « Working with em units can start to get tricky when it comes to inheritance », in other word If we redefine the font size it can be relay difficult if we use to many `em`...

sbrunner:
for example I don't think that a good idea to have the `@map-tools-size` and `app-margin` in `em`...

blattmann:
@sbrunner : I agree. I prefer using "rem" and I think it is worth to replace the "em" step by step with "rem", but then we should maybe also remove the base font size:

> Important to Know:
Some designers use rem based layouts in conjunction with a fixed px unit font-size setting on the html element. This is typically done so that a change of font size on the html element can scale overall the design up or down.

> I strongly advise against this as it overrides the font size the html element inherits from the user’s browser settings. Hence this prevents the settings from having any effect, and removes the user’s ability to optimize for best viewing.

> If you do want to change the font size on the html element, do so with an em or rem value as the root font size will then still be a multiple of the user’s browser font size setting.

> This will still allow you to scale your design up or down by changing your html element’s font size, but you’ll preserve the effect of user browser settings.

sbrunner:
I agree :-)

pgiraud:
We have to be careful though with bootstrap variables. To be tested.

sbrunner:
+1 @pgiraud, it should be done after the 2.0 release :-)

ybolognini:
+1 for the "done after the 2.0", thank you @sbrunner :)

PR #699, opened 2016-02-18
Mobile: Search field for wide screen or landscape adjusted

blattmann:
For mobile with a wide screen or mobile in landscape mode: Search input field is aligned left and clear button (cross) is aligned right.

### DIFF
--- a/contribs/gmf/less/mobile.less
+++ b/contribs/gmf/less/mobile.less
@@ -239,6 +239,20 @@ button[ngeo-mobile-geolocation] {
   right: @app-margin;
   left: auto;
 }
+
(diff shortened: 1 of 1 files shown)


pgiraud:
I get a weird display on tablet.
screenshot from 2016-02-18 12 44 07

blattmann:
Let's see, if this behaviour remains after the previous changes are merged.

pgiraud:
Which previous changes do you expect to modify this behavior?

blattmann:
There was a change for the input width and several changes more that might remove that behaviour if you did reuse these changes. As already mentioned in Hangouts I did not have any strange behaviour in my tests before so I expect it to be solved.

Issue #539, opened 2016-01-21
Documentations issues


sbrunner:
Missing example:
* [x] https://camptocamp.github.io/ngeo/master/apidoc/gmf.themeselectorDirective.html
* [x] https://camptocamp.github.io/ngeo/master/apidoc/ngeo.popupDirective.html

Missing some @htmlAttribute:
* [ ] https://camptocamp.github.io/ngeo/master/apidoc/gmf.layertreeDirective.html
* [ ] https://camptocamp.github.io/ngeo/master/apidoc/gmf.mapDirective.html
* [ ] https://camptocamp.github.io/ngeo/master/apidoc/gmf.mobileBackgroundLayerSelectorDirective.html
* [ ] https://camptocamp.github.io/ngeo/master/apidoc/ngeo.btngroupDirective.html
* [ ] https://camptocamp.github.io/ngeo/master/apidoc/ngeo.controlDirective.html
* [ ] https://camptocamp.github.io/ngeo/master/apidoc/ngeo.desktopGeolocationDirective.html
* [ ] https://camptocamp.github.io/ngeo/master/apidoc/ngeo.filereaderDirective.html
* [ ] https://camptocamp.github.io/ngeo/master/apidoc/ngeo.layertreeDirective.html
* [ ] https://camptocamp.github.io/ngeo/master/apidoc/ngeo.mapDirective.html
* [ ] https://camptocamp.github.io/ngeo/master/apidoc/ngeo.mobileGeolocationDirective.html
* [ ] https://camptocamp.github.io/ngeo/master/apidoc/ngeo.profileDirective.html
* [ ] https://camptocamp.github.io/ngeo/master/apidoc/ngeo.recenterDirective.html
* [ ] https://camptocamp.github.io/ngeo/master/apidoc/ngeo.scaleselectorDirective.html
* [ ] https://camptocamp.github.io/ngeo/master/apidoc/ngeo.searchDirective.html
* [ ] https://camptocamp.github.io/ngeo/master/apidoc/ngeo.sortableDirective.html

PR #701, opened 2016-02-18
Do the source link to github

sbrunner:
Removes the source copy

related to #539


sbrunner:
See result at: http://camptocamp.github.io/ngeo/sources-links/apidoc/gmf.authenticationDirective.html

gberaudo:
Great!

PR #703, opened 2016-02-18
Separate same type of layers by some groups

ger-benjamin:
To fix: https://github.com/camptocamp/ngeo/issues/673

Example: https://ger-benjamin.github.io/ngeo/master/examples/contribs/gmf/apps/mobile/index.html

That can (and should probably) be improved by implementing the same system in the ngeo background manager (cause it can only manage one background for now). But this can be done later.


ger-benjamin:
@fredj perhaps that change something for the search..? (activation of a layer)

fredj:
LGTM

PR #715, opened 2016-02-19
Update OpenLayers to current master

tsauerwein:
To follow the OpenLayers master a couple of small changes were required because:

 * a style function can now return a single style and not only an array of styles (see https://github.com/openlayers/ol3/pull/4401)
 * and because a fill style can now take a canvas gradient or pattern (see https://github.com/openlayers/ol3/pull/4632).

### DIFF
--- a/src/services/print.js
+++ b/src/services/print.js
@@ -415,16 +415,24 @@ ngeo.Print.prototype.encodeVectorLayer_ = function(arr, layer, resolution) {
     var geometryType = geometry.getType();
     var geojsonFeature = geojsonFormat.writeFeatureObject(feature);
 
-    var styles = null;
+    var styleData = null;
     var styleFunction = feature.getStyleFunction();
     if (goog.isDef(styleFunction)) {
-      styles = styleFunction.call(feature, resolution);
+      styleData = styleFunction.call(feature, resolution);
     } else {
       styleFunction = layer.getStyleFunction();
       if (goog.isDef(styleFunction)) {
-        styles = styleFunction.call(layer, feature, resolution);
+        styleData = styleFunction.call(layer, feature, resolution);
       }
     }
+
+    /**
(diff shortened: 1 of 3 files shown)


src/services/print.js
```diff
@@ -415,23 +415,30 @@ ngeo.Print.prototype.encodeVectorLayer_ = function(arr, layer, resolution) {
...
+    goog.asserts.assert(goog.isArray(styles));
+
     if (!goog.isNull(styles) && styles.length > 0) {
       geojsonFeatures.push(geojsonFeature);
       if (goog.isNull(geojsonFeature.properties)) {
         geojsonFeature.properties = {};
       }
       for (var j = 0, jj = styles.length; j < jj; ++j) {
-        var style = styles[j];
+        style = styles[j];
```
sbrunner: I'm not fan about reuse a variable ...

tsauerwein: True, me neither. Will change.

sbrunner:
Just a small comment, also that's looks good :-)

tsauerwein:
Thanks for the review, Stéphane!

@fredj @gberaudo I had to use `goog.isArray` instead of `Array.isArray` in the asserts because otherwise the compiler did not understand that it's an array. Maybe you have an idea?

fredj:
> @fredj @gberaudo I had to use goog.isArray instead of Array.isArray in the asserts because otherwise the compiler did not understand that it's an array. Maybe you have an idea?

This is a known issue but it's now fixed in closure-compiler

tsauerwein:
> This is a known issue but it's now fixed in closure-compiler

But we are not yet using this new version?

fredj:
> But we are not yet using this new version?

no, this change is very new and there's no release yet

tsauerwein:
Ok, thanks for the info, Fred!

Issue #707, opened 2016-02-19
Search field should always be aligned left and cross aligned right


ybolognini:
It looks like there are now three different behaviors:
* Small screen mobile (like Galaxy S4 Portrait) where search field + cross are centered
* Large screen mobile (like Galaxy S4 Landscape or iPhone 6) where search field is aligned left and cross is aligned right
* Tablet version

We should have only two : 
* Mobile with search field is aligned left and cross is aligned right
* Tablet

Screenshot shows search field + cross centered: should never happen
![galaxy_s4_search](https://cloud.githubusercontent.com/assets/1180264/13175207/1e8f2ce0-d709-11e5-9f5c-d2fdc30ed4f5.jpg)

eleu:
don't forget to set the label on "Ongoing" ;)

blattmann:
I have no rights to change the labels ;)
Thanks for setting it on "Ongoing".

eleu:
all ok then, we will set it.

Issue #710, opened 2016-02-19
More space left of search field


ybolognini:
Space before search field should be the same as space right of cross.

![search_field_space](https://cloud.githubusercontent.com/assets/1180264/13175663/4c3400aa-d70c-11e5-8429-cb369db150ff.jpg)

Issue #709, opened 2016-02-19
On tablet, search cross should be aligned right


ybolognini:
With same space as everywhere else.

![search_keyword_tablet](https://cloud.githubusercontent.com/assets/1180264/13175565/89a8c7fa-d70b-11e5-93aa-daf934e1b4fa.jpg)

Issue #708, opened 2016-02-19
Search keyword is displayed next to suggested result


ybolognini:
When typing in search field, suggested result is displayed next to keyword. See screenshot.

It looks better on tablet but I think there's the same problem (offset of 1 pixel?)

![search_keyword](https://cloud.githubusercontent.com/assets/1180264/13175537/56415fb2-d70b-11e5-9054-b66b31ec8afa.jpg)
![search_keyword_tablet](https://cloud.githubusercontent.com/assets/1180264/13175538/5654fb76-d70b-11e5-8703-04b91e120539.jpg)

PR #729, opened 2016-02-22
Search field is now always aligned left and cross aligned right.

blattmann:
Must fix: #707, #708, #709, #710

Search field and cross adjusted. Only 2 modes available now:
* Mobile view < 768px
* Tablet view > 768px

Example:
https://blattmann.github.io/ngeo/707_search_field_alignments/examples/contribs/gmf/apps/mobile/

### DIFF
--- a/contribs/gmf/less/mobile.less
+++ b/contribs/gmf/less/mobile.less
@@ -13,8 +13,8 @@
 @map-tools-size: 3em;
 @app-margin: 1em;
 
-@search-input-size: 200px;
-@search-input-width: ceil(@search-input-size + 9);
+@search-input-size: calc(~"100% - 7.75em");
+@search-input-size-clear: 200px;
 
 main>* {
   position: absolute;
@@ -118,7 +118,7 @@ span.twitter-typeahead {
     position: absolute;
     top: 0;
     left: 50%;
-    margin-left: ~"calc("@search-input-size / 2 ~" - " @map-tools-size~")";;
+    margin-left: ~"calc("@search-input-size-clear / 2 ~" - " @map-tools-size~")";
     height: @map-tools-size;
     width: @map-tools-size;
     line-height: @map-tools-size;
@@ -139,9 +139,6 @@ span.twitter-typeahead {
     outline: none;
     &:focus {
       box-shadow: none;
-    }
-    &.ng-pristine.ng-untouched.ng-valid.tt-hint {
-        left: 1em !important;
     }
   }
 }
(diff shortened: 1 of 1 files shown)


contribs/gmf/less/mobile.less
```diff
@@ -118,7 +118,7 @@ span.twitter-typeahead {
     position: absolute;
     top: 0;
     left: 50%;
-    margin-left: ~"calc("@search-input-size / 2 ~" - " @map-tools-size~")";;
+    margin-left: ~"calc("@search-input-size-clear / 2 ~" - " @map-tools-size~")";
```
ger-benjamin: The `~"..." ` is used to say to "less" to not interpret this content. That means that the browser will do the calculation. When we can, we prefer that less do the calculation, so our application is a little bit faster.

In this case, all units are in px (the "2" can be in pixel or other thing, that doesn't change the result) so less can do the complete job and I pretty sur that we can write this line like this:
`margin-left: calc(@search-input-size-clear / 2 - @map-tools-size);`

That's easier to read and faster for the browser.

blattmann:
@ger-benjamin :
Ok, if we do not need to calculate this value in the browser, we do not need to escape this line.
Regarding the performance we should use then:

`margin-left: ceil(@search-input-size / 2 - @map-tools-size);`

This produces a calculated value. E.g. `margin-left: 4em;` instead of `margin-left: calc(4em);`

What do you think?

ger-benjamin:
There is a lot of "min-device-width: @screen-sm-min" or max-device-width: @screen-sm-max rules.
That's hard to understand. 

Can you just move and regroups your code to have:
 - A default part first (for mobile but it's the default, we don't need @"screen-sm-..." rules)
- A part only for the tablette, starting with "@media (min-width: @screen-sm-min) {" and that overrides css rules that must be overrided (no needs of a rule for landscape and a rule for portrait, that's the same behaviors, so only one rule for tablette, at the end of this file).

(quick example, not working for anything)

```
input {
  width: @search-input-size;
  padding: 0 5px;
  border-width: 0;
  border-radius: 0;
  height: @map-tools-size - .2em;
}
/** ... **/

// For tablet only
@media (min-width: @screen-sm-min) {
 input {
    width: 100%
  }
}
/** ... **/
```

ger-benjamin:
> This produces a calculated value. E.g. margin-left: 4em; instead of margin-left: calc(4em);

That's a good idea !

ger-benjamin:
Looks good on the apps/mobile but please, take the time to understand my two comments and address them.

eleu:
note: addresses #707

ger-benjamin:
Clean on #743

Issue #684, opened 2016-02-17
Measures options in menu should be labels with checkbox


ybolognini:
When user wants to select Coordinate or Line mesures, menu entry should look like select backgrounds layer: text with checkbox. Buttons should be replaced.

PR #735, opened 2016-02-23
Use checkbox for measure buttons

ger-benjamin:
Fix: https://github.com/camptocamp/ngeo/issues/684

Example: https://ger-benjamin.github.io/ngeo/measure_options_layout/examples/contribs/gmf/apps/mobile/index.html?lang=en

### DIFF
--- a/contribs/gmf/apps/mobile/index.html
+++ b/contribs/gmf/apps/mobile/index.html
@@ -108,11 +108,11 @@
       <div id="measure-tools" class="slide" data-header-title="Measure tools">
         <a ngeo-btn
           ng-click="mainCtrl.hideNav()"
-          class="btn btn-default measure-tools-button"
+          class="measure-tools-button"
           ng-model="mainCtrl.measurePointActive">Coordinate</a>
         <a ngeo-btn
           ng-click="mainCtrl.hideNav()"
-          class="btn btn-default measure-tools-button"
+          class="measure-tools-button"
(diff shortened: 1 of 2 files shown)


pgiraud:
The result looks perfect to me.
Wouldn't it better to change the HTML to avoid using .btn and .btn-default at all?

ger-benjamin:
Yes, but that seems generated by the ngeo-btn directive...

ger-benjamin:
Done

pgiraud:
I would do more job in the HTML.
https://github.com/camptocamp/ngeo/compare/master...pgiraud:measure_options_layout
What do you think?

sbrunner:
Less we have in the HTML better it is, we shouldn't forget that the HTML is copy on all the client application, than he will be harder to maintain ...

sbrunner:
Be had a call with @pgiraud and finally it's not so hard HTML changes, it's OK for me :-)

contribs/gmf/apps/mobile/index.html
```diff
@@ -108,11 +108,11 @@
       <div id="measure-tools" class="slide" data-header-title="Measure tools">
         <a ngeo-btn
           ng-click="mainCtrl.hideNav()"
-          class="btn btn-default measure-tools-button"
+          class="measure-tools-button"
           ng-model="mainCtrl.measurePointActive">Coordinate</a>
```
sbrunner: Missing translate

PR #743, opened 2016-02-23
Clean and fix search issues

ger-benjamin:
Supersedes #729
Must:
* [x] Fix: #707
* [x] Fix: #708 
* [x] Fix: #709 
* [x] Fix: #710
* + Clean this old file

Example: https://ger-benjamin.github.io/ngeo/debug_and_clean_search/examples/contribs/gmf/apps/mobile/?lang=en

We probably can do somethings better with all these not justified "em" units... But I think that good enough for now.
I've moved the "search" part in another less file. But we should done the same thing on desktop later.

### DIFF
--- a/contribs/gmf/less/desktop.less
+++ b/contribs/gmf/less/desktop.less
@@ -140,7 +140,7 @@ main {
           margin-bottom: 0;
         }
         .search-label {
-          color: @search-result-font-color;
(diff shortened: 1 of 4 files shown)


ger-benjamin:
Please review

sbrunner:
Looks good to me

Issue #733, opened 2016-02-23
The translate methods don't work for the gettext collector


sbrunner:
This kind of thing don't works with `angular-gettext`
https://github.com/camptocamp/ngeo/blob/master/contribs/gmf/src/directives/mobilemeasurelength.js#L260-L267

we should use the syntax:
```html
<tag translate>text</tag>
```
or
```html
<tag attribute="{{'text'|translate}}" ...>
```

But the method `gettextCatalog_.getString()` should be used for (the `translate` don't works):
* Theme names
* Node names
* Background layers names
* Raster layer names

sbrunner:
I just modify the description...

PR #747, opened 2016-02-23
Use `translate` filter

tsauerwein:
Closes https://github.com/camptocamp/ngeo/issues/733

@sbrunner, please check if this includes everything you had in mind with #733.

### DIFF
--- a/contribs/gmf/src/directives/partials/mobilemeasurelength.html
+++ b/contribs/gmf/src/directives/partials/mobilemeasurelength.html
@@ -2,29 +2,29 @@
    ng-if="ctrl.drawing && (!ctrl.valid)"
    ng-click="ctrl.addPoint()">
      <span class="fa fa-check"></span>
-     {{ ::ctrl.translate('Set as starting point') }}
+     {{'Set as starting point' | translate}}
 </a>
 <a class="btn btn-default"
    ng-if="ctrl.dirty"
    ng-click="ctrl.addPoint()">
      <span class="fa fa-plus"></span>
-     {{ ::ctrl.translate('Add new point') }}
+     {{'Add new point' | translate}}
 </a>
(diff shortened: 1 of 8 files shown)


contribs/gmf/src/directives/partials/mobilemeasurelength.html
```diff
@@ -2,29 +2,29 @@
    ng-if="ctrl.drawing && (!ctrl.valid)"
    ng-click="ctrl.addPoint()">
      <span class="fa fa-check"></span>
-     {{ ::ctrl.translate('Set as starting point') }}
+     {{::'Set as starting point' | translate}}
```
sbrunner: The collecting is working with that?

sbrunner: Effectively it should :-)
Any reason about the `::`?

sbrunner:
An other point to fix...
https://github.com/tsauerwein/ngeo/blob/i18n-measure/contribs/gmf/src/directives/authenticationdirective.js#L282

Also that's looks good :-)

contribs/gmf/src/directives/partials/mobilemeasurelength.html
tsauerwein: > Any reason about the ::?

Good point. We can't use one-time-binding because otherwise the text would not be updated when changing the language. I will remove it.

tsauerwein:
For the authentication directive I changed `this.translate(...)` to `this.gettextCatalog.getString(...)`. This is recognized by angular-gettext-tools.

sbrunner:
@tsauerwein, OK, Excellent :-)
Then that looks good to merge :-) (When Travis is happy)

tsauerwein:
Thanks for the review!

PR #749, opened 2016-02-24
Handle add_group and add_layer search result

fredj:
TODO:
 * display a notification message if the selected group is already present in the layer tree.
 * check if it works in compiled mode
 * add_layer: 
   * if the layer exists in the current theme, set visibility to true


fredj:
The remaining issues will be addressed later

ger-benjamin:
> The remaining issues will be addressed later

Here: https://github.com/camptocamp/ngeo/issues/883

And some `@type {Object}` must be changed to `@type {GmfThemesNode}` but I'll do this very soon in another PR.

ger-benjamin:
Thanks, Please merge when travis passes

Issue #754, opened 2016-02-24
[Mobile/Tablet] Layout details - queryer


ger-benjamin:
In the display queryer:
* [x] On small screens, texts recovers each others.
* [x] Boundingby properties should not be displayed (for this point, see also https://github.com/camptocamp/ngeo/issues/756)

PR #755, opened 2016-02-24
Query small details

ger-benjamin:
Fix: https://github.com/camptocamp/ngeo/issues/754

Demo: https://ger-benjamin.github.io/ngeo/query-small-details/examples/contribs/gmf/apps/mobile/index.html?lang=en

For the second commit (quick fix for the demo), we need a better solution. See issue https://github.com/camptocamp/ngeo/issues/756


sbrunner:
Looks good :-)

PR #761, opened 2016-02-25
ngTouch is required for gmf thus for desktop as well

pgiraud:
Currently desktop app doesn't load correctly.
`angular-touch.js` needs to be loaded for desktop app since it's required by the `gmf` module.

Please review.


sbrunner:
Looks good :-)

fgravin:
As  `ngTouch` is not used for desktop, it is useless to load it.
I would separate `app.module` into 2 modules for each platform respectivly, and inject the good module dependencies from each one.

PR #763, opened 2016-02-25
Don't fail on gh-pages cleanup

ger-benjamin:
@sbrunner Something like that ?


sbrunner:
Looks good, merge :-)

PR #786, opened 2016-03-01
Use GMF fulltext-search web-service for ngeo search example

### LINKED ISSUE #775: Hosted ngeo-search does not work (CORS issue)
http://camptocamp.github.io/ngeo/master/examples/search.html

Could https://geomapfish-demo.camptocamp.net/2.0/wsgi/fulltextsearch also be used in this example?

tsauerwein:
closes #775

### DIFF
--- a/examples/search.js
+++ b/examples/search.js
@@ -158,9 +158,9 @@ app.SearchController.prototype.createVectorLayer_ = function() {
  * @private
  */
 app.SearchController.prototype.createAndInitBloodhound_ = function(ngeoCreateGeoJSONBloodhound) {
-  var url = 'http://devv3.geoportail.lu/main/wsgi/fulltextsearch?query=%QUERY';
-  var bloodhound = ngeoCreateGeoJSONBloodhound(url, undefined,
-                                               ol.proj.get('EPSG:3857'));
+  var url = 'https://geomapfish-demo.camptocamp.net/2.0/wsgi/fulltextsearch?query=%QUERY';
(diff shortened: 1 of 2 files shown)


examples/search.js
```diff
@@ -194,6 +194,11 @@ app.module.controller('AppSearchController', app.SearchController);
  * @constructor
  */
 app.MainController = function() {
+  proj4.defs('EPSG:21781',
```
fredj: you can `goog.require('gmf.proj.EPSG21781');`instead

tsauerwein: There are more places that redefine 21781, I will replace them.

But what about using stuff from the gmf module in a pure ngeo example?

fredj:
> But what about using stuff from the gmf module in a pure ngeo example?

Yes, that's a good point ...

examples/search.js
sbrunner: Shouldn't that mean that the projections should be in ngeo?

tsauerwein: > Shouldn't that mean that the projections should be in ngeo?

That would make sense!

fredj: > Shouldn't that mean that the projections should be in ngeo?

+1

fredj:
Please merge and open an issue to move `gmf.proj.EPSG21781` to `ngeo`

tsauerwein:
Thanks, for the review. New issue is here: https://github.com/camptocamp/ngeo/issues/789

PR #790, opened 2016-03-01
Add missing decimals, displayHelpTooltip and sketchStyle properties to measure options

fredj:


tsauerwein:
LGTM

Would it be useful to document these properties?

Issue #767, opened 2016-02-25
The blank layer should be collected in the strings to translate


sbrunner:
https://github.com/camptocamp/ngeo/blob/169c902ec5bc633438778791a9d983fdfe1e2192/contribs/gmf/src/services/themesservice.js#L148

PR #793, opened 2016-03-01
Mark label for blank layer as translatable

tsauerwein:
Closes https://github.com/camptocamp/ngeo/issues/767


sbrunner:
Is it collected?
If yes, it's so simple why don't we try it :-)
If yes merge :-)

tsauerwein:
Sure, it shows up in the pot file.

sbrunner:
Thanks :-)

Issue #789, opened 2016-03-01
Move `gmf.proj.EPSG21781` into ngeo module


tsauerwein:
* Move `gmf.proj.EPSG21781` into ngeo module
 * Remove manual `proj4.defs('EPSG:21781', ...` definitions in examples

PR #801, opened 2016-03-02
Whiteout when search is active for mobile

sbrunner:
Replace #789 

When the search field is active, hide anything else (for small screens).

Demo: https://tsauerwein.github.io/ngeo/search-overlay/examples/contribs/gmf/apps/mobile/

Closes https://github.com/camptocamp/ngeo/issues/778

### DIFF
--- a/contribs/gmf/less/mobile-nav.less
+++ b/contribs/gmf/less/mobile-nav.less
@@ -24,9 +24,9 @@ main {
   .overlay {
     /* shadow layer visible when navigation is active */
     position: absolute;
-    z-index: @above-content-index;
-    height: 100%;
-    width: 100%;
+    z-index: @above-search-index;
+    height: 100vh;
+    width: 100vw;
     top: 0;
     left: 0;
     cursor: pointer;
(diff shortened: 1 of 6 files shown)


tsauerwein:
The PR now pointing on 2.0 makes the Travis build fail:

    The command "if [ ${TRAVIS_PULL_REQUEST} != "false" ] ; then git diff --check ${TRAVIS_BRANCH} -- ; fi" exited with 128.

https://travis-ci.org/camptocamp/ngeo/jobs/113077736#L345

sbrunner:
I will have a look :-)

sbrunner:
Will be fixed by: #803

tsauerwein:
Rebased to fix the Travis build

contribs/gmf/less/mobile-nav.less
```diff
@@ -44,6 +44,30 @@ main {
   }
 }
 
+.search-overlay {
+  /* shadow layer visible when search is active */
+  position: absolute;
+  z-index: @above-menus-index;
+  height: 100%;
+  width: 100%;
```
sbrunner: I prefer:
100vh
100vw

ybolognini:
@tsauerwein @pgiraud Following discussion today, please keep left and right buttons

contribs/gmf/less/mobile-nav.less
tsauerwein: Does it make a difference with `position: absolute`? Or what advantages does it have?

tsauerwein:
PR updated, the buttons are now visible when the search is active.

contribs/gmf/less/mobile-nav.less
sbrunner: Using less deferent units maps the addition directly with less possible in more cases.
And I prefer using vw and vh against % because we shouldn't worry about the parent div size.
finally your right that in this case it don't change anything :-)

ybolognini:
Looks great, thanks

contribs/gmf/less/mobile-nav.less
tsauerwein: Thanks, for the explanation. I replaced it with vw/vh.

sbrunner:
Looks good for me :-)

pgiraud:
:+1:

tsauerwein:
Thanks for the reviews.

PR #805, opened 2016-03-02
Print dashed lines

rmichaelis:


src/services/print.js
```diff
@@ -619,6 +619,10 @@ ngeo.Print.prototype.encodeVectorStyleStroke_ = function(symbolizer, strokeStyle
     symbolizer.strokeColor = goog.color.rgbArrayToHex(strokeColorRgba);
     symbolizer.strokeOpacity = strokeColorRgba[3];
   }
+  var strokeDashstyle = strokeStyle.getLineDash();
+  if (goog.isDefAndNotNull(strokeDashstyle)) {
```
gberaudo: You should use `strokeDashstyle !== undefined` instead of goog.

rmichaelis: Done. Thanks for the tip.
Why did you decide these last weeks to avoid the use of goog.isDef ?

gberaudo: We want the code to use as much as possible standard javascript.

fredj:
+1

fredj:
And this correct to have `strokeDashstyle !== null` because `strokeStyle.getLineDash()`return an arraw or null

PR #806, opened 2016-03-02
Gets the l10n strings from the demo to the apps

### LINKED ISSUE #788: Get the localisation from the demo application for the apps

sbrunner:
Fix #788

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -64,7 +64,8 @@ GIT_REMOTE_NAME ?= origin
 
 # i18n
 L10N_LANGUAGES = fr de it
-L10N_PO_FILES = $(addprefix c2cgeoportal/locale/,$(addsuffix /LC_MESSAGES/c2cgeoportal.po, $(L10N_LANGUAGES)))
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -678,13 +680,20 @@ transifex-init: .build/dev-requirements.timestamp .tx/config \
 	.build/python-venv/bin/tx pull -l $* --force
 	$(TOUCHBACK_TXRC)
 
+.PRECIOUS: .build/locale/%/LC_MESSAGES/demo.po
+.build/locale/%/LC_MESSAGES/demo.po:
+	mkdir -p $(dir $@)
+	wget -O $@ https://raw.githubusercontent.com/camptocamp/demo_geomapfish/master/demo/locale/$*/LC_MESSAGES/demo-client.po
```
ger-benjamin: I hope that doesn't create some problems like in https://github.com/camptocamp/ngeo/blob/master/buildtools/cleanup-ghpages.py#L14

ger-benjamin:
Please merge

Makefile
sbrunner: I Hope, It it's the case we should rework on it :-)

PR #807, opened 2016-03-02
Don't removes anything at the end of the build

sbrunner:
move `contribs/gmf/build/%.closure.js` to `.build/%.js` (needed only during the build)


ger-benjamin:
Seems alright

PR #815, opened 2016-03-04
Search for coordinates with marker

blattmann:
* parsing the value from the search input field
* looking for some geocode
* if a geocode was found it will be attached to the URL in `getBloodhoudRemoteOptions_` and the marker should be set on the map in `matchCoordinate_`


ger-benjamin:
See: https://github.com/camptocamp/ngeo/pull/855

PR #824, opened 2016-03-04
Format booth projection as the same

### LINKED ISSUE #797: Uniformise projections
This should be reformatted an in 21781:
https://github.com/camptocamp/ngeo/blob/master/src/proj/epsg2056projection.js#L6-L9

This should be removed:
https://github.com/camptocamp/ngeo/blob/master/src/proj/epsg21781projection.js#L20
also change here:
https://github.com/camptocamp/ngeo/blob/master/contribs/gmf/examples/mobilebackgroundlayerselector.js#L44

sbrunner:
Fixes  #797


fredj:
+1

PR #832, opened 2016-03-07
Use 'key in obj' instead of 'goog.object.containsKey(obj, key)'

fredj:


sbrunner:
Looks good :-)

gberaudo:
+1. Can be simplified by removing the tests for `undefined`:
`undefined in {} -> false`

fredj:
thanks @gberaudo, code updated

sbrunner:
... I don't see any changes ...

gberaudo:
There is actually a closure compilation issue when using `undefined`.
Thanks @fredj, LGTM for merging.

PR #837, opened 2016-03-07
Use '!== undefined' instead of angular.isDefined()

### LINKED ISSUE #791: Stop using angular.isDefined
We should use `obj !== undefined` instead of `angular.isDefined(obj)`

ger-benjamin:
Fix #791

### DIFF
--- a/examples/layerorder.js
+++ b/examples/layerorder.js
@@ -139,7 +139,7 @@ app.MainController = function($scope, ngeoDecorateLayer, ngeoSyncArrays) {
  * @export
  */
 app.MainController.prototype.toggleRoadsLayer = function(val) {
-  if (!angular.isDefined(val)) {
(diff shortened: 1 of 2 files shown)


examples/layerorder.js
```diff
@@ -139,7 +139,7 @@ app.MainController = function($scope, ngeoDecorateLayer, ngeoSyncArrays) {
  * @export
  */
 app.MainController.prototype.toggleRoadsLayer = function(val) {
-  if (!angular.isDefined(val)) {
+  if (val === undefined) {
```
fredj: `val !== undefined`

fredj:
+1

examples/layerorder.js
ger-benjamin: Nop :walking:

PR #839, opened 2016-03-07
Fix LayerHelper group creation

adube:
This PR fixes the ngeo LayerHelper `createBasicGroup` method, allowing it to actually use the already created groups and return them.  Inside a `forEach`, calling `return` didn't actually did the return of the `createBasicGroup` method, but rather the anonymous function used for the forEach.


adube:
This is ready for review.  @fredj maybe you could look at this ?   Thanks.

fredj:
LGTM, thanks

PR #843, opened 2016-03-08
Replace export files with @export

### LINKED ISSUE #799: Use export notation
Instead of having export files in the `exports` directory, we should use the `@export` notation

tsauerwein:
Closes #799


fredj:
Looks very very good to me ! Thanks Tobias

tsauerwein:
Thanks for the review!

PR #846, opened 2016-03-08
Get the demo l10n also for the english

sbrunner:


fredj:
And what if you completely remove `contribs/gmf/build/gmf-en.json`?
isn't the `contribs/gmf/build/gmf-%.json` target sufficient ?

sbrunner:
I suspected that it will failed when we get the en location from Transifex, but finally it should work :-)

fredj:
Nice, please merge when travis is green

Issue #712, opened 2016-02-19
On iOS, button icons aren't centered horizontaly


ybolognini:
Icons inside buttons (data, tools, geolocation) aren't centered on iPad.

pgiraud:
![screenshot from 2016-02-22 10 56 04](https://cloud.githubusercontent.com/assets/319774/13215127/911b9a9a-d953-11e5-95f3-154e15fa7f7a.png)

The icons are a bit offset to the right. Zoom +/- are correctly centered.

ebelo:
Attention: only to handle in this ticket the 3 icones: geolocation, tool and menu

pgiraud:
Title and description modified accordingly.

blattmann:
I have no rights to change the labels ;)
Thanks for setting it on "Ongoing".

eleu:
can't reproduce this on https://camptocamp.github.io/ngeo/master/examples/contribs/gmf/apps/mobile/ any more -> already fixed?

eleu:
only visible on ipad real device. postponed for the moment.

ger-benjamin:
Exacte. I can say that this issue is visible only with "imported" fonts, like fontAwesome **and** our gmf-icons. Perhaps it's something in the definition of the font but I don't found anybody that has the same problem...

ger-benjamin:
As say in the title, that's on IOS, not only on Ipad

PR #849, opened 2016-03-08
Make sure buttons are centered horizontally on iOS

tsauerwein:
Demo: https://tsauerwein.github.io/ngeo/ios-center-button/examples/contribs/gmf/apps/mobile/index.html

Closes https://github.com/camptocamp/ngeo/issues/712

### DIFF
--- a/contribs/gmf/less/mobile.less
+++ b/contribs/gmf/less/mobile.less
@@ -23,6 +23,7 @@ main>* {
 main button {
   height: @map-tools-size;
   width: @map-tools-size;
+  padding: 0;
 }
 
 gmf-map {


sbrunner:
Isn't it an issue of the reset.css?
Should the button be added in this block?
https://github.com/shannonmoeller/reset-css/blob/master/reset.css#L6-L18

sbrunner:
More investigation:
* normalise.css don't reset it: https://github.com/necolas/normalize.css/blob/master/normalize.css
* yui2 do it: https://github.com/yui/yui2/blob/master/sandbox/reset/reset.css

...

tsauerwein:
I wouldn't set it for all buttons. For buttons in forms you probably want to keep the default browser-specific style.

sbrunner:
OK, then that's looks good for me :-)

fredj:
+1, travis fail because of #848

tsauerwein:
Thanks for the info, Fred!

I will merge when Travis passes.

Issue #835, opened 2016-03-07
Some correction in the documentation:


sbrunner:
Some correction in the documentation:

http://camptocamp.github.io/ngeo/master/apidoc/ngeo.controlDirective.html
change description to: 
```
Provides a directive can be used to add a control to the map using a DOM element.
```
http://camptocamp.github.io/ngeo/master/apidoc/ngeo.layertreeDirective.html add missing '>' in the examples

http://camptocamp.github.io/ngeo/master/apidoc/ngeo.sortableDirective.html
```
Provides the "ngeoSortable" directive. This directive allows drag-and-dropping DOM items between them. The directive also changes the order of elements in the array it is given.
```
=>
```
Provides a directive that allows drag-and-dropping DOM items between them. It also changes the order of elements in the given array.
```

replace "ngeo-sortable" by `ngeo-sortable` and "ngeoSortable" by `ngeoSortable`

http://camptocamp.github.io/ngeo/master/apidoc/gmf.mobiledisplayqueriesDirective.html 
```
Provide a "gmf mobile display queries" directive.

This directive displays results of the ngeoQueryResult and shows related features on the map using the ngeoFeatureOverlayMgr.

You can override the default directive's template by setting the value 'gmfMobiledisplayqueriesTemplateUrl'.
```
=>
```
Provide a directive to displays results of the {@link ngeo.queryResult} and shows related features on the map using the {@link ngeo.FeatureOverlayMgr}.

You can override the default directive's template by setting the value `gmfMobiledisplayqueriesTemplateUrl`.
```

http://camptocamp.github.io/ngeo/master/apidoc/gmf.mobileMeasureLengthDirective.html
```
Provide a "mobile measure" directive.
```
=>
```
Provide a directive to do a length measure on the mobile devices.
```

http://camptocamp.github.io/ngeo/master/apidoc/gmf.mobileMeasurePointDirective.html
```
Provide a "mobile measure" directive.
```
=>
```
Provide a directive to do a point (coordinate and élévation) measure on the mobile devices.
```

http://camptocamp.github.io/ngeo/master/apidoc/gmf.searchDirective.html
```
A "search" directive that allows ...
```
=>
```
A directive that allows ...
```

```
This directive uses the ngeoFeatureOverlayMgr to create a feature overlay for drawing features on the map. The application is responsible to initialize the ngeoFeatureOverlayMgr with the map.
```
=>
```
[...]

PR #851, opened 2016-03-08
Correction in the documentation

blattmann:
#835

Some correction in the documentation performed.


sbrunner:
Thanks, but you should also rebase on the new branch

Issue #834, opened 2016-03-07
Add two way link from api to example


sbrunner:
For example add a tow way link between 
http://camptocamp.github.io/ngeo/master/examples/control.html
and
http://camptocamp.github.io/ngeo/master/apidoc/ngeo.controlDirective.html

blattmann:
Should I create the PR for this task also for branch 2.0?

sbrunner:
Yes :-)

blattmann:
Thanks ;)

PR #853, opened 2016-03-09
Two way links from api to example added

blattmann:
Related issue: #834


blattmann:
The PR did not use the selected Branch 2.0. Will close and reopen.

Issue #696, opened 2016-02-18
Search for coordinates


ybolognini:
Add search for coordinates.

When searching for coordinates, the map recenters. Add a marker to the center, just like the highlight in case of search results.

Spec for general coordinate search and marker: https://github.com/camptocamp/c2cgeoportal/wiki/Spec-%23696-Search-for-coordinates-with-marker

sbrunner:
same for points

fgravin:
Related to https://github.com/camptocamp/c2cgeoportal/issues/2044

ger-benjamin:
Close with #855

PR #855, opened 2016-03-09
Search coordinates

ger-benjamin:
Ref to: https://github.com/camptocamp/ngeo/issues/696 and https://github.com/camptocamp/c2cgeoportal/issues/2044
Specs:  https://github.com/camptocamp/c2cgeoportal/wiki/Spec-%23696-Search-for-coordinates-with-marker

Example: https://ger-benjamin.github.io/ngeo/search_coordinates/examples/contribs/gmf/apps/mobile

### DIFF
--- a/contribs/gmf/apps/desktop/index.html
+++ b/contribs/gmf/apps/desktop/index.html
@@ -87,6 +87,7 @@
           <gmf-search gmf-search-map="mainCtrl.map"
             gmf-search-datasources="mainCtrl.searchDatasources"
             gmf-search-currenttheme="mainCtrl.theme"
+            gmf-search-coordinatesprojections="mainCtrl.searchCoordinatesProjections"
             gmf-search-clearbutton="true">
           </gmf-search>
         </div>
--- a/contribs/gmf/src/directives/search.js
+++ b/contribs/gmf/src/directives/search.js
@@ -3,6 +3,7 @@ goog.provide('gmf.searchDirective');
 
 goog.require('gmf');
 goog.require('gmf.Themes');
+goog.require('ngeo.AutoProjection');
 goog.require('ngeo.CreateGeoJSONBloodhound');
 goog.require('ngeo.FeatureOverlay');
 goog.require('ngeo.FeatureOverlayMgr');
@@ -12,7 +13,9 @@ goog.require('ngeo.FeatureOverlayMgr');
  * @suppress {extraRequire}
  */
 goog.require('ngeo.searchDirective');
+goog.require('ol.Feature');
 goog.require('ol.Map');
+goog.require('ol.geom.Point');
 goog.require('ol.proj');
 
 
@@ -44,6 +47,7 @@ gmf.module.value('gmfSearchTemplateUrl',
  *      <gmf-search gmf-search-map="ctrl.map"
  *        gmf-search-datasources="ctrl.searchDatasources"
  *        gmf-search-currenttheme="ctrl.theme"
+ *        gmf-search-coordinatesprojections="ctrl.searchCoordinatesProjections"
  *        gmf-search-clearbutton="true">
  *      </gmf-search>
  *      <script>
@@ -60,6 +64,7 @@ gmf.module.value('gmfSearchTemplateUrl',
  *      <gmf-search gmf-search-map="ctrl.map"
  *        gmf-search-datasources="ctrl.searchDatasources"
  *        gmf-search-currenttheme="ctrl.theme"
+ *        gmf-search-coordinatesprojections="ctrl.searchCoordinatesProjections"
  *        gmf-search-clearbutton="true">
  *      </gmf-search>
  *      <script>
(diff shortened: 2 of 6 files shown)


fredj:
Instead of having `exports/services/autoprojection.js` add a `@export` to the methods you want to export

src/services/autoprojection.js
```diff
@@ -0,0 +1,85 @@
...
+ * TODO
+ * @param {ol.Coordinate} coordinates TODO
+ * @param {ol.proj.Projection} projection TODO
+ * @param {Array.<ol.proj.Projection>=} opt_projections TODO
+ * @return {?ol.Coordinate} TODO
+ * @export
+ */
+ngeo.AutoProjection.prototype.tryProjections = function(coordinates, projection,
+    opt_projections) {
+  var extent = projection.getExtent();
```
sbrunner: I think that we should try to get the extent from the view ...

src/services/autoprojection.js
```diff
@@ -0,0 +1,85 @@
...
+ngeo.AutoProjection = function() {};
+
+
+/**
+ * TODO
+ * @param {string} str TODO
+ * @return {?ol.Coordinate} TODO
+ * @export
+ */
+ngeo.AutoProjection.prototype.stringToCoordinates = function(str) {
```
sbrunner: If str is `' '` I can't predict that the code will do...
I think that if one of the `parseFloat` return `NaN`, this function should return `null`.

tsauerwein: It would be good to have unit-tests for that.

ger-benjamin:
Normally, the style used for the marker must be customizable but I think that this point must be in another PR / issue. Because I think that if we be able to customize the "coordinates" marker, we can be able to customize the "normal search" style too.

sbrunner:
We should use the extent set on the view if it's define, if not we will use the projection extent :-)
We don't want to recenter on a point that not possible to use as a center ...
Thanks for the rename :-)

ger-benjamin:
Done. I've added a "extent" parameter. More parameters but this is more logical at the end I think.

src/services/autoprojection.js
```diff
@@ -0,0 +1,92 @@
...
+ngeo.AutoProjection.prototype.tryProjections = function(coordinates,
+    extent, viewProjection, opt_projections) {
+  var position;
+  if (opt_projections === undefined) {
+    opt_projections = [viewProjection];
+  }
+  opt_projections.some(function(projection) {
+    position = ol.proj.transform(coordinates, projection, viewProjection);
+    if (ol.extent.containsCoordinate(extent, position)) {
+      return true;
```
sbrunner: not `return projection`?

src/services/autoprojection.js
```diff
@@ -0,0 +1,92 @@
...
+    }
+    position = null;
+  });
+  return position;
```
sbrunner: can just be `return null`...

contribs/gmf/apps/desktop/index.html
```diff
@@ -87,6 +87,7 @@
           <gmf-search gmf-search-map="mainCtrl.map"
             gmf-search-datasources="mainCtrl.searchDatasources"
             gmf-search-currenttheme="mainCtrl.theme"
+            gmf-search-coordinatesprojections="'EPSG:21781,EPSG:2056,EPSG:4326'"
```
sbrunner: Shouldn't be in the abstractController?

src/services/autoprojection.js
ger-benjamin: No, that's in a "some" function. I use `return true` as a `break` in my loop.

src/services/autoprojection.js
ger-benjamin: No, that's in a "some" function. I use `return true` as a `break` in my loop.

contribs/gmf/src/directives/search.js
```diff
@@ -3,17 +3,22 @@ goog.provide('gmf.searchDirective');
...
 goog.require('ngeo.FeatureOverlayMgr');
+goog.require('ol.Feature');
+goog.require('ol.geom.Point');
 /**
  * This goog.require is needed because it provides 'ngeo-search' used in
  * the template.
  * @suppress {extraRequire}
  */
 goog.require('ngeo.searchDirective');
+//goog.require('ol.Feature');
```
sbrunner: should be removed

contribs/gmf/src/directives/search.js
```diff
@@ -3,17 +3,22 @@ goog.provide('gmf.searchDirective');
...
 goog.require('ol.Map');
 goog.require('ol.proj');
+//goog.require('ol.geom.Point');
```
sbrunner: also

sbrunner:
Two comments to remove also that's looks good for me :-)

tsauerwein:
Thanks for adding the tests!

sbrunner:
@tsauerwein +1

sbrunner:
Thanks, merge :-)

PR #863, opened 2016-03-11
Ensure tooltips are dismissed even after click

pgiraud:
The default configuration for tooltip is "hover focus". With `focus`, the tooltip may stay displayed when clicking on a button until it looses focus.
This for example happen in the desktop application when clicking on a tool button on the right side bar.


sbrunner:
Looks good :-)

PR #871, opened 2016-03-14
[Work in progress] Load the gmfThemes in gmf.searchDirective via the $injector

fredj:
Simplifies the search example

### DIFF
--- a/contribs/gmf/examples/search.js
+++ b/contribs/gmf/examples/search.js
@@ -17,19 +17,14 @@ var app = {};
 /** @type {!angular.Module} **/
 app.module = angular.module('app', ['gmf']);
 
-app.module.constant('gmfTreeUrl', 'data/themes.json');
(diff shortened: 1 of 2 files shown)


contribs/gmf/examples/search.js
```diff
@@ -17,25 +17,22 @@ var app = {};
 /** @type {!angular.Module} **/
 app.module = angular.module('app', ['gmf']);
 
-app.module.constant('gmfTreeUrl', 'data/themes.json');
```
sbrunner: No mode defined?

fgravin:
Why not

fredj:
Note: a loadTheme() in gmf.SearchController is probably needed

fgravin:
Why are you doing all of this for the search firective ? (init the layermanager, load themes etc..)

PR #873, opened 2016-03-14
Set groupValues and groupActions in search example

fredj:
Don't allow to select theme search actions (add_theme, add_group or add_layer)


fredj:
Please review

fgravin:
LG

fredj:
Thanks

PR #878, opened 2016-03-14
Remove the need to call gmf.Themes#loadThemes before any other functions

### LINKED ISSUE #877: theme service and loadThemes
It's currently not possible to have:
```js
gmfThemes.getThemeObject().then(function() {
  console.log('loaded');
});
gmfThemes.loadThemes();
```
Because the promise is created in `loadThemes`.

The workaround is to the the `gmf.ThemesEventType.LOAD` event.

fredj:
And remove the now unneeded `gmf.ThemesEventType.LOAD` event

fixes #877

### DIFF
--- a/contribs/gmf/src/services/themesservice.js
+++ b/contribs/gmf/src/services/themesservice.js
@@ -1,9 +1,7 @@
 goog.provide('gmf.Themes');
-goog.provide('gmf.ThemesEventType');
 
 goog.require('gmf');
 goog.require('goog.array');
-goog.require('goog.asserts');
 goog.require('ngeo.LayerHelper');
 goog.require('ol.events.EventTarget');
 goog.require('ol.layer.Tile');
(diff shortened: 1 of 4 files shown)


fgravin:
LGTM

contribs/gmf/src/services/themesservice.js
```diff
@@ -212,20 +204,18 @@ gmf.Themes.prototype.getThemesObject = function() {
...
-   */
-  var promiseSuccessFn = function(resp) {
-    this.dispatchEvent(gmf.ThemesEventType.LOAD);
-    return /** @type {gmf.ThemesResponse} */ (resp.data);
-  }.bind(this);
 
-  this.promise_ = this.$http_.get(this.treeUrl_, {
+  var deferred = this.deferred_;
+
+  this.$http_.get(this.treeUrl_, {
```
gberaudo: What should happen when the `loadTheme` method is called several times?

fredj: Bad things ...
That's probably an argument to move the theme loading in the constructor

PR #879, opened 2016-03-14
Fix themes.json - mixed should have a boolean value

adube:
This PR fixes the `"mixed"` values inside the themes.json file that were strings to boolean.

The official theme service returns a boolean for `"mixed"`.

### DIFF
--- a/contribs/gmf/examples/data/themes.json
+++ b/contribs/gmf/examples/data/themes.json
@@ -7,7 +7,7 @@
       "icon": "https://geomapfish-demo.camptocamp.net/2.0/wsgi/project/722c806f06024a77858af5f2433a27cc/img/enseignement.jpeg",
       "children": [
         {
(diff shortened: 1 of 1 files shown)


adube:
@fgravin Please, review.

fredj:
LGTM

fredj:
there are other "true" and "false" in this file (`wfsSupport`, `isChecked`, ...)

adube:
Yes, but not all of them are correctly declared in the `gmf-themes.js` extern file, plus they are used as string when compared, i.e. `if (value == 'true')`.  I wouldn't touch them as part of this fix, as it could become much bigger of an issue.

I agree that this needs to be consistent, but this should be part of a separate task where both the server and client should be fixed.  Do you agree ?

fredj:
Thanks @adube

PR #881, opened 2016-03-14
Allow the scales array to be refreshed

rmichaelis:
The goal of this PR is to allow to the list of scales to be refreshed.
The use case in our app is that the list of available scales is theme dependent. When a user switches the current theme, then the number of scales could vary.

My first thought was to listen on the maxzoom and minzoom property of the view and to limit the scales using these values. But none of both are observable. Therefore I've chosen this solution but I'm not sure, I've chosen the best way to achieve this. Could you please review and give advices.

### DIFF
--- a/src/directives/scaleselector.js
+++ b/src/directives/scaleselector.js
@@ -117,12 +117,12 @@ ngeo.ScaleselectorController = function($scope, $element, $attrs) {
    */
   this.zoomLevels;
 
-  $scope.$watchCollection(goog.bind(function() {
-    return this.scales;
-  }, this), goog.bind(function(newNames) {
+  $scope.$watch(function() {
+    return Object.keys(this.scales).length;
+  }.bind(this), function(newLength) {
     this.zoomLevels = Object.keys(this.scales).map(Number);
     this.zoomLevels.sort(ol.array.numberSafeCompareFunction);
-  }, this));
+  }.bind(this));
 
   var mapExpr = $attrs['ngeoScaleselectorMap'];
 
--- a/src/directives/scaleselector.js
+++ b/src/directives/scaleselector.js
@@ -111,14 +111,18 @@ ngeo.ScaleselectorController = function($scope, $element, $attrs) {
       ($scope.$eval(scalesExpr));
   goog.asserts.assert(this.scales !== undefined);
 
-  var zoomLevels = Object.keys(this.scales).map(Number);
-  zoomLevels.sort(ol.array.numberSafeCompareFunction);
-
   /**
    * @type {Array.<number>}
    * @export
    */
-  this.zoomLevels = zoomLevels;
+  this.zoomLevels;
+
+  $scope.$watchCollection(goog.bind(function() {
+    return this.scales;
+  }, this), goog.bind(function(newNames) {
+    this.zoomLevels = Object.keys(this.scales).map(Number);
+    this.zoomLevels.sort(ol.array.numberSafeCompareFunction);
+  }, this));
 
   var mapExpr = $attrs['ngeoScaleselectorMap'];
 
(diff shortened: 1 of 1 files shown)


rmichaelis:
Any news about this ?

src/directives/scaleselector.js
```diff
@@ -111,14 +111,18 @@ ngeo.ScaleselectorController = function($scope, $element, $attrs) {
...
-  zoomLevels.sort(ol.array.numberSafeCompareFunction);
-
   /**
    * @type {Array.<number>}
    * @export
    */
-  this.zoomLevels = zoomLevels;
+  this.zoomLevels;
+
+  $scope.$watchCollection(goog.bind(function() {
```
fgravin: how can `scales` can change ?
Does it change its reference, does it just add or remove item from the array ?

rmichaelis: I add/remove properties from the ``scale`` object

fgravin: Ok, so a slight improvement could be to watch the `Object.keys(this.scales).length` and use a simple `$watch`.
`$watchCollection` watch all properties of your object on all digest loops, can sometimes be avoided.

Could you check you have same result ?

src/directives/scaleselector.js
```diff
@@ -111,14 +111,18 @@ ngeo.ScaleselectorController = function($scope, $element, $attrs) {
...
+    return this.scales;
+  }, this), goog.bind(function(newNames) {
```
fgravin: `fn(){}.bind(this)` is preffered.

src/directives/scaleselector.js
rmichaelis: As long as I don't have the same array length with different keys, your propsal works fine. 
As I don't have such usecase, it looks good to me.
I push the new version.

rmichaelis:
Do you have any comments on the last commit ?

src/directives/scaleselector.js
```diff
@@ -117,12 +117,12 @@ ngeo.ScaleselectorController = function($scope, $element, $attrs) {
    */
   this.zoomLevels;
 
-  $scope.$watchCollection(goog.bind(function() {
-    return this.scales;
-  }, this), goog.bind(function(newNames) {
+  $scope.$watch(goog.bind(function() {
```
fgravin: remove `goog.bind`

rmichaelis:
Should be better now.

src/directives/scaleselector.js
```diff
@@ -243,6 +247,7 @@ ngeo.ScaleselectorController.prototype.handleResolutionChange_ = function(e) {
  */
 ngeo.ScaleselectorController.prototype.handleViewChange_ = function(e) {
   this.registerResolutionChangeListener_();
+  this.handleResolutionChange_(null);
```
fgravin: Why is this needed now and not before ?

rmichaelis: I think that was already needed before, but the problem was not detected.

As soon as the view has been changed, the current scale could be different.
When the view changes, the function ``registerResolutionChangeListener_`` is called and registers a listener on the  ``change:resolution`` event but the event is not fired and ``this.currentScale`` is not initialized with the new current scale belonging to the new view.

fgravin: Then is `registerResolutionChangeListener_` still usefull ? 
I don't really understand what it does, seems like it register/unregister on all view changes ..

rmichaelis: Yes it's still useful.
The ``change:view`` event is only fired when a new view is associated to the map. 
The ``change:view`` event is not fired when something into the view changes (i.e. center, resolution, ...)

Then when a new view is associated to the map, first we have to unregister the event ``change:resolution`` associated  to the old view object and then register a new listener on event ``change:resolution``  to the new view. 

Is it now more understandable ?

fgravin: Yes ok thanks

Looks good then

fgravin:
Thanks for the work @rmichaelis

PR #886, opened 2016-03-15
Use 'GmfThemesNode' instead of 'Object' in gmf themesservice

ger-benjamin:
Needed by https://github.com/camptocamp/ngeo/issues/779


fredj:
LGTM

PR #891, opened 2016-03-16
adding ng-if on span.layerIcon to avoid useless DOM node creation

### LINKED ISSUE #887: Add layer icon element only when relevant
Currently a span.layerIcon is added for every single node in the layer tree.
AFAIK, groups (nodes with children) never use a layer icon. So instead of adding an element which is then not displayed, we'd better change the partial so that the element is not added for groups.

@sbrunner Can you please confirm that groups never have icons?

olivierSemet:
closes #887


pgiraud:
Please merge once Travis is OK.

olivierSemet:
I do not have the rights to merge the PR

PR #893, opened 2016-03-16
Clean the CSS (add --clean-css=--s0 lessc argument)

sbrunner:


fredj:
LGTM

gberaudo:
What about generating a sourcemap with `--source-map` or `--source-map-inline-sources` option?
https://github.com/jakubpawlowicz/clean-css

sbrunner:
+1 for the second (for me the first one will probably not working)

gberaudo:
Thanks, +1.

sbrunner:
Merging without source-map (I don't succeed to make it working :-( )

gberaudo:
Thanks @sbrunner for taking the time to try source maps and for the feedback.

PR #903, opened 2016-03-17
Use translate filter in layertree partial

fredj:


sbrunner:
Shouldn't be on 2.0 ?

fredj:
> Shouldn't be on 2.0 ?

see #906

sbrunner:
Thanks :-)

PR #906, opened 2016-03-17
Use translate filter in layertree partial

fredj:


sbrunner:
Looks good for me :-)

fredj:
thanks

PR #911, opened 2016-03-18
Update bg layer selector directive name

pgiraud:


sbrunner:
Looks good for me :-)

pgiraud:
It looks like this pull request breaks the desktop application. I now get a "Cannot read property 'getLayers' of null" error.

Issue #869, opened 2016-03-14
Search - Can customize feature style


ger-benjamin:
Currently, we display searched features with the default overlay style (blue border, blue/transparent fill). In CGXP we can choose the default style and we have a color picker.

Remaining questions for this issue:
- Should we create a colorpicker component ?
- Should we be able to configure a specific style for coordinates search (like a cross instead of the default circle) ?

sbrunner:
Yes two times

fgravin:
> Yes two times

means yes for both ?

ybolognini:
> Should we create a colorpicker component ?

Yes

> Should we be able to configure a specific style for coordinates search (like a cross instead of the default circle) ?

Yes
 
>> Yes two times
> means yes for both ?

Yes

fgravin:
I've created a color picker directive.
#928

fgravin:
Where are the specs ?
Seems the work of @ger-benjamin allow to have different colors for each group of the search, was it wanted ?

So, when the user choose a color from the color picker, does it overload all group styles ?

Can the user manually set a color for the cross if coordinates as well ?

ger-benjamin:
Perhaps that I've went too far. It seems that allowing to have different colors for each group reflet any request (even if I like it).

The original request is to reproduce what are currently existing in GMF 1.6 I've identified:
 - Users can set a default color.
 - Users can choose a color on a colorpicker (on the fly, at any moment)
 - Users can set the style for the default cross on coordinates search.

Perhaps I've missed something...

fgravin:
> Users can set a default color
> Users can set the style for the default cross on coordinates search

By user you mean the platform administrator (with custom params) ?

fgravin:
Could we have more info about the spec @sbrunner @ybolognini ?

Shall we keep the possibility to have one style per search group ???

ger-benjamin:
- Admin can set a default color.
 - Any users can choose a color on a colorpicker (on the fly, at any moment)
 - Admin can set the style for the default cross on coordinates search.

sbrunner:
> Shall we keep the possibility to have one style per search group ???
no

+1 @ger-benjamin

pgiraud:
Design specs to be discussed in WIKI page https://github.com/camptocamp/c2cgeoportal/wiki/Specs-%23869---Color-of-search-result

ybolognini:
Specifications are now ready for validation!

[...]

PR #922, opened 2016-03-21
Can set style of search overlay

ger-benjamin:
Fix: https://github.com/camptocamp/ngeo/issues/869 (any color-picker in this PR)

Examples (desktop and mobile currently errored on github, wip): 
 - https://ger-benjamin.github.io/ngeo/style_search_overlay/examples/contribs/gmf/apps/mobile
 - https://ger-benjamin.github.io/ngeo/style_search_overlay/examples/contribs/gmf/apps/desktop
 - https://ger-benjamin.github.io/ngeo/style_search_overlay/examples/contribs/gmf/search.html

Try to search: "lausanne" or "600000 200000"

Ready for review.

### DIFF
--- a/contribs/gmf/src/directives/search.js
+++ b/contribs/gmf/src/directives/search.js
@@ -17,6 +17,11 @@ goog.require('ol.Feature');
 goog.require('ol.Map');
 goog.require('ol.geom.Point');
 goog.require('ol.proj');
+goog.require('ol.style.Circle');
+goog.require('ol.style.Fill');
+goog.require('ol.style.RegularShape');
+goog.require('ol.style.Stroke');
+goog.require('ol.style.Style');
 
 
 gmf.module.value('gmfSearchTemplateUrl',
@@ -46,6 +51,7 @@ gmf.module.value('gmfSearchTemplateUrl',
  * Example flat results:
  *
  *      <gmf-search gmf-search-map="ctrl.map"
+ *        gmf-search-styles="ctrl.searchStyles"
  *        gmf-search-datasources="ctrl.searchDatasources"
  *        gmf-search-coordinatesprojections="ctrl.searchCoordinatesProjections"
  *        gmf-search-clearbutton="true">
@@ -62,6 +68,7 @@ gmf.module.value('gmfSearchTemplateUrl',
  * Example with categories:
  *
  *      <gmf-search gmf-search-map="ctrl.map"
+ *        gmf-search-styles="ctrl.searchStyles"
  *        gmf-search-datasources="ctrl.searchDatasources"
  *        gmf-search-coordinatesprojections="ctrl.searchCoordinatesProjections"
  *        gmf-search-clearbutton="true">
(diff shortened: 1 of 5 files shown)


contribs/gmf/src/directives/search.js
```diff
@@ -228,6 +234,13 @@ gmf.SearchController = function($scope, $compile, $timeout, gettextCatalog,
    */
   this.featureOverlay_ = ngeoFeatureOverlayMgr.getFeatureOverlay();
 
+  var featureStyle = this.scope_['featuresStyle'];
+  if (featureStyle instanceof ol.style.Style ||
+      featureStyle instanceof Array ||
+      featureStyle instanceof Function) {
+    this.featureOverlay_.setStyle(featureStyle);
```
fgravin: Are we sure we have only features from search in the feature overlay manager when search is active ?
Why not puting the style in the feature ?

fgravin: The style is meant to be changed by the user, this setter doesn't handle this case.

fgravin:
I don't think this gonna cover the need.
I think that the search directive should internally manage a dynamic style (color) for its feature.

Also, it seems that you rely on feature overlay style: we'll have conflicts with other features in the overlay manager

ger-benjamin:
> Also, it seems that you rely on feature overlay style: we'll have conflicts with other features in the overlay manager

I don't think so. overlayManager works with groups. See --> https://github.com/camptocamp/ngeo/blob/master/src/services/featureoverlay.js#L141

> The style is meant to be changed by the user, this setter doesn't handle this case. 
I think that the search directive should internally manage a dynamic style (color) for its feature.

You are right. I'll do something better (I will be quick, I've an idea)

ger-benjamin:
I think that's more flexible now (better in all cases).

(Examples are not up to date but the result is the same).

fgravin:
What are the exact expectations (specs) for this PR ?
The need i hear of was *dynamic change of the search feature color*
I think we also have *be able to set a style for the point if coordinates*

What else ?

contribs/gmf/src/directives/search.js
```diff
@@ -517,6 +533,42 @@ gmf.SearchController.prototype.createSearchCoordinates_ = function(view) {
...
+    image: new ol.style.Circle({radius: 5, fill: fill, stroke: stroke})
+  });
+  var customStyles = this.scope_['featuresStyles'] || {};
+  goog.object.extend(styles, customStyles);
+  var defaultStyle = styles['default'];
+  if (feature === null) {
+    return defaultStyle;
+  }
+  var kind = feature.get('layer_name');
+  return styles[kind] || defaultStyle;
```
fgravin: This doesn't resolve the dynamic part, does it ?
You still load once a style object

contribs/gmf/src/directives/search.js
```diff
@@ -517,6 +533,42 @@ gmf.SearchController.prototype.createSearchCoordinates_ = function(view) {
...
+    stroke: stroke,
+    image: new ol.style.Circle({radius: 5, fill: fill, stroke: stroke})
+  });
+  var customStyles = this.scope_['featuresStyles'] || {};
+  goog.object.extend(styles, customStyles);
+  var defaultStyle = styles['default'];
+  if (feature === null) {
+    return defaultStyle;
+  }
+  var kind = feature.get('layer_name');
```
sbrunner: I think that we should manage when the `layer_name` is not provided ...

fgravin:
Closing.
See https://github.com/camptocamp/ngeo/pull/951 for further work

PR #927, opened 2016-03-22
Add ngeo popover component

olivierSemet:
PR for the new ngeo component popover. The first goal of this component is to display actions capabilities for groups of layers or individual layers into the layertree but there is no limitation for other use at the moment.

**Behavior of the component:**

- Clicking on the anchor tag **SHOULD** toogle the popover
- If multiple anchors exist, only **one** popover can be displayed at a time
- Each anchor is linked to its own popover, meaning it exists as much popovers as anchors
- Clicking into the popover **SHOULD NOT** dismiss it
- Clicking somewhere else **SHOULD** dismiss the popover

Check the example on my GH pages [here](http://oliviersemet.github.io/ngeo/adding-popover-component/examples/popover.html)

### DIFF
--- a/src/directives/popover.js
+++ b/src/directives/popover.js
@@ -1,0 +1,147 @@
+goog.provide('ngeo.popoverDirective');
+goog.provide('ngeo.popoverAnchorDirective');
+goog.provide('ngeo.popoverContentDirective');
+goog.provide('ngeo.PopoverController');
+
+goog.require('ngeo');
+
+/**
+ * Provides a directive used to display a Bootstrap popover.
+ *
+ *<div ngeo-popover>
+ *  <a ngeo-popover-anchor class="btn btn-info">anchor 1</a>
+ *  <div ngeo-popover-content="">
+ *    <ul>
+ *      <li>action 1:
+ *        <input type="range"/>
+ *      </li>
+ *    </ul>
+ *  </div>
+ *</div>
+ * @ngdoc directive
+ * @ngInject
+ * @ngname ngeoPopover
+ * @return {angular.Directive} The Directive Definition Object.
+ */
+ngeo.popoverDirective = function() {
+  return {
+    restrict: 'A',
+    scope : true,
+    controller: 'NgeoPopoverController',
+    link : function(scope, elem, attrs, ngeoPopoverCtrl) {
+
+      ngeoPopoverCtrl.anchorElm.on('hidden.bs.popover', function() {
+        /**
+         * @type {{inState : Object}}
+         */
+        var popover = ngeoPopoverCtrl.anchorElm.data('bs.popover');
+        popover['inState'].click = false;
+      });
+
+      ngeoPopoverCtrl.anchorElm.on('inserted.bs.popover', function() {
+        ngeoPopoverCtrl.shown = true;
+        ngeoPopoverCtrl.bodyElm.parent().on('click', function(e) {
+          e.stopPropagation();
+        })
+      });
+
+      ngeoPopoverCtrl.anchorElm.popover({
+        container: 'body',
+        html: true,
+        content: ngeoPopoverCtrl.bodyElm
+      });
+
+      scope.$on('$destroy', function() {
+        ngeoPopoverCtrl.anchorElm.popover('destroy');
+        ngeoPopoverCtrl.anchorElm.unbind('inserted.bs.popover');
+        ngeoPopoverCtrl.anchorElm.unbind('hidden.bs.popover');
+      })
+    }
+  }
+};
+
+/**
+ * @ngdoc directive
+ * @ngInject
+ * @ngname ngeoPopoverAnchor
+ * @return {angular.Directive} The Directive Definition Object
+ */
+ngeo.popoverAnchorDirective = function() {
+  return {
+    restrict: 'A',
+    require: '^^ngeoPopover',
+    link: function(scope, elem, attrs, ngeoPopoverCtrl) {
+      ngeoPopoverCtrl.anchorElm = elem;
+    }
+  }
+};
+
+/**
+ * @ngdoc directive
+ * @ngInject
+ * @ngname ngeoPopoverContent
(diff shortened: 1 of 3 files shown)


src/directives/popover.js
```diff
@@ -0,0 +1,137 @@
...
+ *</div>
+ * @ngdoc directive
+ * @ngInject
+ * @ngname ngeoPopover
+ * @return {angular.Directive} The Directive Definition Object.
+ */
+ngeo.popoverDirective = function() {
+  return {
+    restrict: 'A',
+    scope: {},
```
fgravin: scope should not be isolate, otherwise we can't manage inner content of the popover from upper controllers.

src/directives/popover.js
```diff
@@ -0,0 +1,137 @@
...
+    controller: 'NgeoPopoverController',
+    link: function(scope, elem, attrs, controller) {
+      function onClick(clickEvent) {
```
fgravin: put the logic in the controller for more consistency (remove the `link` function).

src/directives/popover.js
```diff
@@ -0,0 +1,137 @@
...
+/**
+ * @ngdoc directive
+ * @ngInject
+ * @ngname ngeoPopoverAnchor
+ * @return {angular.Directive} The Directive Definition Object
+ */
+ngeo.popoverAnchorDirective = function() {
+  return {
+    restrict: 'A',
+    require: ['^^ngeoPopover'],
```
fgravin: `^` is better i think.

fgravin: Well no :)

src/directives/popover.js
```diff
@@ -0,0 +1,137 @@
...
+    scope: {},
+    link: function(scope, elem, attrs, controllers) {
+      var ngeoPopoverCtrl = controllers[0];
+      ngeoPopoverCtrl.anchorElm = elem;
+      elem.on('inserted.bs.popover', function(event) {
+        ngeoPopoverCtrl.shown = true;
+        angular.element('.popover').on('click', function(e) {
```
fgravin: You are registering an event on all `.popover` elements, not only on yours.

src/directives/popover.js
```diff
@@ -0,0 +1,137 @@
...
+          e.stopPropagation();
+        })
+      });
+
+      elem.on('hidden.bs.popover', function(event) {
+        /**
+         * @type {{inState : Object}}
```
fgravin: Good

fgravin:
Please try in your example to generate inner content of your popover from mainController model.
Fix scope inheritance from issues you get.

src/directives/popover.js
```diff
@@ -0,0 +1,137 @@
...
+ */
+ngeo.popoverContentDirective = function() {
+  return {
+    restrict: 'A',
+    transclude: true,
+    require: ['^^ngeoPopover'],
+    scope: {},
+    link: function(scope, elem, attrs, controllers, transclude) {
+      var ngeoPopoverCtrl = controllers[0];
+      ngeoPopoverCtrl.bodyElm = transclude();
```
fgravin: Can you try without transclude and just set `ngeoPopoverCtrl.bodyElm` from element first child.
Translude is actually used for something specfic that is not your usecase.
I wonder if the transclude function doesn't ensure the content is compiled though.
To test.

fgravin:
The layertree will change a lot so the popover will be generated and destroyed, be sure you clean everything well when they are destroyed.

src/directives/popover.js
fgravin: is this event unregistered on `scope.$destroy` ?

fgravin:
Good work, I'v made some comments though.

In the end, i think that one directive should have been enough for the whole thing, your approach is good but might introduce useless complexity.

olivierSemet:
@fgravin following your advices, I pushed a new version.

fgravin:
Looks good thanks @olivierSemet.
I'll wait the PSC validates the spec to merge it.

src/directives/popover.js
```diff
@@ -0,0 +1,148 @@
...
+        })
+      });
+
+      ngeoPopoverCtrl.anchorElm.popover({
+        container: 'body',
+        html: true,
+        content: ngeoPopoverCtrl.bodyElm
+      });
+
+      scope.$on('$destroy', function() {
```
fgravin: Need to check it's needed (wait to integrate it in desktop app for that).
The element will be removed anyway, i hope jquery unregister all events when removing an element.

src/directives/popover.js
```diff
@@ -0,0 +1,147 @@
...
+ */
+ngeo.popoverDirective = function() {
+  return {
+    restrict: 'A',
+    scope : true,
+    controller: 'NgeoPopoverController',
+    link : function(scope, elem, attrs, controller) {
+      var ngeoPopoverCtrl = controller;
+
+      ngeoPopoverCtrl.anchorElm .on('hidden.bs.popover', function() {
```
tsauerwein: minor: space in front of ` .on`

src/directives/popover.js
```diff
@@ -0,0 +1,147 @@
...
+
+      ngeoPopoverCtrl.anchorElm.popover({
+        container: 'body',
+        html: true,
+        content: ngeoPopoverCtrl.bodyElm
+      });
+
+      scope.$on('$destroy', function() {
+        ngeoPopoverCtrl.anchorElm .unbind('inserted.bs.popover');
+        ngeoPopoverCtrl.anchorElm .unbind('hidden.bs.popover');
```
tsauerwein: same here: space between `anchorElm` and ` .unbind`

I am surprised that the linter does not catch this.

olivierSemet: Thanks, it should fix the problem. I think the error was triggered by the option _--check_ of _git diff_ regarding the man _git-check_

fgravin:
Good work.
We know that there is a potential limitation when `require` is not correctly handled cause of the `transclude` linked function usage.
But merging cause we didn't find good solution for now.

PR #933, opened 2016-03-24
Add new ngeo-zoom-delta atribute to ngeo.recenterDirective

fredj:
To be able to create custom zoom buttons


fgravin:
LGTM
Why the default ol zoom button are not satisfying ?

sbrunner:
Difficult to style exactly as we want when we don't control the dom, potential i18n issue also ...

PR #934, opened 2016-03-24
[wip] DIY zoom buttons

fredj:
work in progress

### DIFF
--- a/contribs/gmf/apps/desktop/index.html
+++ b/contribs/gmf/apps/desktop/index.html
@@ -83,6 +83,14 @@
         </div>
       </div>
       <div class="map-container">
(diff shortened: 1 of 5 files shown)


tsauerwein:
Are we going to re-implement all ol3 controls because they don't support i18n?

sbrunner:
When it's writing 10 lines to escape style and i18n issues yes without hesitation

contribs/gmf/apps/desktop/index.html
```diff
@@ -83,6 +83,14 @@
         </div>
       </div>
       <div class="map-container">
+        <div class="zoom-buttons" ngeo-recenter ngeo-recenter-map="::mainCtrl.map">
```
sbrunner: Shouldn't we named it as map-buttons, because on mobile we as have the geolocation button in the div...

PR #938, opened 2016-03-24
Don't create query sources for WMTS layers without wmsUrl

fredj:


sbrunner:
Thanks, that looks good :-)

fredj:
Thanks for the review

PR #955, opened 2016-03-29
Add gmf feature style directive

adube:
This PR introduces a `gmf-featurestyle` directive and an example featuring it.  It also introduces a `gmf.FeatureHelper` service that is responsible of giving utility methods to create, get and set the style for a feature using its inner properties.  It also contains methods to calculate and return a measure string from a feature geometry.

Wiki: https://github.com/camptocamp/c2cgeoportal/wiki/Spec-%231658-Measures-and-redlining

## Tasks remaining to do

 * [x] Deal with circle measurements, i.e. how should we do it properly.  If we do not support `ol.geom.Circle` as type of geometry, then we need to adjust what has been made.  **Answer**: we'll keep circles as polygons, as before.
 * [x] Travis build pass
 * [x] Code review

## Live example

 * http://adube.github.io/ngeo/edit-feature-style/examples/contribs/gmf/featurestyle.html?map_x=-8003263&map_y=6362007&map_zoom=6

### DIFF
--- a/contribs/gmf/src/directives/featurestyle.js
+++ b/contribs/gmf/src/directives/featurestyle.js
@@ -1,0 +1,284 @@
+goog.provide('gmf.FeaturestyleController');
+goog.provide('gmf.featurestyleDirective');
+
+goog.require('gmf');
+goog.require('ngeo.FeatureHelper');
+/** @suppress {extraRequire} */
+goog.require('ngeo.colorpickerDirective');
+
+
+/**
+ * Directive used to set the style of a vector feature. The options depend
+ * on the type of geometry.
+ * Example:
+ *
+ *     <gmf-featurestyle
+ *         gmf-featurestyle-feature="ctrl.selectedFeature">
+ *     </gmf-featurestyle>
+ *
+ * @htmlAttribute {ol.Feature} gmf-featurestyle-feature The feature.
+ * @return {angular.Directive} The directive specs.
+ * @ngInject
+ * @ngdoc directive
+ * @ngname gmfFeaturestyle
+ */
+gmf.featurestyleDirective = function() {
+  return {
+    controller: 'GmfFeaturestyleController',
+    scope: {
+      'feature': '=gmfFeaturestyleFeature'
+    },
+    bindToController: true,
+    controllerAs: 'fsCtrl',
+    templateUrl: gmf.baseTemplateUrl + '/featurestyle.html'
+  };
+};
+
+gmf.module.directive('gmfFeaturestyle', gmf.featurestyleDirective);
+
+
+/**
+ * @param {!angular.Scope} $scope Angular scope.
+ * @param {ngeo.FeatureHelper} ngeoFeatureHelper Gmf feature helper service.
+ * @constructor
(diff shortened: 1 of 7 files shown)


adube:
@fgravin Ready for review.

pgiraud:
@adube Let's forget about  the "icon" that was supposed to be displayed ahead of measure. I don't this is required. Can you please remove the bordered text in you example?

contribs/gmf/src/directives/featurestyle.js
```diff
@@ -0,0 +1,322 @@
...
+  if (previousFeature) {
+    this.type = null;
+    this.color = null;
+    this.measure = null;
+  }
+
+  if (newFeature) {
+    keys.push(
+        ol.events.listen(
+            newFeature,
```
fgravin: Code could be a bit factored.

```js

[gmf.FeatureProperties.ANGLE, gmf.FeatureProperties.COLOR, ...].forEach(function(propName) {
keys.push(
         ol.events.listen(
             newFeature,
             ol.Object.getChangeEventType(propName),
             this.handleFeatureChange_,
             this
         )
     );
}.bind(this)
```

fgravin:
Looks good in the approach.
Maybe worth it to improve the css (slider etc..), it's gonna be usefull for desktop integration.

Have you watched the `FeatureHash` format ?
https://github.com/camptocamp/ngeo/blob/master/src/ol-ext/format/featurehash.js
It does parse and serialize styles from/to properties, you need to have a look and be sure everything is coherant and no code is duplicated. Thanks.

adube:
* removal of "icon": OK
 * improve css: OK (for slider, for the rest it's already fine, I think)
 * `FeatureHash`: I'll need to discuss this more with you, @fgravin.

adube:
@fgravin I made all the requested/discussed modifications, with the exception of the modifications to the `FeatureHash`.  It will be easier to make those fixes when we implement the Permalink with it, in an other PR.

Please merge if everything looks okay (once Travis passes).

adube:
(You may need to clean your browser cache to see the changes in the live example).

src/ngeo.js
```diff
@@ -13,3 +13,33 @@ ngeo.module = angular.module('ngeo', []);
  * @type {string}
  */
 ngeo.baseTemplateUrl = 'ngeo';
+
+
+/**
+ * @enum {string}
+ */
+ngeo.FeatureProperties = {
```
fgravin: shouldn' start with upper case, it's not a class

adube: I'll fix this.  Sorry, I'm used to OL3 doing it this way, see: https://github.com/openlayers/ol3/blob/master/src/ol/layer/layerbase.js#L12

fgravin: Well maybe i'm wrong ..

fgravin: keep it like this then.

adube:
Thanks for the review.

src/ngeo.js
tsauerwein: For enums upper-case is fine. :+1:

PR #959, opened 2016-03-30
Get rid of goog.array.* functions

fredj:
Use the ol3 functions instead

### DIFF
--- a/src/services/layerHelper.js
+++ b/src/services/layerHelper.js
@@ -2,6 +2,7 @@ goog.provide('ngeo.LayerHelper');
 
 goog.require('ngeo');
 goog.require('ol.Collection');
+goog.require('ol.array');
 goog.require('ol.format.WMTSCapabilities');
 goog.require('ol.layer.Group');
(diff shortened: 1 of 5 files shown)


src/services/layerHelper.js
```diff
@@ -92,7 +93,7 @@ ngeo.LayerHelper.prototype.createWMTSLayerFromCapabilitites = function(
 
       // Add styles from capabilities as param of the layer
       var layers = result['Contents']['Layer'];
-      var l = goog.array.find(layers, function(elt, index, array) {
+      var l = ol.array.find(layers, function(elt, index, array) {
```
fgravin: `find` is in `Array.prototype` also.

fredj: yes but unfortunately not available for IE (es6) :
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/find

fgravin:
LGTM

fredj:
Thanks for the review

Issue #967, opened 2016-03-30
Alignment of subgroups in group


rbovard:
There is an alignment issue with subgroups in group (we cannot see that a group is inside another one).

fredj:
fixed with #981

Issue #963, opened 2016-03-30
WMTS layers is considere as WMS layers in the querry...


sbrunner:
On http://geomapfish-workshop.camptocamp.net/apm/mobile/theme/OSM?debug
we get an error on https://github.com/camptocamp/ngeo/blob/2.0/src/services/query.js#L210

for WMTS layers...

fredj:
#938 should be backported to the 2.0 branch

Issue #961, opened 2016-03-30
FTS does not work with coordinates


gnerred:
When a coordinate pair is entered, a network request is send, instead of recentering the map

sbrunner:
This will be a feature of the 2.1...

PR #971, opened 2016-03-31
Merge master into 2.0

fredj:
Should fix:
 - #961
 - #963
 - #967


sbrunner:
We don't want to merge fill master in 2.0!

fgravin:
Yep, cannot we wherry-pick the needed commits ?
Otherwise, when we do bug fixing, we should do them into 2.0 and merge 2.0 into master.

PR #975, opened 2016-04-01
Add ngeo draw feature directive

adube:
This PR introduces a `ngeo-drawfeature` directive and an example featuring it.

## Todo

 * [x] deal with i18n values
 * [x] use contants for feature properties and geometry types once #955 is merged
 * [x] set the default style of the features when drawn once #955 is merged
 * [x] prevent the features and overlay that are added in the measure interactions to be shown, as those conflict with the actual drawn features.
 * [x] code review
 * [x] merge into one commit
 * [x] travis build passes

## Live example

 * http://adube.github.io/ngeo/edit-feature-draw/examples/drawfeature.html

### DIFF
--- a/contribs/gmf/src/directives/partials/search.html
+++ b/contribs/gmf/src/directives/partials/search.html
@@ -8,4 +8,5 @@
 <span class="clear-button ng-hide"
         ng-hide="!ctrl.clearButton || ctrl.input_value == ''"
         ng-click="ctrl.onClearButton()"></span>
+  <input data-ng-model="color" data-ng-change="ctrl.setStyleColor(color)"></input>
 </div>
--- a/src/directives/drawfeature.js
+++ b/src/directives/drawfeature.js
@@ -1,0 +1,299 @@
+goog.provide('ngeo.DrawfeatureController');
+goog.provide('ngeo.drawfeatureDirective');
+
+goog.require('ngeo');
+goog.require('ngeo.DecorateInteraction');
+goog.require('ngeo.FeatureHelper');
+/** @suppress {extraRequire} */
+goog.require('ngeo.btngroupDirective');
+/** @suppress {extraRequire} */
+goog.require('ngeo.drawpointDirective');
+/** @suppress {extraRequire} */
+goog.require('ngeo.drawrectangleDirective');
+/** @suppress {extraRequire} */
+goog.require('ngeo.drawtextDirective');
+/** @suppress {extraRequire} */
+goog.require('ngeo.measureareaDirective');
+/** @suppress {extraRequire} */
+goog.require('ngeo.measureazimutDirective');
+/** @suppress {extraRequire} */
+goog.require('ngeo.measurelengthDirective');
+goog.require('ol.Feature');
+
+/**
+ * Directive used to draw vector features on a map.
+ * Example:
+ *
+ *     <ngeo-drawfeature
+ *         ngeo-btn-group
+ *         class="btn-group"
+ *         ngeo-drawfeature-active="ctrl.drawActive"
+ *         ngeo-drawfeature-map="::ctrl.map">
+ *       <a
+ *         href
+ *         translate
+ *         ngeo-btn
+ *         ngeo-drawpoint
+ *         class="btn btn-default ngeo-drawfeature-point"
+ *         ng-class="{active: dfCtrl.drawPoint.active}"
+ *         ng-model="dfCtrl.drawPoint.active"></a>
+ *       <a
+ *         href
+ *         translate
+ *         ngeo-btn
+ *         ngeo-measurelength
+ *         class="btn btn-default ngeo-drawfeature-linestring"
+ *         ng-class="{active: dfCtrl.measureLength.active}"
+ *         ng-model="dfCtrl.measureLength.active"></a>
+ *       <a
+ *         href
+ *         translate
+ *         ngeo-btn
+ *         ngeo-measurearea
+ *         class="btn btn-default ngeo-drawfeature-polygon"
+ *         ng-class="{active: dfCtrl.measureArea.active}"
+ *         ng-model="dfCtrl.measureArea.active"></a>
+ *       <a
+ *         href
+ *         translate
+ *         ngeo-btn
+ *         ngeo-measureazimut
+ *         class="btn btn-default ngeo-drawfeature-circle"
+ *         ng-class="{active: dfCtrl.measureAzimut.active}"
+ *         ng-model="dfCtrl.measureAzimut.active"></a>
+ *       <a
+ *         href
+ *         translate
+ *         ngeo-btn
+ *         ngeo-drawrectangle
+ *         class="btn btn-default ngeo-drawfeature-rectangle"
+ *         ng-class="{active: dfCtrl.drawRectangle.active}"
+ *         ng-model="dfCtrl.drawRectangle.active"></a>
+ *       <a
+ *         href
+ *         translate
+ *         ngeo-btn
+ *         ngeo-drawtext
+ *         class="btn btn-default ngeo-drawfeature-text"
+ *         ng-class="{active: dfCtrl.drawText.active}"
+ *         ng-model="dfCtrl.drawText.active"></a>
+ *     </ngeo-drawfeature>
+ *
+ * @htmlAttribute {boolean} ngeo-drawfeature-active Whether the directive is
+ *     active or not.
+ * @htmlAttribute {ol.Map} ngeo-drawfeature-map The map.
+ * @return {angular.Directive} The directive specs.
+ * @ngInject
+ * @ngdoc directive
+ * @ngname ngeoDrawfeature
+ */
+ngeo.drawfeatureDirective = function() {
+  return {
+    controller: 'ngeoDrawfeatureController',
+    scope: true,
+    bindToController: {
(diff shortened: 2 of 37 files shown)


adube:
@fgravin The directive currently contains lots of copy/pasted code from many places: Luxembourg, ngeo examples, etc.  The i18n text contained within the directive are a result of that.  Would you please tell me what strategy I should use for the i18n ?

If you want to review what's in progress, go ahead.  Keep in mind what's remaining to do (see in the description).

fgravin:
Looks good for what i've seen so far.

> Would you please tell me what strategy I should use for the i18n ?

Ok i will think about it.

fgravin:
> The i18n text contained within the directive are a result of that. Would you please tell me what strategy I should use for the i18n

See https://github.com/Geoportail-Luxembourg/geoportailv3/blob/30272dc5426b6c2d60d7a9aa8d5d38843ca93360/geoportailv3/static/js/measure/measuredirective.js#L138

You need to pass some html to the constructor options.
This html must be compiled (`$compile`) with the directive `translate`, this way the string will be replaced on the fly if the language change.

adube:
Understood.  I'm on it.

adube:
@fgravin This is now ready for review.  The demo is up-to-date too.

contribs/gmf/src/directives/drawfeature.js
```diff
@@ -0,0 +1,399 @@
...
+ * @ngInject
+ * @ngdoc directive
+ * @ngname gmfDrawfeature
+ */
+gmf.drawfeatureDirective = function() {
+  return {
+    controller: 'GmfDrawfeatureController',
+    scope: {
+      'active': '=gmfDrawfeatureActive',
+      'map': '=gmfDrawfeatureMap'
```
fgravin: use `>` instead (one way binding)

contribs/gmf/src/directives/drawfeature.js
```diff
@@ -0,0 +1,399 @@
...
+   */
+  this.drawText = new ol.interaction.Draw({
+    type: ol.geom.GeometryType.POINT
+  });
+
+  var drawText = this.drawText;
+  this.interactions_.push(drawText);
+  drawText.setActive(false);
+  ngeoDecorateInteraction(drawText);
+  this.map.addInteraction(drawText);
```
fgravin:
```js
  interaction.setActive(false);
  ngeoDecorateInteraction(interaction);
  this.map.addInteraction(interaction);
```

maybe this could be factorized in a `this.interactions_.forEach()` ?
It's a bit heavy.

contribs/gmf/src/directives/drawfeature.js
```diff
@@ -0,0 +1,399 @@
...
+      name = this.gettextCatalog_.getString('LineString');
+      break;
+    case ngeo.GeometryType.POLYGON:
+      name = this.gettextCatalog_.getString('Polygon');
+      break;
+    case ngeo.GeometryType.RECTANGLE:
+      name = this.gettextCatalog_.getString('Rectangle');
+      feature.set(prop.IS_RECTANGLE, true);
+      break;
+    default:
```
fgravin:
```js
default:
  name = this.gettextCatalog_.getString(type)
```
and removing point, linestring, polygon cases ??

contribs/gmf/src/directives/drawfeature.js
```diff
@@ -0,0 +1,399 @@
...
+};
+
+
+/**
+ * @param {ol.ObjectEvent} event Event.
+ * @private
+ */
+gmf.DrawfeatureController.prototype.handleActiveChange_ = function(event) {
+  var active;
+  this.interactions_.some(function(interaction) {
```
fgravin: use the return value of the `some` function instead of a new `var active`.
Something like
```js
this.active = this.interactions_.some(function(interaction) {
  return interaction.getActive();
});
```

fgravin:
Some minor comments, the overall looks good.
But i wonder if this shouldn't go into ngeo in a more customisable fashion.

contribs/gmf/src/directives/drawfeature.js
adube: I don't understand this comment.

adube: Ah, okay I get it.  The constants currently have all lower case values.  I suppose it wouldn't hurt to make them camel case.  For `LineString`, that means that the constant should be renamed `LINE_STRING`, I suppose.

fgravin: name = this.gettextCatalog_.getString('LineString');

This line could go in default ? As you are doing it for all

adube:
Thanks for the review.  I'll address the changes required.

Please, let me know if this should go into ngeo instead.

fgravin:
What i have thought of in the beginning was a directive draw like this, with the controller logic.
Then, one small directive per drawing type, requiring the draw directive, accessing its controller.

Each small directive register itself in the main controller, and adds its specific handlers.
Common handlers and whole logic would be in main draw directive, no template for this one.
It would go into ngeo.

What do you think ?

adube:
@fgravin and I discussed about this.  It makes sense to move the draw directive in ngeo.  I'm on it.

adube:
@fgravin Travis fails.  I don't understand why.

sbrunner:
You see that the error is in the animation example ... https://travis-ci.org/camptocamp/ngeo/jobs/120962154#L546,, I also don't understand why it failed ...

adube:
@fgravin Now, each draw/measure directive have its own file.

I still have the build issue when making `make check-examples` (see Travis build error).  I still don't know what's going on wrong.  That prevents me from updating the live example.  If you have a minute to check what's wrong I'd appreciate it.

fgravin:
Do a `make hosted-example`and test your example in compiled mode (localhost hosted-examples ..).
You'll see an error, with a link, open the link if not to long, it can give help.

Otherwise, try to remove your subdirective from the template, see if you still have the error. Try to debug step by step by removing some code to see what's going on.
You don't miss any `@ngInject` ?

adube:
@fgravin Thanks for the tip.  All other examples but mine were throwing an error.  That was because of the addition of `gettext` as dependency in ngeo.  Now, all examples properly include the .js file needed.

With this, we're good to go.

adube:
Live example updated.

src/directives/drawfeature.js
```diff
@@ -0,0 +1,298 @@
...
+   * @type {Array.<ol.interaction.Interaction>}
+   * @private
+   */
+  this.interactions_ = [];
+
+  /**
+   * @type {ol.interaction.Draw}
+   * @export
+   */
+  this.drawPoint;
```
fgravin: I think we should remove it from here.
Cause the main directive doesn't know what interaction it can have.
Do you agree ?

adube: Mmm... I liked having those listed here.  Those 6 draw/measure directives can only be used within this main drawFeature directive, and those properties identify them correctly.  If one looks at this file only, one will understand the different interactions that can be bound from here.

I'm not against removing them.  I just find it useful.  Also, I haven't tested but this makes the properties correctly exported by closure, I think.  Otherwise, we would probably need to do `drawFeatureCtrl['drawPoint'] = drawPoint`.

Bottom line: I'm in favour of keeping them.  What do you think ?

fgravin:
Can you rebase please ?

adube:
Done.

src/directives/measureazimut.js
```diff
@@ -0,0 +1,76 @@
...
+    require: '^^ngeoDrawfeature',
+    /**
+     * @param {!angular.Scope} $scope Scope.
+     * @param {angular.JQLite} element Element.
+     * @param {angular.Attributes} attrs Attributes.
+     * @param {ngeo.DrawfeatureController} drawFeatureCtrl Controller.
+     */
+    link: function($scope, element, attrs, drawFeatureCtrl) {
+
+      var helpMsg = gettext('Click to start drawing azimut');
```
fgravin: I don't really get it, i know it comes from Lux.
What is `helpMsg` value, the translation of the key ?

Does `translate` directive below is usefull ? I think something is redundant here.

adube: You tell me.

contribs/gmf/src/directives/partials/search.html
```diff
@@ -8,4 +8,5 @@
 <span class="clear-button ng-hide"
         ng-hide="!ctrl.clearButton || ctrl.input_value == ''"
         ng-click="ctrl.onClearButton()"></span>
+  <input data-ng-model="color" data-ng-change="ctrl.setStyleColor(color)"></input>
```
ger-benjamin: @adube Can you explain me what is that ? That creates a field below the search field that seems not used.
Is that an error ? What's the aim ?

pgiraud: Maybe related to this https://github.com/camptocamp/c2cgeoportal/wiki/Specs-%23869---Color-of-search-result

adube: This was part of a patch that @fgravin did for me.  I hadn't noticed it, but it shouldn't have been added.  You can remove this.

ger-benjamin: Thanks for these information, I will remove that temporarily.

PR #980, opened 2016-04-04
Don't create query sources for WMTS layers without wmsUrl

### LINKED ISSUE #963: WMTS layers is considere as WMS layers in the querry...

fredj:
fixes #963 

Cherry-pick #938 into the 2.0 branch


sbrunner:
Looks good, merge when Travis is green

Issue #654, opened 2016-02-16
Set default theme in mobile / desktop apps


fredj:
Spec: https://github.com/camptocamp/c2cgeoportal/wiki/Spec-%23654-Functionalities-in-the-ngeo-application

Ideally, the default theme should come from the `themes` service

fredj:
@tsauerwein I don't think #655 fixes this issue: the default theme is still the (inexistant) `main`

tsauerwein:
Sorry, I referenced the wrong issue.

sbrunner:
With the GeoMapFish this kind of things come from the functionalities, than I don't think that the better way to get it from the theme web service.

For me it's better to put it in a separate web service or in the authentication web service.

What do you think?

sbrunner:
Concerned configuration:
* `default_theme` the default theme.
* `default_basemap` the default background layer.
* `authorized_plugins` the list of restricted plugins that the user can access.
* `queryer_layers` the layers available in the query builder.

fredj:
https://github.com/camptocamp/c2cgeoportal/pull/2456

PR #984, opened 2016-04-05
Get the default basemap from the user functionalities

### LINKED ISSUE #964: Set default background layer
Currently we cannot set the default background layer on mobile.

fredj:
Fixes #964
See #654

### DIFF
--- a/contribs/gmf/apps/desktop/index.html
+++ b/contribs/gmf/apps/desktop/index.html
@@ -12,12 +12,6 @@
       </div>
     </header>
     <main>
-      <!-- using the mobile background layer selector (hidden) temporarily
-           here to get a background-layer -->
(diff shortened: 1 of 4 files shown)


sbrunner:
Looks good but shouldn't be documented in the directive?

sbrunner:
Can it be on the 2.0 branch?

fredj:
I'm not sure that `gmf.MobileBackgroundlayerselectorController` is the right place to check the default background.

What about choosing the value (in permalink and gmf.User) in the abstract controller and passing the value to the background selector?

fredj:
Ready to be reviewed

sbrunner:
Looks good for me :-)

fredj:
Thanks for the review

PR #986, opened 2016-04-05
Add the modify circle interaction.

jlap:
This commit adds the modify circle interaction, along with an example of how it works.

[The interaction was taken from the luxembourg app](https://github.com/Geoportail-Luxembourg/geoportailv3/blob/master/geoportailv3/static/js/draw/modifycircleinteraction.js).

A live example is available here: http://adube.github.io/ngeo/modify-circle/examples/modifycircle.html

### DIFF
--- a/externs/ngeox.js
+++ b/externs/ngeox.js
@@ -258,6 +258,43 @@ ngeox.format.FeatureHashOptions.prototype.properties;
  */
 ngeox.interaction;
 
+
+/**
+ * @typedef {{
+ *     pixelTolerance: (number|undefined),
+ *     style: (ol.style.Style|Array.<ol.style.Style>|ol.style.StyleFunction|undefined),
+ *     features: ol.Collection.<ol.Feature>,
+ *     wrapX: (boolean|undefined)}}
+ * @api
+ */
+ngeox.interaction.ModifyOptions;
+
+
+/**
+ * Pixel tolerance for considering the pointer close enough to a segment or
+ * vertex for editing.
+ * @type {number|undefined}
+ * @api
+ */
+ngeox.interaction.ModifyOptions.prototype.pixelTolerance;
+
+
+/**
+ * Style used for the features being modified.
+ * @type {ol.style.Style|Array.<ol.style.Style>|ol.style.StyleFunction|undefined}
+ * @api
+ */
+ngeox.interaction.ModifyOptions.prototype.style;
+
+
+/**
+ * The features the interaction works on.
+ * @type {ol.Collection.<ol.Feature>}
+ * @api
+ */
+ngeox.interaction.ModifyOptions.prototype.features;
+
+
 /**
  * Interactions for measure tools.
  * @typedef {{
--- a/externs/ngeox.js
+++ b/externs/ngeox.js
@@ -258,43 +258,6 @@ ngeox.format.FeatureHashOptions.prototype.properties;
  */
 ngeox.interaction;
 
-
-/**
- * @typedef {{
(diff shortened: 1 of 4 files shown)


jlap:
@fgravin Could you take a look at my travis error? I don't know how to fix it.

fgravin:
The message is explicit, it doesn't find your constructor in build mode. It is probably because you haven't exported your class, so the name are minified on compiling.

1. Add `@export` in your constructor api doc
2. You need to define a type for your interaction options, see `ngeox.interaction.MeasureOptions` in ngeox.js

jlap:
Thanks @fgravin , the missing export was the problem. I now have a different build error, that only occurs on travis. Building the example locally works for me now (it didn't with the last error), but travis now throws a very cryptic error message, do you have an idea what it means?

Thanks for the help!

fgravin:
> Building the example locally works

I'm perplex :)
Add the gettext.js file in your example.
```html
<script src="../node_modules/angular-gettext/dist/angular-gettext.js"></script>
```
Tell me if it fixes it.

jlap:
That worked thanks!

Here's a live example: http://adube.github.io/ngeo/modify-circle/examples/modifycircle.html

jlap:
@fgravin This PR is now ready for review

sbrunner:
The live example looks good :-)

externs/ngeox.js
```diff
@@ -258,6 +258,43 @@ ngeox.format.FeatureHashOptions.prototype.properties;
...
+
+/**
+ * @typedef {{
+ *     pixelTolerance: (number|undefined),
+ *     style: (ol.style.Style|Array.<ol.style.Style>|ol.style.StyleFunction|undefined),
+ *     features: ol.Collection.<ol.Feature>,
+ *     wrapX: (boolean|undefined)}}
+ * @api
+ */
+ngeox.interaction.ModifyOptions;
```
fgravin: should be `ModifyCircleOptions`, there gonna be other modify interactions.

fgravin:
One comment that you have to change please.
Other wise looks good thanks @jlap

externs/ngeox.js
jlap: These are not specific to the circle modification interation though, it's actually just about the same as the `olx.interaction.ModifyOptions`, and I think all modify interactions which inherit from ol's modify interactions will need these options.

(Actually I don't know why just using `olx.interaction.ModifyOptions` wasn't working, since it's really all I needed)

fgravin: ok so maybe it was enough with `olx.interaction.ModifyOptions` ??
This issue you got was just from the `@export` thing, maybe you should revert this then ?

jlap: Oh right, I'm trying it now, I think you're right. I had done it in another commit after the export one because I still had an error, but that must have been the missing ngettext error.

jlap: It worked!

fgravin: ok so update your PR please.

jlap: It is updated, it passed the travis

fgravin: i may have mixed something, weren't we talking about removing the type `ngeox.interaction.ModifyOptions` to use the type `olx.interaction.ModifyOptions` ?

jlap: Yes we were. We're currently in a comment thread in an outdated commit of the diff, if you go to https://github.com/camptocamp/ngeo/pull/986/files it'll be up-to-date.

I'll squash the commits of the PR into a single one once you say this is ready to merge.

PR #995, opened 2016-04-07
Only create background layers once

### LINKED ISSUE #994: gmf.Themes#getBgLayers must be idempotent
The layers are created every time the function is called.

Example:
```js
[1,2].forEach(function(step) {
  gmfThemes.getBgLayers().then(function(layers) {
    layers.forEach(function(layer) {
      console.log('call', step, goog.getUid(layer));
    });
  });
});
```
output is:
```
call 1 133
call 1 130
call 2 135
call 2 131
```

but must be:
```
call 1 133
call 1 130
call 2 133
call 2 130
```

fredj:
Fixes #994


sbrunner:
Looks good for me :-)

fredj:
Thanks for the review

Issue #965, opened 2016-03-30
HTML code in attributes


rbovard:
The HTML code in attributes is not interpreted on mobile.

Example:
`<a href="https://github.com/" target="_blank">Cliquer ici</a>`

tsauerwein:
@sbrunner Can the values in feature attributes always be [trusted](https://docs.angularjs.org/api/ng/service/$sce#trustAs)? Or can queries be made on user provided layers? In that case we would have to [sanitize](https://docs.angularjs.org/api/ngSanitize/service/$sanitize) the values.

sbrunner:
Good question, should we add a metatata with the list of the trsted (HTML) attributes?
For me it's better.

tsauerwein:
> ... list of the trusted (HTML) attributes?

Would this be per layer or global? Global had the risk that a user-provided layer could have an attribute with the same name. Per layer it would be sufficient to flag the whole layer as *trusted* (I think there would be no need to list each trusted attribute of a layer).

For now I would suggest to go with the *sanitize-everything* approach (see https://github.com/camptocamp/ngeo/pull/996). If we see that this causes a problem, we can still introduce a trusted flag in the metadata.

sbrunner:
Per layer, we don't have any global metadata

> I think there would be no need to list each trusted attribute of a layer

It's exactly what I say :-)

> can still introduce a trusted flag in the metadata.

List of attributes...

tsauerwein:
Fixed with https://github.com/camptocamp/ngeo/pull/996

PR #996, opened 2016-04-07
Support HTML in feature attributes (query popup)

tsauerwein:
This PR proposes to use [angular-sanitize](https://docs.angularjs.org/api/ng/directive/ngBindHtml) to sanitize feature attributes so that HTML code is rendered as HTML.

Closes https://github.com/camptocamp/ngeo/issues/965


fredj:
LGTM

tsauerwein:
@sbrunner Also ok for you?

sbrunner:
OK from now :-)

tsauerwein:
Thanks for the review.

PR #1002, opened 2016-04-08
Fix event unlistening in gmf-featurestyle

adube:
This PR fixes the un-listening of events in the `gmf-featurestyle` directive that are bound to the feature.


adube:
@fgravin Ready for review.

fgravin:
Good catch thanks.

Issue #972, opened 2016-03-31
The default value of isChecked should be false


sbrunner:
https://github.com/camptocamp/ngeo/blob/568e21f275cca0f14edc713d6f7f2279575341db/contribs/gmf/src/directives/layertree.js#L303 and the documentation

tsauerwein:
Fixed with https://github.com/camptocamp/ngeo/issues/1013

PR #1013, opened 2016-04-11
Hide layers by default

tsauerwein:
Only show a layer when `isChecked` is `'true'` in the layer meta-data.

Closes https://github.com/camptocamp/ngeo/issues/972


sbrunner:
Looks good, special thanks for the tests :-)

tsauerwein:
Thanks for the review.

Issue #594, opened 2016-02-04
mobile - automatically show the tree when theme is chosen


pgiraud:
Note: please do not work on this until validated.

- [ ] Create a new `gmf-mobile-nav-back` which requires the `gmf-mobile-nav` directive.
- [ ] Add a `navigateBack` or `goBack` in the `gmf-mobile-nav` directive's controller.

The directive watches the expression (`mainCtrl.theme` in our case) and calls the `back` method when appropriate (ie. when the theme selector panel is active).

tsauerwein:
Fixed with https://github.com/camptocamp/ngeo/pull/655

ger-benjamin:
Re-open with https://github.com/camptocamp/ngeo/pull/888

Because` gmf-mobile-navback` is based on a `watch` function (that watch the theme) and:
>Now the "currentTheme" is the theme in the TreeManager. This theme (the object) "never change". >Only its properties change. So I should use a watchCollection.

PR #1015, opened 2016-04-11
Fix going back after theme selection

tsauerwein:
Listen on `theme.name` instead of `theme`. Because theme is an object that never changes, only its values change.
See https://github.com/camptocamp/ngeo/issues/594#issuecomment-197804863

Closes https://github.com/camptocamp/ngeo/issues/594 again


sbrunner:
Looks good :-)

Issue #1011, opened 2016-04-11
Insert MIXED node layers as layer.Group in Map


fgravin:
The MIXED node layers are inserted in the map as flat layers.
They should be inserted as a layer.Group, making easier to manage the group position in map layers array.

- [x] Add groupLayer to map
- [ ] Manage permalink
- [ ] Add a test

fgravin:
To be changed here:
https://github.com/camptocamp/ngeo/blob/master/contribs/gmf/src/directives/layertree.js#L479

Also impact permalink service
https://github.com/camptocamp/ngeo/blob/master/contribs/gmf/src/services/permalink.js#L578

PR #1018, opened 2016-04-12
Insert MIXED node layers as layer.Group in Map

olivierSemet:
Related to #1011 
PING @fgravin


contribs/gmf/src/directives/layertree.js
```diff
@@ -273,14 +276,14 @@ gmf.LayertreeController.prototype.getLayerCaseMixedGroup_ = function(node) {
   for (i = 0; i < subNodes.length; i++) {
     subNode = subNodes[i];
     // Create all sublayers include wms layers;
-    layer = this.getLayer(subNode, 1, true);
+    layer = this.getLayer(subNode, 1, true, node);
```
fgravin: I wonder if all of this is still usefull.
@ger-benjamin is it still usefull to add the Image layers to the group in one shot here, while we could add them when they come from the `getLayers` function ?

contribs/gmf/src/directives/layertree.js
```diff
@@ -241,7 +242,9 @@ gmf.LayertreeController.prototype.getLayer = function(node, opt_depth,
     var isMerged = type === gmf.Themes.NodeType.NOT_MIXED_GROUP;
     layer.set('isMerged', isMerged);
 
-    this.dataLayerGroup_.getLayers().insertAt(0, layer);
+    if (!parentNode || gmf.Themes.getNodeType(parentNode) !== gmf.Themes.NodeType.MIXED_GROUP) {
```
fgravin: I think here, if the parent node is a mixed, then we should add the layer to parentNode.layer (group) instead of in the map. 
This way, everything is build on the fly, that would remove the logic from `getLayerCaseMixedGroup_`.

contribs/gmf/src/directives/layertree.js
```diff
@@ -202,11 +202,12 @@ gmf.LayertreeController = function($http, $sce, $scope, ngeoCreatePopup,
  * @param {GmfThemesNode} node Layer tree node.
  * @param {number=} opt_depth ngeo layertree node depth.
  * @param {boolean=} opt_createWMS True to allow create wms layer.
+ * @param {GmfThemesNode=} parentNode parent node of the first parameter.
  * @return {ol.layer.Base} The OpenLayers layer or group for the node.
  * @export
  */
 gmf.LayertreeController.prototype.getLayer = function(node, opt_depth,
-        opt_createWMS) {
+        opt_createWMS, parentNode) {
```
fgravin: The parentNode should come from the controller, when the main ngeo directive calls the template.
This should be modified in https://github.com/camptocamp/ngeo/blob/master/src/directives/layertree.js#L211
Do you agree @ger-benjamin

ger-benjamin:
I think that that's not the right way.
Please wait before to continue.

... I write something...

contribs/gmf/src/services/permalink.js
```diff
@@ -579,7 +579,13 @@ gmf.Permalink.prototype.initLayers_ = function() {
   //     to any change happening to the existing layers and any added or
   //     removed ones.
   layers.forEach(function(layer) {
-    this.registerLayer_(layer, true);
+    if (layer instanceof ol.layer.Group) {
+      layer.getLayers().forEach(function(l) {
+        this.registerLayer_(l, true);
```
fgravin: Good, but they are never unregistered
See `handleLayersRemove_`

contribs/gmf/src/services/permalink.js
```diff
@@ -579,7 +579,13 @@ gmf.Permalink.prototype.initLayers_ = function() {
...
+      }, this);
+    } else {
+      this.registerLayer_(layer, true);
+    }
```
adube: Managing this upon initialization is not enough (I think).  Later on, if layers are added/removed to the 'data' layer group or even to the inner sub group layers, the handler methods (that call registerLayer_ and unregisterLayer) won't support adding/removing `ol.layer.Group` instances.   I think the management should occur at that level, i.e. in those methods.  They could be responsible of calling themselves (registerLayer_ should call registerLayer_ and so forth with unregisterLayer_) when dealing with a group.  That way, we would also support infinite number of layer levels.

The ADD and REMOVE events also need to be registered for the collections of layers of those groups (I'm pretty sure it's required to do so).

ger-benjamin:
I think that the getLayer function should only manage groups. So, delet all after https://github.com/camptocamp/ngeo/blob/master/contribs/gmf/src/directives/layertree.js#L274

Then the function "getLayerCaseMixedGroup_" should call another function that create all child layers, set their visibility, params, etc, then put theme in a layergroup add this layer group in the map and return this group as layer. Perhaps all these operation must be done in a new function `addLayerOnMap`.

The function getLayerCaseNotMixedGroup_ should do the same but for not mixed group perhaps the current result of this one can use the function "addLayerOnMap".

In summary, there is no use now to complexifiy call 'manually' (in the gmf layertree) the `gmf.getLayer` function

olivierSemet:
@ger-benjamin @adube @fgravin 
Following your comments, I pushed a new version.

contribs/gmf/src/directives/layertree.js
```diff
@@ -181,14 +179,42 @@ gmf.LayertreeController = function($http, $sce, $scope, ngeoCreatePopup,
...
+/**
  * Create and return a layer corresponding to the ngeo layertree's node.
  * This function will only create a layer for each "top-level" (depth 1) groups.
  *
  * On "not mixed" type nodes, the returned layer will be an ol.layer.Image (WMS)
  * with each name of node's children as LAYERS parameters.
  *
- * On "mixed" type node, the returned  layer will be an ol.layer.Group with
- * a collection of layers that corresponds to each children of the node.
+ * On "mixed" type node, the returned  layer will be an ol.layer.Group
```
fgravin: extra space

contribs/gmf/src/directives/layertree.js
```diff
@@ -200,58 +226,50 @@ gmf.LayertreeController = function($http, $sce, $scope, ngeoCreatePopup,
...
  * @param {GmfThemesNode} node Layer tree node.
+ * @param {ngeo.LayertreeController} parentCtrl parent controller of the node
  * @param {number=} opt_depth ngeo layertree node depth.
  * @param {boolean=} opt_createWMS True to allow create wms layer.
  * @return {ol.layer.Base} The OpenLayers layer or group for the node.
  * @export
  */
-gmf.LayertreeController.prototype.getLayer = function(node, opt_depth,
-        opt_createWMS) {
+gmf.LayertreeController.prototype.getLayer = function(node, parentCtrl, opt_depth, opt_createWMS) {
```
fgravin: is `opt_createWMS` ever used ?

contribs/gmf/src/directives/layertree.js
```diff
@@ -23,8 +22,7 @@ gmf.module.value('gmfLayertreeTemplate',
       var subTemplateUrl = gmf.baseTemplateUrl + '/layertree.html';
       return '<div ngeo-layertree="gmfLayertreeCtrl.tree" ' +
           'ngeo-layertree-map="gmfLayertreeCtrl.map" ' +
-          'ngeo-layertree-nodelayer="gmfLayertreeCtrl.getLayer(node, depth)" ' +
-          'ngeo-layertree-listeners="gmfLayertreeCtrl.listeners(treeScope, ' +
```
fgravin: why ???

contribs/gmf/src/directives/layertree.js
fgravin: is `depth` really optional ?

fgravin:
Some minor comments but it looks more to what i expected. Good work thanks.

However, you did not manage correctly the permaling changes, read @adube comments cause you are not doing things properly.

Also, we must add more tests.

contribs/gmf/src/directives/layertree.js
olivierSemet: Sorry this is a copy/past mistake...

fgravin:
Ok looks good.
I would like @adube  to review and confirm that the permalink thing works.
Thanks

contribs/gmf/src/directives/layertree.js
```diff
@@ -181,14 +179,42 @@ gmf.LayertreeController = function($http, $sce, $scope, ngeoCreatePopup,
...
 /**
+ * LayertreeController.prototype.prepareLayer_ - inject metadata into the layer
+ * @private
+ * @param {GmfThemesNode} node Layer tree node.
+ * @param {ol.layer.Base} layer The OpenLayers layer or group for the node.
+ */
+gmf.LayertreeController.prototype.prepareLayer_ = function(node, layer) {
+  var type = gmf.Themes.getNodeType(node);
+  var ids = this.getNodeIds_(node);
+  if (goog.isDefAndNotNull(layer)) {
```
adube: if the layer can be `null` or `undefined`, it should be defined as such in the param section above, as such:

`@param {?ol.layer.Base=} opt_layer The...`

adube: Also, a `truthy` comparison should be enough: `if (layer) {`

contribs/gmf/src/directives/layertree.js
```diff
@@ -181,14 +179,42 @@ gmf.LayertreeController = function($http, $sce, $scope, ngeoCreatePopup,
...
+    layer.set('querySourceIds', ids);
+    layer.set('layerName', node.name);
+
+    var isMerged = type === gmf.Themes.NodeType.NOT_MIXED_GROUP;
+    layer.set('isMerged', isMerged);
+
+    // If layer is 'unchecked', set it to invisible.
+    var metadata = node.metadata;
+    if (node.children === undefined && goog.isDefAndNotNull(metadata)) {
```
adube: thruthy

contribs/gmf/src/services/permalink.js
```diff
@@ -633,7 +645,13 @@ gmf.Permalink.prototype.handleLayersAdd_ = function(evt) {
...
   var layer = evt.element;
   goog.asserts.assertInstanceof(layer, ol.layer.Base);
-  this.unregisterLayer_(layer);
+  if (layer instanceof ol.layer.Group) {
+    layer.getLayers().forEach(function(l) {
+      this.unregisterLayer_(l);
+    }, this);
+  } else {
+    this.unregisterLayer_(layer);
+  }
```
adube: I'm still not sure if this could work for the same reasons explained in my previous comment about this.  Thoughts ?

adube:
I added some minor comments and a major one.

Would it also be possible to have a github.io live example to be able to test it ?

olivierSemet:
@adube check the example [here](https://oliviersemet.github.io/ngeo/1011-insert-mixed-node-layer-as-layer-group-in-map/examples/contribs/gmf/apps/desktop/)

contribs/gmf/src/services/permalink.js
olivierSemet: Agreed, need more recursion

fgravin:
So ?

adube:
The recursion of addition / removal of layers doesn't seem to have been fixed yet, thus why I hadn't tested yet.

I did for what's currently available.  The setting of properties work, but I can't fully test the example.  When trying to reload the page, we get 404 errors.

The LayerTree example seems to be missing some CSS too: https://oliviersemet.github.io/ngeo/1011-insert-mixed-node-layer-as-layer-group-in-map/examples/contribs/gmf/layertree.html

fgravin:
> The setting of properties work, but I can't fully test the example. When trying to reload the page, we get 404 errors.

Yeah, you cannot reload apps, because the theme go into path and break the path. So you need to remove everything in path after `desktop`

adube:
Something's wrong.  I toggled off the visibility of all layers, created the url without the theme then used it.  I still have one layer visible where it shouldn't be.

Here's the url you can test with: https://oliviersemet.github.io/ngeo/1011-insert-mixed-node-layer-as-layer-group-in-map/examples/contribs/gmf/apps/desktop/theme/OSM?lang=fr&baselayer_ref=map&tree_enable_osm_time=false&tree_enable_osm_scale=false&tree_enable_osm_open=false&tree_group_layers_Layers&tree_group_layers_Group

fgravin:
@olivierSemet is writing tests. We must have tests that cover all of this.

olivierSemet:
@adube the bug you are talking about is known and described at #1029. So this bug already exists in the master branch. We should fix it but IMO this is out of the scope of this PR.

olivierSemet:
@adube, @fgravin I pushed a new version which should fix the group recursion problem. I also provided a new unit test file for the permalink service. It tests the recursion.

contribs/gmf/src/services/permalink.js
```diff
@@ -645,6 +651,12 @@ gmf.Permalink.prototype.handleLayersRemove_ = function(evt) {
  */
 gmf.Permalink.prototype.registerLayer_ = function(layer, opt_init) {
 
+  if (layer instanceof ol.layer.Group) {
+    layer.getLayers().forEach(function(layer) {
+      this.registerLayer_(layer, opt_init);
+    }, this)
+  }
+
```
adube: When a layer is `ol.layer.Group`, the code carries on after looping inside its layers, which means that events are registered for the `ol.layer.Group`.  Is this needed ?

contribs/gmf/src/services/permalink.js
```diff
@@ -692,6 +704,11 @@ gmf.Permalink.prototype.registerLayer_ = function(layer, opt_init) {
  * @private
  */
 gmf.Permalink.prototype.unregisterLayer_ = function(layer) {
+
+  if (layer instanceof ol.layer.Group) {
+    layer.getLayers().forEach(this.unregisterLayer_, this);
+  }
+
```
adube: Same here.

contribs/gmf/test/spec/controllers/gmflayertreecontroller.spec.js
```diff
@@ -0,0 +1,92 @@
...
+      layer : null
+    };
+
+    var layer = gmfLayertreeCtrl.getLayer(nonMixedNode, parentTreeCtrl, 1);
+
+    expect(layer).toBeDefined();
+    expect(layer instanceof ol.layer.Image).toBeTruthy();
+    expect(gmfLayertreeCtrl.prepareLayer_).toHaveBeenCalledWith(nonMixedNode, layer);
+  })
+
```
fgravin: Add a test for level 2 node that have NON-MIXED parent.
layer should be null.
Could be part of future work

fgravin:
I closely supervised the gmf layer tree refining. I think it is a very good improvment, thanks @olivierSemet 
Hopefully it won't break anything.

Very important to add test, really appreciable.
Thanks for this good work.

contribs/gmf/src/services/permalink.js
olivierSemet: Changing visibility at group level is possible so I would say yes.

contribs/gmf/src/services/permalink.js
fgravin: Thanks @adube 
You are right group should not be listen.
@olivierSemet will fix this and fix his test in an other PR.

contribs/gmf/src/services/permalink.js
olivierSemet: Ok @adube, I talked with @fgravin about it. You're right, there is no need to put a listener on ol.layer.Group objects. I going to fix it on a new PR right now. I'll let you know when it's done.

adube: Understood.

Issue #968, opened 2016-03-30
Placement of action icons on long layers


rbovard:
If a layer name is too long, the action icons are moved down and overlaps under items

![image](https://cloud.githubusercontent.com/assets/4549666/14146524/374c0dda-f698-11e5-99fa-066871b585a8.png)

In this case, the icons are for the layer `mb_statio_public_restreint`

tsauerwein:
Should the title be cut off ('mb_statio_public_...')? Or should the remaining characters be displayed on a new line?

olivierSemet:
I investigated on the second solution @tsauerwein  and found a way to display on multiple lines the name when its too long. But the solution implies a certain amount of CSS modification (aka not a straightforward solution)

tsauerwein:
@olivierSemet Do you still have that code in a branch?

sbrunner:
On CGXP we use the first solution ...

olivierSemet:
@tsauerwein yes I have. I'll push it on my GH and I'll ping you when it's done

tsauerwein:
> On CGXP we use the first solution ...

@sbrunner  Should it be done like in cgxp?

sbrunner:
The CGXP version is good enough, we will do the simplest version :-)

sbrunner:
Finally in cgxp we don't add the ellipsis, we just make that the icons don't move and don't be covered...

sbrunner:
I just added a layer with the name `Layer with very very very very very long name` but it don't works, see: #1017

olivierSemet:
I can see the layer but this is not a long enough name :-) 
 
<img width="284" alt="capture d ecran 2016-04-12 a 11 46 19 1" src="https://cloud.githubusercontent.com/assets/8366681/14456250/3f3f5568-00a4-11e6-8eb4-2eb82360c4f7.png">

PR #1007, opened 2016-04-11
Remove ticks on Layertree's leaves

olivierSemet:
Related to issue #1006 

PING @pgiraud

PR #1020, opened 2016-04-12
Truncate long layer names

tsauerwein:
Long layer names are truncated using `text-overflow: ellipsis`, the title attribute contains the full name.

gmf-layer-names

Closes https://github.com/camptocamp/ngeo/issues/968

Note that this PR is made against master and not 2.0 as confirmed with @ybolognini.

### DIFF
--- a/contribs/gmf/less/layertree.less
+++ b/contribs/gmf/less/layertree.less
@@ -106,7 +106,14 @@
   }
 
   .name {
-    white-space: normal;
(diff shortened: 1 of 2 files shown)


pgiraud:
This should be adjusted once "state" (tick) icon is removed. See https://github.com/camptocamp/ngeo/pull/1007

sbrunner:
Looks good

tsauerwein:
Thanks for the reviews.

Issue #450, opened 2015-12-08
run tests in a real browser window


fredj:
Find a simple way to be able to run all the unit tests in a browser.

tsauerwein:
I guess you know that you can launch the tests in the browser by setting `browsers: ['Chrome']` in the Karma conf? You will also also have to install `karma-chrome-launcher', see also in the [docs](http://karma-runner.github.io/0.13/config/browsers.html) or [here](http://stackoverflow.com/questions/15906903/debug-in-browser-using-testacular-now-karma). But it is not as nice as the runner we have for ol.

sbrunner:
This don't allow to run the test on a chrome browser with a debugger ...

tsauerwein:
> This don't allow to run the test on a chrome browser with a debugger ...

https://github.com/camptocamp/ngeo/pull/1025 allows to do exactly that.

sbrunner:
Sorry I read your documentation and it ream fixed by #1025, not by your comment ... :-)

PR #1025, opened 2016-04-13
Make running the unit tests less painful

tsauerwein:
This PR suggests to add a target `test-debug` to the Makefile which starts the unit tests in Chrome, so that the debugger can be used.

closes https://github.com/camptocamp/ngeo/issues/450


fgravin:
Nice one, LGTM

tsauerwein:
Thanks for the review.

PR #1043, opened 2016-04-19
[WIP] Modify rectangle interaction

jlap:
This commit adds the modify rectangle interaction, along with an example of how it works.

The interactions has two layers, one layers with the points which you can drag, and a second one with the polygon that represents the box.

A live example is available here: 
http://adube.github.io/ngeo/modify-rectangle/examples/modifyrectangle.html

However, there is still missing a way of knowing you can interact with the rectangle, I was thinking of either a hover effect in the style, or a different cursor pointer, but I wanted to hear some opinions before taking action. With the modify circle, what was done is the point only appears when you hover the circle, but for the rectangles, I went with [the spec](https://github.com/camptocamp/c2cgeoportal/wiki/Spec-%231658-Measures-and-redlining), where we can see all four corners of the rectangle.

I see three solutions for this:
- Changing the cursor on hover
- Changing the style on hover
- Only display the point on hover (like the modify circle)

I'd appreciate some opinions, @fgravin ! You can check out the code if you want, above changes aren't going to be huge, I'd appreciate it just to make sure this is all on the right track.


fgravin:
Ok very good.
For the style, the modify Interaction already has a style to display Vertices. 
Best option is that the hover on vertices highligh the corner.

jlap:
@fgravin what do you think of this?: http://adube.github.io/ngeo/modify-rectangle/examples/modifyrectangle.html

fgravin:
Very good.
I merge it, you'll work back on it if needed.
Please integrate it in drawFeature example, with the Vertex Style.
Thanks @jlap

PR #1044, opened 2016-04-19
Add ngeo-exportfeatures directive, used in gmf-drawfeature

adube:
This PR introduces the ability to `export` features that are drawn using the `gmf-drawfeature` directive.  

 * The `ngeo.FeatureHelper` is responsible of executing the writing of features and download as well.
 * A `ngeo-exportfeatures` directive is responsible of the UI aspect.  It will also be responsible to show a drop-down list if more than one format is defined.
 * The `ngeoExportFeatureFormats` constant will be responsible of defining in what formats we want the exportation to be supported.  Defaults to `KML` only, if not defined.

## Todo

 * [x] support KML export
 * [x] support GPX
 * [x] show drop-down menu if more than one format defined
 * [x] review
 * [x] apply reviewer's corrections
 * [x] reproject when writing features

We won't use template instead of jQuery to build menu items, because the menu has to be a sibling of the button. That can't work with pure Angular.

## Live demo

 * http://adube.github.io/ngeo/gmf-drawfeature-export/examples/contribs/gmf/drawfeature.html

### DIFF
--- a/src/directives/exportfeatures.js
+++ b/src/directives/exportfeatures.js
@@ -1,0 +1,214 @@
+goog.provide('ngeo.ExportfeaturesController');
+goog.provide('ngeo.exportfeaturesDirective');
+
+goog.require('ngeo');
+
+
+/**
+ * Directive used to export vector features in different types of format.
+ * To configure which formats to use, define the `ngeoExportFeatureFormats`
+ * constant, as such:
+ *
+ *     app.module.constant('ngeoExportFeatureFormats', [
+ *         ngeo.FeatureHelper.FormatType.KML,
+ *         ngeo.FeatureHelper.FormatType.GPX
+ *     ]);
+ *
+ * Example:
+ *
+ *     <button
+ *       ngeo-exportfeatures
+ *       ngeo-exportfeatures-features="ctrl.features"
+ *       class="btn btn-link">Export</button>
+ *
+ * @htmlAttribute {ol.Collection.<ol.Feature>} ngeo-exportfeatures-features The
+ *     features to export
+ * @return {angular.Directive} The directive specs.
+ * @ngInject
+ * @ngdoc directive
+ * @ngname ngeoExportfeatures
+ */
+ngeo.exportfeaturesDirective = function() {
+  return {
+    controller: 'ngeoExportfeaturesController',
+    scope: true,
+    bindToController: {
+      'features': '=ngeoExportfeaturesFeatures'
+    },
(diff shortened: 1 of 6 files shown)


adube:
@fgravin Ready for review.  See also the example.

src/directives/exportfeatures.js
```diff
@@ -0,0 +1,214 @@
...
+   */
+  this.menu_ = null;
+
+  /**
+   * @type {Array.<jQuery>}
+   * @private
+   */
+  this.items_ = [];
+
+  // build the drop-down menu and items of there's more than one format
```
fgravin: if

src/directives/exportfeatures.js
```diff
@@ -0,0 +1,214 @@
...
+  if (formats.length > 1) {
```
fgravin: why not using angular template ????

src/services/featurehelper.js
```diff
@@ -446,6 +450,104 @@ ngeo.FeatureHelper.prototype.getStrokeProperty = function(feature) {
...
+    clone.setProperties(feature.getProperties());
+    clone.setGeometry(clone.getGeometry().clone());
+    if (this.projection_) {
+      clone.getGeometry().transform(this.projection_, ol.proj.get('EPSG:4326'));
+    }
+    this.setStyle(clone, false);
+    clones.push(clone);
+  }, this);
+
+  var data = format.writeFeatures(clones);
```
fgravin: http://openlayers.org/en/master/apidoc/ol.format.GPX.html#writeFeatures

Is the clone needed ?

```js
format.writeFeatures(features, {
  dataProjection: 'EPSG:4326',
  featureProjection: this.projection_
})
```
doesn't do it properly ?

fgravin:
Good, see my comments.

src/services/featurehelper.js
adube: For transforming to 'EPSG:4326', the above is an excellent idea, but we still need to clone the features.  The exportation may occur while a feature is selected, therefore it may have the "select" style applied (halo, vertex).

I think it's okay to leave the cloning there.

src/directives/exportfeatures.js
adube: I don't have the reflexes to use Angular as it's not always obvious for me to use it when it can be.  In this case, I don't see how it can be used, but that's just a lack of understanding on my part. Would you please explain this to me ???? :)

PR #1045, opened 2016-04-19
Permalink service should not add listeners on ol.layer.Group

olivierSemet:
@adube @fgravin related to #1018  comments, I pushed a new version avoiding listeners when layer is a ol.layer.Group


fgravin:
Ok sounds good to me.
Waiting for @adube confirmation.

adube:
+1.  Great work.

fgravin:
Thanks for review

Issue #1029, opened 2016-04-13
Layertree's state is not consistent after page reload (localStorage does not reflect the tree state)


olivierSemet:
Bug recipe:

1. go to [gmf app](https://camptocamp.github.io/ngeo/master/examples/contribs/gmf/apps/desktop/) remove all keys in your localStorage

1. Reload the page, you should have  the following state for the layertree
<img width="290" alt="capture d ecran 2016-04-13 a 15 30 18" src="https://cloud.githubusercontent.com/assets/8366681/14494746/c2ad383c-018c-11e6-9738-1cae02a76395.png">

1. Disable all first level nodes by clicking on them. you should have the following state:
<img width="301" alt="capture d ecran 2016-04-13 a 15 32 57" src="https://cloud.githubusercontent.com/assets/8366681/14494827/1f92968c-018d-11e6-9320-d6cae974a2b0.png">

1. Reload the page. State expected, same as 3. but found:
<img width="302" alt="capture d ecran 2016-04-13 a 15 35 28" src="https://cloud.githubusercontent.com/assets/8366681/14494892/722b15cc-018d-11e6-9c06-d6395fb1cd28.png">
 
See below the localStorage after those actions:

<img width="711" alt="capture d ecran 2016-04-13 a 15 37 14" src="https://cloud.githubusercontent.com/assets/8366681/14494976/ccc53314-018d-11e6-8c43-29ce3946b14e.png">

1. There is an empty value for tree_group_layers_Group and tree_group_layers_Layers groups nodes
2. There is no key for the OSM function group node (only for its children) (and the bug is here?)

fgravin:
Thanks for reporting.
@olivierSemet could we have a test that point this issue out ?

fredj:
@olivierSemet fixed with #1048 ?

olivierSemet:
@adube yes

fredj:
thanks, closing this issue

PR #1048, opened 2016-04-20
Layertree state is not consistent with localstorage

olivierSemet:
related to #1029


olivierSemet:
@fgravin @pgiraud could you review please?

fgravin:
Ok looks good, thanks for adding tests.
I just would like a quick review from @adube to be sure everything is fine.

adube:
Good catch.  Thanks for the fix.

+1

PR #1051, opened 2016-04-21
Fix bug in built version of the modify rectangle interaction

jlap:
This commit fixes a bug in the modify rectangle that only occurs in the built version.

Vector sources usually raise an assert error when trying to add a feature to them that was already added, but it turns out that this error is not raised in the built version (I think that may be a bug in ol?).

So what this commit does is check manually if the feature has already been added.

This fixes the issue when if you select and unselect the rectangle repeatedly, the feature would sometimes be added in double and not removed.

Here is a link to the working interaction, integrated in the draw feature example:
http://jlap.github.io/ngeo/modify-rectangle-bug-build/examples/contribs/gmf/drawfeature.html

@fgravin this is ready to review.

### DIFF
--- a/src/ol-ext/interaction/modifyrectangle.js
+++ b/src/ol-ext/interaction/modifyrectangle.js
@@ -136,13 +136,26 @@ ngeo.interaction.ModifyRectangle.prototype.addFeature_ = function(feature) {
     var boxSource = this.vectorBoxes_.getSource();
     var pointSource = this.vectorPoints_.getSource();
 
-    try {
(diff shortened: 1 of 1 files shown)


tsauerwein:
> Vector sources usually raise an assert error when trying to add a feature to them that was already added, but it turns out that this error is not raised in the built version (I think that may be a bug in ol?).

The assertions are removed in the built version.

jlap:
I actually found a better way to fix this while working on the rotate, closing for now.

PR #1053, opened 2016-04-21
Missing check icon

pgiraud:
When we decided to remove the "check" icons (tick) in the layertree we did not pay attention to the fact that the "check" icon was used for the theme selector (in flush mode).
This pull request re-adds the check-icon.
It also removes the css rule for check-o icon which is not used anymore and for which the svg icon has been removed.

Please review.


sbrunner:
Thanks, merge :-)

PR #1054, opened 2016-04-21
Don't add legend block if not required

pgiraud:
We currently add the legend block even if no legend is to be displayed.
This pull request fixes this by using the same `ng-if` statement as for the legend icon button.

Please review.


fgravin:
I think that extra parenthesis are useless.
`::` is just detected as first charcters of a watchable expression, no matter what comes after.

But look goods

pgiraud:
Parenthesis removed.
Merging.

PR #1060, opened 2016-04-25
Fix mobile-query example

tsauerwein:
When running locally with `make serve`, the mobile-query example failed with:

> Error: [$injector:unpr] Unknown provider: ngeoLayerHelperProvider <- ngeoLayerHelper <- ngeoQuery <- ngeoMobileQueryDirective

[query.js](https://github.com/camptocamp/ngeo/blob/master/src/services/query.js) was missing a require to `ngeo.LayerHelper`. Unfortunately this is not detected by the tests because the examples are run in built-mode.

### DIFF
--- a/src/services/query.js
+++ b/src/services/query.js
@@ -1,6 +1,7 @@
 goog.provide('ngeo.Query');
 
 goog.require('ngeo');
+goog.require('ngeo.LayerHelper');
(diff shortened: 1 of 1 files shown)


fgravin:
Good catch thanks you can merge

PR #1064, opened 2016-04-26
Display action icons on mouse hover #closes 1052

olivierSemet:
@pgiraud 
check live example [here](https://oliviersemet.github.io/ngeo/1052-tree-actions-buttons-only-displayed-on-hover/examples/contribs/gmf/apps/desktop/)
IMO we should add some extra UI to clarify which layer item is currently hovered


pgiraud:
LGTM. Merging.

PR #1065, opened 2016-04-26
Remove unused gmf.Permalink from gmf.LayertreeController

fredj:


fredj:
ping @sbrunner

sbrunner:
I don't know the exact impart, but it seam that looks good :-)

fredj:
thanks for the review

PR #1066, opened 2016-04-26
Download legend only when display is requested close #1055

### LINKED ISSUE #1055: Legend images should be downloaded only when displayed
Currently we add DOM elements even for legend block that are hidden (collapsed) by default. This may send a lot of http requests at page load even if the user never displays the legend.
It would be better to load the images only when required.

olivierSemet:
@pgiraud coud you review? Thank you


pgiraud:
I'm not a fan of relying on a "click" to check whether the "img" node should be added or not. What if we decide to expand the legend block programmaticaly?
Unfortunately I have no better idea for the moment.

fredj:
And what about using an external library ? https://github.com/Pentiado/angular-lazy-img

olivierSemet:
@pgiraud  I hesitated for a long time. I took the solution with less changes as possible but handle it programmatically  is easy as well. 

@fredj  IMO adding an extra lib for a such simple problem is not a good solution

sbrunner:
@gberaudo , @fredj what's the status of this Pull Request?

pgiraud:
This pull request is in my opinion too hacky. I'd prefer to leave it unmerged for the moment.

fredj:
same here

sbrunner:
Than we can close it?

PR #1070, opened 2016-04-26
Layertree's right button use an absolute position close #1068

### LINKED ISSUE #1068: layertree's right buttons should use an absolute position
Currently, right buttons of each node are floating to the right. Since the number of buttons per node can be different, the righ buttons should be positioned to an absolute position (on the right)

olivierSemet:


olivierSemet:
@pgiraud  could you check? Thank you

contribs/gmf/less/layertree.less
```diff
@@ -69,15 +70,16 @@
     opacity: 0.5;
   }
 
-  .leaf {
+  .leaf, .group {
```
pgiraud: Please add a line break. For readability and consistency.

olivierSemet: between .leaf and .group?

pgiraud: Yes.
In general, we add a line break after the commas in the less files.

olivierSemet: Ok, I just took the example just above to write those lines. There is no line break. I'll put a line break.

fredj:
Can we merge this ? ping @pgiraud

PR #1072, opened 2016-04-27
Add missing semicolons

fredj:
Found with ESLint


tsauerwein:
LGTM

Should we add a check to our ESLint config?

fredj:
I've proposed to add this check in openlayers/eslint-config-openlayers but we can add it in our .eslintrc

fredj:
Thanks for the review

PR #1075, opened 2016-04-27
Minor goog cleanup

fredj:
- Use Array.isArray instead of goog.isArray
 - Use arr.forEach instead of goog.array.forEach


tsauerwein:
Thanks, please feel free to merge.

fredj:
Thanks for the review

PR #1078, opened 2016-04-27
Rotation interaction

jlap:
This commit adds a new Rotation interaction that allows the user to rotate features.

The integration with the draw feature example is included in this commit.

The modify rectangle had to be modified a bit as the calculation I used didn't work when the rectangle was on an angle.

The rotate interaction works similarly to the translate interaction:
- Activating the interaction adds an icon at the center of the polygon or line.
- Pressing the escape key deactivates the interaction.
- When you're done with your rotation (on mouse up), the interaction gets deactivated.

You can try it out in two places:
- Stand alone example: http://jlap.github.io/ngeo/rotation/examples/rotate.html
- Integrated with the draw feature: http://jlap.github.io/ngeo/rotation/examples/contribs/gmf/drawfeature.html

**Things to do before merging**:

 - [X] Decide between the two rotation behavior
 - [ ] Review

### DIFF
--- a/contribs/gmf/src/directives/drawfeature.js
+++ b/contribs/gmf/src/directives/drawfeature.js
@@ -370,6 +370,10 @@ gmf.DrawfeatureController.prototype.handleActiveChange_ = function(active) {
         ol.interaction.TranslateEventType.TRANSLATEEND,
         this.handleTranslateEnd_, this));
 
+    keys.push(ol.events.listen(this.rotate_,
+        ngeo.RotateEventType.ROTATEEND,
+        this.handleRotateEnd_, this));
+
     toolMgr.registerTool(drawUid, this.drawToolActivate, false);
     toolMgr.registerTool(drawUid, this.mapSelectToolActivate, true);
 
@@ -537,6 +541,16 @@ gmf.DrawfeatureController.prototype.handleMenuActionClick_ = function(evt) {
  */
 gmf.DrawfeatureController.prototype.handleTranslateEnd_ = function(evt) {
   this.translate_.setActive(false);
+  this.scope_.$apply();
+};
+
+
+/**
+ * @param {ngeo.RotateEvent} evt Event.
+ * @private
+ */
+gmf.DrawfeatureController.prototype.handleRotateEnd_ = function(evt) {
+  this.rotate_.setActive(false);
   this.scope_.$apply();
 };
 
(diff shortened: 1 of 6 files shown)


ybolognini:
I prefer first behavior, so the user can continue to pan while in rotate mode.

Two things I saw while testing:
 * There are some bugs with interacting with vertices while action menu is open
 * I think the menu should close when interacting with vertices (move, delete vertices). @pgiraud what do you think?
 * Looks like one cannot resize a circle anymore?

jlap:
> There are some bugs with interacting with vertices while action menu is open

What kind of bugs?

> Looks like one cannot resize a circle anymore?

Eeh that's weird, I'm taking a look into it, it bugs so in master too so it'll be fixed in a separate PR.

ybolognini:
One bug:
 * Draw a new polygon with at least 4 vertices
 * Click on a vertex to delete it
 * Menu opens (I think it shouldn't, but that's another story)
 * Vertex is deleted but I can still add a vertex to the "ghost" line (line that isn't displayed anymore)

Another bug (similar):
 * Draw a new triangle
 * Click on feature to open menu
 * Add a vertex and move it away
 * Sometimes it works, sometimes the geometry is broken (extra vertices added, ghost lines just like first bug, etc.). Try to add several vertices (always with open menu) and you'll see

jlap:
Oh yes I can see it now, thank you for the explanation.

jlap:
For the problem with the circle I opened PR #1082 

For the issue with the menu, @adube could you take a look?

adube:
I'll create one issue per bug raised.

adube:
See: #1083, #1084.

pgiraud:
With both options, once the rotation is finished, it's impossible to drag a vertex. If you try all the map moves. At the second try, it's then possible.

An other issue is clearly related to the actions menu. I'll report it in a dedicated issue.

jlap:
@pgiraud I fixed the issue with the vectrex dragging at the end of the rotation. I applied the same method as the translate since the translate was working.

You can try it out at: http://jlap.github.io/ngeo/rotation/examples/contribs/gmf/drawfeature.html?map_x=-1575214&map_y=2739503&map_zoom=4 

With that out of the way, since nobody else gave an opinion on that subject, I'll assume everybody agrees with @ybolognini that the interaction should only work by clicking the polygon, as to not prevent the user from dragging the map.

With that decision taken and that last bug fixed, this PR is no longer a work in progress, and is ready for review, @fredj 

Thanks a lot everyone!

fredj:
The branch has conflicts with master, please rebase

jlap:
@fredj It's up to date now!

fredj:
Thanks @jlap

PR #1082, opened 2016-04-28
Fix the resizing of a circle in modify circle

jlap:
This fixes the bug where you couldn't resize a circle anymore.

This was caused by using the string 'isCircle' at some places, and a defined constant in other places, so when compiling some got minimized and some didn't. Now everything uses the constant, so the example here:
http://jlap.github.io/ngeo/fix-modify-circle/examples/contribs/gmf/drawfeature.html
and here:
http://jlap.github.io/ngeo/fix-modify-circle/examples/modifycircle.html

Are working again.

### DIFF
--- a/examples/modifycircle.js
+++ b/examples/modifycircle.js
@@ -51,14 +51,16 @@ app.MainController = function() {
    */
   this.features = new ol.Collection();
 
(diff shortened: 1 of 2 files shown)


fredj:
Thanks

PR #1087, opened 2016-04-28
Add new ngeo.DecorateLayerLoading service

fredj:
Expose a new `loading` property that is true when the source(s) is(are) loading and false otherwise.
Supports tiles and images layers.

### DIFF
--- a/src/services/decoratelayerloading.js
+++ b/src/services/decoratelayerloading.js
@@ -1,0 +1,82 @@
+goog.provide('ngeo.DecorateLayerLoading');
+
+goog.require('goog.asserts');
+goog.require('ngeo');
+
+
+/**
+ * Provides a function that adds a 'loading 'property (using
+ * `Object.defineProperty`) to an ol.layer.Group or a layer with
+ * an ol.source.Tile or an ol.source.Image source.
+ * This property is true when the layer is loading and false otherwise.
+ *
+ * Example:
+ *
+ *      <span ng-if="layer.loading">please wait</span>
+ *
+ * @typedef {function(ol.layer.Base, angular.Scope)}
+ * @ngdoc service
+ * @ngname ngeoDecorateLayerLoading
+ */
+ngeo.DecorateLayerLoading;
+
+
+/**
+ * @param {ol.layer.Base} layer Layer to decorate.
+ * @param {angular.Scope} $scope Scope.
+ */
+ngeo.decorateLayerLoading = function(layer, $scope) {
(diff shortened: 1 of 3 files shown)


fredj:
Thanks for any review

src/services/decoratelayerloading.js
```diff
@@ -0,0 +1,79 @@
...
+ * @param {angular.Scope} $scope Scope.
+ */
+ngeo.decorateLayerLoading = function(layer, $scope) {
+  goog.asserts.assertInstanceof(layer, ol.layer.Base);
+
+  var sources = [];
+  if (layer instanceof ol.layer.Group) {
+    // layer group
+    sources = layer.getLayersArray().map(function(layer) {
+      goog.asserts.assert(layer instanceof ol.layer.Layer);
```
tsauerwein: We don't have to support groups in groups?

src/services/decoratelayerloading.js
```diff
@@ -0,0 +1,79 @@
+goog.provide('ngeo.DecorateLayerLoading');
+
+goog.require('goog.asserts');
+goog.require('ngeo');
+
+
+/**
+ * Provides a function that adds a 'loading 'property (using
+ * `Object.defineProperty`) to the layer.
```
tsauerwein: Do you maybe want to mention that the service is only supposed to be used for layer with tile and image sources?

tsauerwein:
LGTM :+1:

src/services/decoratelayerloading.js
```diff
@@ -0,0 +1,79 @@
...
+    sources = layer.getLayersArray().map(function(layer) {
+      goog.asserts.assert(layer instanceof ol.layer.Layer);
+      return layer.getSource();
+    });
+  } else {
+    goog.asserts.assert(layer instanceof ol.layer.Layer);
+    sources = [layer.getSource()];
+  }
+
+  layer.set('load_count', 0);
```
gberaudo: Have you considered setting the property silently here and in the rest of the file?

src/services/decoratelayerloading.js
sbrunner: Yes we should support it!

src/services/decoratelayerloading.js
fredj: Yes, I will add a note

src/services/decoratelayerloading.js
fredj: Yes, good remark, thanks

fredj:
Thanks for the review

PR #1097, opened 2016-04-29
Fix master - actual state...

sbrunner:


sbrunner:
Needs help to fix the tests...

contribs/gmf/test/spec/services/authenticationservice.spec.js
```diff
@@ -48,7 +48,8 @@ describe('gmf.Authentication', function() {
     gmfAuthentication.load_();
     $httpBackend.flush();
 
-    expect(spy.calls.length).toBe(1);
+    console.log(spy.calls)
```
fredj: to be removed

sbrunner:
Just steal one issue with layer opacity ..., does anyone have an idea?

contribs/gmf/src/directives/layertree.js
```diff
@@ -258,11 +258,13 @@ gmf.LayertreeController.prototype.getLayer = function(node, parentCtrl, depth) {
         layer = this.getLayerCaseNotMixedGroup_(node);
         this.prepareLayer_(node, layer);
         break;
-      // no default
+      default:
+        throw 'node wrong type: ' + type;
```
fredj: `goog.asserts.fail(''node wrong type: ' + type')` instead ?

sbrunner: For me it should be an exception, not an assert ...

gberaudo: For exceptions, you should do `throw new Error('node wrong type: ' + type);`.

sbrunner: Fixed, thanks :-)

fredj:
+1

PR #1121, opened 2016-05-04
Fix the templates path when the URL changed

sbrunner:
demo: http://camptocamp.github.io/ngeo/template/examples/contribs/gmf/apps/desktop/index.html


fredj:
+1

Issue #1088, opened 2016-04-28
[Edit] Menu open/close behavior


ybolognini:
* [x] Action menu opens on right-click and long-press on mobile (so it won't open on geometry selection)
 * [x] Action menu should disappear when user clicks anywhere else or when an action is selected
 * [x] Remove "Action" title and cross to close the menu

adube:
>  Action menu should disappear when user clicks anywhere else or when an action is selected

This is already the case, right ?

> Remove "Action" title and cross to close the menu

Okay, simple enough.


> The action menu shouldn't appear when user clicks on a vertex

Okay.

PR #1129, opened 2016-05-04
Menu open/close behavior

jlap:
This commit is for the issue #1088 

The example is live at: http://jlap.github.io/ngeo/menu-behavior/examples/contribs/gmf/drawfeature.html?map_x=576979&map_y=1613530&map_zoom=3

The example displays the new behaviors as described in #1088: 

* Action menu opens on right-click and long-press on mobile (so it won't open on geometry selection)
* Action menu should disappear when user clicks anywhere else or when an action is selected
* Remove "Action" title and cross to close the menu

This PR is ready for review!

### DIFF
--- a/contribs/gmf/src/directives/drawfeature.js
+++ b/contribs/gmf/src/directives/drawfeature.js
@@ -198,8 +198,7 @@ gmf.DrawfeatureController = function($scope, $timeout, gettextCatalog,
       cls: 'fa fa-trash-o',
       label: gettextCatalog.getString('Delete'),
       name: gmf.DrawfeatureController.MenuActionType.DELETE
-    }],
-    title: gettextCatalog.getString('Actions')
+    }]
   });
   this.map.addOverlay(this.menu_);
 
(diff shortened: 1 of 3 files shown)


contribs/gmf/src/directives/drawfeature.js
```diff
@@ -409,14 +414,41 @@ gmf.DrawfeatureController.prototype.handleFeaturesRemove_ = function(evt) {
 gmf.DrawfeatureController.prototype.handleMapSelectActiveChange_ = function(
     active) {
 
+  var mapDiv = this.map.getTarget();
```
fredj: use `this.map.getTargetElement()` instead

contribs/gmf/src/directives/drawfeature.js
```diff
@@ -427,7 +459,66 @@ gmf.DrawfeatureController.prototype.handleMapSelectActiveChange_ = function(
...
+  clearTimeout(this.longPressTimeout_);
+};
+
+
+/**
+ * @param {Event} evt Event.
+ * @private
+ */
+gmf.DrawfeatureController.prototype.handleMapContextMenu_ = function(evt) {
+  evt.preventDefault();
```
fredj: preventDefault should only be called if there's a feature under the cursor

PR #1136, opened 2016-05-06
Removes unneeded ol.css

sbrunner:
Do we relay needs this css ?


fredj:
The css is used for the scaleline and the zoom button (at least)

sbrunner:
OK, thanks :-)

PR #1150, opened 2016-05-09
Don't display the reorder handle if there's one layer group

### LINKED ISSUE #1081: Don't display the reorder handles if there's is only one group
thought ?

fredj:
fixes #1081


sbrunner:
Looks good :-)

fredj:
Thanks for the review

PR #1151, opened 2016-05-09
Fix contextmenu/longpress unlisten

fredj:
#1129 follow-up


fredj:
otherwise the context menu is active when the draw panel is closed

fredj:
Thanks for any review

fredj:
I will merge soon, last chance to comments ...

pgiraud:
If I understand correctly it's actually the last argument that is of interest. The "scope" one, right?
This looks good to me. Please merge.

fredj:
An event handler must be un-registered with the same arguments as it was registered.
The problem here was that the handler was registered with the `false` and `this` arguments with un-registered without.

fredj:
Thanks for the review !

PR #1161, opened 2016-05-10
Styling and icon for out of scale nodes

pgiraud:
I created an icon, and I hope it's clear enough.
The text is now italic when the layer is out of scale.
IMO the icon is intrusive and no really relevant when the node is not active so I hide it in this case.

Demos:
https://pgiraud.github.io/ngeo/layertree_out_of_scale/examples/contribs/gmf/apps/desktop
https://pgiraud.github.io/ngeo/layertree_out_of_scale/examples/contribs/gmf/apps/mobile

Remaining issues:
Positioning of elements is not perfect and the icon will be visible if the name of the layer is very long.

Please review.


contribs/gmf/less/icons.less
```diff
@@ -53,4 +53,8 @@
   &.gmf-icon-text:after {
     content: '\e906';
   }
+
+  &.gmf-icon-search-go:after {
+    content: '\e908';
```
olivierSemet: should use the double dot :: for pseudo element

pgiraud: Indeed.
Let's be consistent for now and change all the selectors once for all.
Can you please open an issue for addressing this right after that?

contribs/gmf/src/directives/layertree.js
```diff
@@ -758,15 +758,22 @@ gmf.LayertreeController.prototype.removeNode = function(node) {
...
+gmf.LayertreeController.prototype.zoomToResolution = function(treeCtrl) {
+  var node = /** @type {GmfThemesNode} */ (treeCtrl.node);
   var view = this.map.getView();
   var resolution = node.minResolutionHint || node.maxResolutionHint;
   if (resolution !== undefined) {
     view.setResolution(resolution);
   }
+
+  // automatically activate layer if not already active
+  if (this.getNodeState(treeCtrl) != 'on') {
```
olivierSemet: It will never be true since the only way to call zoomToResolution for now is by clicking on the zoom icon, which is only displayed if gmfLayertreeCtrl.getNodeState(layertreeCtrl) == 'on'

contribs/gmf/less/icons.less
olivierSemet: Ok

contribs/gmf/src/directives/layertree.js
pgiraud: Good catch thanks. I added this before I decided to show the icon only when the node is active.

sbrunner:
The icon don't mean anything at the standard zoom..., should the arrow be moved on the right?

pgiraud:
> The icon don't mean anything at the standard zoom..., should the arrow be moved on the right?

You're right. I tried to change it a bit. I hope this is a bit clearer.
screenshot from 2016-05-10 11 40 10

sbrunner:
@pgiraud Yes, wonderful :-)

fredj:
LGTM

ybolognini:
Nice! 
Could we add a little space between layer name and go-to-scale icon?

pgiraud:
Space added (with an arbitrary value chosen). Merging.

Issue #1147, opened 2016-05-09
Layertree desktop fixes


ybolognini:
* [x] -later- Larger grab zone for reorder (now two pixels wide, should be at least 4)
* [ ] -later- When dragging a group to the bottom, the dashed zone is displayed and then disappear forever. It should stay visible
* [x] -todo- Icon go-to-scale isn't as specified: icon should be magnifier and positionned after layer name
* [x] -todo- When layer out of scale, should its name be in italics
* [x] -todo- Tooltips (for "show/hide legend" and "zoom to scale") should be implemented using bootstrap. See for instance print/rotate widget
* [x] When we click on « show legend » the popover should be hidden
* [x] https://github.com/camptocamp/ngeo/pull/1155#issuecomment-218100623

sbrunner:
see also: https://github.com/camptocamp/ngeo/pull/1155#issuecomment-218100623

pgiraud:
Last remaining item should be addressed in a different issue. #1347.
Closing.

PR #1166, opened 2016-05-10
 Larger grab zone for reorder

olivierSemet:
Here is my proposal to get a larger grab zone for reorder (check todo tasks #1147) without changing the layertree design.
The idea is to display a zone (div) in absolute position with a larger width than the handler icon. Therefore the handler icon is only display for the design purpose. The real "handling zone"  is the div itself (so we can change easily its size).

- desktop demo: [here](https://oliviersemet.github.io/ngeo/increase-size-of-drag-and-drop-handle/examples/contribs/gmf/apps/desktop/)
-mobile demo: [here](https://oliviersemet.github.io/ngeo/increase-size-of-drag-and-drop-handle/examples/contribs/gmf/apps/mobile/)

ping @fredj @pgiraud @sbrunner  if the solution looks good for you.

### DIFF
--- a/contribs/gmf/less/desktoplayertree.less
+++ b/contribs/gmf/less/desktoplayertree.less
@@ -27,7 +27,7 @@ gmf-layertree {
     }
 
     &:hover {
-      .sortable-handle {
+      .sortable-handle-icon {
         visibility: visible;
       }
     }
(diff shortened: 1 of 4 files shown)


olivierSemet:
I have fancier solution. WIP.

sbrunner:
It's OK, merging

PR #1167, opened 2016-05-10
The legend is now a boolean

sbrunner:


fredj:
https://github.com/camptocamp/ngeo/blob/master/contribs/gmf/examples/data/themes.json needs to be updated

sbrunner:
Why do we have a copy?

fredj:
the files is used by the examples in contribs/gmf/examples/

sbrunner:
Why thy don't use the service directly, or is it created at the build time?

PR #1168, opened 2016-05-10
Replace ol.Collection function calls in partial with function in controller

### LINKED ISSUE #1145: Desktop app - features list don't appear
In the desktop app, when features are already drawn on the map but none is selected, there should be a list of features in the panel along with "export" and "delete all" links.
The panel is currently empty.

Note: this works as expected in the draw feature example.

screenshot from 2016-05-09 15 03 23

fredj:
fixes #1145


sbrunner:
Looks good :-)

fredj:
Thanks for the review

Issue #1163, opened 2016-05-10
Permalink regressions


ger-benjamin:
Tested on demo 2.1, last commit:  "Add isExpanded" May 9

Not working anymore:

1. With mixed groups, layers state are not mirrored in the URL. (try it with WMTS layers as with WMS layers in mixed groups (theme "edit" for instance). URL is not not updated on activation/deactivation of layers).
2. Opacity is not mirrored in the URL (I think that's not required, please, comfirm)

PR #1187, opened 2016-05-10
Properly register group layers in permalink service

tsauerwein:
The moment the group layer for mixed layers is registered with the permalink service, the group does not have any layers yet. They are only added later when the directive in the layer tree is compiled (?).

By setting up a listener on the layer collection of each group layer, all layers can be registered.

Closes https://github.com/camptocamp/ngeo/issues/1163


fredj:
LGTM

tsauerwein:
Thanks for the review.

PR #1188, opened 2016-05-10
Fix OL3 version (previous no longer exist due to push --force)

adube:
This PR fixes the current issue with OL3 version used.  The previous commit no longer exists probably due to a `git push --force`.


adube:
@fredj Ready for review / merge.

adube:
The reason Travis fails is unrelated to the fix... but somewhat related to the upgrade of OpenLayers.

fredj:
sorry for the push force :-(

PR #1184, opened 2016-05-10
Fix the ol3 commit

sbrunner:

PR #1180, opened 2016-05-10
Don't add a vertex on right click

fredj:
Fixes #1085 

work in progress

PR #1190, opened 2016-05-11
Fix openlayers build

fredj:
fixes #1184 
fixes #1188

### DIFF
--- a/package.json
+++ b/package.json
@@ -52,7 +52,7 @@
     "less-plugin-autoprefix": "1.5.1",
     "less-plugin-clean-css": "1.5.1",
     "nomnom": "1.8.1",
-    "openlayers": "fredj/ol3#afb2084",
+    "openlayers": "fredj/ol3#6d5dcde",
     "phantomjs-prebuilt": "2.1.7",
     "proj4": "2.3.14",
     "temp": "0.8.3",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -52,7 +52,7 @@
     "less-plugin-autoprefix": "1.5.1",
     "less-plugin-clean-css": "1.5.1",
     "nomnom": "1.8.1",
-    "openlayers": "fredj/ol3#afb2084",
+    "openlayers": "fredj/ol3#b49546d",
```
tsauerwein: That branch still includes the change in ol3 that removed the provide for `ol.style.StyleFunction`, right?
See https://github.com/camptocamp/ngeo/pull/1180#issuecomment-218204325

fredj: non, this branch does not include this change: see https://github.com/fredj/ol3/tree/modify_condition_alt

tsauerwein: Yes, but commit https://github.com/fredj/ol3/commit/b49546d is in branch https://github.com/fredj/ol3/tree/modify_condition which includes the change. :)

fredj: oups, forget to push this change ...

tsauerwein:
Thanks! LGTM

fredj:
why travis didn't start ?

PR #1197, opened 2016-05-11
[WIP] Use goog.i18n.NumberFormat to format measurements

sbrunner:
#1192 part b


fredj:
@sbrunner do we still need this ?

PR #1198, opened 2016-05-11
I18n fixes in layertree and login / print panel

fredj:


sbrunner:
Looks good :-)

fredj:
Thanks for the review

Issue #1115, opened 2016-05-04
gmf-drawfeature: selection issue


fredj:
with contribs/gmf/examples/drawfeature.html:
 * draw a rectangle
 * draw a polygon
 * mouse over the polygon
 * the modification handles are displayed on the rectangle

pgiraud:
I'm pretty sure this is related to #1117.

fredj:
yes, it was related to #1117 

fixed with #1123

PR #1200, opened 2016-05-11
UI layertree legend

olivierSemet:
Following @pgiraud advices on issue: #1115
- "hide legend" link a is bit smaller,
- Use "@link-color" for the "hide legend" link
- pointer cursor only on the "hide legend" link

demo: https://oliviersemet.github.io/ngeo/pr-layertree-ui-improvement/examples/contribs/gmf/apps/desktop/index.html

@pgiraud @ybolognini @sbrunner

### DIFF
--- a/contribs/gmf/less/layertree.less
+++ b/contribs/gmf/less/layertree.less
@@ -186,17 +186,20 @@
   .legend {
     margin-left: @app-margin + @half-app-margin;
     position: relative;
+    border: 1px solid @main-bg-color;
     background-color: lighten(@main-bg-color, 8%);
     a {
       display: none;
       position: absolute;
-      color: @brand-info;
       right: @half-app-margin;
-      text-decoration: underline;
-      font-size: 10;
+      font-size: 1.1rem;
 
       &::before {
         display: none;
+      }
+
+      &:hover {
+        text-decoration: underline;
       }
     }
   }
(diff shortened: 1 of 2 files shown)


contribs/gmf/less/layertree.less
```diff
@@ -186,14 +186,15 @@
...
     background-color: lighten(@main-bg-color, 8%);
     a {
       display: none;
       position: absolute;
-      color: @brand-info;
+      color: @color-light;
       right: @half-app-margin;
       text-decoration: underline;
-      font-size: 10;
+      font-size: 10px;
```
sbrunner: Avoid using px, we should prefer rem...

sbrunner:
Should we have the underline on the « hide legend » only on the hover?

contribs/gmf/less/layertree.less
olivierSemet: agreed

olivierSemet:
@sbrunner This is not specified in https://github.com/camptocamp/c2cgeoportal/wiki/Spec-%231708-Layertree-desktop#legend

Let me know if you really want this behavior.

sbrunner:
What do you think @pgiraud?

pgiraud:
I agree, underline on hover only.

Also, why not use the default link colors?

I think that it's due to the fact that we already have a `.gmf-layertree-node a` class. Maybe we should be more specific.

fredj:
is it good to merge?

olivierSemet:
LGTM

pgiraud:
OK for me. Merging.

PR #1204, opened 2016-05-11
Use node.metadata.legendImage by default

### LINKED ISSUE #1086: Metadata legendImage is not used in layertree
When a metadata "legendImage" is defined (like "http://example.com/images/an_image.gif"), it must be taken instead of send a getLegendGraphic request.

olivierSemet:
Close #1086
@ger-benjamin @sbrunner  those little changes do the work?
For now there is no node in the layertree with the metadata attribut legendImage for testing.

### DIFF
--- a/contribs/gmf/src/directives/layertree.js
+++ b/contribs/gmf/src/directives/layertree.js
@@ -675,6 +675,10 @@ gmf.LayertreeController.prototype.getLegendURL = function(treeCtrl) {
     return null;
   }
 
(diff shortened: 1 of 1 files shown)


sbrunner:
Looks good :-)

ger-benjamin:
Not tested but that should work...

fredj:
legendImage must be quoted: ['legendImage']

olivierSemet:
@fredj you're right thanks

PR #1205, opened 2016-05-11
Use the primary & secondary colors for range inputs

pgiraud:
The colors are manually set but should instead use the colors defined in `vars.less` otherwise we'll loose coherence if the integrators want to customize the colors.


sbrunner:
Looks good :-)

PR #1207, opened 2016-05-11
All links should have the same color

### LINKED ISSUE #1206: Links color is not visible enough
Color chosen for the links may not be strong enough. The links appear as if they were disabled.

capture d ecran 2016-05-11 a 16 05 01

pgiraud:
Also removes unused special-link variable
Fixes #1206.


fredj:
+1

Issue #1100, opened 2016-05-02
Improve the behavior of the ngeo popover component


olivierSemet:
- [x] The Popover should be destroyed when a scroll on the layertree occurs

- [x] The Popover should be destroyed when a drag and drop occurs

olivierSemet:
@fredj  could we close this issue?

PR #1208, opened 2016-05-11
Dismiss popover if a drag and drop event occurs

olivierSemet:
Here is the issue. The dragListGroup component (from closure and used by the ngeo sortable directive) listen on a 'mousdown' event. When the event occurs, the component clone the node (in our case a gmf-Layertree node). So the DOM element dragged is a clone.

Definition of a click event:

> The click event is fired when a pointing device button (usually a mouse button) is pressed and released on a single element. 

Therefore when the device button is released, the element behind the pointer is not the same as when the mousdown event occurred. So neither the original element nor the cloned one won't trigger a click event.

That's why the click event do not occur and will never occur. 
We can't use a clic event.

@pgiraud @fredj Let me know if you think I missed something

Demo: https://oliviersemet.github.io/ngeo/pr-dismiss-popover-on-drag-and-drop-event/examples/contribs/gmf/apps/desktop/index.html

source: [click definition](https://developer.mozilla.org/fr/docs/Web/Events/click)

### DIFF
--- a/src/directives/popover.js
+++ b/src/directives/popover.js
@@ -41,7 +41,7 @@ ngeo.popoverDirective = function() {
       ngeoPopoverCtrl.anchorElm.on('inserted.bs.popover', function() {
         ngeoPopoverCtrl.bodyElm.show();
         ngeoPopoverCtrl.shown = true;
-        ngeoPopoverCtrl.bodyElm.parent().on('click', function(e) {
+        ngeoPopoverCtrl.bodyElm.parent().on('mousedown', function(e) {
           e.stopPropagation();
         });
       });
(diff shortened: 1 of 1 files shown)


olivierSemet:
fixes #1100

fredj:
The problem with this solution is that the opacity slider can't be "slided": as soon as the thumb is clicked the popover is closed

olivierSemet:
@fredj  I forgot to change a click handler to a mousedown handler, should be ok now

fredj:
still not working in your demo, please update it

fredj:
- open the menu
 - mousedown the slider thumb
 - the menu is closed

olivierSemet:
ok thanks @fredj

olivierSemet:
@fredj  I updated the gh pages. The code is fine, the demo just didn't use the right one

fredj:
demo looks good, thanks

Issue #1160, opened 2016-05-10
Selecting scale in print makes strange zoom / scale choice behaviour


ger-benjamin:
Tested on demo 2.1, last commit:  "Add isExpanded" May 9

In print panel, play with the scale and select 1:25'000. Result scale is 1:10'000.
Choose again 1:25'000. Result scale is 1:25'000 but nothing has changed.
You must zoom in/out on the map to get a correct value of scale.

ger-benjamin:
(Perhaps that a part of the solution: https://github.com/camptocamp/ngeo/pull/1203)

fredj:
yes, it's probably the same thing

PR #1212, opened 2016-05-12
Constrain resolution on print set scale

ger-benjamin:
Fix: #1160 
Example: https://ger-benjamin.github.io/ngeo/print_scale/examples/contribs/gmf/apps/desktop/index.html


ger-benjamin:
I've tried with the additional parameters "..., 0, 1" but that not works...

ger-benjamin:
:dolphin:

fredj:
LGTM

PR #1214, opened 2016-05-12
Simplify the mask calculation if the map rotate

### LINKED ISSUE #1213: Simplify the mask calculation if the map rotate
Simplify this: https://github.com/camptocamp/ngeo/blob/master/src/services/printutils.js#L89-L112 (use old system) when rotation function doesn't exist.

ger-benjamin:
Fix #1213 

Example: 
 - http://ger-benjamin.github.io/ngeo/simple_calculation/examples/contribs/gmf/apps/desktop_alt (use rotation)
 - http://ger-benjamin.github.io/ngeo/simple_calculation/examples/contribs/gmf/apps/desktop (don't use rotation)


ger-benjamin:
(I just move logic into two differents private functions)

fredj:
thanks @ger-benjamin

PR #1226, opened 2016-05-17
Remove angular.$q service from ngeo.Notification

adube:
This PR removes the angular.$q service from the ngeo notification service since it's not used at all.


adube:
Ready for review.

fredj:
thanks (I should have noticed that in the review ...)

adube:
No problem.  I should have noticed that one too earlier.

PR #1237, opened 2016-05-17
Using node.layers property instead of node.name fix #1017

### LINKED ISSUE #1017: When a layer have an "name" different than the "WMS layer name" it doesn't work
see for example the layer named: `Layer with very very very very very long name`
with layers: `restaurant`

olivierSemet:
Fix #1017 

@sbrunner @fredj  please review both and carefully.

live example: https://oliviersemet.github.io/ngeo/fix-no-request-when-layer-name-differente-from-node-name/examples/contribs/gmf/apps/desktop/index.html

I think, there is some extra work to have cleaner code (check the getNodeSate function from layertree directive) but I suggest to go step by step. This PR should fix the issue #1017 

Also we should test a node with multiple layers to have a concret case for node.layers = "layer_1, layer_2 ..."

### DIFF
--- a/contribs/gmf/src/directives/layertree.js
+++ b/contribs/gmf/src/directives/layertree.js
@@ -279,7 +279,7 @@ gmf.LayertreeController.prototype.getLayer = function(node, parentCtrl, depth) {
       break;
     case gmf.Themes.NodeType.WMS:
       var url = node.url || this.gmfWmsUrl_;
-      layer = this.layerHelper_.createBasicWMSLayer(url, node.name,
+      layer = this.layerHelper_.createBasicWMSLayer(url, node.layers,
               node.serverType);
       break;
     default:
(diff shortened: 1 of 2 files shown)


sbrunner:
I try to test but the application is completely broken, in the theme OSM contains only 3 layers ...

I also add a layer named two_layer with two layers :-)

olivierSemet:
@sbrunner yes thanks I saw it. Closure compilation troubles :( 
WIP :)

sbrunner:
I don't think so, as I see a MIXED_GROUP can't have any sub groups. it should be fixed ...

olivierSemet:
@sbrunner  you added a node in  "OSM Function" group called "osm_time". I don't now why yet but this node throw an error "Error: node wrong type: NotMixedGroup"

olivierSemet:
Ok but this is not the subject of this issue then

sbrunner:
True, I notice it here because It hide the result...

olivierSemet:
@sbrunner could you undo your changes? I need compatible tree to work. thanks

sbrunner:
I create a temporary fix for that...
https://github.com/camptocamp/ngeo/pull/1249

sbrunner:
Merged :-)

sbrunner:
Should be rebase to get #1249 and to be testable :-)

fredj:
LGTM, let's merge this PR; we can open new issue if we have missed something

PR #1242, opened 2016-05-18
Adding placement property for ngeo popover directive

olivierSemet:
- We'll need it to embed time slider into the popover
@fredj @pgiraud


fredj:
thanks

PR #1244, opened 2016-05-18
Use notification service in desktop geolocation

adube:
This PR introduces the usability of the `ngeo.Notification` service within the `ngeo.DesktopGeolocation` directive. Instead of listening to the event and manage the errors at the application level, those are now managed directly in the geolocation service.

## Todo

 * [ ] review

## Live example

After clicking the "Toggle" button, when the browser asks to share your location, choose "Block".  You'll see the error message being displayed.

 * https://adube.github.io/ngeo/notification-on-geolocation-failure-desktop/examples/desktopgeolocation.html

### DIFF
--- a/examples/desktopgeolocation.html
+++ b/examples/desktopgeolocation.html
@@ -16,6 +16,13 @@
       button[ngeo-desktop-geolocation]:after {
         content: 'Toggle'
       }
+      .ngeo-notification {
+        left: 50%;
+        margin: 0 0 0 -100px;
+        position: absolute;
+        top: 0;
+        width: 200px;
+      }
     </style>
   </head>
   <body ng-controller="MainController as ctrl">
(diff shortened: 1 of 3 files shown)


sbrunner:
On Firefox If I responds not now I don't have any message...
Can it be solved?

fredj:
> On Firefox If I responds not now I don't have any message...
Can it be solved?

Note that there is absolutely no way to determine if the user responded "not now".
(I've faced the same issue in an other project)

tsauerwein:
See https://github.com/adube/ngeo/blob/b19a64761d1efaf9a3dec95cc796205318c3516a/src/directives/desktopgeolocation.js#L220-L226

I guess a notification should also be shown there.

adube:
Thanks for the review.  I'll make the fix suggested by @tsauerwein .

adube:
Live example updated.  As @fredj said, nothing happens when one chooses 'not now' in FF.

sbrunner:
Same for me, On Firefox nothing happens, than as discussed with @fredj  this code block should be removed ...

fredj:
we discussed about the "not now" issue with @sbrunner and we think that the hack in `ngeo.DesktopGeolocationController.prototype.setPosition_` should be removed.

@adube can you please remove the all the block in the function ?

adube:
Yes, I'm on it.

adube:
Done.  Live demo updated.

sbrunner:
Looks goof for me :-)

fredj:
thanks @adube

PR #1245, opened 2016-05-18
Use notification service in gmf authentication directive

adube:
This PR introduces the usability of the `ngeo.Notification` service within the `gmf.Authentication` directive.  The message are displayed where they used to be.

## Todo

 * [ ] review

## Live example

Try:

 * login with bad credentials
 * change password with bad credentials

in this example:

 * https://adube.github.io/ngeo/notification-on-authentication-failure/examples/contribs/gmf/authentication.html

### DIFF
--- a/contribs/gmf/src/directives/authenticationdirective.js
+++ b/contribs/gmf/src/directives/authenticationdirective.js
@@ -3,6 +3,7 @@ goog.provide('gmf.authenticationDirective');
 
 goog.require('gmf');
 goog.require('gmf.Authentication');
+goog.require('ngeo.Notification');
 /** @suppress {extraRequire} */
 goog.require('ngeo.modalDirective');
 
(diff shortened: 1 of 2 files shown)


sbrunner:
Two comments not directly related:
- The message « Could not connect. » is not user friendly... shouldn't be something like « Wrong user name or password. »?
- The empty field message should use the same kind of message...

Also that's looks good :-)

adube:
> The message « Could not connect. » is not user friendly... shouldn't be something like « Wrong user name or password. »?

This is out of scope of this PR, indeed.

>The empty field message should use the same kind of message...

I'm not sure where those messages come from.  The browser ?  Boostrap ?  Angular ?  I'd like to have help regarding this if we're to change this as well.  They are automatically shown because the fields are required.  Should we remove the requirement ?

sbrunner:
1. => #1250

2. From the browser because we set the HTML attribute `required="required"`

adube:
Okay, thanks.

Should this be covered in this PR ?  If so, what's the plan.  Do we remove `required="required"` ?

sbrunner:
> Should this be covered in this PR ? 

As you want...

> If so, what's the plan. Do we remove required="required" ?

Yes :-), add add a custom message :-)

adube:
I'll make an other PR.

sbrunner:
@fredj can we merge this pull request?

fredj:
yes, merging

PR #1246, opened 2016-05-18
Adding datepicker and slider from angular-ui

olivierSemet:
@fredj


package.json
```diff
@@ -42,6 +42,9 @@
     "graceful-fs": "4.1.4",
     "jasmine-core": "2.4.1",
     "jquery": "2.2.3",
+    "angular-ui-date": "1.0.1",
+    "angular-ui-slider": "0.3.2",
+    "jquery-ui": "git+https://github.com/jquery/jquery-ui.git#1.11.4",
```
fredj: I think you can simplify this to: ```"jquery-ui": "jquery/jquery-ui.git#1.11.4",``` (untested)

olivierSemet: No, npm jquery-ui package is useful when using tools like browserify.
https://www.npmjs.com/package/jquery-ui

sbrunner: The version on npm didn't works?

olivierSemet: It's made to be used with Browserify

fredj: `jquery/jquery-ui.git#1.11.4` is the github syntax; that's what we use for openlayers

fredj:
Maybe we should create a custom build of jQueryUI and commit it.
Online build: http://jqueryui.com/download/#!version=1.11.4&components=1110000000001000010000000000000000000

olivierSemet:
@fredj  I thought about doing that but in order to be as closely as your dependency management configuration, I took the decision to use npm. The solution of making a custom build is good as the build doesn't not have to change often"

olivierSemet:
@fredj  custom build does not embed i18n datepicker js files...

olivierSemet:
@fredj  @sbrunner  can we exclude 'third-party' from folders checked by Travis?

fredj:
#1269 should fix this issue, ping @sbrunner

sbrunner:
@fredj I can't see the relation...
We should ass a specific rule for the `contribs/gmf/third-party/jquery-ui/jquery-ui.css` file ...

fredj:
@sbrunner travis fails because the css files contains tabs instead of spaces.
This is specified in `.gitattributes`

fredj:
@olivierSemet so we include all jquery-ui ? Is it still possible to build only with the datepicker and slider ?=

olivierSemet:
@fredj  no check the banner of jquer-ui file:

> 
/*! jQuery UI - v1.11.4 - 2016-05-24
* http://jqueryui.com
* Includes: core.js, widget.js, mouse.js, position.js, datepicker.js, slider.js
* Copyright jQuery Foundation and other contributors; Licensed MIT */

fredj:
ok, got it.

sbrunner:
Better solution to don't disable all checks:
https://github.com/camptocamp/ngeo/commit/b280756295acf664215c969ecfc04c80a70b352e

olivierSemet:
@sbrunner could we merged?

sbrunner:
Thanks, merge :-)

PR #1257, opened 2016-05-19
Fix gmf authentication incorrect login i18n

### LINKED ISSUE #1250: Could not connect...
The message « Could not connect. » is not user friendly... shouldn't be something like « Wrong user name or password. »?

adube:
Fixes #1250.


sbrunner:
Thanks :-)

PR #1258, opened 2016-05-19
Fix #977 - Use notification in all gmf authentication err

adube:
This PR makes sure that all errors messages within the gmf authentication service are shown using the notification service.

## Todo

 * [ ] review

## Live example

 * https://adube.github.io/ngeo/gmf-auth-fix-login-err/examples/contribs/gmf/authentication.html


sbrunner:
Looks good but it should be rebased :-)

adube:
Rebased.

PR #1259, opened 2016-05-19
Show coordinates above the point instead of directly on

adube:
This PR makes sure that the point coordinates that are drawn with the draw tools are shown above the point instead of directly on it.

## Todo

 * [ ] review

## Live example

 * http://adube.github.io/ngeo/feature-helper-point-coordinates-above/examples/contribs/gmf/drawfeature.html?rl_features


sbrunner:
Looks good for me @fredj any comments?

fredj:
thanks @adube

PR #1262, opened 2016-05-20
Add options to control measure labels: X, Y

adube:
This PR introduces options to choose the labels to show instead of the default `X` and `Y` when showing point measurements, i.e. when showing coordinates.

## Todo

 * [ ] review

## Live example

 * https://adube.github.io/ngeo/measure-labels-x-y/examples/contribs/gmf/apps/desktop_alt/?lang=fr&rl_features=Fp(yff7-55hI~n*Point%25201%27c*%2523DB4436%27a*0%27o*0.2%27m*true%27s*10%27k*1)&map_x=630880&map_y=202064&map_zoom=3

### DIFF
--- a/contribs/gmf/apps/desktop/js/controller.js
+++ b/contribs/gmf/apps/desktop/js/controller.js
@@ -26,6 +26,9 @@ app.module.constant('ngeoExportFeatureFormats', [
   ngeo.FeatureHelper.FormatType.GPX
 ]);
 
+app.module.constant('ngeoMeasureLabelX', 'E');
+
+app.module.constant('ngeoMeasureLabelY', 'N');
 
 app.module.constant('ngeoQueryOptions', {
   'limit': 20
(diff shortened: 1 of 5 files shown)


adube:
Ready for review.

ger-benjamin:
And what if someone wants a prefix on the mesure X or Y ?
The problematic looks a little like this: https://github.com/camptocamp/ngeo/pull/1124 (file coordinateformat.js)
I've the feeling that we'll define these kind of 'label' too often.

I think that we have a X, a Y and that we must be able to choose a filter to format that. Angular filters (https://docs.angularjs.org/guide/filter) was created for these cases no ?.

What do you think @fredj (both PR)

sbrunner:
Shouldn't be the `E` and `N` be localized?

fredj:
> I think that we have a X, a Y and that we must be able to choose a filter to format that. Angular filters (https://docs.angularjs.org/guide/filter) was created for these cases no ?.

yes, we must create an angular format function

sbrunner:
@adube if it's OK for you @ger-benjamin will get this Pull request because there is some coverage with the pull request #1124 (he get also this one :-) )

adube:
@sbrunner No problem.  @ger-benjamin go a-head and pick this one up.

ger-benjamin:
I'll open another PR

PR #1267, opened 2016-05-24
Use openlayers v3.16.0

fredj:


sbrunner:
Looks good, merge when Travis is green

PR #1269, opened 2016-05-24
Move the text configuration to .editorconfig

fredj:


sbrunner:
What's this .editorconfig?

fredj:
http://editorconfig.org/
We this this system in ol3

sbrunner:
No relation with git, then you disable the git checks...

PR #1281, opened 2016-05-25
Don't force WMTS encoding to REST

### LINKED ISSUE #1236: WMTS should also work in KVP
Here we force to use REST:
https://github.com/camptocamp/ngeo/blob/4f5bf0f993ed112945755eb56b275f5b80761877/src/services/layerHelper.js#L96

I think that OpenLayers should be able to detect if the server work in REST or KVP...

fredj:
fixes #1236


ger-benjamin:
LGTM

sbrunner:
+1

Issue #1277, opened 2016-05-24
Remove unneeded calc() in less files


fredj:
there are several `calc` that can be removed from the less files.
Example in `search.less`:
```
left: calc(2 * @app-margin + @map-tools-size);
```
Will be compiled to:
```
left: calc(6rem);
```

PR #1285, opened 2016-05-25
less simplifications

fredj:
Small simplifications in the less files

### DIFF
--- a/contribs/gmf/less/desktop.less
+++ b/contribs/gmf/less/desktop.less
@@ -224,18 +224,18 @@ main {
 
 
 ::-webkit-scrollbar-track {
-    -webkit-border-radius: @app-margin;
-    border-radius: @app-margin;
-    background: @main-bg-color;
+  -webkit-border-radius: @app-margin;
+  border-radius: @app-margin;
+  background: @main-bg-color;
 }
 
 ::-webkit-scrollbar {
-    width: @half-app-margin;
+  width: @half-app-margin;
 }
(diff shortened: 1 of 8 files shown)


ger-benjamin:
Do you plan to fix https://github.com/camptocamp/ngeo/issues/1277 by the way ?

fredj:
Yes, I will also fix #1277

contribs/gmf/less/desktop.less
```diff
@@ -224,18 +224,18 @@ main {
 
 
 ::-webkit-scrollbar-track {
-    -webkit-border-radius: @app-margin;
-    border-radius: @app-margin;
-    background: @main-bg-color;
+  -webkit-border-radius: @app-margin;
```
pgiraud: Won't the auto prefix take care of that?

contribs/gmf/less/layertree.less
```diff
@@ -34,10 +36,10 @@
       color : @color-light;
       display: inline-block;
       &::before {
-        content: "\f054";
+        content: @fa-var-chevron-right;
```
pgiraud: Awesome. I didn't know there were variable for the unicode codes in font-awesome. Nice catch.

contribs/gmf/less/desktop.less
fredj: you're right, I will change that

fredj:
Thanks for the review

PR #1287, opened 2016-05-25
Enhance ngeo.Popup, add example

adube:
This PR introduces new methods in the `ngeo.Popup`:

 * `setUrl`: set the content as iframe using an url
 * `setWidth`, `setHeight` and `setSize`: forces the popup to use the defined size
 * `addClass`: add an extra CSS class name to allow more customization of popups
 * `setAutoDestroy`: when set, the popup gets automatically destroyed when closed, which is useful to do some automatic cleanup.
 * `open`: a single method from which one can quickly and easily open a new popup with properties set in an object of options.  It is an alternative to calling each other methods separately.

It also introduces an example to show what's possible to do with the popup factory.

## Todo

 * [ ] review

## Live example

 * http://adube.github.io/ngeo/window-with-iframe/examples/popup-service.html

### DIFF
--- a/examples/popup-service.html
+++ b/examples/popup-service.html
@@ -1,0 +1,62 @@
+<!DOCTYPE html>
+<html ng-app='app'>
+  <head>
+    <title>ngeoCreatePopup service factory example</title>
+    <meta charset="utf-8">
+    <meta name="viewport"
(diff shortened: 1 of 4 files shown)


adube:
FYI, this is ready for review.

PR #1289, opened 2016-05-25
Remove unecessary twbs row/col nesting

pgiraud:
With this pull request I remove an unnecessary level of nesting of "row" & "col-span-12" elements in the tools content panel.
Please review.


fredj:
LGTM (travis didn't start but I think it's safe to merge)

PR #1293, opened 2016-05-26
Fixing tools panel layout in IE

### LINKED ISSUE #1141: Right panel is always visible in IE
The space available to display the right tools panel (green empty column) is always visible on IE.
bs_win7_ie_11 0

pgiraud:
Fixes #1141

### DIFF
--- a/contribs/gmf/less/desktop.less
+++ b/contribs/gmf/less/desktop.less
@@ -149,11 +149,12 @@ main {
 
   .tools-content {
     width: @right-panel-width;
-    margin-right: -@right-panel-width + @toolbar-width;
+    margin-right: -@right-panel-width;
     transition: margin-right 0.2s ease;
+    float: right;
 
     &.active {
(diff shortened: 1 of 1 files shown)


contribs/gmf/less/desktop.less
```diff
@@ -184,6 +185,7 @@ main {
     float: right;
     height: 100%;
     position: relative;
+    z-index: 2;
```
ger-benjamin: @content-index ? https://github.com/camptocamp/ngeo/blob/master/contribs/gmf/less/vars.less#L18

pgiraud: Nope, I don't want to use this variable here since it doesn't mean anything in this context. IMO, using variables for indexes is useful for elements with absolute positioning and that are in the same container.

ger-benjamin:
>  it doesn't mean anything in this context.

Ok, I agree.

fredj:
+1

PR #1301, opened 2016-05-26
Add ngeo-disclaimer service and example

adube:
This PR introduces the `ngeo.Disclaimer` service. Since it would have had some method in common with the `ngeo.Notification` service, both extend the same class to avoid copy/pasting code that would be the same.

This PR is the first step towards supporting disclaimers inside the map for layers.  It doesn't do that yet.  This PR is only the section for NGEO.  The layers will be managed in the next PR, for GMF.

## Todo

 * [x] review

## Live example

 * http://adube.github.io/ngeo/disclaimer/examples/disclaimer.html

### DIFF
--- a/src/services/disclaimer.js
+++ b/src/services/disclaimer.js
@@ -1,0 +1,153 @@
+goog.provide('ngeo.Disclaimer');
+
+goog.require('goog.asserts');
+goog.require('ngeo');
+goog.require('ngeo.Message');
+
+
+/**
+ * Provides methods to display any sort of messages, disclaimers, errors,
+ * etc. Requires Bootstrap library (both CSS and JS) to display the alerts
+ * properly.
+ *
+ * @constructor
+ * @extends {ngeo.Message}
+ * @ngdoc service
+ * @ngname ngeoDisclaimer
+ * @ngInject
+ */
+ngeo.Disclaimer = function() {
+
+  goog.base(this);
+
+  var container = angular.element('<div class="ngeo-disclaimer"></div>');
+  angular.element(document.body).append(container);
+
+  /**
+   * @type {angular.JQLite}
+   * @private
+   */
+  this.container_ = container;
+
+  /**
+   * Cache of messages.
+   * @type {Object.<string, angular.JQLite>}
+   * @private
+   */
+  this.messages_ = {};
+
+};
+goog.inherits(ngeo.Disclaimer, ngeo.Message);
+
+
+/**
+ * Show disclaimer message string or object or list of disclamer message
+ * strings or objects.
+ * @param {string|Array.<string>|ngeox.Message|Array.<ngeox.Message>}
+ *     object A message or list of messages as text or configuration objects.
+ * @export
+ */
+ngeo.Disclaimer.prototype.alert = function(object) {
+  this.show(object);
+};
+
+
+/**
+ * Close disclaimer message string or object or list of disclamer message
+ * strings or objects.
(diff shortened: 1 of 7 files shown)


src/services/disclaimer.js
```diff
@@ -0,0 +1,153 @@
...
+  var type = message.type;
+  goog.asserts.assertString(type, 'Type should be set.');
+
+  var uid = this.getMessageUid_(message);
+  if (this.messages_[uid] !== undefined) {
+    return;
+  }
+
+  var classNames = ['alert', 'fade', 'alert-dismissible'];
+  switch (type) {
```
sbrunner: Any reason to have disclaimer type?

sbrunner:
Except my question it looks good :-)

src/services/disclaimer.js
adube: Because it was asked in the specs, see: https://github.com/camptocamp/c2cgeoportal/wiki/Spec-%231769-Disclaimer#architecture

>Manage different colors like the notifier

src/services/disclaimer.js
```diff
@@ -0,0 +1,153 @@
...
+ * @extends {ngeo.Message}
+ * @ngdoc service
+ * @ngname ngeoDisclaimer
+ * @ngInject
+ */
+ngeo.Disclaimer = function() {
+
+  goog.base(this);
+
+  var container = angular.element('<div class="ngeo-disclaimer"></div>');
```
adube: @pgiraud: I used `"ngeo-disclaimer"` instead of `"disclaimer"`.  I recommend using the extra namespace, here.

src/services/disclaimer.js
```diff
@@ -0,0 +1,153 @@
...
+  }
+
+  var el = angular.element(
+    '<div role="alert" class="' + classNames.join(' ') + '"></div>');
+  var button = angular.element(
+    '<button type="button" class="close" data-dismiss="alert" aria-label="' +
+    'Close' +  // FIXME i18n
+    '"><span aria-hidden="true">&times;</span></button>');
+  var msg = angular.element('<span />').html(message.msg);
+  el.append(button).append(msg);
```
pgiraud: It looks weird to me to have that much of HTML code in a service.

adube: What do you suggest ?

src/services/disclaimer.js
sbrunner: OK :-/

src/services/disclaimer.js
pgiraud: I don't have anything to suggest actually.
Only that it may have deserved a dedicated directive. It may not be worth.

adube: I thought about it too at first.  Here's my original thoughts about it.

A directive may not be injected.  I wanted to create something that I would inject from anywhere and ask it to show my messages.  A service was the ideal pick for that.

The next step for me is to create a directive in GMF that would define a container to hold the messages for the map.  I'd simply always ask the disclaimer service to show messages in which I define the target of that directive.

An alternative to all this would have been to keep the messages "as-is" in the service and don't manage the DOM.  We could then have a disclaimer directive listening to the messages added in there to display them accordingly.

Thoughts ?

adube:
Do you have any more comments regarding this ?  Is the review completed ?

fredj:
I will merge when travis is green

adube:
Great, thanks.

PR #1314, opened 2016-05-30
Use map-tools-color variable for the .nav-trigger color

fredj:


ger-benjamin:
+1

fredj:
Thanks for the review

PR #1322, opened 2016-06-01
Pass the server type to the ol.source.ImageWMS constructor

fredj:
Request high DPI image from the WMS server

### DIFF
--- a/src/services/layerHelper.js
+++ b/src/services/layerHelper.js
@@ -48,20 +48,24 @@ ngeo.LayerHelper.GROUP_KEY = 'groupName';
  * @param {string} sourceURL The source URL.
  * @param {string} sourceLayersName A dot separated names string.
  * @param {string=} opt_serverType Type of the server ("mapserver",
- *     "geoserver", qgisserver, …).
+ *     "geoserver", "qgisserver", …).
  * @return {ol.layer.Image} WMS Layer.
  * @export
  */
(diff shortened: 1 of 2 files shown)


src/services/layerHelper.js
```diff
@@ -48,7 +48,7 @@ ngeo.LayerHelper.GROUP_KEY = 'groupName';
  * @param {string} sourceURL The source URL.
  * @param {string} sourceLayersName A dot separated names string.
  * @param {string=} opt_serverType Type of the server ("mapserver",
- *     "geoserver", qgisserver, …).
+ *     "geoserver", "qgis", …).
```
sbrunner: It's qgisserver not qgis ..., see: http://mapfish.github.io/mapfish-print-doc/layers.html#Wms%20Layer
If somewere it's not like that it's an issue ...

fredj: it's qgis not qgisserver in ol3 :-)
https://github.com/openlayers/ol3/blob/master/src/ol/source/wmssource.js

sbrunner: Then we have an issue ...

sbrunner: For me ol3 is wrong :-( 
see: http://docs.qgis.org/2.8/fr/docs/user_manual/preamble/whats_new.html#qgis-server

ger-benjamin:
Looks good

fredj:
Thanks for the review

PR #1326, opened 2016-06-01
Admin value isChecked is now taken in account in all cases

ger-benjamin:
Fix: https://github.com/camptocamp/c2cgeoportal/issues/2219 (Fix 2.0: must be merged on the 2.0 only., nothing to do in 2.1 (already works))

Example: https://ger-benjamin.github.io/ngeo/admin_value_2.0/examples/contribs/gmf/apps/mobile/theme/OSM?lang=en&baselayer_ref=map

(For example, hotel is not checked, some layer in the same group are checked and so on. You can compare with the admin interface of the demo 2.0 (ui_metadatas): https://geomapfish-demo.camptocamp.net/2.0/wsgi/admin/

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,8 +8,8 @@
 /contribs/gmf/fonts/gmf-icons.ttf
 /contribs/gmf/fonts/gmf-icons.woff
 /.tx/config
-/.tx/.tx/ngeo.ngeo-*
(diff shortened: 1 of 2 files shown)


sbrunner:
Looks good :-)

PR #1338, opened 2016-06-06
Layout harmonization

### LINKED ISSUE #1134: Too much cursor change in layer tree
sans titre 1

When hovering group nodes in the layer tree, the cursor changes too much.
We should at least try to prevent the "select text" cursor.

### LINKED ISSUE #1302: Legend icons are squeezed
screenshot from 2016-05-27 09 59 59

pgiraud:
With this pull request, a lot of work is done on the layer tree CSS to harmonize the different sizes.
Font size gets a bit larger without loosing space.
Also the different nodes in the tree now use the flexbox display so that elements are better aligned and distributed. The computed width of labels for example is a lot less hackish and takes the really available space.
Flexbox has an other advantage: no blank space is automatically added between elements. So this fixes #1134.

I also reduced the nesting when possible.

Legend icons size (actually ratio) is taken into account correctly. The layout actually adapts to it. Icons are not cropped to squeezed. Only max limits for width and height are fixed. Fixes #1302.

After / Before
screenshot from 2016-06-06 09 17 46

On mobile, the font size is also bigger which makes it easier for users. Spaces between navigation items is also more consistent.
The flexbox display is also used and allows us to use several lines for long labels instead of an ellipsis. Elements are still correctly vertically aligned.

After / Before
screenshot from 2016-06-06 09 24 12

Please have a look at individual commits for more detailed information.

### DIFF
--- a/contribs/gmf/less/desktoplayertree.less
+++ b/contribs/gmf/less/desktoplayertree.less
@@ -14,9 +14,8 @@ gmf-layertree {
 .gmf-layertree-node {
 
   .right-buttons {
-    visibility: hidden;
+    display: none;
     cursor: pointer;
-    right: @app-margin;
   }
 
   &.depth-1 {
@@ -68,11 +67,15 @@ gmf-layertree {
   .group:hover,
   .leaf:hover {
     .right-buttons {
-      visibility: visible;
+      display: block;
     }
   }
 
-  .leaf .legend:hover {
+  .leaf {
+    padding-right: @half-app-margin;
+  }
+
+  .legend:hover {
     a {
       cursor: pointer;
       display: inline-block;
@@ -81,6 +84,10 @@ gmf-layertree {
 
   .legend-button {
     display: none;
+  }
+
+  .name {
+    white-space: nowrap;
   }
 
   // leave space for the drag handle
@@ -97,6 +104,7 @@ gmf-layertree {
     top: 0;
     height: 100%;
     width : @app-margin;
+    margin: 0;
   }
 
   .sortable-handle-icon {
--- a/contribs/gmf/less/desktoplayertree.less
+++ b/contribs/gmf/less/desktoplayertree.less
@@ -77,7 +77,8 @@ gmf-layertree {
     display: none;
   }
 
-  a.expand-node.fa {
+  // leave space for the drag handle
+  .depth-1 a.expand-node.fa {
     margin-left: @half-app-margin;
   }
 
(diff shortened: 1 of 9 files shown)


contribs/gmf/less/desktoplayertree.less
```diff
@@ -11,8 +13,8 @@ gmf-layertree {
 
 .gmf-layertree-node {
 
-  .rightButtons {
-    visibility: hidden;
+  .right-buttons {
+    display: none;
```
sbrunner: Why changing `visibility: visible/hidden;` by `display: *`?

contribs/gmf/less/layertree.less
```diff
@@ -75,8 +51,15 @@
 
   .leaf,
   .group {
-    font-size: @font-size-small;
+    font-size: ceil(@font-size-base * 0.9);
```
sbrunner: Is it a good practice to use ceil?

contribs/gmf/less/desktoplayertree.less
pgiraud: Because I want the space to be available for the label when the buttons are hidden.

pgiraud:
demos:
https://pgiraud.github.io/ngeo/layout_harmonization/examples/contribs/gmf/apps/desktop/
https://pgiraud.github.io/ngeo/layout_harmonization/examples/contribs/gmf/apps/mobile/

contribs/gmf/less/layertree.less
pgiraud: This is what bootstrap use. I don't have any other good reason to do it.

sbrunner: From where does the `font-size-base` come from? from bootstrap?
Doesn't is just be equals to 0.9rem?

sbrunner:
I just have a look on the code and I put some questions :-)
Also that's looks good for me :-)

contribs/gmf/less/layertree.less
pgiraud: @font-size-base indeed comes from bootstrap.
It's set to 14px by default otherwize we decide to change it.

pgiraud:
Travis build is now fixed. Can please someone give this a final review?

contribs/gmf/less/layertree.less
sbrunner: And what about the ceil?

pgiraud: This is what bootstrap use (for font-size-small for example). I don't have any other good reason to do it.

fredj:
LGTM, please merge

PR #1339, opened 2016-06-06
Use 90% of the viewport width on small mobile devices

pgiraud:
#1338 should be merged first.
The idea is to use as much space as possible for navigation panels on small devices on portrait mode.
screenshot from 2016-06-06 09 38 57

This unfortunately breaks background and theme selectors layout since they use fixed sizes.

### DIFF
--- a/contribs/gmf/less/vars.less
+++ b/contribs/gmf/less/vars.less
@@ -22,7 +22,7 @@
 @above-search-index: 6;
 
 
-@nav-width: 26rem;
+@nav-width: 32rem;
 
 @border-radius-base: 0;
 @border-color: black;
--- a/contribs/gmf/less/mobile-nav.less
+++ b/contribs/gmf/less/mobile-nav.less
@@ -230,6 +230,31 @@ nav.nav-right {
   }
 }
 
+// For small devices (in portrait mode) we want the navigation menus to take
+// 90% of the viewport width
+@media (max-width: @screen-xs) {
+  main {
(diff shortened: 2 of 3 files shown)


contribs/gmf/less/vars.less
```diff
@@ -22,7 +22,7 @@
 @above-search-index: 6;
 
 
-@nav-width: 26rem;
+@nav-width: 400px;
```
sbrunner: Why using px?

pgiraud: Because `26rem` doesn't mean anything for a panel width. `rem`s should be use for fonts or typography only IMHO.

sbrunner: I don't agree, px should be avoid as possible, and rem should be used every where is possible, with that we can easy do a choerant zoom all the application by only changing the root font!

ger-benjamin: You aim is to use 90% of the screen ? Can't you use `90vh` ?

pgiraud: I'm already using 90vw but only for a small device in portrait mode.
However I agree with Stéphane to continue using rem for the width for larger screens. I changed the value from 26 to 32. It was too small.

pgiraud:
Layout in background and theme selectors is now fixed.
Please review again.

sbrunner:
Looks good for me :-)

PR #1340, opened 2016-06-06
Add elevation components

ger-benjamin:
Part of: https://github.com/camptocamp/c2cgeoportal/issues/1668

Example: https://ger-benjamin.github.io/ngeo/elevation/examples/contribs/gmf/elevation.html

Example apidoc: https://ger-benjamin.github.io/ngeo/elevation/apidoc/gmf.elevationDirective.html

I've tried to exclude the dropdown of the directive. I think that's not a part of it. So we can display an elevation on any `<element>` and we don't need a partial. But that add two attributes in the html declaration of the directive (`gmf-elevation-elevation` and `gmf-elavation-layer`) and a watch (the layer)

### DIFF
--- a/contribs/gmf/src/directives/elevation.js
+++ b/contribs/gmf/src/directives/elevation.js
@@ -16,9 +16,10 @@ goog.require('ngeo.Debounce');
  * Example:
  *
  *      <span gmf-elevation
- *            gmf-elevation-active="elvationActive"
+ *            gmf-elevation-active="elevationActive"
  *            gmf-elevation-elevation="elevationValue"
  *            gmf-elevation-layer="mainCtrl.elevationLayer"
+ *            gmf-elevation-layers="::mainCtrl.elevationLayers"
  *            gmf-elevation-map="::mainCtrl.map">
  *            {{elevationValue | number:2}}m
  *      </span>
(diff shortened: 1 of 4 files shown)


sbrunner:
The layers parameter should be used...
https://camptocamp.github.io/c2cgeoportal/master/developer/webservices.html#raster

ger-benjamin:
@sbrunner Done. Now only the elevation layer choosed will be sent to the server. (Note that in the measure tool (which use also the getAltitude service), this parameters is never set)

ger-benjamin:
Please review

sbrunner:
Thanks @ger-benjamin, issue added #1343 :-)

contribs/gmf/src/directives/elevation.js
```diff
@@ -0,0 +1,214 @@
...
+ *      </span>
+ *
+ *
+ * @htmlAttribute {boolean} gmf-elevation-active A boolean to set active or
+ *     deactive the component.
+ * @htmlAttribute {number} gmf-elevation-elevation The value to set with the
+ *     elevation value.
+ * @htmlAttribute {string?} gmf-elevation-layer Optional elevation layer to
+ *     display. If not provided, display the first value.
+ * @htmlAttribute {string?} gmf-elevation-layers Optional list of layers to
```
fredj: type is `{Array.<string>}`

fredj:
LGTM

PR #1352, opened 2016-06-08
Infobar

ger-benjamin:
Revival of #1128 

Fix: https://github.com/camptocamp/c2cgeoportal/issues/1668 (without "contextual information" because `Display contextual information on right click. (still to be designed!)`)

Specs: https://github.com/camptocamp/c2cgeoportal/wiki/Spec-%231668-Map-(Info-bar)

Examples: 
 - https://ger-benjamin.github.io/ngeo/infobar/examples/contribs/gmf/apps/desktop
 - https://ger-benjamin.github.io/ngeo/infobar/examples/contribs/gmf/apps/desktop_alt (without choice for elevation)

Must be done in another PR (2.1?):
 - Display contextual information on right click (still to be designed, ping @ybolognini ) 
 - For a same zoom level, scales are not the same in the print panel and in the infobar (perhaps that's normal but this is strange to see).

Must be done in another PR (2.2):
 - Uniforms scaleselector, elevation and mouseposition. Because: 
  - `scaleselector` is old, doesn't use filter for scales, use partial with dropdown included and with one option to configure it (dropup).
  - `mouseposition` use partial with dropdown but without any option to configure it.
  - `elevation` doesn't include any partial so the dropdown is full configurable but the `index.html` is less compact.

Other bugs found and issues that will be open (2.1):
 - We should define what must be defined in the `src/controller` and what must be defined in the `apps controller`.

For the design of the buttons, it was too [hard / manual] to use a dropdown button on the side of a component and the same style for the component but without having the button effect. So I make a unique button (As proposed by @pgiraud).

### DIFF
--- a/contribs/gmf/apps/desktop/index.html
+++ b/contribs/gmf/apps/desktop/index.html
@@ -192,7 +192,7 @@
         module.constant('langUrls', langUrls);
         module.constant('authenticationBaseUrl', 'https://geomapfish-demo.camptocamp.net/2.1/wsgi');
         module.constant('fulltextsearchUrl', 'https://geomapfish-demo.camptocamp.net/2.1/wsgi/fulltextsearch?limit=30&partitionlimit=5');
-        module.constant('gmfAltitudeUrl', 'https://geomapfish-demo.camptocamp.net/2.1/wsgi/raster');
+        module.constant('gmfRasterUrl', 'https://geomapfish-demo.camptocamp.net/2.1/wsgi/raster');
         module.constant('gmfPrintUrl', 'https://geomapfish-demo.camptocamp.net/2.1/wsgi/printproxy');
         module.constant('gmfWmsUrl', 'https://geomapfish-demo.camptocamp.net/2.1/wsgi/mapserv_proxy');
         module.constant('gmfTreeUrl', 'https://geomapfish-demo.camptocamp.net/2.1/wsgi/themes?version=2&background=background');
--- a/contribs/gmf/apps/desktop/index.html
+++ b/contribs/gmf/apps/desktop/index.html
@@ -111,7 +111,7 @@
           </gmf-backgroundlayerselector>
         </div>
         <gmf-map class="gmf-map" gmf-map-map="mainCtrl.map"></gmf-map>
-        
+
         <!--infobar-->
         <div class="footer" ng-class="{'active': mainCtrl.info}">
           <button class="btn btn-sm fa map-info ng-cloak" ng-click="mainCtrl.info = !mainCtrl.info"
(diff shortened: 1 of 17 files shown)


contribs/gmf/src/controllers/abstractdesktop.js
```diff
@@ -111,6 +122,65 @@ gmf.AbstractDesktopController = function(config, $scope, $injector) {
...
+    '9': $sce.trustAsHtml('1&nbsp;:&nbsp;250'),
+    '10': $sce.trustAsHtml('1&nbsp;:&nbsp;100'),
+    '11': $sce.trustAsHtml('1&nbsp;:&nbsp;50')
+  };
+
+  /**
+   * @type {Array.<string>}
+   * @export
+   */
+  this.elevationLayers = ['aster', 'srtm'];
```
ger-benjamin: @fredj Do you know how I could translate that in a good way ?

contribs/gmf/apps/desktop/index.html
```diff
@@ -111,6 +111,49 @@
...
+                 gmf-mouseposition-projections="::mainCtrl.mousePositionProjections">
+            </gmf-mouseposition>
+            <span
+              gmf-elevation
+              gmf-elevation-active="mainCtrl.info"
+              gmf-elevation-elevation="elevationValue"
+              gmf-elevation-layer="mainCtrl.elevationLayer"
+              gmf-elevation-layers="::mainCtrl.elevationLayers"
+              gmf-elevation-map="mainCtrl.map">
+            </span>
```
pgiraud: Do you need a specific `<span>`? Isn't it possible to use the directive directly within the div below?

src/directives/partials/scaleselector.html
```diff
@@ -1,12 +1,14 @@
...
-  <ul class="dropdown-menu" role="menu">
-    <li ng-repeat="zoomLevel in ::scaleselectorCtrl.zoomLevels">
-      <a href ng-click="scaleselectorCtrl.changeZoom(zoomLevel)"
-         ng-bind-html="scaleselectorCtrl.getScale(zoomLevel)">
-      </a>
-    </li>
-  </ul>
+<div class="form-group form-group-sm dropdown" ng-class="::{'dropup': scaleselectorCtrl.options.dropup}">
+  <div class="btn-group btn-block">
+      <button type="button" class="btn btn-default form-control" data-toggle="dropdown" aria-expanded="false">
```
pgiraud: Indentation doesn't look correct

pgiraud: You need to add the `dropdown-toggle` class name otherwise the corners are squared on the right.
screenshot from 2016-06-08 11 48 45

contribs/gmf/src/controllers/abstractdesktop.js
```diff
@@ -111,6 +122,65 @@ gmf.AbstractDesktopController = function(config, $scope, $injector) {
...
+    '3': $sce.trustAsHtml('1&nbsp;:&nbsp;20\'000'),
+    '4': $sce.trustAsHtml('1&nbsp;:&nbsp;10\'000'),
+    '5': $sce.trustAsHtml('1&nbsp;:&nbsp;5\'000'),
+    '6': $sce.trustAsHtml('1&nbsp;:&nbsp;2\'000'),
+    '7': $sce.trustAsHtml('1&nbsp;:&nbsp;1\'000'),
+    '8': $sce.trustAsHtml('1&nbsp;:&nbsp;500'),
+    '9': $sce.trustAsHtml('1&nbsp;:&nbsp;250'),
+    '10': $sce.trustAsHtml('1&nbsp;:&nbsp;100'),
+    '11': $sce.trustAsHtml('1&nbsp;:&nbsp;50')
+  };
```
sbrunner: specific project data, => should be in the apps

contribs/gmf/src/controllers/abstractdesktop.js
sbrunner: specific project data, => should be in the apps

sbrunner:
Shouldn't we keep the last coordinate when the mouse exit the map?

pgiraud:
> Shouldn't we keep the last coordinate when the mouse exit the map?

@sbrunner I don't think so. Any argument in favour of keeping the last coordinate?

pgiraud:
> For the design of the buttons, it was too [hard / manual] to use a dropdown button on the side of a component and the same style for the component but without having the button effect. So I make a unique button (As proposed by @pgiraud).

I really want to give this a try because I'm not satisfied with the look&feel. Alignment (distance to edge) of left and right icons is visually weird.

pgiraud:
Can we make it so the info bar is hidden by default in desktop app and visible by default in alternative app?

contribs/gmf/less/desktop.less
```diff
@@ -64,6 +64,84 @@ main {
...
+    }
+    .form-group{
+      margin-bottom: 0;
+    }
+  }
+
+  [ngeo-scaleselector] {
+    min-width: 12rem;
+    .dropdown-menu {
+      min-width: 12rem;
```
pgiraud: I don't think we need to try to adapt the menu width to the corresponding button. Dropdown menus automatically set their size depending on their content. This is enough IMO.
It's also more code to maintain.

sbrunner:
> Any argument in favour of keeping the last coordinate?

For me it's relay strange when I want to change the projection and suddenly the coordinates dispars...

contribs/gmf/src/controllers/abstractdesktop.js
sbrunner: @ger-benjamin the translation should be done in the template, the string collecting will already be done on the server configuration.

contribs/gmf/less/desktop.less
ger-benjamin: That's true but only for dropdown with size > 160px (~=16rem here).
I don't know why but bootstrap defin a `min-width: 160px` on its `.dropdown-menu`.

ger-benjamin:
> I really want to give this a try because I'm not satisfied with the look&feel. Alignment (distance to edge) of left and right icons is visually weird.

No problem for me, I just think (based on my tests) that we will write too much "less" code to counter bootstrap comportment. But perhaps I'm wrong.

> Can we make it so the info bar is hidden by default in desktop app and visible by default in alternative app?

I'm on it.

fredj:
> For me it's relay strange when I want to change the projection and suddenly the coordinates dispars...

It's how it's done in ol3: http://openlayers.org/en/latest/examples/mouse-position.html

ger-benjamin:
@sbrunner :
> specific project data, => should be in the apps

Some others configurations must be in apps i think. I'll open an issue with that after closing this one.

@pgiraud 
> Can we make it so the info bar is hidden by default in desktop app and visible by default in alternative app?

It seems that defaults values in the controller are not take in count in the apps concerning the state of our panels. Try with the tool panel. You can set its value that defines if it should be open or not. Whatever you choose, that has no effect. I'll open a bug with that.

ger-benjamin:
All is done except the issues in the description of this PR and the test by pgiraud.
Examples are up to date.
Thanks all. (Something else ?)

contribs/gmf/apps/desktop/index.html
```diff
@@ -152,6 +192,7 @@
         module.constant('langUrls', langUrls);
         module.constant('authenticationBaseUrl', 'https://geomapfish-demo.camptocamp.net/2.1/wsgi');
         module.constant('fulltextsearchUrl', 'https://geomapfish-demo.camptocamp.net/2.1/wsgi/fulltextsearch?limit=30&partitionlimit=5');
+        module.constant('gmfAltitudeUrl', 'https://geomapfish-demo.camptocamp.net/2.1/wsgi/raster');
```
sbrunner: It can have other attribute than the altitude, then `gmfRasterUrl` will be a better name, see http://sitn.ne.ch/theme/main ...

ger-benjamin: Okay...  we need to discuss about that. Because Altitude service must also take another name. And even the elevation directive should be renamed. 
In all cases, I'll rename the constant directly after this PR (as discussed).

sbrunner:
Looks good for me :-)

Issue #1351, opened 2016-06-08
datepicker images


fredj:
The images for the datepicker are not present in the demo.
https://camptocamp.github.io/ngeo/master/examples/contribs/gmf/apps/desktop_alt

olivierSemet:
@sbrunner 

Does the rule I add for that is not correct?

> .build/examples-hosted/contribs/gmf/build/images/: contribs/gmf/third-party/jquery-ui/images/
	mkdir -p $@
	cp -r $< $@

On my side make examples-hosted adds the missing folder

sbrunner:
It looks looks correct but it don't works ...

fredj:
The images are in the gh-pages branch but there's an extra `images` directory:
https://github.com/camptocamp/ngeo/tree/gh-pages/master/examples/contribs/gmf/build/images/images

PR #1355, opened 2016-06-08
Copy the JQuery UI images to the right folder

sbrunner:
demo: https://camptocamp.github.io/ngeo/jquery-images/examples/contribs/gmf/apps/desktop_alt/?lang=fr&baselayer_ref=map

fix: #1351


fredj:
The same should be done for the other rule in the Makefile (line 381)

sbrunner:
Thanks, done :-)

PR #1367, opened 2016-06-09
Reordering layertree groups enhancement

### LINKED ISSUE #1347: Groups reorder enhancement - dashed zone at the bottom
This is a nice-to-have.
When dragging a group to the bottom, the dashed zone is displayed and then disappear forever. It should stay visible.

olivierSemet:
Close #1347

demo: https://oliviersemet.github.io/ngeo/reordering-layers-enhancement/examples/contribs/gmf/apps/desktop/index.html

@pgiraud @fredj

### DIFF
--- a/src/directives/sortable.js
+++ b/src/directives/sortable.js
@@ -76,6 +76,11 @@ ngeo.sortableDirective = function($timeout) {
            */
           var dragListGroup = null;
 
+          /**
+           * @type {number}
+           */
+          var onDragListHeightRatio = 1.5;
+
           scope.$watchCollection(function() {
             return sortable;
           }, function() {
@@ -88,6 +93,7 @@ ngeo.sortableDirective = function($timeout) {
            */
           function resetUpDragDrop() {
             var children = element.children();
+            var dragHoverClass;
             for (var i = 0; i < children.length; ++i) {
               angular.element(children[i]).data('idx', i);
             }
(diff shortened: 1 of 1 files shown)


src/directives/sortable.js
```diff
@@ -96,9 +102,13 @@ ngeo.sortableDirective = function($timeout) {
...
             }
 
+            if (options['dragHoverClass'] !== undefined) {
+              dragHoverClass = options['dragHoverClass'];
+            }
+
             dragListGroup = new goog.fx.DragListGroup();
             dragListGroup.addDragList(element[0],
-                goog.fx.DragListDirection.DOWN);
+                goog.fx.DragListDirection.DOWN, true, dragHoverClass);
```
pgiraud: Any example of a usage of the `dragHoverClass` options? Looks like it's not used at all.
I'd rather add it if we really need it.

pgiraud:
We shouldn't do this dynamically in javascript. We should be able to do this in pure css. The dragList element should take all the available height.
I'll work on this one.

src/directives/sortable.js
olivierSemet: No I used it in a former solution. The class name passed to through this option is added to the list when the user start to drag and till the end of the drag event. My first solution was to increase the height of the list with an additional CSS class. I though that this parameter might be useful in the future.

pgiraud:
Superseded by #1375.

PR #1370, opened 2016-06-09
From eslint-config-openlayers version 5 semi is no more needed

sbrunner:
fix: https://github.com/camptocamp/ngeo/pull/1369#issuecomment-224894583


fredj:
LGTM

PR #1372, opened 2016-06-09
Add the role in the print requests

sbrunner:
fix #1182

PR #1378, opened 2016-06-13
Don't recreate the getCapabilities promise

### LINKED ISSUE #1142: print capabilities should not be requested each time the print panel is open
To be discussed.
Since we're using a promise, there should be only one request to the server, IMO.

pgiraud:
Fixes #1142.
Please review.


fredj:
LGTM

fredj:
See also #1372

PR #1401, opened 2016-06-15
Use fs-extra as fs and remove usage of graceful-fs

fredj:
See https://github.com/openlayers/ol3/pull/5151


fredj:
todo: remove https://github.com/camptocamp/c2cgeoportal/blob/master/c2cgeoportal/scripts/import_ngeo_apps.py#L159

sbrunner:
Thanks, merge when Travis is happy :-)

fredj:
Thanks for the review

PR #1411, opened 2016-06-15
Harmonize radius

pgiraud:
Removed all remaining rounded corners.
Demo: https://pgiraud.github.io/ngeo/radius/examples/contribs/gmf/apps/desktop

Actually the radius is configurable. Set to 0 by default.


fredj:
+1

PR #1412, opened 2016-06-15
Replace label by placeholder in feature style form

pgiraud:
Removes the "name" label in the drawing style editor and replaces it by a placeholder.


ger-benjamin:
+1

Issue #1350, opened 2016-06-07
Mobile measures tooltip displayed on top of search results


ybolognini:
When measuring on mobile and then starting a search, measures tooltip stays visible on top of results.

PR #1426, opened 2016-06-17
Define z-index for mobile measures tooltip

marionb:
- place the measures tool-tip on the level of the content (fixes issue #1350)
- [example](https://marionb.github.io/ngeo/mobile-measures-tooltip/examples/contribs/gmf/apps/mobile/?lang=en)


fredj:
Thanks !

PR #1428, opened 2016-06-17
Infobar layout

pgiraud:
With this pull request I improved the look&feel of the info bar.

The not exhaustive list of improvements is:
 - alignments are more consistent,
 - icons where removed because they take a lot of space and are not self explanatory,
 - more space is added around component, this is more airy,
 - info bar don't break layout if screen width is small,
 - background selector position depends on whether the info bar is displayed or not.

screenshot from 2016-06-17 12 00 32

Please also notice that I changed the behavior of the elevation directive. Now we call the service with only one layer at a time. And the elevation is displayed only when the mouse cursor is not moving. No elevation is displayed when the mouse is out of the map.

Démos:
https://pgiraud.github.io/ngeo/infobar_layout/examples/contribs/gmf/apps/desktop
https://pgiraud.github.io/ngeo/infobar_layout/examples/contribs/gmf/apps/desktop_alt

### DIFF
--- a/contribs/gmf/apps/desktop/index.html
+++ b/contribs/gmf/apps/desktop/index.html
@@ -93,7 +93,7 @@
           </div>
         </div>
       </div>
-      <div class="map-container">
+      <div class="map-container" ng-class="{'infobar-active': mainCtrl.showInfobar}">
         <gmf-search gmf-search-map="mainCtrl.map"
           class="search"
           gmf-search-datasources="mainCtrl.searchDatasources"
@@ -134,7 +134,7 @@
                     gmf-elevation-map="mainCtrl.map">
               <a type="button" class="btn btn-default dropdown-toggle btn-sm" data-toggle="dropdown" aria-expanded="false">
                 <span class="elevation-value">
-                  {{elevationValue | number}}<span ng-show="elevationValue">m</span>
+                  {{elevationValue | number}}<span ng-show="elevationValue"> m</span>
                   <span ng-show="elevationLoading" class="fa fa-spinner"></span>
                   <span ng-show="!elevationValue && !elevationLoading" translate>Elevation…</span>
                 </span><span class="caret"></span>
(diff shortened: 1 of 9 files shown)


sbrunner:
https://github.com/pgiraud/ngeo/blob/0d4b4a97981a06bd0acef0c7a9fedbc10201312b/contribs/gmf/apps/desktop/js/controller.js#L74
No more used?
I found stage to display "Elevation" when we wait for the elevation at least it shoudn't ends with "..." =>  "Elevation..."?

pgiraud:
> https://github.com/pgiraud/ngeo/blob/0d4b4a97981a06bd0acef0c7a9fedbc10201312b/contribs/gmf/apps/desktop/js/controller.js#L74
> No more used?

It is still used. For the list of available layers in the dropdown for elevation.



> I found stage to display "Elevation" when we wait for the elevation at least it shoudn't ends with "..." => "Elevation..."?

Changed.

sbrunner:
The result looks good, but having 60 lines to include the infobar is too much...
We should put more code in the component!
But this can be handle in an other pull request...

tonio:
Minor, but the scale caret is not aligned the same way as elevation & coordinates…

pgiraud:
> Minor, but the scale caret is not aligned the same way as elevation & coordinates…

Fixed.

contribs/gmf/apps/desktop/index.html
```diff
@@ -118,40 +118,42 @@
...
-                    <a href ng-click="mainCtrl.selectedElevationLayer = elevationItem">
-                      <span class="fa fa-fw" ng-class="{'fa-check': mainCtrl.selectedElevationLayer === elevationItem}"></span>
-                      {{::elevationItem}}
-                    </a>
-                  </li>
-                </ul>
-              </div>
+              <a type="button" class="btn btn-default dropdown-toggle btn-sm" data-toggle="dropdown" aria-expanded="false">
+                <span class="elevation-value">
+                  {{elevationValue | number}}<span ng-show="elevationValue">m</span>
```
fredj: Include a space between the number and the `m` ?

fredj:
LGTM

PR #1439, opened 2016-06-20
Add XSDAttribute format

adube:
This PR introduces the `ngeo.format.XSDAttribute` format, which allows the reading of attribute definitions in XSD and convert them into a list of attribute definitions, i.e. `Array.<ngeox.Attribute>`.

This PR introduces an example which is at a very "early" state. In the up-coming PRs, it will be updated to use the `ngeo-attributes` directive.  Don't mind it too much for now.

## Todo

 * [x] review 
 * [x] apply reviewer's comments / corrections

## Live example

 * http://adube.github.io/ngeo/xsd-format/examples/attributes.html


adube:
@fredj Ready for review.

adube:
Will do.

adube:
I left the one in the constructor, though.

fredj:
I'm now not so sure about how to determine if an attribute is required or not (`nillable` vs `minOccurs`) ...

After reading some random stackoverflow posts, I think that we should use `minOccurs > 0` for `required`.

@sbrunner what do you think ? Do we have some use cases ?

src/ol-ext/format/xsdattribute.js
```diff
@@ -0,0 +1,146 @@
...
+  var nillable = node.getAttribute('nillable');
+  var required = !(nillable === true || nillable === 'true');
+
+  var attribute = {
+    name: name,
+    required: required
+  };
+
+  var type = node.getAttribute('type');
+  if (type) {
```
fredj: The type is required (in `ngeox.Attribute`); please remove this test and add a `goog.asserts.assert` just like you did for `name`

adube: Aye aye, I'm on it.

src/ol-ext/format/xsdattribute.js
```diff
@@ -0,0 +1,146 @@
...
+      var choices = [];
+      for (var i = 0, ii = enumerations.length; i < ii; i++) {
+        choices.push(enumerations[i].getAttribute('value'));
+      }
+      attribute.choices = choices;
+    } else {
+      return null;
+    }
+  }
+
```
fredj: please add `goog.asserts.assert(attribute.type);` here.

src/ol-ext/format/xsdattribute.js
adube: No wait, it's not always defined.  Yeah, you saw it too...

sbrunner:
+1

fredj:
@sbrunner do you know any project when an attribute is required? to know if we must use `nillable` or `minOccurs`

sbrunner:
I think that we actually don't manage that...

fredj:
ok, let's keep `nillable`, we can change it later

PR #1440, opened 2016-06-20
Add ngeo-attributes directive

adube:
This PR introduces the `ngeo-attributes` directive, which produces a form to edit the properties of a feature.

This PR doesn't include a date picker. It should come in the next PR.

## Todo

 * [x] wait for #1439 to be merged
 * [x] review
 * [x] apply reviewer's required corrections

## Live example

 * http://adube.github.io/ngeo/ngeo-attributes/examples/attributes.html

### DIFF
--- a/examples/attributes.html
+++ b/examples/attributes.html
@@ -8,9 +8,6 @@
     <meta name="mobile-web-app-capable" content="yes">
     <link rel="stylesheet" href="../node_modules/openlayers/css/ol.css" type="text/css">
     <link rel="stylesheet" href="../node_modules/bootstrap/dist/css/bootstrap.css" type="text/css">
-    <style>
-
-    </style>
   </head>
   <body ng-controller="MainController as ctrl">
 
--- a/examples/attributes.html
+++ b/examples/attributes.html
@@ -7,10 +7,40 @@
           content="initial-scale=1.0, user-scalable=no, width=device-width">
     <meta name="mobile-web-app-capable" content="yes">
     <link rel="stylesheet" href="../node_modules/openlayers/css/ol.css" type="text/css">
+    <link rel="stylesheet" href="../node_modules/bootstrap/dist/css/bootstrap.css" type="text/css">
+    <style>
+
+    </style>
   </head>
   <body ng-controller="MainController as ctrl">
-    <p id="desc">This example shows how to use the <code>ngeo-attributes</code> directive.</p>
-    <code id="list-of-attributes"></code>
+
+    <div class="container">
+      <div class="row">
+        <div class="col-sm-6">
+          <p id="desc">
+            This example shows how to use the <code>ngeo-attributes</code>
+            directive, which binds a form to the inner properties of a
+            <code>ol.Feature</code> object. Changes made through the form are
+            applied to the feature and vice-versa. Try playing with the form
+            or click
+            <input
+                type="button"
+                ng-click="ctrl.updateName()"
+                value="Change name" >
+            </input>
+            to change the name.
+          </p>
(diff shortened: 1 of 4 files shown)


adube:
You can still see the approach I took and validate if I'm on the right track...

adube:
Ready for review.

adube:
@fredj Ready for review.

fredj:
reviewing now

examples/attributes.html
```diff
@@ -7,10 +7,40 @@
           content="initial-scale=1.0, user-scalable=no, width=device-width">
     <meta name="mobile-web-app-capable" content="yes">
     <link rel="stylesheet" href="../node_modules/openlayers/css/ol.css" type="text/css">
+    <link rel="stylesheet" href="../node_modules/bootstrap/dist/css/bootstrap.css" type="text/css">
+    <style>
```
fredj: `<style>` not used, please remove

src/directives/attributes.js
```diff
@@ -0,0 +1,149 @@
...
+   * above does the opposite, i.e. it listens to the feature inner properties
+   * changes and apply them to the form. To prevent circular issues, while
+   * applying changes coming from the form, this flag is set. While set, changes
+   * from the feature inner properties are ignored.
+   * @type {boolean}
+   * @private
+   */
+  this.updating_ = false;
+
+  $scope.$on('$destroy', this.handleDestroy_.bind(this));
```
fredj: :+1:

src/directives/attributes.js
```diff
@@ -0,0 +1,149 @@
...
+
+};
+
+
+/**
+ * Called when an input node value changes
+ * @param {string} name Attribute name
+ */
```
fredj: `@export` annotation is probably missing (the function is used in the html template)

src/directives/attributes.js
```diff
@@ -0,0 +1,149 @@
...
+
+goog.require('ngeo');
+goog.require('ngeo.EventHelper');
+
+
+/**
+ * Directive used to render the attributes of a feature into a form.
+ * Example:
+ *
+ * Todo...
```
fredj: to be done ...

src/directives/attributes.js
```diff
@@ -0,0 +1,149 @@
...
+   * The list of attributes to create the form with.
+   * @type {Array.<ngeox.Attribute>}
+   * @export
+   */
+  this.attributes;
+
+  /**
+   * The feature containing the values.
+   * @type {ol.Feature}
+   * @export
```
fredj: `@export` probably not needed

adube:
Thank you very much for the review.  I'll apply the required corrections.

src/directives/attributes.js
adube: I think it is.  `this.feature` gets initialized by the `bindToController`, so unless we export `this.feature`, it would be renamed and the rest of the code would fail.

Something we could do is make the inner property private:

```
 /**
  * The feature containing the values.
  * @type {ol.Feature}
  * @private
  */
this.feature_ = this['feature'];
```

but then again the inner property "feature" gets set already... so why not use it ?

Bottom-line: I suggest we leave this as-is.  What do you think ?

adube:
Fixed.  Ready for merge (if you agree with my previous comments).

src/directives/attributes.js
fredj: You're right, I didn't thought about `bindToController`

adube:
@fredj Could this be merged soon ?

fredj:
thanks @adube

PR #1444, opened 2016-06-21
Uniformize font size

pgiraud:
With this pull request, font size gets uniformized. A smaller font is used everywhere.
Class names prefixed with "-sm" should be avoided as much as possible from now on.

Demo: https://pgiraud.github.io/ngeo/font-size-harmonization/examples/contribs/gmf/apps/desktop

### DIFF
--- a/contribs/gmf/less/desktop.less
+++ b/contribs/gmf/less/desktop.less
@@ -18,9 +18,12 @@
 @left-panel-width: 32rem;
 @right-panel-width: 28rem;
 @topbar-height: 4.5rem;
-@toolbar-width: 4.5rem;
 @border-color: darken(@brand-primary, @standard-variation);
 @search-width: 8 * @map-tools-size;
+@font-size-base: 13px;
+@padding-base-vertical: 5px;
+@padding-base-horizontal: 10px;
+@form-group-margin-bottom: 10px;
 
 html, body {
   position: relative;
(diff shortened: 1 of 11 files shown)


contribs/gmf/less/desktop.less
```diff
@@ -20,6 +20,10 @@
 @toolbar-width: 4.5rem;
 @border-color: darken(@brand-primary, @standard-variation);
 @search-width: 8 * @map-tools-size;
+@font-size-base: 13px;
+@padding-base-vertical: 5px;
+@padding-base-horizontal: 10px;
+@form-group-margin-bottom: 10px;
```
sbrunner: I think that we should use rem to be directly compatible with Bootstrap 4...

pgiraud: Bootstrap stable release is not intended to be released sometimes soon. Let's keep pixels for now to match Bootstrap 3 for now. It should be easy to change this when we migrate to 4.

sbrunner: So we make the migration harder :-/

pgiraud: It will be probably not so much work to change those variables compared to the other required modifications when upgrading.

sbrunner: I don't agree, everything we can directly to is better!

pgiraud: In bootstrap 4, they are still using a pixel value for the font-size of the html element. But they will introduce a new `font-size-root` which doesn't exist yet.
https://github.com/twbs/bootstrap/blob/v4-dev/scss/_variables.scss#L166-L172

Also variables for the padding have completely changed in v4. They will called "spacers". The current code won't be relevant in v4 so I'm not really sure we should pay to much attention to it right now.

pgiraud:
As discussed with @ybolognini this can be merged.
px vs. rem considerations can be addressed later if really needed.

PR #1445, opened 2016-06-21
Position of bg selector, disclaimer, features window relative to info…

pgiraud:
… bar

With this pull request background selector button, disclaimer and query results window are positioned relatively to the info bar. They move if the info bar is opened. And the disclaimer are positioned next to the background selector.

screenshot from 2016-06-21 17 10 42
screenshot from 2016-06-21 17 10 30

Demo: https://pgiraud.github.io/ngeo/disclaimer/examples/contribs/gmf/apps/desktop


fredj:
+1, thanks

PR #1446, opened 2016-06-22
Add gmf-mobile-nav-back-on-click directive

### LINKED ISSUE #1344: Mobile: selecting currently selected theme doesn't close themes panel
On mobile
 * select a theme
 * remove a group
 * open theme selector
 * click on the same theme

--> removed group is added, which is good, but selector doesn't close (it should). This works on desktop.

fredj:
And use it to close the gmf-themeselector on mobile.

Fixes #1344

### DIFF
--- a/contribs/gmf/src/directives/mobilenav.js
+++ b/contribs/gmf/src/directives/mobilenav.js
@@ -286,3 +286,44 @@ gmf.mobileNavBackDirective = function() {
 };
 
 gmf.module.directive('gmfMobileNavBack', gmf.mobileNavBackDirective);
+
+
+/**
+ * A directive to be used in conjunction with {@link gmf.mobileNavDirective}.
+ * The directive can be set on a slide element of {@link gmf.mobileNavDirective}.
+ * When the element is clicked, the navigation returns to the previous slide if
+ * the slide is currently active.
+ *
+ * Example:
+ *
+ *    <nav class="nav-left" gmf-mobile-nav>
+ *      ...
+ *      <gmf-themeselector
+ *         gmf-mobile-nav-back-on-click
+ *         gmf-themeselector-currenttheme="mainCtrl.theme">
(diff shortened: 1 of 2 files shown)


contribs/gmf/src/directives/mobilenav.js
```diff
@@ -286,3 +286,44 @@ gmf.mobileNavBackDirective = function() {
...
+ *
+ * Example:
+ *
+ *    <nav class="nav-left" gmf-mobile-nav>
+ *      ...
+ *      <gmf-themeselector
+ *         gmf-mobile-nav-back-on-click
+ *         gmf-themeselector-currenttheme="mainCtrl.theme">
+ *      </gmf-themeselector>
+ *
```
sbrunner: No `athmAttributes`?

fredj: This directive has no html attributes

pgiraud:
This is a pretty elegant workaround.
+1

contribs/gmf/src/directives/mobilenav.js
sbrunner: Sorry I just see the attributes from the theme selector, add ..., forget my comment :-)

fredj:
Thanks for the review

pgiraud:
@fredj did you check if there are still cases where we want to keep the previous `gmf-mobile-nav-back` directive?

fredj:
yes, `gmf-mobile-nav-back` is not used anymore, do you think we should remove it ?

pgiraud:
If it's not used, I'm in favour to remove it. It adds unrequired complexity. This "back-click" directive is much more easy to use and understand and is useful in many contexts.

fredj:
ok, and the code is in the git history if needed

fredj:
I will create a PR

PR #1456, opened 2016-06-23
Convert string value for `m` in to a boolean

### LINKED ISSUE #1418: Drawings - Display length/area not checked correctly
When a user draws a line/polygon and chooses to display the length/area and then reloads the page, the label is correctly displayed on the map. However the checkbox in the style editor isn't checked.

screenshot from 2016-06-16 09 05 48

marionb:
The string value used for the feature attribute showMeasure is now turned in to a boolean.
(fixes #1418)

### DIFF
--- a/contribs/gmf/src/services/permalink.js
+++ b/contribs/gmf/src/services/permalink.js
@@ -234,7 +234,21 @@ gmf.Permalink = function($timeout, ngeoBackgroundLayerMgr, ngeoDebounce,
    */
   this.featureHashFormat_ = new ngeo.format.FeatureHash({
     setStyle: false,
-    encodeStyles: false
+    encodeStyles: false,
+    propertiesType: {
+      'fillColor': ngeo.FeatureProperties.COLOR,
+      'fillOpacity': ngeo.FeatureProperties.OPACITY,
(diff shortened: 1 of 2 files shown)


sbrunner:
This is a big hack, the property should be declared as boolean...
And this part
https://github.com/marionb/ngeo/blob/8436fb54701698ac515bb4555fd5993c8889aa83/src/ol-ext/format/featurehash.js#L58-L69
shouldn't be in this class, but in the redligning ...

sbrunner:
Add something like a `propertiesType` here: https://github.com/camptocamp/ngeo/blob/6039de98f964223b63ecbd57a45dbfee5d69e2db/contribs/gmf/src/services/permalink.js#L237

sbrunner:
Exellent, tha last thing to do is to replace that:
https://github.com/marionb/ngeo/blob/f9d860e13771d16a48cb9432ca4afe59ebf73b3e/src/ol-ext/format/featurehash.js#L57-L70:
by: 
```js
ngeo.format.FeatureHashLegacyProperties_ = {}
```
Then it it's working and Travis tests pass for me it's ok to merge :-)

PR #1482, opened 2016-06-27
Fixing some permalink issues

olivierSemet:
For now:

- The permalink is not fully generated during the initialization phase and therefore does not reflect the state from the localStorage one. 

 - - [ ] This should be fixed now.

- The permalink initialization does not take care of mixed groups inside a mixed group 

 - - [ ] This should be fixed now.

Demo: https://oliviersemet.github.io/ngeo/permalink-enhancement/examples/contribs/gmf/apps/desktop_alt/

**Empty you localStorage and reload before testing the app** 

@pgiraud @fredj @sbrunner Could you review and check ?

### DIFF
--- a/contribs/gmf/src/directives/layertree.js
+++ b/contribs/gmf/src/directives/layertree.js
@@ -712,14 +712,10 @@ gmf.LayertreeController.prototype.updateWMSLayerState_ = function(layer,
   } else {
     layer.setVisible(true);
     var source = /** @type {ol.source.ImageWMS} */ (layer.getSource());
-    var prevNames = source.getParams()['LAYERS'];
-    var prevTime = source.getParams()['TIME'];
-    if (names !== prevNames || opt_time !== prevTime) {
-      if (opt_time) {
-        source.updateParams({'LAYERS': names, 'TIME' : opt_time});
-      } else {
-        source.updateParams({'LAYERS': names});
-      }
+    if (opt_time) {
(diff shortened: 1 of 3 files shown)


sbrunner:
The generated permalink is wrong:
when I enable the layer `osm_open` I see:
`tree_enable_osm_open=true`
but it should be (see: https://github.com/camptocamp/c2cgeoportal/issues/1657):
`tree_group_layers_osm_function=osm_open,...`
This should work at least on reading, If it's better to change it you should add the gorup name =>
`tree_enable_osm_function_osm_open=true`...

sbrunner:
I just see that's was wrong before ...

olivierSemet:
Not I sure I understood:

> tree_enable_<layer_name>. Is the layer enabled for external layers (present only if it's not the default)

Is the localStorage/url queryParameterkey key used for external layers like osm_open (it's in mixed group). So regarding the spec [#1657](https://github.com/camptocamp/c2cgeoportal/issues/1657) it seems to be good.

> tree_group_layers_<group_name>: List of visible layers (present only if it's not the default)


Is the localStorage/url queryParameterkey key used for internal layers like those containing in "groups" for example

olivierSemet:
The link for the demo was not the good one. It's good now

sbrunner:
Sorry, I miss that's was a mixed group :-)

olivierSemet:
We can close this PR, I'll produce a global one. Thanks

PR #1483, opened 2016-06-28
Generate template cache for components

gberaudo:
The key for components is prefix/component/partial.html


sbrunner:
Looks good :-)

PR #1486, opened 2016-06-29
Add parameter showNotification to determin when to call notifyCantAddGroups_

### LINKED ISSUE #1385: The message "group already added" should never be display
When we get a search result with a geometry and an add_layer.

marionb:
fixes #1385

### DIFF
--- a/contribs/gmf/src/directives/search.js
+++ b/contribs/gmf/src/directives/search.js
@@ -751,7 +751,7 @@ gmf.SearchController.select_ = function(event, feature, dataset) {
             groupActions.indexOf('add_layer') >= 0) {
         // FIXME: Set the layer visible again (Issue also in the
         // treemanager service).
-        this.gmfTreeManager_.addGroupByLayerName(actionData, true, false);
+        this.gmfTreeManager_.addGroupByLayerName(actionData, true, true);
       }
     }
   }
--- a/contribs/gmf/src/directives/search.js
+++ b/contribs/gmf/src/directives/search.js
@@ -751,7 +751,7 @@ gmf.SearchController.select_ = function(event, feature, dataset) {
             groupActions.indexOf('add_layer') >= 0) {
         // FIXME: Set the layer visible again (Issue also in the
         // treemanager service).
-        this.gmfTreeManager_.addGroupByLayerName(actionData, true);
+        this.gmfTreeManager_.addGroupByLayerName(actionData, true, false);
       }
     }
   }
--- a/contribs/gmf/src/directives/search.js
+++ b/contribs/gmf/src/directives/search.js
@@ -736,6 +736,8 @@ gmf.SearchController.prototype.blur = function() {
  */
 gmf.SearchController.select_ = function(event, feature, dataset) {
   var actions = feature.get('actions');
+  var featureGeometry = /** @type {ol.geom.SimpleGeometry} */
+      (feature.getGeometry());
   if (actions) {
     var groupActions = /** @type {Array.<string>} */ (
         this.datasources_[0].groupActions);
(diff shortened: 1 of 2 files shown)


contribs/gmf/src/services/treemanager.js
```diff
@@ -114,10 +114,12 @@ gmf.TreeManager.prototype.addTheme = function(theme) {
  * already in the tree.
  * @param{Array.<GmfThemesNode>} groups An array of gmf theme nodes.
  * @param{boolean=} opt_add if true, force to use the 'add' mode this time.
+ * @param{boolean=} showNotification if false notifyCantAddGroups_ is not called
  * @export
  */
-gmf.TreeManager.prototype.addGroups = function(groups, opt_add) {
+gmf.TreeManager.prototype.addGroups = function(groups, opt_add, showNotification) {
```
tsauerwein: Optional parameters should be prefixed with `opt_`.

contribs/gmf/src/services/treemanager.js
```diff
@@ -114,10 +114,12 @@ gmf.TreeManager.prototype.addTheme = function(theme) {
...
   var groupNotAdded = [];
+  showNotification = typeof showNotification !== 'undefined' ? showNotification : true;
```
tsauerwein: This could be:

     showNotification = showNotification !== 'undefined' ? showNotification : true;

contribs/gmf/src/services/treemanager.js
```diff
@@ -126,7 +128,7 @@ gmf.TreeManager.prototype.addGroups = function(groups, opt_add) {
       groupNotAdded.push(group);
     }
   }.bind(this));
-  if (groupNotAdded.length > 0) {
+  if (groupNotAdded.length > 0 && showNotification === true) {
```
tsauerwein: We usually just do: `if (... & showNotification)`

pgiraud:
Instead of a `showNotification`, I'd rather be in favour of using a parameter that is false by default. `opt_silent` makes more sense to me.
See for example https://github.com/openlayers/ol3/blob/3935b7db241f47236841616775af0309fd917b0d/src/ol/object.js#L197

This way you would have less code to change. You wouldn't need to convert the parameter to a boolean.
```
if (groupNotAdded.length > 0 && !opt_silent)
```

contribs/gmf/src/directives/search.js
```diff
@@ -751,7 +751,7 @@ gmf.SearchController.select_ = function(event, feature, dataset) {
             groupActions.indexOf('add_layer') >= 0) {
         // FIXME: Set the layer visible again (Issue also in the
         // treemanager service).
-        this.gmfTreeManager_.addGroupByLayerName(actionData, true);
+        this.gmfTreeManager_.addGroupByLayerName(actionData, true, false);
```
sbrunner: It should be `false` only if the geometry is defined :-)

contribs/gmf/src/services/treemanager.js
```diff
@@ -126,7 +127,7 @@ gmf.TreeManager.prototype.addGroups = function(groups, opt_add) {
       groupNotAdded.push(group);
     }
   }.bind(this));
-  if (groupNotAdded.length > 0) {
+  if (groupNotAdded.length > 0 & !opt_silent) {
```
pgiraud: &&

contribs/gmf/src/directives/search.js
```diff
@@ -751,7 +751,7 @@ gmf.SearchController.select_ = function(event, feature, dataset) {
             groupActions.indexOf('add_layer') >= 0) {
         // FIXME: Set the layer visible again (Issue also in the
         // treemanager service).
-        this.gmfTreeManager_.addGroupByLayerName(actionData, true);
+        this.gmfTreeManager_.addGroupByLayerName(actionData, true, true);
```
sbrunner: ... it should be conditional ...

sbrunner:
Looks good to me :-)

PR #1359, opened 2016-06-08
Displaying loader when source is not available for a layer

olivierSemet:
closes #1114 
closes #1057 

demo:
- desktop app: https://oliviersemet.github.io/ngeo/source-not-available-wmts-layer/examples/contribs/gmf/apps/desktop/index.html

- mobile app: https://oliviersemet.github.io/ngeo/source-not-available-wmts-layer/examples/contribs/gmf/apps/mobile/index.html

@fredj @pgiraud

PR #1487, opened 2016-06-29
Remove specific style for no-source

### LINKED ISSUE #1057: "Source not available" displayed temporarily
In the desktop app, a "Source not available" appears furtively in layertree for the last two layers when loading the cadastre theme.

### LINKED ISSUE #1114: Don't put string in the less files
It's not possible to localized them
Found error:
https://github.com/camptocamp/ngeo/blob/master/contribs/gmf/less/layertree.less#L156

pgiraud:
ie. for when the WMTS capabilities are loading

Supersedes #1359
Fixes #1114
Fixes #1057 

After discussion with @ybolognini, we decided to fix this issue by simply not showing anything to the user when the capabilities are loading. The capabilities request doesn't last long enough to require a loader to be displayed.
Managing this in a more elegant way is a nice to have but requires more work.

Please review.

### DIFF
--- a/contribs/gmf/less/layertree.less
+++ b/contribs/gmf/less/layertree.less
@@ -139,13 +139,6 @@
     }
   }
 
-  .no-source {
-    opacity: 0.3;
-    &::after {
-      content: "(source not available)";
-    }
-  }
-
   .out-of-resolution {
     .name {
       font-style: italic;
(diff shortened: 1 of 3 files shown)


fredj:
LGTM

pgiraud:
Thanks for the review.

PR #1488, opened 2016-06-29
Fix query service

### LINKED ISSUE #1392: Query on layers that aren't selected
When a layer is deselected, query is still possible and gives results.

fredj:
Fixes #1392 

@adube, @ger-benjamin I need a review on this.

To reproduce the issue:
 - https://camptocamp.github.io/ngeo/master/examples/contribs/gmf/apps/desktop
 - select "Learning" theme
 - Unselect the "Bus stop" node
 - query the map: the `LAYERS` and `QUERY_LAYERS` in the `GetFeatureInfo` is `bus_stop,information`

### DIFF
--- a/contribs/gmf/src/services/querymanager.js
+++ b/contribs/gmf/src/services/querymanager.js
@@ -128,8 +128,9 @@ gmf.QueryManager.prototype.createSources_ = function(node) {
           }
         }, this);
         layers = childLayerNames.join(',');
-      } else if (node.type === 'WMS') {
-        validateLayerParams = true;
+        if (node.type === 'WMS' && childLayerNames.length == 1) {
+          validateLayerParams = true;
+        }
       }
 
       var source = {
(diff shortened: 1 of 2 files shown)


sbrunner:
Can it be on 2.0?

fredj:
> Can it be on 2.0?

yes, I will backport it for 2.0 if this fix is correct

sbrunner:
OK :-)

ger-benjamin:
Nice catch and seems to work fine (But I don't have worked on this part of the code)

fredj:
Thanks for the review

contribs/gmf/src/services/querymanager.js
```diff
@@ -128,8 +128,9 @@ gmf.QueryManager.prototype.createSources_ = function(node) {
           }
         }, this);
         layers = childLayerNames.join(',');
-      } else if (node.type === 'WMS') {
-        validateLayerParams = true;
+        if (node.type === 'WMS' && childLayerNames.length == 1) {
+          validateLayerParams = true;
+        }
```
adube: I hadn't noticed that nodes with childLayers could have the 'WMS' type.  The logic still works here, so it's fine.

adube:
+1 for me.

fredj:
#1490

PR #1494, opened 2016-06-29
Add ngeo-createfeature directive

adube:
This PR introduces the `ngeo-createfeature` directive, which allows the creation of vector features (point, line, polygon) using draw or measure interaction from a map.  Drawn features are added to a collection of features.

It is similar to the `ngeo-drawfeature` directive, but has the following things different:

 * it only support drawing one type of geometry at a time (`ngeo-drawfeature` allows 6 types: point, line, polygon, text, circle, rectangle)
 * it doesn't set any style to the feature
 * it will be used for the purpose of editing features from real layers, while the `ngeo-drawfeature` only draw "sketchs" that are not saved

## Todo

 * [ ] review

## Live example

 * http://adube.github.io/ngeo/ngeo-createfeature/examples/createfeature.html

### DIFF
--- a/examples/createfeature.html
+++ b/examples/createfeature.html
@@ -1,0 +1,124 @@
+<!DOCTYPE html>
+<html ng-app='app'>
+  <head>
+    <title>Create Feature Example</title>
+    <meta charset="utf-8">
+    <meta name="viewport"
+          content="initial-scale=1.0, user-scalable=no, width=device-width">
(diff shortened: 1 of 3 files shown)


adube:
@fredj Ready for review.

fredj:
looks very good, thanks

adube:
Thanks!

Issue #985, opened 2016-04-05
Change background layer on theme change


fredj:
On theme change, the background layer could also change, `functionalities.default_basemap` must be checked.

PR #1495, opened 2016-06-29
Change the background on theme change

fredj:


sbrunner:
Looks good :-)

fredj:
There is some duplicated code with the block bellow (change background depending to the user); it needs to be refactored. This can be done later

fredj:
fixes #985

Issue #390, opened 2015-11-16
Keep bindToController with scope: true directives when needed


fgravin:
Just a post to discuss about Angular bindToController feature.

Regarding: https://github.com/camptocamp/ngeo/blame/master/src/directives/layertree.js#L144

For example, I think whole section

    var nodeExpr = $attrs['ngeoLayertree'];
    this.node = undefined;
    $scope.$watch(nodeExpr, goog.bind(function(newVal, oldVal) {
        this.node = newVal;
    }, this));

Could be replaced by 
    
    bindToController : { node: '=ngeoLayertree'}

in the directive definition.
We might consider this option for future devs.

You want to use `scope: true` for your directives, but think of using `bindToController: {}` syntax to preserve the power of the `Controller as` feature and to keep bindings with parent scope wihout isolated scope definitions.

fredj:
+1
See also http://toddmotto.com/no-scope-soup-bind-to-controller-angularjs/

fredj:
It's sad that the angular's documentation does not mention it ...

fredj:
In the [angular externs](https://github.com/google/closure-compiler/blob/master/contrib/externs/angular-1.4.js#L434), `bindToController` can be `boolean|!Object<string, string>` so we are able to use this syntax

fredj:
Angular 1.5.0 now officially support this. see: http://angularjs.blogspot.ch/2016/02/angular-150-ennoblement-facilitation.html

ybolognini:
What's the status?

ger-benjamin:
We already use that is some directive, like here : https://github.com/camptocamp/ngeo/blob/master/src/directives/colorpicker.js (see `colors` attributes (both in the directive and in the controller) and the bindToController statement (in the directive).)

But obviously not in old directives (made before AngularJS 1.5) like: https://github.com/camptocamp/ngeo/blob/master/contribs/gmf/src/directives/search.js (see the `clearbutton` for instance).

I've update the guidelines. See: #1501 Can I close this with this PR ?

PR #1501, opened 2016-06-30
Update guidelines

ger-benjamin:
Fix: #390

### DIFF
--- a/docs/guidelines.md
+++ b/docs/guidelines.md
@@ -106,17 +106,46 @@ ngeo.exampleDirective = function(…) {
   return {
     restrict: 'A',
     scope: {
-      'm': '=ngeoExampleMap'
+      map: '=ngeoExampleMap'
     }
     controller: function() {
-      var m = this['m'];
+      var m = this['map'];
+      // Then, for Closure-Compiler, assert and type this value.
       // …
     },
     controllerAs: 'ctrl',
-    bindToController: true,
     // …
   });
 ```
+
+But if you bind your directive to the controller, you must
+use the `.` notation. See the example below :
+
+```js
+ngeo.exampleDirective = function(…) {
+  return {
+    restrict: 'A',
+    scope: {
+      map: '=ngeoExampleMap'
+    }
+    bindToController: true,
+    controller: function() {
+      /**
+       * This value will always be the value of 'ngeoExampleMap' and vice versa.
+       * The '@type …' and the '@export' below are only for Closure-Compiler.
+       * @type {ol.Map}
+       * @export
+       */
+      this.map;
+      // …
+    },
+    controllerAs: 'ctrl',
+    // …
+  });
+```
+
+Note also that if you use the `&` binding, you will still get a function
+instead of your ol.Map but directly in the scope of your controller.
 
 ## Directive name or attributes names in the DOM
 
@@ -145,7 +174,7 @@ And in the directive, a scope like this:
 ngeo.mobileMyexampleDirective = function(…) {
   return {
     scope: {
-      't': '=ngeoMobileMyexampleTitle'
+      t: '=ngeoMobileMyexampleTitle'
     },
     // …
   };


docs/guidelines.md
```diff
@@ -113,7 +113,30 @@ ngeo.exampleDirective = function(…) {
       // …
     },
     controllerAs: 'ctrl',
+    // …
+  });
+```
+
+But if you bind your directive to the controller and you declare your attribute
+as a two ways binding attrribute, you can use the `.` notation. See the example
```
tsauerwein: > you can use the `.` notation

Are you sure this works correctly with the Closure Compiler?

ger-benjamin: Yes. For instance, like here : https://github.com/camptocamp/ngeo/blob/master/src/directives/colorpicker.js (see colors attributes (both in the directive and in the controller) and the bindToController statement (in the directive).)

fredj: `binding attribute,` instead of `binding attrribute`

tsauerwein: How does the compiler know that the properties defined in `scope` relate to the instance properties? I think it is by chance that it works for the example you have linked to.

tsauerwein: More or less *by chance*. My guess is that the properties in `scope` are not renamed because there are other properties with the same name that are exported or in an externs file. @gberaudo, what do you think?

gberaudo: Thanks @ger-benjamin for having a stab at this section. I discussed it with @sbrunner and we wanted to clarify it.

What you propose is almost the good way. You need to change it to:
```
/**
  @type {ol.Map}
  @export
*/
this.m;
```

The above code `declares` a typed variable called `m` and asks closure not to rename it.
It is then possible to access the variable using the dotted notation AND closure will check all access are valid.

Now, why you code worked?
Closure rename the variables even if their size is optimal (m is already 1 character).
So the only explanation I see would be that there exists somewhere a type with an attribute named `m`.
That is a bit surprising though.

@ger-benjamin, could you change to the code above and remove the part about using the bracket notation?

gberaudo: @tsauerwein, I think the code was missing `bindToController: true`. There is no way the property in the scope can be accessed through the controller without this.

tsauerwein: Also note that in the [colorpicker code](https://github.com/camptocamp/ngeo/blob/master/src/directives/colorpicker.js#L39-L41) the properties for `scope` are not given as strings.

ger-benjamin: I haven't added the 
``
/**
  @type {ol.Map}
  @export
*/
``

part because it was not existing in the previous example. This part concern AngularJS, not Closure at all. But I can add it in both part and explain that.

And I use `m` because it was as is in the previous example but I'll rename it as `map` that will be clearer. 

>Also note that in the colorpicker code the properties for scope are not given as strings.

Thanks for that. (Even if that works, that's not (longer?) necessary)

sbrunner:
@tsauerwein Is it ok for you?

docs/guidelines.md
```diff
@@ -106,18 +106,47 @@ ngeo.exampleDirective = function(…) {
...
+      var m = this['map'];
+      // Then, for Closure-Compiler, assert and type this value.
       // …
     },
     controllerAs: 'ctrl',
+    // …
+  });
+```
+
+But if you bind your directive to the controller, you will can
```
gberaudo: `you will can` -> `you must`

docs/guidelines.md
```diff
@@ -106,18 +106,47 @@ ngeo.exampleDirective = function(…) {
...
+ngeo.exampleDirective = function(…) {
+  return {
+    restrict: 'A',
+    scope: {
+      map: '=ngeoExampleMap'
+    }
     bindToController: true,
+    controller: function() {
+      /**
+       * This value will be always the value of 'ngeoExampleMap' and vice versa.
```
gberaudo: will always be

docs/guidelines.md
```diff
@@ -106,18 +106,47 @@ ngeo.exampleDirective = function(…) {
...
+       * The two '@' under are only for Closure-Compiler.
```
gberaudo: The `@export` below is [...]

docs/guidelines.md
```diff
@@ -106,18 +106,47 @@ ngeo.exampleDirective = function(…) {
...
+       */
+      this.map;
+      // …
+    },
+    controllerAs: 'ctrl',
     // …
   });
 ```
 
+Note also that if you use the `&` binding, you still will get a function
```
gberaudo: you will still

fredj:
@tsauerwein please merge this pull request if you're happy with these changes

PR #1505, opened 2016-06-30
Fix gmf EditFeature service header

adube:
This PR fixes the gmf.EditFeature service to use the correct header in the POST, PUT and DELETE requests it sends.

## Todo

 * [ ] solve the `goog is not defined` error in the example
 * [ ] review

## Live example

 * https://adube.github.io/ngeo/fix-gmf-editfeature-service/examples/contribs/gmf/editfeature.html


adube:
Mmm, not sure why but I get a `goog is not defined` in the example due the the assertions that are made.  I thought those were supposed to be automatically removed.

What is the way to fix this ?

fredj:
Look like our examples doesn't support `goog.asserts`. Only `goog.provide` and `goog.require` are [stripped](https://github.com/camptocamp/ngeo/blob/master/Makefile#L536).

What about replacing `goog.asserts` with `console.assert`?

PR #1512, opened 2016-07-01
Add tooltip to search color picker

### LINKED ISSUE #1420: Add tooltip to the color chooser button and popover
See 665b68e

The tooltip content is not translated; `{{'Change the color of the search result'|translate}}` is displayed instead.

pgiraud:
Fixes #1420.
`data-original-title` is to be used instead of `data-title`.

### DIFF
--- a/contribs/gmf/src/directives/partials/search.html
+++ b/contribs/gmf/src/directives/partials/search.html
@@ -11,7 +11,9 @@
   </span>
 </div>
 <span ng-show="ctrl.colorchooser && ctrl.displayColorPicker" ngeo-popover ngeo-popover-placement="bottom">
-  <button class="color-button fa fa-tint" ngeo-popover-anchor></button>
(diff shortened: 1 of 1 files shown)


pgiraud:
Actually, I think that the header in the popover is enough. There's no real need for a tooltip for this tool.
Opinions?

fredj:
I also think that the popover header is enough

pgiraud:
Unfortunately the tooltip header doesn't show up in built mode. I'll investigate.

pgiraud:
For an unknown reason the text I put in the tooltip is also used as a title for the popover. I changed the pull request so that no tooltip appears but the title in the popover is kept.
https://pgiraud.github.io/ngeo/search_color_tooltip/examples/contribs/gmf/apps/desktop_alt

fredj:
+1, thanks

PR #1516, opened 2016-07-04
Use console.assert instead of goog.asserts.assert in editfeature example

fredj:
See https://github.com/camptocamp/ngeo/pull/1505#issuecomment-230225282


pgiraud:

fredj:
Thanks for the review

PR #1523, opened 2016-07-04
Mobile login nav

### LINKED ISSUE #1450: mobile login and nav state
https://camptocamp.github.io/ngeo/master/examples/contribs/gmf/apps/mobile

Open the right nav, select login and login. The right nav closes.
When the rigth nav is opened again, the login panel is displayed; IMHO the "main" menu should be displayed instead (the menu with measure and login).

It's the same when a background layer is selected

What do you think ?

pgiraud:
Those changes allow to nav back to main menu in right navigation when the user is successfully logged in.

Fixes #1450.

I first reverted 079cd0cd5634e0c9ab6a1227f87da6d2ad9f832d to reintroduce `mobileNavBackDirective` but changed it so that it takes a boolean expression.

I also had to change the scope for the authentication directive. I'm not super happy with this change but couldn't find any better way.

Please review.

### DIFF
--- a/contribs/gmf/src/directives/mobilenav.js
+++ b/contribs/gmf/src/directives/mobilenav.js
@@ -243,6 +243,52 @@ gmf.MobileNavController.prototype.backIfActive = function(element) {
 
 /**
  * A directive to be used in conjunction with {@link gmf.mobileNavDirective}.
+ * The directive can be set on a slide element of {@link gmf.mobileNavDirective}
+ * with an expression. When the value of the expression changes and becomes
+ * true, the navigation returns to the previous slide, if the slide is
+ * currently active.
+ *
+ * Example:
+ *
+ *    <nav class="nav-left" gmf-mobile-nav>
+ *      ...
+ *      <gmf-authentication class="slide"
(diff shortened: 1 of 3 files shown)


contribs/gmf/src/directives/mobilenav.js
```diff
@@ -243,6 +243,56 @@ gmf.MobileNavController.prototype.backIfActive = function(element) {
...
+ * Example:
+ *
+ *    <nav class="nav-left" gmf-mobile-nav>
+ *      ...
+ *      <gmf-themeselector
+ *         id="themes"
+ *         class="slide"
+ *         data-header-title="Themes"
+ *         gmf-themeselector-currenttheme="mainCtrl.theme"
+ *         gmf-mobile-nav-back="mainCtrl.theme">
```
fredj: Should be changed to something like `gmf-mobile-nav-back="mainCtrl.theme.name !== 'OSM'"` (boolean value)

pgiraud:
Changed the doc.

sbrunner:
Looks good merge :-)

PR #1525, opened 2016-07-05
Use the WMTS dimensions to build the source

### LINKED ISSUE #1524: Use the WMTS dimensions
The WMTS nodes may contains a `dimensions` object.
This object should be used when the `ol.source.WMTS` instance is created.

fredj:
fixes #1524


sbrunner:
Looks good to me :-)

PR #1532, opened 2016-07-06
Display an error when the themes contains some error

sbrunner:
Partially fix https://github.com/camptocamp/c2cgeoportal/issues/2361

### DIFF
--- a/contribs/gmf/src/services/themesservice.js
+++ b/contribs/gmf/src/services/themesservice.js
@@ -60,7 +60,6 @@ gmf.Themes = function($http, $injector, $q, ngeoLayerHelper, gettextCatalog) {
    * @private
    */
   this.treeUrl_ = null;
-
   if ($injector.has('gmfTreeUrl')) {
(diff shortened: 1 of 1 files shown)


contribs/gmf/src/services/themesservice.js
```diff
@@ -349,8 +357,16 @@ gmf.Themes.prototype.loadThemes = function(opt_roleId) {
     cache: false,
     withCredentials: true
   }).then(function(response) {
+    if (response.data['errors'].length != 0) {
+      var message = 'The themes contains some errors:\n' +
+        response.data['errors'].join('\n');
+      console.error(message);
+      if (this.ngeoLocation_ !== null && this.ngeoLocation_.hasParam('debug')) {
+        window.alert(message);
```
fredj: What about always using `console.error` ?

sbrunner: To always easily view that we have some errors in the theme..., why not?

fredj:
please merge

PR #1537, opened 2016-07-07
Make the ngeo.Location service available to child classes

fredj:
I need to access the service from a child class


sbrunner:
Looks good :-)

fredj:
Thanks for the review

Issue #781, opened 2016-03-01
Query open by default


ybolognini:
We had a talk during one PSC meeting about if query result window should be expanded or not by default. Question wasn't fully resolved, it looks like for some of you it should be expanded and for some not.

We propose to add a configuration variable to decide if result window should be expanded or not after a query.

ybolognini:
ping @rbovard @gnerred @svarim @vvmalbre @mriedo @gsollberger @svavoi @cedricmoullet @kalbermattenm

Please read description and validate

gnerred:
where is description ?

ybolognini:
@gnerred  Sorry I mean the first comment, the description of the issue

gnerred:
ok sorry, I thought there was more detailed documentation. ok for me

rbovard:
Okay for me.

svavoi:
Okay for us.

vvmalbre:
Ok for us.

PR #1543, opened 2016-07-08
Add configuration variable to decide if the query result window is open or closed.

marionb:
fixes issue #781.

(Example: https://marionb.github.io/ngeo/default_open_query/examples/contribs/gmf/apps/mobile/index.html)


ger-benjamin:
Ok for me, just ensure you that the state must not be reinitialized with the defaultcollapsed value when the dispalquerywindow is closed (hidden). If the current behavior is correct, please merge.

PR #1545, opened 2016-07-08
Fix click event listener in ngeo-recenter

fredj:
Otherwise this code doesn't work:
```
<button ngeo-extent="[-1898084, 4676723, 3972279, 8590299]" translate>Move to Europe</button>
```


sbrunner:
Looks good :-)

fredj:
Thanks for the review

PR #1557, opened 2016-07-11
Move datepicker from gmf to ngeo

jlap:
This commits moves the datepicker directive from gmf to ngeo.

The wmstime service was splitted in two for this. Now we have the time service in ngeo, and the wmstime service in gmf. The wms time service has the formatting functions, while the time service has simple option setting functions and and utc utility function.

For all of this to work, jquery-ui needs to be a dependency of ngeo instead of gmf, and angular ui-date too. So the third-party folder was moved, and I had to tweak the Makefile a bit for it to work.

Also the ui-date library was included globally in gmf, so that's what I did here too, and that's the bulk of the file changes we see here.

Here the link to the original datepicker example, I kept it since the format is a bit different. It'll visually work exactly as it used to, but the example uses the new ngeo datepicker instead of the gmf one that was deleted:
http://jlap.github.io/ngeo/datepicker/examples/contribs/gmf/datepicker.html

And here's and example right in ngeo:
http://jlap.github.io/ngeo/datepicker/examples/datepicker.html

A nice addition could be to add date formatting functions to the ngeo.time class, but I think that might be out of scope for the ticket.


jlap:
Oh forgot to notify you, @fredj this is ready for review !

sbrunner:
For me that's looks good but it should be rebased...

jlap:
It's rebased now!

sbrunner:
Thanks :-)

PR #1561, opened 2016-07-12
Avoid querying identical sources Fix #1535

### LINKED ISSUE #1535: If two identical layers are on the map, query doesn't work
Steps to reproduce:
- Open OSM theme
- Active at least two same layers in different groups. Like "osm_time_r_dp" in groupe "Layers" and groupe osm_function.
- Do a query on "osm_time_r_dp"
- Navigate between results (animation bug and errors in your console)

From where the bug come ?
 - In the ngeo query component, the component count all features from all layers source but keep and share only distinct sources (no multiple sources with the same id). The total number of features don't match with the reality. Difficulty, in this component, new sources (from others groups) can come later and the number of feature will grow even ifb these new sources will be not really added.
 - In the "displayquerywindow" component, the next/previous function rely on the counted number of features

olivierSemet:
Related to #1535 

**Demo**
- https://oliviersemet.github.io/ngeo/avoiding-identical-request/examples/contribs/gmf/apps/desktop/

@ger-benjamin @sbrunner please could you have a look and tell me if this is a step in the right direction?

### DIFF
--- a/src/services/query.js
+++ b/src/services/query.js
@@ -430,7 +430,7 @@ ngeo.Query.prototype.getQueryableSources_ = function(map, wfsOnly) {
         if (!wfsItemsByUrl[url]) {
           wfsItemsByUrl[url] = [];
         }
-        wfsItemsByUrl[url].push(item);
+        this.pushSourceIfUnique_(item, wfsItemsByUrl[url]);
       } else if (!wfsOnly) {
         // use WMF GetFeatureInfo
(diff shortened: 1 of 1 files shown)


sbrunner:
I don't know if there is a better way to have a unique list , otherwise that's looks good :-)
@ger-benjamin merge if that's looks good for you :-)

ger-benjamin:
Looks good to me.

PR #1562, opened 2016-07-12
Use the current target in slide-in click handler

fredj:
Otherwise the following doesn't work:
```
<a href data-toggle="slide-in" translate>foo</a>
```
The `translate` directive adds and extra `span` and it's this span that respond
to the click event.

### DIFF
--- a/contribs/gmf/src/directives/mobilenav.js
+++ b/contribs/gmf/src/directives/mobilenav.js
@@ -128,7 +128,7 @@ gmf.MobileNavController.prototype.init = function(element) {
   element.find('[data-toggle=slide-in]').on('click', function(evt) {
 
     // the element to slide out is the div.slide parent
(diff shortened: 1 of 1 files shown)


fredj:
If we have:
```
<a href data-toggle="slide-in">foo</a>
```
In the click handle: `evt.target ` is the `a` and  `evt.currentTarget ` is the `a`.

And if we have:
```
<a href data-toggle="slide-in" translate>foo</a>
```
Transformed by the translate directive to:
```
<a href data-toggle="slide-in" ><span>foo</span></a>
```

In the click handle: `evt.target ` is the `span` and  `evt.currentTarget ` is the `a`.

pgiraud:
Excellent!

fredj:
Thanks for the review

PR #1569, opened 2016-07-14
Don't use node 0.10 anymore

sbrunner:


.travis.yml
```diff
@@ -17,7 +17,7 @@ addons:
 
 node_js:
 - "4"
-- "0.10"
+- "6"
```
fredj: version 5 instead of 6 ?

fredj:
LGTM

.travis.yml
sbrunner: For me the version 5 is a development version ...

PR #1573, opened 2016-07-14
Display azimuth of circles in the redlining tools

jlap:
This commit is a work in progress for the following ticket: https://github.com/camptocamp/c2cgeoportal/issues/2258

This change replaces the Display surface with a "Display azimuth and radius" checkbox in the gmf feature style.

When you draw a new circle in the gmf drawfeature, you can now check that new box and the azimuth will be displayed in the corner nearest to the azimuth angle. It'll also display the line of the radius and it's length. It'll look like this:

image

To be able to do this the circle feature had to "remember" where it's radius is, so it's saved as a property of the feature. I tried keeping it as a geometry collection but it caused a ton of problem, notably in the permalink when it tries to hash the feature.

You can test it out here: http://jlap.github.io/ngeo/azimuth/examples/contribs/gmf/drawfeature.html

This is incomplete work, as you cannot modify the azimuth. So if you modify a circle it's radius line will hang being awkwardly. That'll be the next step to complete the ticket.

I am available tomorrow, then I will be unavailable for a couple of weeks (back August 1st), so this PR could be merged as-is, picked up by somebody else, or stay open until I come back, @ybolognini

### DIFF
--- a/contribs/gmf/src/directives/partials/featurestyle.html
+++ b/contribs/gmf/src/directives/partials/featurestyle.html
@@ -119,7 +119,7 @@
           for="gmf-featurestyle-showmeasure"
           class="control-label">
         <span ng-switch-when="Circle">
-          {{'Display surface' | translate}}
+          {{'Display azimuth and radius' | translate}}
         </span>
         <span ng-switch-when="Polygon">
           {{'Display surface' | translate}}
--- a/contribs/gmf/examples/featurestyle.html
+++ b/contribs/gmf/examples/featurestyle.html
@@ -91,6 +91,7 @@
   <body ng-controller="MainController as ctrl">
     <gmf-map gmf-map-map="ctrl.map"></gmf-map>
     <gmf-featurestyle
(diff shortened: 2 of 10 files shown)


jlap:
@fredj This is ready for review!

contribs/gmf/src/directives/partials/featurestyle.html
```diff
@@ -119,7 +119,7 @@
           for="gmf-featurestyle-showmeasure"
           class="control-label">
         <span ng-switch-when="Circle">
-          {{'Display surface' | translate}}
+          {{'Display azimuth and radius' | translate}}
```
sbrunner: `fsCtrl.measure` contains only the azimut, the radius should be added :-)

jlap:
@sbrunner That's right, I just switched it around and should be alright!

sbrunner:
For me that's looks good, I let someone else to review again and merge :-)

ybolognini:
@pgiraud , could we close this PR`?

PR #1585, opened 2016-07-20
Add gmf-editfeature and gmf-editfeatureselector directives

adube:
This is PR introduces the `gmf-editfeature` and `gmf-editfeatureselector` directives. There are still components and features that are missing from the SPECs, but I'm creating this early PR to be able to have the work in progress reviewed.

## Todo

 * [x] review

## Live example

 * https://adube.github.io/ngeo/gmf-editfeature-directive/examples/contribs/gmf/editfeatureselector.html

## Not included

This PR doesn't include the following features.  They will come in separate tickets:

 * refresh the WMS layer after a save has been made
 * measure while drawing
 * save button should be "gray" until modifications are maded
 * geometry modifications
 * validate required fields
 * have an option for the pixel buffer
 * confirm before delete

Also, there's currently a CORS error when deleting.

### DIFF
--- a/contribs/gmf/src/directives/editfeature.js
+++ b/contribs/gmf/src/directives/editfeature.js
@@ -19,7 +19,7 @@ goog.require('ol.format.GeoJSON');
 
 /**
  * Directive used to insert, modify and delete features from a single layer.
- * It allows you can modify the geometry of the feature in addition to its
+ * It allows you to modify the geometry of the feature in addition to its
  * attributes.
  *
  * In order to modify or delete a feature, you must click on the map at the
@@ -36,9 +36,12 @@ goog.require('ol.format.GeoJSON');
  *         gmf-editfeature-vector="::ctrl.vectorLayer">
  *     </gmf-editfeature>
  *
+ * @htmlAttribute {GmfThemesNode} gmf-editfeature-layer The GMF node of the
+ *     editable layer.
  * @htmlAttribute {ol.Map} gmf-editfeature-map The map.
+ * @htmlAttribute {ol.layer.Vector} gmf-editfeature-vector The vector layer in
+ *     which to draw the vector features.
  * @return {angular.Directive} The directive specs.
- * @ngInject
  * @ngdoc directive
  * @ngname gmfEditfeature
  */
@@ -108,7 +111,7 @@ gmf.EditfeatureController = function($scope, $timeout, gmfEditFeature,
 
   /**
    * @type {gmf.EditFeature}
-   * @export
+   * @private
    */
   this.editFeatureService_ = gmfEditFeature;
 
@@ -259,9 +262,7 @@ gmf.EditfeatureController.prototype.delete = function() {
   );
 
   // (2) Reset selected feature
-  this.feature = null;
-  this.features.clear();
-
+  this.cancel();
 };
 
 
@@ -273,9 +274,7 @@ gmf.EditfeatureController.prototype.delete = function() {
 gmf.EditfeatureController.prototype.handleEditFeature_ = function(resp) {
   var features = new ol.format.GeoJSON().readFeatures(resp.data);
   if (features.length) {
-    var feature = features[0];
-    var id = feature.getId();
-    this.feature.setId(id);
+    this.feature.setId(features[0].getId());
   }
 };
 
@@ -292,21 +291,19 @@ gmf.EditfeatureController.prototype.handleDeleteFeature_ = function(resp) {
 
 /**
  * @param {Array.<ngeox.Attribute>} attributes Attributes.
- * @export
+ * @private
  */
 gmf.EditfeatureController.prototype.setAttributes_ = function(attributes) {
-  this.timeout_(function() {
-    // Set attributes
-    this.attributes = attributes;
+  // Set attributes
+  this.attributes = attributes;
 
-    // Get geom type from attributes and set
-    var geomAttr = ngeo.format.XSDAttribute.getGeometryAttribute(
-      this.attributes
-    );
-    if (geomAttr && geomAttr.geomType) {
-      this.geomType = geomAttr.geomType;
-    }
-  }.bind(this), 0);
+  // Get geom type from attributes and set
+  var geomAttr = ngeo.format.XSDAttribute.getGeometryAttribute(
+    this.attributes
+  );
+  if (geomAttr && geomAttr.geomType) {
+    this.geomType = geomAttr.geomType;
+  }
 };
 
 
@@ -356,8 +353,7 @@ gmf.EditfeatureController.prototype.toggle_ = function(active) {
     this.createActive = false;
     //this.modify_.setActive(false);
     this.mapSelectActive = false;
-    this.feature = null;
-    this.features.clear();
+    this.cancel();
   }
 
 };
@@ -402,8 +398,7 @@ gmf.EditfeatureController.prototype.handleMapClick_ = function(evt) {
     this.handleGetFeatures_.bind(this));
 
   // (2) Clear any previously selected feature
-  this.feature = null;
-  this.features.clear();
+  this.cancel();
 
   // (3) Pending
   //this.pending = true;
--- a/contribs/gmf/src/directives/editfeature.js
+++ b/contribs/gmf/src/directives/editfeature.js
@@ -1,0 +1,446 @@
+goog.provide('gmf.EditfeatureController');
+goog.provide('gmf.editfeatureDirective');
+
+
+goog.require('gmf');
+goog.require('gmf.EditFeature');
+goog.require('gmf.XSDAttributes');
+/** @suppress {extraRequire} */
+goog.require('ngeo.attributesDirective');
+/** @suppress {extraRequire} */
+goog.require('ngeo.btnDirective');
+/** @suppress {extraRequire} */
+goog.require('ngeo.createfeatureDirective');
+goog.require('ngeo.EventHelper');
+goog.require('ngeo.ToolActivate');
+goog.require('ngeo.ToolActivateMgr');
+goog.require('ol.format.GeoJSON');
+
+
+/**
+ * Directive used to insert, modify and delete features from a single layer.
+ * It allows you can modify the geometry of the feature in addition to its
+ * attributes.
+ *
+ * In order to modify or delete a feature, you must click on the map at the
+ * location of the feature to select it first.
+ *
+ * In order to create a new feature, you use the "Draw" button and digitalize
+ * the feature on the map.
+ *
+ * Example:
+ *
+ *     <gmf-editfeature
+ *         gmf-editfeature-layer="::ctrl.layer">
+ *         gmf-editfeature-map="::ctrl.map"
+ *         gmf-editfeature-vector="::ctrl.vectorLayer">
+ *     </gmf-editfeature>
+ *
+ * @htmlAttribute {ol.Map} gmf-editfeature-map The map.
+ * @return {angular.Directive} The directive specs.
+ * @ngInject
+ * @ngdoc directive
+ * @ngname gmfEditfeature
+ */
+gmf.editfeatureDirective = function() {
+  return {
+    controller: 'GmfEditfeatureController',
+    scope: {
+      'layer': '=gmfEditfeatureLayer',
+      'map': '<gmfEditfeatureMap',
+      'vectorLayer': '<gmfEditfeatureVector'
+    },
+    bindToController: true,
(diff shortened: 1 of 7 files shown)


adube:
@ger-benjamin This PR is now ready for review. You'll notice in the example that there's a lot of things missing from the SPECs. I listed them in the description above. I'll work on each one of them individually on separate PRs to make this one smaller.

After saving, you can refresh the page to see if your modifications. Also, note that the geometry can't be edited yet.

ger-benjamin:
@adube I'll be on holiday in ~one hour. So please ask... (who @ybolognini)  ?

adube:
@ger-benjamin Understood.  I'll contactt @ybolognini by e-mail.

adube:
@pgiraud Hi Pierre.  Would you please take care of reviewing this PR ?  Thanks.

contribs/gmf/src/directives/editfeature.js
```diff
@@ -0,0 +1,446 @@
...
+goog.require('ngeo.createfeatureDirective');
+goog.require('ngeo.EventHelper');
+goog.require('ngeo.ToolActivate');
+goog.require('ngeo.ToolActivateMgr');
+goog.require('ol.format.GeoJSON');
+
+
+/**
+ * Directive used to insert, modify and delete features from a single layer.
+ * It allows you can modify the geometry of the feature in addition to its
```
pgiraud: It allows you "to"

contribs/gmf/src/directives/editfeature.js
```diff
@@ -0,0 +1,446 @@
...
+ *
+ * Example:
+ *
+ *     <gmf-editfeature
+ *         gmf-editfeature-layer="::ctrl.layer">
+ *         gmf-editfeature-map="::ctrl.map"
+ *         gmf-editfeature-vector="::ctrl.vectorLayer">
+ *     </gmf-editfeature>
+ *
+ * @htmlAttribute {ol.Map} gmf-editfeature-map The map.
```
pgiraud: There are missing @htmlAttribute

contribs/gmf/src/directives/editfeature.js
```diff
@@ -0,0 +1,446 @@
...
+ * @return {angular.Directive} The directive specs.
+ * @ngInject
```
pgiraud: I don't think you need @ngInject here.

contribs/gmf/src/directives/editfeature.js
```diff
@@ -0,0 +1,446 @@
...
+ * @ngname GmfEditfeatureController
+ */
+gmf.EditfeatureController = function($scope, $timeout, gmfEditFeature,
+    gmfXSDAttributes, ngeoEventHelper, ngeoToolActivateMgr) {
+
+  /**
+   * @type {GmfThemesNode}
+   * @export
+   */
+  this.layer;
```
pgiraud: Do you really need to export this?
I don't see it used in the partial.

contribs/gmf/src/directives/editfeature.js
```diff
@@ -0,0 +1,446 @@
...
+   * @type {ol.Map}
+   * @export
+   */
+  this.map;
+
+  /**
+   * @type {ol.layer.Vector}
+   * @export
+   */
+  this.vectorLayer;
```
pgiraud: Do you really need to export this?
I don't see it used in the partial.

contribs/gmf/src/directives/editfeature.js
```diff
@@ -0,0 +1,446 @@
...
+   * @type {GmfThemesNode}
+   * @export
+   */
+  this.layer;
+
+  /**
+   * @type {ol.Map}
+   * @export
+   */
+  this.map;
```
pgiraud: Do you really need to export this?
I don't see it used in the partial.

contribs/gmf/src/directives/editfeature.js
```diff
@@ -0,0 +1,446 @@
...
+   * @type {angular.$timeout}
+   * @private
+   */
+  this.timeout_ = $timeout;
+
+  /**
+   * @type {gmf.EditFeature}
+   * @export
+   */
+  this.editFeatureService_ = gmfEditFeature;
```
pgiraud: Do you really need to export this?
I don't see it used in the partial.

pgiraud: @private?

contribs/gmf/src/directives/editfeature.js
```diff
@@ -0,0 +1,446 @@
...
+
+    //toolMgr.unregisterTool(otherUid, this.createToolActivate);
+    //toolMgr.unregisterTool(otherUid, this.modifyToolActivate);
+
+    this.createActive = false;
+    //this.modify_.setActive(false);
+    this.mapSelectActive = false;
+    this.feature = null;
+    this.features.clear();
+  }
```
pgiraud: Please remove the un-necessary comments.

contribs/gmf/src/directives/editfeature.js
```diff
@@ -0,0 +1,446 @@
...
+  this.editFeatureService_.deleteFeature(
+    this.layer.id,
+    this.feature
+  ).then(
+    this.handleDeleteFeature_.bind(this)
+  );
+
+  // (2) Reset selected feature
+  this.feature = null;
+  this.features.clear();
```
pgiraud: Why not calling `cancel` instead?

contribs/gmf/src/directives/editfeature.js
```diff
@@ -0,0 +1,446 @@
...
+ * Called after an insert, update or delete request.
+ * @param {angular.$http.Response} resp Ajax response.
+ * @private
+ */
+gmf.EditfeatureController.prototype.handleEditFeature_ = function(resp) {
+  var features = new ol.format.GeoJSON().readFeatures(resp.data);
+  if (features.length) {
+    var feature = features[0];
+    var id = feature.getId();
+    this.feature.setId(id);
```
pgiraud: Can't this be simplified to `this.feature.setId(features[0].getId())`?

contribs/gmf/src/directives/editfeature.js
```diff
@@ -0,0 +1,446 @@
...
+ * @private
+ */
+gmf.EditfeatureController.prototype.handleDeleteFeature_ = function(resp) {
+  console.log('ok, deleted');
+};
+
+
+/**
+ * @param {Array.<ngeox.Attribute>} attributes Attributes.
+ * @export
```
pgiraud: @private?

contribs/gmf/src/directives/editfeature.js
```diff
@@ -0,0 +1,446 @@
...
+    this.attributes = attributes;
+
+    // Get geom type from attributes and set
+    var geomAttr = ngeo.format.XSDAttribute.getGeometryAttribute(
+      this.attributes
+    );
+    if (geomAttr && geomAttr.geomType) {
+      this.geomType = geomAttr.geomType;
+    }
+  }.bind(this), 0);
```
pgiraud: Why is a timeout needed here?

contribs/gmf/src/directives/editfeature.js
```diff
@@ -0,0 +1,446 @@
...
+    toolMgr.unregisterTool(createUid, this.mapSelectToolActivate);
+
+    //toolMgr.unregisterTool(otherUid, this.createToolActivate);
+    //toolMgr.unregisterTool(otherUid, this.modifyToolActivate);
+
+    this.createActive = false;
+    //this.modify_.setActive(false);
+    this.mapSelectActive = false;
+    this.feature = null;
+    this.features.clear();
```
pgiraud: Why not calling `cancel` instead?

contribs/gmf/src/directives/editfeature.js
```diff
@@ -0,0 +1,446 @@
...
+
+  this.editFeatureService_.getFeatures([this.layer.id], extent).then(
+    this.handleGetFeatures_.bind(this));
+
+  // (2) Clear any previously selected feature
+  this.feature = null;
+  this.features.clear();
+
+  // (3) Pending
+  //this.pending = true;
```
pgiraud: Unrequired comment.

contribs/gmf/src/directives/editfeatureselector.js
```diff
@@ -0,0 +1,184 @@
...
+ *         gmf-editfeatureselector-vector="::ctrl.vectorLayer">
+ *     </gmf-editfeatureselector>
+ *
+ * @htmlAttribute {boolean} gmf-editfeatureselector-active Whether the
+ *     directive is active or not.
+ * @htmlAttribute {ol.Map} gmf-editfeatureselector-map The map.
+ * @htmlAttribute {ol.layer.Vector} gmf-editfeatureselector-vector The vector
+ *     layer where the selected or created features are drawn.
+ * @return {angular.Directive} The directive specs.
+ * @ngInject
```
pgiraud: @ngInject may not be required here.

contribs/gmf/src/directives/editfeatureselector.js
```diff
@@ -0,0 +1,184 @@
...
+    this.selectedLayer = value;
+  }
+  return this.layers;
+};
+
+
+/**
+ * @param {GmfThemesNode} node A theme, group or layer node.
+ * @param {Array.<GmfThemesNode>} nodes An Array of nodes.
+ * @export
```
pgiraud: @private?

contribs/gmf/src/directives/editfeature.js
adube: We don't need to export it.  It is the property that is assigned automatically by the directive `scope` and `bindTocontroller` (see above). However, if we don't, the Closure compiler would rename it and therefore the assignement would not work properly.  The other way to avoid having this marked as  `@export` would be to do something like this:

```
this.layer_ = this['layer'];
```

and mark it as `@private`.  But then again, the property would be set twice, which is pretty useless.  This is something I have already explained to Fred a couple weeks ago and he agreed with leaving this "as-is".

What do you think ?

contribs/gmf/src/directives/editfeature.js
adube: See previous comment.

contribs/gmf/src/directives/editfeature.js
adube: In the other 7 PRs, I use those lines, i.e. they get un-commented.  I'll leave this like this.

contribs/gmf/src/directives/editfeature.js
pgiraud: Makes perfect sense. I forgot this case. Thanks for the explanation.

contribs/gmf/src/directives/editfeature.js
adube: I'll make the change.

I usually make the code as "human-readable" as possible.  I.e. it's easier to read the code with 3 lines instead of this one liner.  I do so because, in the end, everything gets simplified and minimized by the Compiler anyway.  What I mean is: the change doesn't really matter in the end.  Thoughts ?

contribs/gmf/src/directives/editfeature.js
adube: I can't remember why I did that.  I removed it and it still works fine.

contribs/gmf/src/directives/editfeature.js
pgiraud: I like when code is easy to read too. But in this case, in my opinion, the code isn't really easier to read with more lines.
You can leave it as is though.

contribs/gmf/src/directives/editfeature.js
adube: Same.  I'm using it in a further PR.  I'll leave this "as-is".

contribs/gmf/src/directives/editfeature.js
pgiraud: There are other places where you use $timeout.
Please make sure that you do it on purpose.

contribs/gmf/src/directives/editfeature.js
adube: I already changed it.  I agree that it's not that hard to read. I'll leave this changed.

contribs/gmf/src/directives/editfeature.js
adube: The other places serve a purpose, I'm sure of this.

adube:
@pgiraud Thank you very much for the review.  I made the suggested fixes and updated the example.  This is ready for merge.

PR #1589, opened 2016-07-21
Enable search layer

### LINKED ISSUE #1220: Search - Add layer don't enable the layer
* Search for Police
* Select it in Add layer

=>

* The group is added as expected
* The layer Police isn't enable, he should be
* The layer Fuel is enable, he shouldn't

marionb:
Fixes #1220 

Note: In  [themesservice.js](https://github.com/marionb/ngeo/blob/d677bebe1a8b2d2e29b2cf1163dddf14d8077533/contribs/gmf/src/services/themesservice.js#L128) I changed the use of `layer.layers` to `layer.name` to be able to also use this for WMTS layers. Is there a particular reason that we should continue using `layer.layers`?

### DIFF
--- a/contribs/gmf/src/services/themesservice.js
+++ b/contribs/gmf/src/services/themesservice.js
@@ -125,7 +125,7 @@ gmf.Themes.findGroupByLayerName = function(themes, name) {
       var group = theme.children[j];
       for (var k = 0, kk = group.children.length; k < kk; k++) {
         var layer = group.children[k];
-        if (layer.layers == name) {
+        if (layer.name == name) {
           return group;
         }
       }
--- a/contribs/gmf/src/services/treemanager.js
+++ b/contribs/gmf/src/services/treemanager.js
@@ -24,7 +24,9 @@ gmf.module.value('gmfTreeManagerModeFlush', true);
  * This service's theme is a GmfThemesNode with only children and a name.
  * Thought to be the tree source of the gmf layertree directive.
  * @constructor
+ * @param {angular.$timeout} $timeout Angular timeout service.
  * @param {angularGettext.Catalog} gettextCatalog Gettext catalog.
+ * @param {ngeo.LayerHelper} ngeoLayerHelper Ngeo Layer Helper.
  * @param {ngeo.Notification} ngeoNotification Ngeo notification service.
  * @param {gmf.Themes} gmfThemes gmf Themes service.
  * @param {boolean} gmfTreeManagerModeFlush Flush mode active?
@@ -33,14 +35,26 @@ gmf.module.value('gmfTreeManagerModeFlush', true);
  * @ngdoc service
  * @ngname gmfTreeManager
  */
-gmf.TreeManager = function(gettextCatalog, ngeoNotification, gmfThemes,
+gmf.TreeManager = function($timeout, gettextCatalog, ngeoLayerHelper, ngeoNotification, gmfThemes,
     gmfTreeManagerModeFlush, ngeoStateManager) {
+
+  /**
+   * @type {angular.$timeout}
+   * @private
+   */
+  this.$timeout_ = $timeout;
 
   /**
    * @type {angularGettext.Catalog}
    * @private
    */
   this.gettextCatalog_ = gettextCatalog;
+
+  /**
+   * @type {ngeo.LayerHelper}
+   * @private
+   */
+  this.layerHelper_ = ngeoLayerHelper;
 
   /**
    * @type {ngeo.Notification}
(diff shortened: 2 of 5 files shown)


sbrunner:
Can we have an URL to test?

marionb:
Yes sorry I forgot to add it: 
https://marionb.github.io/ngeo/enable_search_layer/examples/contribs/gmf/apps/desktop_alt/index.html

sbrunner:
The message Layer group already added shouldn't be displayed ...
Also that's looks good :-)

sbrunner:
My last message is wrong ...

contribs/gmf/src/services/themesservice.js
```diff
@@ -125,7 +125,7 @@ gmf.Themes.findGroupByLayerName = function(themes, name) {
       var group = theme.children[j];
       for (var k = 0, kk = group.children.length; k < kk; k++) {
         var layer = group.children[k];
-        if (layer.layers == name) {
+        if (layer.name == name) {
```
sbrunner: @olivierSemet does it looks good?

olivierSemet: Yes but it's a bit confused. First the name of the function is `findGroupByLayerName` but should be `findGroupByNodeName` to avoid confusion. 
`var layer = group.children[k];` IMO should be `var node = group.children[k];`
In this function, it mixed up the concept of node and layer. This is better to talk about node when manipulating themes and groups and to talk about layer when playing with Openlayers layers.

contribs/gmf/src/services/treemanager.js
```diff
@@ -267,20 +283,77 @@ gmf.TreeManager.prototype.addGroupByName = function(groupName, opt_add) {
...
-gmf.TreeManager.prototype.addGroupByLayerName = function(layerName, opt_add, opt_silent) {
+gmf.TreeManager.prototype.addGroupByLayerName = function(layerName, opt_add, opt_silent, opt_map) {
   this.gmfThemes_.getThemesObject().then(function(themes) {
     var group = gmf.Themes.findGroupByLayerName(themes, layerName);
     if (group) {
-      this.addGroups([group], opt_add, opt_silent);
-      // FIXME: set the layer visible
+      var groupAdded = this.addGroups([group], opt_add, opt_silent);
+      if (opt_map) {
+        this.$timeout_(function() {
```
sbrunner: @pgiraud can you have a look if we can avoid this tmeout?

contribs/gmf/src/services/treemanager.js
```diff
@@ -267,20 +283,77 @@ gmf.TreeManager.prototype.addGroupByName = function(groupName, opt_add) {
...
+
+    if (groupAdded) {
+      activeLayers = layerName;
+    } else {
+      var source = /** @type {ol.source.ImageWMS} */ (layerGroup.getSource());
+      activeLayers = source.getParams()['LAYERS'];
+      activeLayers += activeLayers.includes(layerName) ? '' : ',' + layerName;
+    }
+
+    ngeo.LayerHelper.prototype.updateWMSLayerState(layerGroup, activeLayers);
```
pgiraud: You should inject the service instead of calling a prototype method.

contribs/gmf/src/services/treemanager.js
```diff
@@ -267,20 +283,77 @@ gmf.TreeManager.prototype.addGroupByName = function(groupName, opt_add) {
...
     }
   }.bind(this));
 };
 
 
 /**
+ * Make the layer of a group visible if the group is already in the layertree.
+ * If the group has jus been added set only the layer layerName visible.
+ * @param{string} layerName Name of the layer to set visible.
+ * @param{Object} group Group containg the layer to activate.
```
pgiraud: containing

contribs/gmf/src/services/treemanager.js
```diff
@@ -267,20 +283,77 @@ gmf.TreeManager.prototype.addGroupByName = function(groupName, opt_add) {
...
+        }.bind(this));
+      }
     }
   }.bind(this));
 };
 
 
 /**
+ * Make the layer of a group visible if the group is already in the layertree.
+ * If the group has jus been added set only the layer layerName visible.
```
pgiraud: jsut

pgiraud: just

contribs/gmf/src/services/treemanager.js
pgiraud: It's required since the OL Layer is actually created after a new node (group) is added to the layertree and a LayerTreeController is instanciated. This happens in a digest loop so it's not synchronous.
We unfortunately need the OL layer to be create in `setLayerVisible_`. It would be much better if it would be possible to set the state in each tree node, but we can't.

contribs/gmf/src/services/treemanager.js
```diff
@@ -267,20 +283,77 @@ gmf.TreeManager.prototype.addGroupByName = function(groupName, opt_add) {
...
+ * @param{string} layerName Name of the layer to set visible.
+ * @param{Object} group Group containg the layer to activate.
+ * @param{boolean} groupAdded True if the croup has been newly added. False otherwise
```
pgiraud: group

marionb:
I fixed the typos and injection service from the comments of @pgiraud (thanks). Is there anything else that needs to be checked or is it possible to merge?
(new example: https://marionb.github.io/ngeo/enable_search_layer/examples/contribs/gmf/apps/desktop/)

PR #1592, opened 2016-07-21
Simplify createTextStyle by shortening the list of args

pgiraud:
With this pull request I tried to simplify the helper to create a text style.
Currently this method can take 7 arguments (5 are optional).

This drives us to some code looking like this:
https://github.com/camptocamp/ngeo/blob/master/src/services/featurehelper.js#L211-L219

This is ugly in my opinion.
I propose instead to use only on object argument.

Please review.


sbrunner:
Looks good to me :-)

PR #1595, opened 2016-07-21
Display azimuth of circles in the redlining tools

pgiraud:
Supersedes #1573

I reworked @jlap 's pull request to make it simpler (well I hope).
Instead of storing the radius line geometry, only the azimut is kept as a property of the feature.
Also, the placement of the azimut value on the map is computed using the point which is on the circle instead of a complex computation with an offset from the center.
Also the radius value and the position of the azimut update correctly if the circle is modified.
Finally, the radius length displays along the line (ie. oriented).

Demo:
http://pgiraud.github.io/ngeo/azimuth/examples/contribs/gmf/drawfeature.html
https://pgiraud.github.io/ngeo/azimuth/examples/contribs/gmf/apps/desktop

Still missing: update of the azimut value when modifying the circle.

### DIFF
--- a/contribs/gmf/src/directives/partials/featurestyle.html
+++ b/contribs/gmf/src/directives/partials/featurestyle.html
@@ -119,7 +119,7 @@
           for="gmf-featurestyle-showmeasure"
           class="control-label">
         <span ng-switch-when="Circle">
-          {{'Display surface' | translate}}
+          {{'Display azimuth and radius' | translate}}
         </span>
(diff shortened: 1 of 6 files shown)


pgiraud:
Still a work in progress. Doesn't work in built mode.

pgiraud:
Please review.

ybolognini:
Done UI tests. Looks good

sbrunner:
For me thats looks ok :-)

PR #1598, opened 2016-07-21
Fixing theme name and layers displayed not consistent close #1596

### LINKED ISSUE #1596: Theme switch strange
- Go to https://geomapfish-demo.camptocamp.net/2.1
- Select theme enseignement
- Go to https://geomapfish-demo.camptocamp.net/2.1/theme/OSM

We have the theme OSM in the theme selector but not in the Layer tree

olivierSemet:
Fix #1596

This little modification should do the work. Unfortunately I cannot make a full test because adding the theme in the URL doesn't work in local or gh-pages demos.

I can provide a kind of proof:

**Empty your localStorage before test**

1. Go to [Desktop_alt app](https://oliviersemet.github.io/ngeo/fix-theme-name-and-layers-not-consistent/examples/contribs/gmf/apps/desktop_alt/)
2. Select another theme in the list (anything except OSM), it must add some groups to the layer tree.
3. Go to the [Desktop](https://oliviersemet.github.io/ngeo/fix-theme-name-and-layers-not-consistent/examples/contribs/gmf/apps/desktop/) 

You should have the correct name of the theme and the groups in the layer tree which correspond exactly to the selected theme (Therefore the group added previously is not loaded)

@sbrunner  @pgiraud

### DIFF
--- a/contribs/gmf/src/services/treemanager.js
+++ b/contribs/gmf/src/services/treemanager.js
@@ -111,6 +111,7 @@ gmf.TreeManager.prototype.setModeFlush = function(value) {
  */
 gmf.TreeManager.prototype.addTheme = function(theme, opt_init) {
   var firstLevelNodes = theme.children;
+  var groupsNames, firstLevelNodesNames, isSubset;
   var treeGroups = /** @type {string} */ (this.ngeoStateManager_.getInitialValue(
     gmf.PermalinkParam.TREE_GROUPS));
   if (this.isModeFlush()) {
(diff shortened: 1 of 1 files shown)


sbrunner:
Tested here https://gmf-test.sig.cloud.camptocamp.net/sbrunner21
and it looks to works, thanks :-)

olivierSemet:
The problem with this solution is when the app use the mode flush. In this case, the user can't add some groups to a theme but he can delete one (or more) of them. This case won't work anymore. The deleted group(s) will appear again.

olivierSemet:
WIP

olivierSemet:
@sbrunner  here is a new version which taking care of more scenario:

- (flush and add mode) If a user removed all groups, the tree must be empty
- (flush and add mode) If a user removed some groups, the tree must reflect his changes

**Tricky scenario:**

- (flush mode only)  a user work with a theme (e.g `Learning`). If someone send him a permalink containing the theme e.g `OSM` **but not** the `treeGroups` values (scenario of issue #1596):

  - The app state is "unstable", localStorage containing the key `treeGroups` with values from the `learning` theme but the desired theme is `OSM`.

In this case, the `treeGroups` is checked to determine if it's a subset of the theme groups (In our case `OSM`)

- If yes, groups are taken from the localStorage `treeGroups` values
- IF no, `treeGroups` values are ignored and the theme is loaded with its default groups

sbrunner:
Looks good to me :-)

contribs/gmf/src/services/treemanager.js
```diff
@@ -120,11 +121,36 @@ gmf.TreeManager.prototype.addTheme = function(theme, opt_init) {
...
+    }
+    if (!this.isModeFlush()) {
+      //Flush mode, Init phase and state exists -> first level groups must be read
+      //from the stateManager
+      groupsNames.forEach(function(name) {
+        this.addGroupByName(name, true);
+      }, this);
+    } else {
+      //Mode flush, check if treeGroups is a subset of firstLevelNodes.
+      //NB:If the theme in the URL was differente from the theme of the
```
gberaudo: Extra e in `different`.

sbrunner:
@olivierSemet fix the tipo and I will merge it :-)

olivierSemet:
@gberaudo @sbrunner  thanks for review. It fixed.

sbrunner:
@olivierSemet thanks :-)

PR #1620, opened 2016-07-26
Set style before locale is set

### LINKED ISSUE #1619: Draw feature example - style not correct until the feature is selected
When one loads the gmf draw feature, if there are features loaded (from url or local storage), they are not styled (blue default color) until they are selected.
I think it used to work and it currently works in the desktop app though the style is applied with a little delay.

pgiraud:
Fixes #1619.
Please review.
cc @olivierSemet


sbrunner:
Looks good :-)

PR #1622, opened 2016-07-26
Azimut modify

pgiraud:
With this pull request I cleaned the modify circle by removing some code that looks not necessary to me. I actually figured out that the number of features managed by the modify circle interaction was increasing where there should be only the currently selected feature.

But the main goal is to allow modification of the azimut when modifying the circle.

Demo:
https://pgiraud.github.io/ngeo/azimut_modify/examples/contribs/gmf/apps/desktop
https://pgiraud.github.io/ngeo/azimut_modify/examples/contribs/gmf/drawfeature.html

(Draw a circle, check the "Show radius and azimut", then modify the circle)

### DIFF
--- a/contribs/gmf/src/directives/drawfeature.js
+++ b/contribs/gmf/src/directives/drawfeature.js
@@ -312,9 +312,9 @@ gmf.DrawfeatureController = function($scope, $timeout, gettextCatalog,
       return this.selectedFeature;
     }.bind(this),
     function(newFeature, previousFeature) {
+      this.selectedFeatures.clear();
       if (previousFeature) {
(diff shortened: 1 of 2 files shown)


sbrunner:
Looks good to me :-)

ybolognini:
Looks very good to me! :)
A tiny detail: when azimut value is displayed on the left or on the right of the circle, the value almost touches the square. Try with 99/100/101°. 20/30° values are much more readable.

ybolognini:
Fix: https://github.com/camptocamp/c2cgeoportal/issues/2258

PR #1603, opened 2016-07-22
Fix feature selection issues.

olivierSemet:
Fixes #1471 #1579

[...]

Issue #1579, opened 2016-07-18
When deleting a vertex (simple click), the geometry is deselected


ybolognini:
* Edit a redlining polygon
* Simple click on a vertex
* The vertex is deleted but the feature is deselected, it should remain selected

PR #1623, opened 2016-07-26
Fix feature selection

### LINKED ISSUE #1471: Redlining features cannot be selected by clicking on it
When redlining panel is open, clicking on a feature should select it. It used to work...

pgiraud:
Supersedes #1603.

Fixes #1471 & #1579.

It looks like the filtering using "layer" in `forEachFeatureAtPixel` is not really required since we already ensure that the feature is the from the correct collection.
This means that we don't need to give the `drawFeature` directive a layer. The `ngeoFeature` service is already doing the job.
In summary, we were doing too much. Code has been simplified and it continues to work.

Demo
https://pgiraud.github.io/ngeo/fix_feature_selection/examples/contribs/gmf/apps/desktop
https://pgiraud.github.io/ngeo/fix_feature_selection/examples/contribs/gmf/drawfeature.html

### DIFF
--- a/contribs/gmf/examples/drawfeature.html
+++ b/contribs/gmf/examples/drawfeature.html
@@ -197,7 +197,6 @@
     <gmf-drawfeature
       ng-show="ctrl.drawFeatureActive === true"
       gmf-drawfeature-active="ctrl.drawFeatureActive"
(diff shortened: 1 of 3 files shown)


sbrunner:
Looks good :-)

olivierSemet:
Better than #1603 . LGTM.

PR #1624, opened 2016-07-26
Increase azimut label offset

pgiraud:
@ybolognini does it work better for you?
Demo
https://pgiraud.github.io/ngeo/azimut_offset/examples/contribs/gmf/apps/desktop


ybolognini:
I does @pgiraud, thanks

PR #1630, opened 2016-07-27
Add ui-date in ngeo-attributes

adube:
This PR introduces the use of the `ui-date` directive in the `ngeo-attributes` directive for both `date` and `datetime` attribute types. Note that we should definitively use something else for `datetime`, but this should come in an other PR.

## Todo

 * [ ] review

## Live demo

 * https://adube.github.io/ngeo/xsd-datetime/examples/contribs/gmf/editfeatureselector.html


adube:
@pgiraud I think this is ready for review.  I think we should discuss what to do to support a "date - time" picker.  Thoughts ?

pgiraud:
LGTM. Merging.

PR #1636, opened 2016-07-28
Add gmf-editfeature in desktop and desktop_alt apps

adube:
This PR introduces the `gmf-editfeature` directive in the **dektop** and **desktop_alt** applications.

## Todo

 * [x] Issue - both the query and edit tools can be active at the same time. You can see this in action in the desktop app.
 * [x] review

## Live demo

You need to be logged in to see the editing tools.  Also, you need to have the "Editing" theme in order to have layers in the drop down list.

 * (Desktop Alternate) https://adube.github.io/ngeo/gmf-editfeature-app/examples/contribs/gmf/apps/desktop_alt/index.html?baselayer_ref=map&lang=fr&map_x=537500&map_y=152851&map_zoom=3&tree_groups=Edit&tree_group_layers_Edit=line%2Cpolygon%2Cpoint

 * (Desktop) https://adube.github.io/ngeo/gmf-editfeature-app/examples/contribs/gmf/apps/desktop/

### DIFF
--- a/contribs/gmf/src/controllers/abstractdesktop.js
+++ b/contribs/gmf/src/controllers/abstractdesktop.js
@@ -158,70 +158,9 @@ gmf.AbstractDesktopController = function(config, $scope, $injector) {
       features: new ol.Collection()
     }),
     style: function(feature, resolution) {
-      // (1) Style definition depends on geometry type
-      var white = [255, 255, 255, 1];
-      var blue = [0, 153, 255, 1];
-      var width = 3;
-      var styles = [];
-
-      var geom = feature.getGeometry();
-      console.assert(geom);
-      var type = geom.getType();
-
-      if (type === ol.geom.GeometryType.POINT) {
-        styles.push(
-          new ol.style.Style({
-            image: new ol.style.Circle({
-              radius: width * 2,
-              fill: new ol.style.Fill({
-                color: blue
-              }),
-              stroke: new ol.style.Stroke({
-                color: white,
-                width: width / 2
-              })
-            }),
-            zIndex: Infinity
-          })
-        );
-      } else {
-        if (type === ol.geom.GeometryType.LINE_STRING) {
-          styles.push(
-            new ol.style.Style({
-              stroke: new ol.style.Stroke({
-                color: white,
-                width: width + 2
-              })
-            })
-          );
-          styles.push(
-            new ol.style.Style({
-              stroke: new ol.style.Stroke({
-                color: blue,
-                width: width
-              })
-            })
-          );
-        } else {
-          styles.push(
-            new ol.style.Style({
-              stroke: new ol.style.Stroke({
-                color: blue,
-                width: width / 2
-              }),
-              fill: new ol.style.Fill({
-                color: [255, 255, 255, 0.5]
-              })
-            })
-          );
-        }
-
-        // (2) Anything else than 'Point' requires the vertex style as well
-        styles.push(ngeoFeatureHelper.getVertexStyle(true));
-      }
-
-      return styles;
-    }.bind(this)
+      return ngeoFeatureHelper.createEditingStyles(feature);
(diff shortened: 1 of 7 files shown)


adube:
@pgiraud Ready for review.

pgiraud:
I wonder if it wouldn't be better to hide the editing button in the right bar until the user is logged in.

contribs/gmf/src/controllers/abstractdesktop.js
```diff
@@ -129,6 +146,94 @@ gmf.AbstractDesktopController = function(config, $scope, $injector) {
...
+            })
+          );
+        }
+
+        // (2) Anything else than 'Point' requires the vertex style as well
+        styles.push(ngeoFeatureHelper.getVertexStyle(true));
+      }
+
+      return styles;
+    }.bind(this)
```
pgiraud: No need to bind this.

contribs/gmf/src/controllers/abstractdesktop.js
```diff
@@ -129,6 +146,94 @@ gmf.AbstractDesktopController = function(config, $scope, $injector) {
...
+  /**
+   * @type {ol.layer.Vector}
+   * @export
+   */
+  this.editFeatureVectorLayer = new ol.layer.Vector({
+    source: new ol.source.Vector({
+      wrapX: false,
+      features: new ol.Collection()
+    }),
+    style: function(feature, resolution) {
```
pgiraud: Can't we put this method in gmf instead?

pgiraud:
Options in the layer selector should be translated.
The name of the currently edited layer as well.

```patch
--- a/contribs/gmf/src/directives/partials/editfeatureselector.html
+++ b/contribs/gmf/src/directives/partials/editfeatureselector.html
@@ -5,8 +5,8 @@
       ng-switch-when="null"
       ng-model="efsCtrl.getSetLayers"
       ng-model-options="{getterSetter: true}"
-      ng-options="layer.name for layer in efsCtrl.availableLayers">
-    <option value="">-- Choose a layer --</option>
+      ng-options="layer.name | translate for layer in efsCtrl.availableLayers">
+    <option value="" translate>-- Choose a layer --</option>
   </select>
 
   <div ng-switch-default>
@@ -17,7 +17,7 @@
       {{'Stop editing' | translate}}
     </button>
 
-    <h3>{{ efsCtrl.selectedLayer.name }}</h3>
+    <h3>{{ efsCtrl.selectedLayer.name | translate }}</h3>
 
     <gmf-editfeature
```

contribs/gmf/src/controllers/abstractdesktop.js
adube: That's a good idea.  It needs the `ngeoFeatureHelper`, though so I'm not sure where to put it.  Or, we could put it in the `ngeoFeatureHelper`, as one of its methods (i.e. not a static method).

What do you think ?

pgiraud: Something like this https://github.com/camptocamp/ngeo/commit/2ce5c9ea509e12f57f3f44a8ed322ca570e686c1 ?

adube: Yes.  Do you agree ?

pgiraud: I agree.

adube:
All corrections were applied.  Live demo updated.  Ready for merge once Travis is happy.

adube:
>I wonder if it wouldn't be better to hide the editing button in the right bar until the user is logged in.

I haven't done anything regarding this.  I think it's best to have the UI components always there.  At least, it tells users that there's something there and what they need to do to activate it.  Thoughts ?  We could also leave this discussion for an other PR, if needed.

adube:
Travis is happy.

PR #1637, opened 2016-07-28
Add form-control to the layer <select> element

adube:
Adds the `form-control` to the select element of the `gmf-editfeatureselector` directive.

## Todo

 * [ ] There's a small "glitch" when selecting a layer.  The select element stays visible for a short while.  It didn't do that without `form-control`.  What should we do about that ? 

## Live demo

 * https://adube.github.io/ngeo/gmf-editfeature-select-form-control/examples/contribs/gmf/editfeatureselector.html


pgiraud:
Thanks.

PR #1640, opened 2016-07-28
Use 'editable' theme node property as configuration

adube:
This PR uses the `editable` theme node property as configuration for the `gmf-editfeatureselector` directive to determine which layers are editable. Adjust the other examples as well.

In order to see it in action, we need the layer tree.  This requires #1636 to be merged first.

## Todo

 * [x] Wait for #1636 to be merged first
 * [x] Review

## Live example

 * https://adube.github.io/ngeo/gmf-editfeature-editable/examples/contribs/gmf/apps/desktop_alt/index.html?baselayer_ref=map&lang=fr&map_x=537500&map_y=152851&map_zoom=3&tree_groups=Edit&tree_group_layers_Edit=line%2Cpolygon%2Cpoint

### DIFF
--- a/contribs/gmf/src/directives/layertree.js
+++ b/contribs/gmf/src/directives/layertree.js
@@ -987,7 +987,9 @@ gmf.LayertreeController.getLayerNodeIds = function(node, opt_editable) {
   var children = node.children || node;
   if (children && children.length) {
     children.forEach(function(childNode) {
-      ids = ids.concat(gmf.LayertreeController.getLayerNodeIds(childNode));
+      ids = ids.concat(
+        gmf.LayertreeController.getLayerNodeIds(childNode, editable)
+      );
     });
   } else if (node.id !== undefined && (!editable || node.editable)) {
     ids.push(node.id);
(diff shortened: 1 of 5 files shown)


adube:
Rebased onto master, live demo updated and ready for review.

adube:
@pgiraud: This is ready for review.

contribs/gmf/src/directives/layertree.js
```diff
@@ -975,16 +977,19 @@ gmf.LayertreeController.prototype.zoomToResolution = function(treeCtrl) {
...
  * @return {Array.<number|string>} Layer names.
  */
-gmf.LayertreeController.getLayerNodeIds = function(node) {
+gmf.LayertreeController.getLayerNodeIds = function(node, opt_editable) {
+  var editable = opt_editable === true;
   var ids = [];
   var children = node.children || node;
   if (children && children.length) {
     children.forEach(function(childNode) {
       ids = ids.concat(gmf.LayertreeController.getLayerNodeIds(childNode));
```
pgiraud: Don't you want to use opt_editable for the children as well?

adube: I don't understand.  Isn't that already the case ?  If a child node has no children, then we go in `else if` below and they are managed, right ?

pgiraud: When `opt_editable` is true, you want to get the ids of the node which are editable and only them. Calling `getLayerNodeIds` for each child of the current node without the "editable" argument will retrieve the ids of the children even if they're not editable.
I may be wrong but you'll get the editable nodes for the first level but all the nodes for level greater than 2.

adube: Oh, you're right.  I get it now.  I'll fix this right away.

adube:
Fixed, rebased onto master, live example updated and ready for merge.

PR #1643, opened 2016-07-29
Use paint brush icon for redlining

pgiraud:
Demo
https://pgiraud.github.io/ngeo/brush_icon/examples/contribs/gmf/apps/desktop

This lets us use the pencil icon for editing which is more appropriate.


sbrunner:
Looks good :-)

PR #1653, opened 2016-08-02
Move the rateLimitWait option from ngeo to gmf

fredj:


gberaudo:
Why not keep it there?

fredj:
Because this configuration applies to `gmf`, the `ngeo` part must be as generic as possible (no assumption that the remote server can handle this rate)

gberaudo:
OK, I understand now. Please merge.

fredj:
Thanks for the review

PR #1660, opened 2016-08-03
Fix xsd attributes in Firefox

### LINKED ISSUE #1633: ngeo-attributes directive doesn't display anything in firefox
Tested in Firefox 47 on Ubuntu

It looks like the ngRepeat which loops through attributes is empty.

adube:
Fixes #1633.

## Live example:

 * https://adube.github.io/ngeo/fix-xsdattributes-ff/examples/attributes.html

### DIFF
--- a/src/ol-ext/format/xsdattribute.js
+++ b/src/ol-ext/format/xsdattribute.js
@@ -55,6 +55,9 @@ ngeo.format.XSDAttribute.prototype.readFromNode = function(node) {
       'localName should be schema');
 
   var elements = node.getElementsByTagName('element');
+  if (!elements.length) {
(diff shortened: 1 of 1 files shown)


pgiraud:
@fredj What do you think about this? Have you ever heard of problems with FF not reading a document propertly if not given the correct namespace?

fredj:
And what about always using the namespace?

gberaudo:
> And what about always using the namespace?

Is there a requirement that the namespace prefix be `xsd`?

adube:
|And what about always using the namespace?

When using the namespace, it works in FF, but not in Chrome.

PR #1661, opened 2016-08-03
Set the language_in to ECMASCRIPT6_STRICT

fredj:
ping @gberaudo


gberaudo:
+1. This change allow users to choose the language version they like.

fredj:
Thanks for the review

PR #1662, opened 2016-08-03
Examples https

### LINKED ISSUE #1626: Add the following code in the gmf exemples
To avoid CORS error
```
/*
 * Redirect to https
 */
if (window.location.protocol == 'http:' &&
    window.location.port != '3000' &&
    window.location.host != 'localhost') {
  var restOfUrl = window.location.href.substr(5);
  /** @type {Location} */
  window.location = 'https:' + restOfUrl;
}
```

pgiraud:
Fixes #1626 

With this pull request we ensure that hosted examples are redirected to https when run on github.
To make it so, the `https.js` file is loaded in each example in their hosted version. There's only one `https.js` file throughout the application. It is copied adequately to be used both in ngeo, gmf or gmf-apps.

I also changed the `gh-pages` target so that it depends on `examples-hosted`.

Please review.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,3 +14,4 @@
 /contribs/gmf/fonts/FontAwesome*
 /contribs/gmf/fonts/fontawesome*
 /contribs/gmf/cursors/*.cur
+/contribs/gmf/examples/https.js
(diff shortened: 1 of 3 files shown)


pgiraud:
I forgot to mention that the https redirect is required as soon as we need CORS to work correctly.

fredj:
LGTM

PR #1666, opened 2016-08-04
Fix error in gmf-editfeatureselector example

adube:
This PR fixes the gmf-editfeatureselector example because of a missing comment block.

## Live example

 * https://adube.github.io/ngeo/gmf-editfeature-fix-example/examples/contribs/gmf/editfeatureselector.html


adube:
@pgiraud Ready for review.

PR #1668, opened 2016-08-05
Use hasFeatureAtPixel instead of forEachFeatureAtPixel

pgiraud:
Please review.

### DIFF
--- a/src/ol-ext/interaction/modifyrectangle.js
+++ b/src/ol-ext/interaction/modifyrectangle.js
@@ -253,20 +253,12 @@ ngeo.interaction.ModifyRectangle.prototype.handleFeatureRemove_ = function(evt) 
  * @private
  */
 ngeo.interaction.ModifyRectangle.prototype.handleMove_ = function(evt) {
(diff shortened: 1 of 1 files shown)


src/ol-ext/interaction/modifyrectangle.js
```diff
@@ -254,13 +254,10 @@ ngeo.interaction.ModifyRectangle.prototype.handleFeatureRemove_ = function(evt)
...
-  var feature = map.forEachFeatureAtPixel(evt.pixel,
-        function(feature, layer) {
-          return feature;
-        });
+  var hit = map.hasFeatureAtPixel(evt.pixel);
 
-  if (this.vectorPoints_.getVisible() != !!feature) {
-    if (feature) {
+  if (this.vectorPoints_.getVisible() != hit) {
+    if (hit) {
```
gberaudo: The if block can be simplified to `this.vectorPoints_.setVisible(hit);`.

gberaudo:
I added a minor comment, otherwise looks good to me.

PR #1669, opened 2016-08-08
Warn for potential system freeze

tsauerwein:
When re-installing the node modules while "make serve" is running, the system might freeze and you are forced to do a hard reboot. Now you get a warning if a node process is running.


pgiraud:
Excellent. Thanks a lot for taking care of this. Please merge.

tsauerwein:
Thanks for the review!

PR #1670, opened 2016-08-08
Make the search placeholder configurable

fredj:
TODO: the `Search…` string is not translated 

edit: it's now fixed

### DIFF
--- a/contribs/gmf/examples/search.html
+++ b/contribs/gmf/examples/search.html
@@ -62,7 +62,7 @@
         height: 400px;
       }
       span.twitter-typeahead {
-        width: 160px;
+        width: 300px;
       }
       .gmf-search > * {
         float: left;
@@ -82,6 +82,7 @@
       span.twitter-typeahead .tt-menu {
         position: absolute;
         top: 100%;
+        width: 100%;
         left: 0;
         z-index: 1000;
         display: none;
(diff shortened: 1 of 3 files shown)


tsauerwein:
> TODO: the Search… string is not translated

For c2corg we are using `x-translate`. `x-translate` is not extracted as translatable string, but Angular recognizes it.

tsauerwein:
(Though that doesn't work together with `placeholder`.)

tsauerwein:
Other idea: Create a custom filter `translate-no-extracted` which internally uses `translate` but which is not extracted by `gettext`.

fredj:
I have to check but I suspect `gettextCatalog.getString` to happen to early (before the remote json are loaded)

fredj:
`gettextCatalog.getString` was indeed called before the remote json file were processed.

fredj:
The solution was to use `placeholder="{{ctrl.placeholder || 'Search…' | translate}}`

gberaudo:
LGTM

gberaudo:
Out of curiosity, do someone knows why the search directive is not bound to the controller?

fredj:
> Out of curiosity, do someone knows why the search directive is not bound to the controller?

I don't know but I don't see any god reason ...

PR #1671, opened 2016-08-08
Check if the map has a size before calling ol.View#fit

fredj:


gberaudo:
LGTM

fredj:
Thanks for the review

PR #1683, opened 2016-08-11
Don't initialize the ngeoFeatureOverlayMgr in the directive

### LINKED ISSUE #1682: Can't have more than one gmf-search on the same page
If more than one gmf-search are on the same page, the feature overlay doesn't work (no geometry display on the map)

fredj:
Like all others directive, fixes #1682


gberaudo:
LGTM, please merge.

fredj:
Thanks for the review

PR #1688, opened 2016-08-11
Handle one level of background groups

gberaudo:
The group will appear in the background selector.
Selecting the group will activate child layers.

All children of a group are assumed to be plain layers.
Supported dimensions are read from the group.

In addition, trigger a source refresh on dimension change.


sbrunner:
Looks good for me :-)

gberaudo:
Thanks for the review.

Issue #1570, opened 2016-07-14
The layer order is wrong in the getmap


sbrunner:
It should be inverted

pgiraud:
Is it the same problem as exposed in #1510?

sbrunner:
No, #1510 is about the group order, and this one the layer order on the map inside a group (mixed and non mixed).
Example URLs:
* https://camptocamp.github.io/ngeo/master/examples/contribs/gmf/apps/desktop_alt/?lang=fr&baselayer_ref=map&tree_groups=OSM%20functions%20mixed&map_x=646915&map_y=178710&map_zoom=4&tree_enable_osm_open=false&tree_enable_Layer%20with%20very%20very%20very%20very%20very%20long%20name=true&tree_enable_osm_scale=false
* https://camptocamp.github.io/ngeo/master/examples/contribs/gmf/apps/desktop_alt/?lang=fr&baselayer_ref=map&tree_groups=OSM%20functions&map_x=646915&map_y=178710&map_zoom=4&tree_group_layers_OSM%20functions=osm_time%2Csrtm

Issue #1614, opened 2016-07-25
The findGroupByLayerName should recures on all the nodes in the group.


sbrunner:
Here: https://github.com/camptocamp/ngeo/blob/master/contribs/gmf/src/services/themesservice.js#L126

See also: https://github.com/camptocamp/ngeo/pull/1589#discussion_r72020249

PR #1697, opened 2016-08-15
Fix layer order in layertree

ger-benjamin:
Fix: #1570

Demo: https://ger-benjamin.github.io/ngeo/gmfLayerOrder/examples/contribs/gmf/apps/desktop

Please, try around functions of the layertree (search layers, queries) as well.

I've:
 
- Removed redounded call of the "prepareLayer_" function (line 372 was redound with 395)
- "insertAt 0" inseatead of "push" of the wms/wmts layers (mixed groups). (line 394)
- Simplified the toggle function (part "not mixed groups", lines 653~681).
- Reorder wms LAYERS names in not mixed groups (every times read in order, must be added to the map in reverse order. (others changes)

Second commit is for the add layer action (order was wrong with wms layers).

(Create a function that update wms, keep server order AND works in all cases was a headache. I've gave up.)

### DIFF
--- a/contribs/gmf/src/directives/layertree.js
+++ b/contribs/gmf/src/directives/layertree.js
@@ -294,7 +294,7 @@ gmf.LayertreeController.prototype.prepareLayer_ = function(node, layer) {
   var metadata = node.metadata;
   if (isMerged) {
     //Case Non Mixed group -> Hide the layer if all child nodes have isChecked set to false
-    this.getFlatNodes_(node, childNodes);
+    gmf.LayertreeController.getFlatNodes(node, childNodes);
     allChildNodesUnchecked = childNodes.every(function(childNode) {
       return !childNode.metadata || !childNode.metadata['isChecked'];
     });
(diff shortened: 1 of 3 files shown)


contribs/gmf/src/directives/layertree.js
```diff
@@ -419,16 +418,18 @@ gmf.LayertreeController.prototype.getLayerCaseMixedGroup_ = function(node) {
  * @private
  */
 gmf.LayertreeController.prototype.getLayerCaseNotMixedGroup_ = function(node) {
-  var childs = [];
+  var child = [];
```
pgiraud: If it is an array of elements, it should be called "children" instead.

ger-benjamin: Oups, yes, thanks, that was my intention

ger-benjamin:
The gmf.LayertreeController.getFlatNode function will be moved in the gmf.Themes service during the fixe of #1614 . I need this function also in the Theme and that a better place for it (already done, I just wait on this PR)

sbrunner:
Looks good to me :-)

PR #1698, opened 2016-08-15
Fix properties not defined on Object

adube:
In an application where we include ngeo alongside other libraries and in which we compile them all together using Compiler Compiler, we get errors like this one:

    ERR! compile ngeo/ol-ext/format/featurehash.js:590:
    ERROR - Property pointRadius never defined on Object<string,(boolean|number|string)>
    ERR! compile   pointRadius = properties.pointRadius;

I checked if the minimized JavaScript file produced by the build script within ngeo and these properties stay as is, i.e. the compiler don't rename them.  In our app, it would.  This patch fixes this.


fredj:
LGTM

adube:
Thanks

Issue #1436, opened 2016-06-20
change the layer property called 'layerName' by 'nodeName'


olivierSemet:
Currently for each ol.Layer, we add a property called 'layerName' to the layer object:

https://github.com/camptocamp/ngeo/blob/master/contribs/gmf/src/directives/layertree.js#L210

I think this property should be called 'nodeName' instead of 'layerName'. As the doc says, this is

> A 'layerName' parameter with the node name as value 
https://github.com/camptocamp/ngeo/blob/master/contribs/gmf/src/directives/layertree.js#L239

IMO this is a bit ambiguous since we use the node name to fill this property and moreover there are some properties attached to the node ('layers' for WMS, 'layer' for WMTS) which contain the name(s) of the layer(s) for this node.

@fredj  @sbrunner @ger-benjamin  What do you think?

olivierSemet:
This property is used on the following lines:

- https://github.com/camptocamp/ngeo/blob/master/contribs/gmf/src/directives/layertree.js#L210
- https://github.com/camptocamp/ngeo/blob/master/src/services/layerHelper.js#L219
- https://github.com/camptocamp/ngeo/blob/master/contribs/gmf/src/directives/print.js#L761
- https://github.com/camptocamp/ngeo/blob/master/contribs/gmf/src/services/permalink.js#L888
- https://github.com/camptocamp/ngeo/blob/master/contribs/gmf/src/services/permalink.js#L874

fredj:
You're right, `nodeName` is better

sbrunner:
`nodeName` is better but if it's not cleat that it's for a layer only it's better to put `layerNodeName`.

PR #1702, opened 2016-08-16
Set layer property 'layerName' to 'layerNodeName'

ger-benjamin:
Fix: #1436 

Example: https://ger-benjamin.github.io/ngeo/layerNodeName/examples/contribs/gmf/apps/desktop


sbrunner:
Looks good :-)

PR #1705, opened 2016-08-17
OpenLayers optimizations

### LINKED ISSUE #1544: OpenLayers optimisations
- [x] Use the preload infinity on the WMTS base layers
- [x] Download the WMTS base layers on interation and on animation
- [ ] Priorise the tiles downloading for the WMTS base layers

fredj:
see #1544 

Notes:
 - The Infinity preloading applies to all WMTS layers, not only to the background layers.
 - All the tiles are loaded while animating and interacting (but only form the desktop interface)

### DIFF
--- a/contribs/gmf/src/controllers/abstractdesktop.js
+++ b/contribs/gmf/src/controllers/abstractdesktop.js
@@ -94,7 +94,9 @@ gmf.AbstractDesktopController = function(config, $scope, $injector) {
     interactions: config.mapInteractions || ol.interaction.defaults({
       pinchRotate: false,
       altShiftDragRotate: false
-    })
+    }),
+    loadTilesWhileAnimating: true,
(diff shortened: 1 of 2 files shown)


sbrunner:
Looks good, but #1544 should steel be open (or create a new issue) :-)

fredj:
> Looks good, but #1544 should steel be open (or create a new issue) :-)

I've replaced `fixes` with `see`. The last point "Priorise the tiles downloading for the WMTS base layers" is not possible with OpenLayers: all the tiles are loaded with the same priority, it's not possible to have a higher priority by layer

fredj:
@sbrunner I've not activated to preloading for the mobile interface to save the bandwidth, is it ok for you?

sbrunner:
ok :-)

fredj:
Thanks for the review

Issue #1527, opened 2016-07-05
Query should be disabled when the print panel is open


fredj:
The query results on the map must stay

ger-benjamin:
Can't it be closed ?

sbrunner:
We steal have the query in the debugger ...

PR #1706, opened 2016-08-17
Deactivate query on print but keep query infos

ger-benjamin:
Fix: #1527 

Demo: https://ger-benjamin.github.io/ngeo/printandqueries/examples/contribs/gmf/apps/desktop/

(No query tools in the desktop_alt)

I (already) don't remember, the query window should be effectively always visible ?


sbrunner:
Looks good to me :-)

PR #1707, opened 2016-08-17
FindGroupByLayerName search now in sub groups

ger-benjamin:
Fix: #1614

Demo: https://ger-benjamin.github.io/ngeo/1614/examples/contribs/gmf/apps/desktop/

I hope it's correct, I don't really know how I can test in the current demo. But now, this method search the group in all `children`.


sbrunner:
Looks good :-)

Issue #1496, opened 2016-06-30
Change background layer on user change


fredj:

ger-benjamin:
@sbrunner We have multiple possibility for the choice of the backgroundlayer. Can you confirm this priority list:

Check if:

 - There is a background in the permalink
 - If not, then:
 - There is a background in the current theme:
 - If not, then:
 - There is a background in the current user:
 - If not then: use the first available background in the theme.json server object.

ger-benjamin:
Mhhh, my list doesn't make sens. The permalink as always a background layer (except the first time but you will have no user and no selected theme too).

So it must be skipped in some case.
See that in my PR (soon)

PR #1711, opened 2016-08-18
Use user defined background layer

ger-benjamin:
Fix: #1496 

Demo: https://ger-benjamin.github.io/ngeo/userbgLayer/examples/contribs/gmf/apps/desktop

(role admin as a grey bg layer)

(In desktop_alt, theme `OSM` doesn't have a defined background layer. In desktop, it has it)


fredj:
LGTM

PR #1713, opened 2016-08-18
Replace constant by value

### LINKED ISSUE #1703: Use `gmf.module.value` instead of `gmf.module.constant`?
Currently `gmf.module.constant` is used to configure the applications, for example the coordinate formatting: https://github.com/camptocamp/ngeo/blob/c0f235c09c8ddf274fbbae1fec1ed040d8dc45fe/contribs/gmf/src/controllers/abstract.js#L50

The problem is, that all constants that are set in the abstract controllers, can not be changed any more in the client application. Because ... they are constants.

For example in a client application, you could not change [gmfRasterUrl](https://github.com/camptocamp/ngeo/blob/211e68a9cc7e744e214d18aa34db07c7136beca1/contribs/gmf/src/controllers/abstractmobile.js#L33-L35) because it is already set in the abstract controller.

We could use `gmf.module.value` only for the constants that are set in the abstract controllers. But to be consistent, I would propose to use `gmf.module.value` instead of `gmf.module.constant` for everything you should be able to set in a client application.

pgiraud:
Fixes #1703.

I wanted to set the value for the appropriate module but eventually decided that it required too much extra code:

```javascript
var gmfModule = angular.module('gmf');
gmfModule.value('gmfTheValue', ...);
```

### DIFF
--- a/contribs/gmf/examples/authentication.js
+++ b/contribs/gmf/examples/authentication.js
@@ -11,7 +11,7 @@ var app = {};
 app.module = angular.module('app', ['gmf']);
 
 
-app.module.constant(
+app.module.value(
     'authenticationBaseUrl',
     'https://geomapfish-demo.camptocamp.net/2.1/wsgi');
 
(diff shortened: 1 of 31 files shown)


sbrunner:
Looks good :-)

PR #1717, opened 2016-08-18
Remove goog.base from ol-ext files

adube:
This PR removes the `goog.base` methods from the files that are in `src/ol-ext`, i.e. from files that inherits from OpenLayers classes.

This allow those classes to be built with Closure-Util in which the `useOfGoogBase` configuration is not used.


adube:
Ready for review.

fredj:
thanks

PR #1726, opened 2016-08-19
Fix #1721 - Allow vertex deletion in gmf-editfeature

### LINKED ISSUE #1721: During edition, it is impossible to delete a vertex
It looks like single click to delete a vertex (like in redlining) doesn't work for editing.

adube:
This PR is a fix that allows vertice to be deleted on click on the selected vector feature in the `gmf-editfeature` directive.

The trick is to check if a vector feature was clicked before launching a new query. In that case, do nothing, which allows the default behavior of the `ol.interaction.Modify` to occur in case of a vertex click to delete it.

## Todo

 * [ ] review

## Live demo

* https://adube.github.io/ngeo/1721-delete-vertex-fix//examples/contribs/gmf/editfeatureselector.html

### DIFF
--- a/contribs/gmf/src/directives/editfeature.js
+++ b/contribs/gmf/src/directives/editfeature.js
@@ -620,12 +620,39 @@ gmf.EditfeatureController.prototype.handleMapSelectActiveChange_ = function(
 
 
 /**
+ * Called when the map is clicked.
+ *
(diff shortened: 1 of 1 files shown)


adube:
@fredj or @pgiraud - Ready for review.

sbrunner:
Looks good, thanks :-)

adube:
Thank you!

PR #1727, opened 2016-08-19
Fix #1720 - Show modal on feature unsaved changes - gmf-editfeature

### LINKED ISSUE #1720: When a geometry is edited, changes are lost when clicking anywhere else
In Specs below, it is noted that when changes are made there is a warning if feature is going to be unselected:

https://github.com/camptocamp/c2cgeoportal/wiki/Spec-%231662-Editing

This doesn't work:
 * Select a feature
 * Edit it
 * Click anywhere else on the map

The same happens when adding a new geometry. New geometry is lost when clicking elsewhere.

adube:
This PR fixes #1720.  When a feature is being edited in the `gmf-editfeature` directive, if there are unsaved modifications and the user clicks out of the feature, a confirmation modal is shown to ask him or her what to do.

The above behavior also happens when the "cancel" button of the form is clicked.

Finally, when the "continue without saving" or "save" of the confirm modal is clicked, the previously clicked location is managed, i.e. after the modal is closed the location where the click occured will be managed. If an other feature had been clicked, it will be selected.

## Todo

 * [x] review

## Will be done in an other task

* The "stop editing" management
* The selection of an other tool

I'll create 2 separate issues for those.

## Live example

 * https://adube.github.io/ngeo/1720-unsaved-modifications/examples/contribs/gmf/editfeatureselector.html

### DIFF
--- a/contribs/gmf/src/directives/partials/editfeature.html
+++ b/contribs/gmf/src/directives/partials/editfeature.html
@@ -29,39 +29,80 @@
         </span>
       </span>
     </a>
-    <form
-      novalidate
-      name="form"
-      class="form gmf-editfeature-attributes-container"
-      ng-switch-default
-      ng-if="efCtrl.attributes">
-      <ngeo-attributes
-        ngeo-attributes-attributes="::efCtrl.attributes"
-        ngeo-attributes-disabled="efCtrl.pending"
-        ngeo-attributes-feature="::efCtrl.feature">
-      </ngeo-attributes>
-      <input
-        type="submit"
-        value="{{'Save' | translate}}"
-        class="btn btn-sm btn-default gmf-editfeature-btn-save"
-        ng-click="form.$valid && efCtrl.save()"
-        ng-disabled="!efCtrl.dirty"
-        title="{{'Save modifications | translate'}}"></input>
-      <a
-        class="btn btn-sm btn-default gmf-editfeature-btn-cancel"
-        ng-click="efCtrl.cancel()"
-        ng-disabled="efCtrl.pending"
-        title="{{'Cancel modifications | translate'}}"
-        href>{{'Cancel' | translate}}</a>
-      <button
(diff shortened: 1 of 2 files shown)


adube:
@sbrunner Ready for review.  The live demo should arrive soon.

contribs/gmf/src/directives/partials/editfeature.html
```diff
@@ -29,39 +29,80 @@
...
-        ng-click="efCtrl.delete()"
-        ng-disabled="efCtrl.pending"
-        ng-show="efCtrl.featureId"
-        title="{{'Delete this feature | translate'}}">
-        <span class="fa fa-trash"></span>
-        {{'Delete' | translate}}
-      </button>
-    </form>
+    <div ng-switch-default>
+      <ngeo-modal ng-model="efCtrl.unsavedModificationsModalShown">
```
pgiraud: Do you really need to put the modal in the switch?

adube: Putting the modal in the switch reduces the number of watchers.  It is also only required while a feature is selected, so putting it in the switch makes sense to me.  I also used the authentication directive as inspiration, which does that.  I was told that reducing the number of watchers is a priority in Angular, so this kind of wierd practice is a result of that.

What do you think ?

pgiraud: Thanks for the explanation.

adube:
Reminder: please, review.

fredj:
LGTM, @pgiraud is it good for you?

pgiraud:
I don't think it fully addresses what I expected when writing the specs. I need to discuss this with @ybolognini.

pgiraud:
Unfortunately, I detected some (random) cases where the previous action is not correctly handled after the modal is closed.
I'll give this feature a new try by using promises.

PR #1733, opened 2016-08-22
WIP - Support theme reload in gmf-editfeatureselector directive

adube:
This requires more discussion.  Code is incomplete and requires more work.

### DIFF
--- a/contribs/gmf/src/services/themesservice.js
+++ b/contribs/gmf/src/services/themesservice.js
@@ -401,6 +401,9 @@ gmf.Themes.prototype.loadThemes = function(opt_roleId) {
 
   goog.asserts.assert(this.treeUrl_, 'gmfTreeUrl should be defined.');
 
+  this.deferred_ = this.$q_.defer();
+  this.promise_ = this.deferred_.promise;
+
   var deferred = this.deferred_;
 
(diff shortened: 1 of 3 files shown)


adube:
@pgiraud and @fredj You can take a look at this for your discussions.

contribs/gmf/src/services/themesservice.js
```diff
@@ -392,6 +401,9 @@ gmf.Themes.prototype.loadThemes = function(opt_roleId) {
 
   goog.asserts.assert(this.treeUrl_, 'gmfTreeUrl should be defined.');
 
+  this.deferred_ = this.$q_.defer();
+  this.promise_ = this.deferred_.promise;
```
fredj: Not sure it will work if we have:
```
gmfThemes.getThemesObject().then(function(themes) {
  //
});
gmfThemes.loadThemes();
```

The promise used in `getThemesObject` is removed in `loadThemes` and the function is never called.

fredj:
This is more general but I think our architecture is not good.
Instead of getting the themes in the directives (using the `gmfThemes` service) we should instead loading the themes in the controller and passing it the the directives (where we could use `watchCollection` to updated on change)

fredj:
@adube please see #1735

PR #1739, opened 2016-08-23
Rework of UI for editing panel

### LINKED ISSUE #1724: Editing: UI harmonization
Editing would need a bit of UI harmonization: title of layer, attribute labels, etc.

pgiraud:
Fixes #1724 

 - Removed the ":" after the label. The spaces before and after are not the same in the different languages. Also, it's not really required. So I removed them for simplicity.
 - Changed the "chosen" editing layer title look and feel to make it a bit more clear.
 - Added a more clear separation between the chosen editing layer title and the editing attributes section.
 - Used primary color for the "save" button.
 - Added a cross (close) button to deselect the current feature. It has the same behavior as the cancel button actually.

Please review.

After / Before
screenshot from 2016-08-23 14 01 41

Demo:
https://pgiraud.github.io/ngeo/editing_ui/examples/contribs/gmf/apps/desktop?baselayer_ref=map&lang=fr&map_x=561294&map_y=204601&map_zoom=7&rl_features=Fa(57v1-3bcEgC8w*9t!hH90s6*~n*Polygon%25201%27c*%2523DB4436%27a*0%27o*0.2%27m*false%27s*10%27k*1)&tree_groups=Edit

### DIFF
--- a/contribs/gmf/less/desktop.less
+++ b/contribs/gmf/less/desktop.less
@@ -532,9 +532,8 @@ gmf-disclaimer {
 /**
  * GMF EditFeature directive
  */
-.gmf-editfeatureselector-stopediting {
-  float: right;
-}
-.gmf-editfeature-btn-delete {
-  float: right;
+gmf-editfeature > div {
+  border-top: 1px solid #333;
+  margin-top: 1rem;
+  padding-top: 1rem;
 }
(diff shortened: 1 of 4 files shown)


ybolognini:
Much better, thanks!

Only remark: I don't like the bold label. And it takes too much space, I think.

In all other forms (print, redlining), label aren't bold. And I would make the editing form more compact, maybe by removing some space above and below label.

pgiraud:
Bold removed.

sbrunner:
The screenshot are inverted ...

sbrunner:
... that right!

PR #1743, opened 2016-08-24
Order group layers in tree and map.

### LINKED ISSUE #1722: The group order is inverted on the map...

ger-benjamin:
Fix #1722 

Example: https://ger-benjamin.github.io/ngeo/layerGroupOrder/examples/contribs/gmf/apps/desktop_alt

The problem was far more complex than expected. Please read and understand before to comment.

To summarize (it's hard), try to understand this example:

 - If I add theme OSM, I add 4 first level groups at once.
  - The layer at the top of the layer tree will be added first in the map
  - then the second, etc.
 - Then I add the "edit" theme. So one another first level layer group.
  - I should add it at the top of the layertree but angular doesn't add all OSM theme again so it will be added alone.

The problem is that I can't just add layer in the map by a `insertAt(0, layer)` or a `push(layer)`.

 - With `insertAt(0, layer)`, the OSM layer will be ordered correctly on the map but the edit theme will be under all other layers.
 - With `push(layer)` the OSM theme will be in a wrong order, but edit will be right at the top.

So the only realistic solution that I found is to `insertAt(all existing first level group layers - number of layers to add at once, layer)`. But it's tricky and will perhaps complex to manage in the future.

Another solution it to add themes one by one. But that's not easy with angular and I don't find how to do that.

Perhaps I should add commentaries somewhere...

@pgiraud I guess you have worked on the layertree. Can you give me your opinion on theses changes ? Perhaps you have a better solution.

### DIFF
--- a/contribs/gmf/src/services/treemanager.js
+++ b/contribs/gmf/src/services/treemanager.js
@@ -84,6 +84,12 @@ gmf.TreeManager = function($timeout, gettextCatalog, ngeoLayerHelper, ngeoNotifi
   });
 
   /**
+   * @type {number}
+   * @public
+   */
+  this.layersToAddAtOnce = 0;
+
+  /**
    * @type {ngeo.StateManager}
    * @private
    */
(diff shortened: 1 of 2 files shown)


sbrunner:
I verify the demo and it's working well :-)

contribs/gmf/src/services/treemanager.js
```diff
@@ -250,6 +262,7 @@ gmf.TreeManager.prototype.addGroup_ = function(group) {
  */
 gmf.TreeManager.prototype.addCustomGroups = function(groups, opt_add) {
   var groupNotAdded = [];
+  this.groupsToAdd = groups.length;
```
pgiraud: This doesn't seem to used. Am I wrong?

ger-benjamin: Right, that's an old thing. I'll remove it...

pgiraud:
I've tried to find an other way to address this issue but failed.
@ger-benjamin Can you please take a look at my comment?
Otherwise this looks good.

PR #1756, opened 2016-08-25
Fix editfeature selector example

adube:
This PR fixes the edit feature selector example.  The list of layers was empty due to the changes of theme loading using the event.

Also, after logging in, we would normally need to reload the themes, but that's out of scope of this example (in my opinion) and would bring more complexity to it. A simple note about reloading the page after logging in is enough (again, in my opinion).


adube:
Ready for review.

Issue #1692, opened 2016-08-15
Not possible to close discalimer on IPod


marionb:
It is not possible to close the disclaimer on the mobile interface on IPod using Safari browser.
The close button of the disclaimer is only showing partially. This is also the case for the IPad (Safari) however there it can be closed.

ger-benjamin:
You need to be precise but it's possible.

I'll try to fix that.

PR #1761, opened 2016-08-29
Fix too small disclaimer close button

ger-benjamin:
Fix https://github.com/camptocamp/ngeo/issues/1692

Example: https://ger-benjamin.github.io/ngeo/fixclose/examples/contribs/gmf/apps/mobile

Close disclaimer is easier on litlle device and no more cut (a in my opinion, the cross is (more) coherent now).


fredj:
LGTM

PR #1767, opened 2016-08-30
Export the gmf-search input value

fredj:
To be able to use the value outside of the directive

### DIFF
--- a/contribs/gmf/src/directives/search.js
+++ b/contribs/gmf/src/directives/search.js
@@ -119,7 +119,7 @@ gmf.searchDirective = function(gmfSearchTemplateUrl) {
   return {
     restrict: 'E',
     bindToController: {
-      'input_value': '=?gmfSearchInputValue'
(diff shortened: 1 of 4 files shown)


contribs/gmf/src/directives/search.js
```diff
@@ -117,6 +118,9 @@ gmf.module.value('gmfSearchTemplateUrl',
 gmf.searchDirective = function(gmfSearchTemplateUrl) {
   return {
     restrict: 'E',
+    bindToController: {
+      'input_value': '=?gmfSearchInputValue'
```
tsauerwein: `inputValue`? (without underscore)

fredj: it's the variable name that was already in gmf-search (but I can change it)

fredj: renamed to inputValue

tsauerwein:
Thanks! :+1:

fredj:
Thanks for the review

Issue #1764, opened 2016-08-30
i18n error with add_layer, add_group, add_theme


sbrunner:
Two related issue:
 - [ ] Actually it's collected but not localised in French ...
 - [ ] The source should be in English (Transifex consider the sources string to be in English...)

PR #1768, opened 2016-08-30
I18n search actions

ger-benjamin:
Fix: #1764 

Example: https://ger-benjamin.github.io/ngeo/i18naddlayers/examples/contribs/gmf/apps/desktop_alt

Is that okay with this solution ?
I think it's good enough. We don't add new actions all day. If that become problematic with further dev, we can adapt the server part to add actions directly in english.

### DIFF
--- a/contribs/gmf/apps/desktop/index.html
+++ b/contribs/gmf/apps/desktop/index.html
@@ -287,7 +287,11 @@
         module.constant('gmfShortenerCreateUrl', 'https://geomapfish-demo.camptocamp.net/2.1/wsgi/short/create');
         module.constant('gmfSearchGroups', ['osm','district']);
         // Requires that the gmfSearchGroups is specified
-        module.constant('gmfSearchActions', ['add_theme', 'add_group', 'add_layer']);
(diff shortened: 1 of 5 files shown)


ger-benjamin:
Ping

sbrunner:
Shouldn't we change gmfSearchActions` to ?:
```
module.value('gmfSearchActions', [
    { action: 'add_theme', name: gettextCatalog.getString('Add a theme') },
    { action: 'add_group', name: gettextCatalog.getString('Add sub theme') },
    { action: 'add_layer', name: gettextCatalog.getString('Add a layer') }
]);
```

ger-benjamin:
add_theme / add_group / add_layer is wrote like that to match with what is sent by the server.

If we "translate" further, we should "translate" the server part.

ger-benjamin:
You're right. I'll do that soon

ger-benjamin:
@sbrunner Done, but I can't translate the title in the html

ger-benjamin:
Done this time

sbrunner:
Looks good to me :-)

PR #1770, opened 2016-08-30
Skip undefined layers

gberaudo:
This is necessary to avoid bad access farther in OpenLayers 3 code.


gberaudo:
Added one commit to fix the build. No idea why it triggered all of a sudden.

sbrunner:
Looks good :-)

gberaudo:
Thanks for the review.

Issue #1765, opened 2016-08-30
i18n issue in the grid


sbrunner:
- open https://camptocamp.github.io/ngeo/master/examples/contribs/gmf/apps/desktop_alt/?lang=fr&baselayer_ref=map&tree_groups=OSM%20functions%20mixed%2CLayers%2CGroup%2COSM%20functions&map_x=632740&map_y=186582&map_zoom=6&tree_group_layers_OSM%20functions&tree_group_layers_Layers&tree_enable_osm_time_r_dp=false&tree_enable_osm_time_v_dp=false&tree_enable_osm_time_v_s=false&tree_enable_osm_time_r_s=true&tree_enable_osm_scale=false
- do a query on a point 

=> the layer is named `merged_osm_times` but I expectect to be `Temps OSM curseur interval` (it's the case with the popup)

tsauerwein:
Note that there is an explicit definition for `merged_osm_times`, this is not the case for the `displayquerywindow`. See: https://github.com/camptocamp/ngeo/blob/master/contribs/gmf/apps/desktop_alt/js/controller.js#L100

You would have to mark `merged_osm_times` as translatable. This could be done by simply calling `gettextCatalog.gettext('merged_osm_times')`, but I don't know if that is the preferred method in this project?

sbrunner:
+1 to do this event if it will don't change anything because it will not be collected...

PR #1792, opened 2016-09-02
Mark merged layer label as translatable

tsauerwein:
Closes https://github.com/camptocamp/ngeo/issues/1765

Note that this string is no extracted together with the GMF translation strings, like all other i18n strings in `contribs/gmf/apps`.


sbrunner:
thanks, merge :-)

Issue #1766, opened 2016-08-30
The editing shouldn't be desplayed only on users that's allow to edit


sbrunner:
I think that we should see if there is an editable layer in the themes and display the editing tool only if there is at least one editable layer, @pgiraud do you agree?

pgiraud:
With @adube, we had a discussion about this. We agreed that it was not a problem that the editing tool is here even when the user is not connected. The message asking the user to login is already useful.
However, it would be better to have a message telling the user that there's no editable layer in the tree instead of showing an empty select.

sbrunner:
The issue isn't about that...
For most users (login or not) they can't use the editing interface than we don't want to have the icon every time ...

ybolognini:
As discussed during scrum meeting, the goal here is to hide Edit tool when user doesn't have editing rights. In GMF 1.6, editing tools were in a separate application, so there was no problems.

After talking about it with @sbrunner , a good solution would be to hide the Edit tool if the user doesn't have any editable layer in all his themes. Please have a look at technical complexity before implementing.

PR #1793, opened 2016-09-02
Improve UX when no editable layers

tsauerwein:
* Only show edit tool if there is one editable layer in one of the themes (no matter if visible or not).
* Show message instead of an empty select box when no editable layer in the map.

Closes https://github.com/camptocamp/ngeo/issues/1766


sbrunner:
?? the customer ask to don't have the tool icon...

tsauerwein:
Updated PR

sbrunner:
Looks good to me :-)

Issue #1738, opened 2016-08-23
The tool panels are not scrollable


pgiraud:
Some functionalities may not be reachable on small screen with a lot of elements in the tool panels.

![screenshot from 2016-08-23 14 24 14](https://cloud.githubusercontent.com/assets/319774/17891629/5f9bab2a-693d-11e6-8229-7e2750270e93.png)

PR #1794, opened 2016-09-02
Make tool panel scrollable in sidebar

tsauerwein:
This makes the tool panel in the right panel scrollable. At first I tried to keep the panel title fixed, and only make the panel content area scrollable. But this caused several layouting issues, which would require more profound changes. Considering that *usually* the vertical space is sufficient, I would propose to go with this simple but effective solution. If wanted we can still improve it later.

ngeo-scrollable-tool-panel

Closes https://github.com/camptocamp/ngeo/issues/1738


pgiraud:
Nice! Thanks. Please merge.

tsauerwein:
Thanks for the review!

PR #1800, opened 2016-09-05
Make draw/edit context menu take precedence over contextual data

### LINKED ISSUE #1725: Right click popover conflicts with edit contextual menu
When editing or redlining, clicking right click opens contextual menu. But this also opens right click popover with information on clicked point.

In this case only one window should be open: the contextual menu.

pgiraud:
Fixes #1725.

With this pull request a right-click (or long press) on a feature when using the drawing or editing tools the propagation of the event is stopped. This helps preventing the contextual menu (with coordinates, elevation, etc...) to show up when a feature is right clicked.

In order to do that I had to use a different target for the event listener because it's not possible to stop propagation on the same element easily. Using `stopImmediatePropagation` is not possible here since we cannot make sure of the order in which the listeners are added.

Please review.

### DIFF
--- a/contribs/gmf/src/directives/drawfeature.js
+++ b/contribs/gmf/src/directives/drawfeature.js
@@ -531,7 +531,7 @@ gmf.DrawfeatureController.prototype.handleFeaturesRemove_ = function(evt) {
 gmf.DrawfeatureController.prototype.handleMapSelectActiveChange_ = function(
     active) {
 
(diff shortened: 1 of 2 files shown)


fredj:
LGTM

pgiraud:
Demo: https://pgiraud.github.io/ngeo/contextualdata/examples/contribs/gmf/apps/desktop

sbrunner:
@pgiraud can we merge it?

PR #1807, opened 2016-09-06
Fix #1791 - Background layer selector directive must use dimensions

adube:
Fixes #1791.  The `gmf-backgroundlayerselector` directive must use the `dimensions` options.  This PR fixes this in all 3 applications and in the example.

## Todo

 * [ ] review

## Live examples

 * (example) https://adube.github.io/ngeo/1791-fix-apps-bls/examples/contribs/gmf/backgroundlayerselector.html
 * (app) https://adube.github.io/ngeo/1791-fix-apps-bls/examples/contribs/gmf/apps/desktop_alt/index.html

PR #1809, opened 2016-09-06
Fix #1780 - Do not change bg layer on login READY

### LINKED ISSUE #1780: Background map is not conserved in the shortened permalink
When opening the shortened permalink in a new tap the background map is always the standard map (in case of the demo it is *Map*) regardless of weather a different map was used in the original tab where the link was taken from

adube:
Fixes #1780.  Before this fix, the abstract controller of the gmf applications was calling the loading of the background layer twice: first time upon initialization (with the reading of permalink), and a second (unwanted) time when the themes service triggers an event.  We do want to reload the bg layer on LOGIN or LOGOUT, but not on READY, which was the 2nd unwanted call that was overriding the first call that managed the permalink.

This PR doesn't currently have a live example yet due to #1807.

## Todo

 * [x] Wait for #1807 to be fixed
 * [x] Create a live example
 * [ ] Review

## Live example

 * https://adube.github.io/ngeo/1780-fix-permalink-bgmap/examples/contribs/gmf/apps/desktop_alt/index.html?baselayer_ref=asitvd.fond_couleur&lang=fr

### DIFF
--- a/contribs/gmf/src/controllers/abstract.js
+++ b/contribs/gmf/src/controllers/abstract.js
@@ -113,7 +113,9 @@ gmf.AbstractController = function(config, $scope, $injector) {
   var userChange = function(evt) {
     var roleId = (evt.user.username !== null) ? evt.user.role_id : undefined;
     // Reload background layer when login status changes.
-    this.updateCurrentBackgroundLayer_(true);
+    if (evt.type !== gmf.AuthenticationEventType.READY) {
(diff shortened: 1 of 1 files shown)


adube:
Live example created and ready for review.

adube:
Reminder: ready for review.

fredj:
looks good

Issue #1680, opened 2016-08-10
Add ngeo scope to CSS classes


gberaudo:
CSS class names used in ngeo are not enough scoped and thus will cause conflicts.

For example, for the search directive, we use the class ".search" and the CSS defines a rule like:
```css
.search input {
  height: 100%;
}
```

This conflicts with any application or component using an `input` in a `search` class.

As a consequence, ngeo classes should be prefixed like this:

```css
.ngeo-search input {
  height: 100%;
}
```

adube:
Changing all CSS classes in one PR would be a huge change. Is it okay if I create multiple PR each focusing on a single directive / component at a time?  For each one, fixes would be required in:

 * all 3 applications of GMF
 * all the examples of GMF that use that directive / component
 * all the examples of NGEO that use that directive / component

@gberaudo or @sbrunner : What do you think?

sbrunner:
+1 to split in more than one PR
but shouldn't we split by directive / component and also directly change it on the apps?

adube:
@sbrunner Yes, this is what I suggest.

adube:
@gberaudo or @sbrunner  I would like to clarify my intentions.  Please, let me know if you agree.

First, anything that's part of GeoMapFish should have a `gmf-` prefix instead of `ngeo-`.

Also, any class that is defined alongside a directive definition should simply be removed and the definition should use the tag name instead.  For example, the `.search` class name is used here:

        <gmf-search gmf-search-map="mainCtrl.map"
          class="search"
          gmf-search-datasources="mainCtrl.searchDatasources"
          gmf-search-coordinatesprojections="mainCtrl.searchCoordinatesProjections"
          gmf-search-colorchooser="true"
          gmf-search-clearbutton="true">
        </gmf-search>

And the definition currently looks like this (in the .less files):
```

.search {

  ...

  .gmf-search {
    background-color: @map-tools-bg-color;
    border: 1px solid @border-color;
    padding: 0 @app-margin;
    height: @map-tools-size;
    position: relative;
  }

  ...
```

Could be changed like this:

        <gmf-search gmf-search-map="mainCtrl.map"
          gmf-search-datasources="mainCtrl.searchDatasources"
          gmf-search-coordinatesprojections="mainCtrl.searchCoordinatesProjections"
          gmf-search-colorchooser="true"
          gmf-search-clearbutton="true">
        </gmf-search>

```
gmf-search {

  ...

  .gmf-search {
[...]

PR #1816, opened 2016-09-07
#1680 - Fix CSS class name in gmf-mouseposition directive

adube:
This PR fixes the CSS class name for the gmf mouseposition directive.  Changes are made in the example and and desktop application.  See #1680.  Note that this is one among many PR that will come to fix #1680.

The use of an `id` within the directive is not a good practice.  If one wanted to have 2 maps with each their own mouse positions directive, then it wouldn't work because the directive appends the created control using the id of the element defined within the template.  That's not an issue for now, but this could be changed.

Instead of using the id in the CSS, the class name of the control itself is used.

## Todo

 * [ ] review

## Live examples

 * (example) https://adube.github.io/ngeo/1680-css-fix-mouseposition/examples/contribs/gmf/mouseposition.html
 * (desktop app) https://adube.github.io/ngeo/1680-css-fix-mouseposition/examples/contribs/gmf/apps/desktop/index.html

### DIFF
--- a/contribs/gmf/examples/mouseposition.html
+++ b/contribs/gmf/examples/mouseposition.html
@@ -20,7 +20,7 @@
       .dropdown-menu {
         min-width: 260px;
       }
-      #mouse-position {
+      .gmf-mouseposition-control {
         display: inline-block;
         min-width: 200px;
       }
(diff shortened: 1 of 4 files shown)


gberaudo:
LGTM. +100 to stop using ids or global objects (ex: `document.getElementById`).

adube:
@gberaudo Thanks. I'll go ahead and create issues for the `id` matters.  I'll create one per component I stumble upon.

If you wish for me to work on them right away, then you can assign them to me.

PR #1817, opened 2016-09-07
Enhance readability of layertree html templates

adube:
This PR makes the layer tree html templates easier to read by breaking very long lines into one per attribute within a node, if there's more than one.  Some space are also added above child nodes if the parent node has many children, again to improve readability.

Having more lines makes the files heavier, but that shouldn't matter since the templates get compiled in the end, on production.


adube:
Ready for review.

sbrunner:
Looks good :-)

adube:
Fixed.  Thanks.

PR #1820, opened 2016-09-08
Ensure node comments are not messed up after drop at same location

### LINKED ISSUE #1744: All groups are deleted after moving a group
To reproduce:
- OSM theme on "desktop" example
- Keeping mouse button down, start to move top group
- Keeping mouse button down, move the group down and up a little
- Move the group back to top place, release mouse button
- Click on "remove" button, all groups are deleted!

pgiraud:
Fixes #1744.

### DIFF
--- a/src/directives/sortable.js
+++ b/src/directives/sortable.js
@@ -87,6 +87,11 @@ ngeo.sortableDirective = function($timeout) {
            * time the sortable array changes (see $watchCollection above).
            */
           function resetUpDragDrop() {
+            // Save the current nodes in order to restore the state if the node
+            // is dropped at the same place
+            // In this case the comments and element nodes are messed up
+            var savedNodes = element.contents();
+
             var children = element.children();
             for (var i = 0; i < children.length; ++i) {
               angular.element(children[i]).data('idx', i);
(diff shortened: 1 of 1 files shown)


pgiraud:
Sorry still some checks errors to fix.

pgiraud:
Please review.
Demo: https://pgiraud.github.io/ngeo/fix-sortable/examples/contribs/gmf/apps/desktop

ger-benjamin:
There is always an issue. Try to move around the first item then drop it at the second place.
Delete it: all elements after will be deleted.

pgiraud:
OK. My pull request fixes only one part. There are 2 different issues.

@ger-benjamin provided an example showing that things are messed up when dropping an element at the same place.
See https://ger-benjamin.github.io/ngeo/repeatError/examples/contribs/gmf/repeaterror.html

With my patch applied, the "drop-to-same-location" issue is fixed:
https://pgiraud.github.io/ngeo/repeatError/examples/contribs/gmf/repeaterror.html

In the GMF layertree, this is also fixed.

But as @ger-benjamin reported, there's an other case which is not correctly handled. It's a completely separate issue though. Because we use a different property for the `ngeo-sortable` and the `ng-repeat` (`gmfLayertreeCtrl.layers` vs. `layertreeCtrl.node.children`), the `ng-repeat` doesn't detect changes properly and doesn't rebuild the DOM.
With the up-coming refactoring on the tree, I hope we will eventually use the same property for both which will fix the issue.

ger-benjamin:
Yes. So then, for me, this problem is fixed. You can merge.

sbrunner:
@pgiraud up ?

pgiraud:
People from the Luxembourg geoportal application confirmed that https://github.com/Geoportail-Luxembourg/geoportailv3/issues/1423 gets fixed with the current pull request. Merging.

PR #1822, opened 2016-09-08
#1680 - Fix CSS class names in apps <body>

adube:
This PR fixes the CSS class name used in the `<body>` element of the desktop applications.  See #1680.  Note that this is one among many PR that will come to fix #1680.

## Todo

 * [ ] review

## Live examples

No live updated required.


adube:
Ready for review.

PR #1826, opened 2016-09-08
#1680 - Fix CSS class names in gmf share

adube:
This PR fixes the CSS class name for the gmf share directive.  See #1680.

The only class name unique to gmf is not actually used anywhere (no specific CSS definition in the example and apps).


adube:
Ready for review.

PR #1827, opened 2016-09-08
Improve readability of print directive template

adube:
This PR improves the readability of the gmf print directive html template.  The class name is also changed, but it's currently not used anywhere.  Finally, this PR also includes a missing `goog.require` for a missing dependency, which prevented the example from working in a dev environment.


adube:
Ready for review.

sbrunner:
Looks good :-)

Issue #1776, opened 2016-09-01
Not possible to export KML Edge/Win10


marionb:
On Win10 Edge it is not possible to export a KML of the redlining features. When clicking on the export a new blank tab opens but there is no download.

Also when exporting a CSV from the query table nothing is downloaded

(On Win10 Firefox the download worked OK)

marionb:
This issue also appears on Safari latest / Mac El Cap.

PR #1830, opened 2016-09-09
Download files with `file-saver`

tsauerwein:
Use the library [file-saver](https://github.com/eligrey/FileSaver.js) to download KML/GPX/CSV files.

Currently I can't test properly, but file-saver has support for all modern browsers.

Closes https://github.com/camptocamp/ngeo/issues/1776

### DIFF
--- a/src/services/featurehelper.js
+++ b/src/services/featurehelper.js
@@ -5,6 +5,7 @@ goog.require('ngeo');
 goog.require('ngeo.filters');
 goog.require('ngeo.interaction.Measure');
 goog.require('ngeo.interaction.MeasureAzimut');
+goog.require('ngeo.Download');
 goog.require('ol.Feature');
 goog.require('ol.geom.LineString');
 goog.require('ol.geom.MultiPoint');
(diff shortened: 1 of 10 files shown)


src/services/featurehelper.js
```diff
@@ -88,6 +89,13 @@ ngeo.FeatureHelper = function($injector, $filter) {
    */
   this.projection_;
 
+  /**
+   * Download service.
+   * @type {ngeo.Download}
+   * @private
+   */
+  this.download_ = $injector.get('ngeoDownload');
```
gberaudo: Shouldn't you use Angular automatic injection instead of doing it manually here?

tsauerwein: There isn't a clear guideline in ngeo when to use `$injector` and when to use automatic injection.

sbrunner: For me the injector should be used when the dependency is optional (and in the gmf controller for simplification).
If it's not like that we can clarify it later...

fredj:
We can merge this PR and create a new issue for the injector usage

sbrunner:
@tsauerwein up :-)

fredj:
@tsauerwein please rebase and merge

src/services/featurehelper.js
tsauerwein: I opened an  issue: https://github.com/camptocamp/ngeo/issues/1865

tsauerwein:
Tested with BrowserStack on Windows 10 and Edge 14, merging.

PR #1845, opened 2016-09-12
Fix treemanager missing require

adube:
This PR fixes the `gmf.TreeManager` service.

1. It was missing a require of the `ngeo.StateManager`
2. It is using the `gmf.PermalinkParam.TREE_GROUPS` constant but was not requiring the permalink service.  In the end, we can't do that because the permalink service already requires the tree manager as well and we would end up with a circular dependency.  To fix that, the constant `gmf.PermalinkParam` was moved to `gmf.js`.

I found this out while trying to view the theme selector example, which didn't work on dev.

## Todo

 * [ ] review

## Live example

 * https://adube.github.io/ngeo/fix-treemanager/examples/contribs/gmf/themeselector.html


adube:
Thanks for the approval.

PR #1846, opened 2016-09-12
#1680 - Fix active CSS class name in gmf themeselector

adube:
This PR fixes the `active` CSS name defined in the `gmf-themeselector` directive.  It's not used in the applications and was not used in the example either, but since it's still somewhat a "feature" I added it in the example.

## Todo

 * [ ] review


gberaudo:
Thanks

Issue #1786, opened 2016-09-01
Querying area with no feature leads to strange behaviour for layer group "Learning"


marionb:
1. Case (fixed with #1848)
  - Add theme Learning
  - make a box query on some of the bus stop features
  - make a box/point query on an area with no features

  ---> Endless loop for search result

2. Case
  - Add theme Learning
  - make many point queries close to a bus stop feature so that no query result is found
  - make a point query on a bus stop feature so that a query result is found

---> the result in the result grid is perturbed (compare result grid in the image below)
![queryonlearninglayer](https://cloud.githubusercontent.com/assets/1460598/18168327/65e2d138-7055-11e6-9415-8c0062ab58aa.png)

adube:
I'm currently unable to reproduce any of the described cases.  Could someone else try to see if she or he is able to reproduce the issues?  If so, then I would like to have more explanation how to reproduce them, please.

@sbrunner, would you please take care of this?

sbrunner:
Unable to reproduce, @marionb can you test it again?

marionb:
I can reproduce this. I also checker with @ger-benjamin who originally reported the issue and we could also reproduce it on his machine. 

-  I can reproduced the first part using used this [permalink](https://camptocamp.github.io/ngeo/master/examples/contribs/gmf/apps/desktop_alt/?lang=en&map_x=560193&map_y=141696&map_zoom=7&tree_groups=OSM%20functions%20mixed%2CLayers%2CGroup%2COSM%20functions&tree_group_layers_Layers=sustenance%2Centertainment%2Cosm_time%2Cpost_office%2Cpolice%2Ccinema&tree_group_layers_OSM%20functions=aster%2Csrtm%2Chalf_query%2Csustenance%2Centertainment%2Cbank%2Cosm_open%2Cosm_scale%2Cosm_time2%2Cosm_time) on Chrome and Firefox with the following steps:
 1. Do a point query on some of the green point.
  2. Do a box query

- The second part of the issue might be related to https://github.com/camptocamp/ngeo/issues/1833. It can be more easily reproducible on Chrome than Firefox although I have managed to see it appear on Firefox as well.

adube:
@marionb Thanks for the clarifications.  I'm able to reproduce the issue described in the first part locally so I'll be able to debug.

For the second part, that looks related to #1833, indeed.  Let's see if the issue disappears once that is fixed.

PR #1848, opened 2016-09-12
#1786 - Prevent query to do request if no queryable source

adube:
This PR fixes the query service. When no queryable source is found, it doesn't launch new queries, which prevents the spinner to be shown indefinitively.

Fixes part 1 of #1786.

## Todo

 * [x] review
 * [x] 2nd review

## Live demo

 * https://adube.github.io/ngeo/1786-fix-query/examples/contribs/gmf/apps/desktop_alt/index.html?lang=en&map_x=560193&map_y=141696&map_zoom=7&tree_groups=OSM%20functions%20mixed&baselayer_ref=asitvd.fond_gris&tree_enable_osm_time_r_dp=false&tree_enable_osm_time_v_dp=false&tree_enable_osm_time_v_s=false&tree_enable_osm_time_r_s=false&tree_enable_osm_scale=true

### DIFF
--- a/src/services/query.js
+++ b/src/services/query.js
@@ -518,7 +518,7 @@ ngeo.Query.prototype.doGetFeatureInfoRequests_ = function(
     items.forEach(function(item) {
       item['resultSource'].pending = true;
       item['resultSource'].queried = true;
-    }.bind(this));
+    });
 
     var infoFormat = items[0].source.infoFormat;
     var wmsGetFeatureInfoUrl = items[0].source.wmsSource.getGetFeatureInfoUrl(
--- a/src/services/query.js
+++ b/src/services/query.js
@@ -444,9 +444,6 @@ ngeo.Query.prototype.getQueryableSources_ = function(map, wfsOnly) {
         }
       }
 
-      item['resultSource'].pending = true;
-      item['resultSource'].queried = true;
-
       if (item.source.wfsQuery) {
         // use WFS GetFeature
         url = item.source.urlWfs || item.source.wmsSource.getUrl();
@@ -470,8 +467,6 @@ ngeo.Query.prototype.getQueryableSources_ = function(map, wfsOnly) {
           this.pushSourceIfUnique_(item, wmsItemsByUrl[url]);
         } else {
           // TODO - support other kinds of infoFormats
-          item['resultSource'].pending = false;
-          item['resultSource'].queried = false;
         }
       }
     }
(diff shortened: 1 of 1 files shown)


src/services/query.js
```diff
@@ -357,9 +357,22 @@ ngeo.Query.prototype.issue = function(map, object) {
...
+    doneRequest = true;
+  }
+
+  if (Object.keys(/** @type {!Object} */ (sources.wfs)).length) {
+    this.doGetFeatureRequestsWithCoordinate_(sources.wfs, coordinate, map);
+    doneRequest = true;
+  }
+
+  if (doneRequest) {
+    this.updatePendingState_();
```
tsauerwein: [updatePendingState](https://github.com/adube/ngeo/blob/cde75bab6989aa54f1415cb8c13f391831264b4a/src/services/query.js#L791-L799) checks if there are pending sources. Why did you have to introduce this variable `doneRequest`? Does it mean that there are still sources marked as `pending`? Which would be wrong, or not?

adube: I hadn't noticed that, you're right.  I'll make the corrections.

adube: Okay, I understand what's going on, now.  Originally, the code only supported WMS sources.  Once one was found to be "queryable", it was marked "pending" and a request was issued.  When WFS was implemented, the code that managed the getting of the queryable sources was moved to a method: `getQueryableSources_`.  That method still has the remains of what was done before, i.e. it marks found sources as `pending`.

In our case, we have only one source that doesn't support WFS.  So when obtaining it, it's marked as `pending` but in the end there's no request sent... but the `updatePendingState` doesn't behave correctly thereafter and we have the spinner being shown.

I think we should move the pending state of sources away from the  `getQueryableSources_` method.  I'll see what I can do.

adube:
Okay, I made the fix I was talking about.  To summarize, we set the 'pending' state just before sending a request for a source instead of when getting its info.

Live demo updated and ready for a second review.

src/services/query.js
```diff
@@ -519,6 +514,12 @@ ngeo.Query.prototype.doGetFeatureInfoRequests_ = function(
   var resolution = /** @type {number} */(view.getResolution());
 
   angular.forEach(wmsItemsByUrl, function(items) {
+
+    items.forEach(function(item) {
+      item['resultSource'].pending = true;
+      item['resultSource'].queried = true;
+    }.bind(this));
```
fredj: no need to `bind`

adube:
Fixed.  Ready for merge once travis agrees.

fredj:
@tsauerwein please merge if it ok for you

Issue #1841, opened 2016-09-12
Do not use id in MousePosition directive


adube:
In the `gmf-mouseposition` directive, an element defined in its html template has an `id`, which gets used in the controller in a `getElementById.getElementById` to get the element.  That is not a good practice, as the id is unique and if one wanted to have 2 of these directives in the same page then that wouldn't work.

The task at hand is about fixing this issue.

adube:
@gberaudo agrees, see: https://github.com/camptocamp/ngeo/pull/1816

adube:
@sbrunner This issue is similar to #1842.  Would you like me to work on it too?

sbrunner:
ok :-)

adube:
Fixed by #1849.

PR #1849, opened 2016-09-12
#1841 - Use class instead of id in MousePosition

adube:
This PR fixes the mouse position directive in gmf to no longer use an `id` of the element where to render the mouse position control.  Instead, a CSS class name is used with the angular `$element` service.

## Todo

 * [ ] review

## Live example

 * https://adube.github.io/ngeo/1841-mouseposition-id/examples/contribs/gmf/mouseposition.html


gberaudo:
Great!

PR #1850, opened 2016-09-12
Fix order of background layers

gberaudo:


sbrunner:
Looks good, thanks :-)

PR #1853, opened 2016-09-14
Avoid text flickering in display query window

### LINKED ISSUE #1694: Double text in query result
When sliding from one query result to the next one the title and the text of the next result covers the the text of the previous result. Then in a next step the previous result disappears and the next one shows properly.
This is happening on the mobile application for iPod Safari and iPad Safari and Chrome

pgiraud:
Fixes #1694.

Demo: https://pgiraud.github.io/ngeo/query_window_animation_flickr/examples/contribs/gmf/apps/mobile


fredj:
LGTM

PR #1854, opened 2016-09-14
Fix prev/next inactive state

pgiraud:
There were a regression introduced with this change https://github.com/camptocamp/ngeo/commit/867bf981ae3d7728e3697e5c369504c85c4b6b55#diff-df063e76ea0d193b20ec188cb6b50b55R68

The previous arrow never got inactive anymore.
screenshot from 2016-09-14 12 10 23

With the current pull request, I simplify the way we handle inactivation of the arrows by using already existing bootstrap's classes. The benefit is also that the button is really disabled (not only visually).

Demo: https://pgiraud.github.io/ngeo/query_window_arrows/examples/contribs/gmf/apps/mobile/?baselayer_ref=asitvd.fond_gris&lang=fr&map_x=558817&map_y=156629&map_zoom=2&tree_groups=Layers

### DIFF
--- a/contribs/gmf/less/displayquerywindow.less
+++ b/contribs/gmf/less/displayquerywindow.less
@@ -155,9 +155,6 @@
         content: @fa-var-chevron-right;
       }
     }
-    .gmf-displayquerywindow-inactive {
-      color: @main-bg-color;
-    }
     button {
       width: inherit;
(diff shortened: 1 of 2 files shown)


fredj:
nice simplification, +1

PR #1813, opened 2016-09-07
[WIP] - manage 'deactivation' with unsaved changes in gmf-editfeature

adube:
This is work in progress for #1741.  It currently doesn't fix the issue and needs more work.

PR #1856, opened 2016-09-14
Manage 'deactivation' with unsaved changes in gmf-editfeature

### LINKED ISSUE #1741: Manage unsaved modifications when gmf-editfeatureselector becomes inactive
When the gmf-editfeatureselector directive becomes inactive, it currently cancels everything without any confirmation, Any unsaved modifications are lost. We should have a confirmation message, like what we have when we click on the "cancel" button of the form.

Requires: #1727

## Details

We will need add some sort of "before deactivate" mechanism to prevent the activation of the other tools.  To verbalize: we need to prevent any tool that is member of the same group to activate if the tool being deactivated should be prevented to be deactivated, for example in the case of unsaved changes.

fredj:
replace #1813 
fixes #1741

### DIFF
--- a/contribs/gmf/src/directives/editfeature.js
+++ b/contribs/gmf/src/directives/editfeature.js
@@ -44,6 +44,7 @@ goog.require('ol.style.Text');
  * Example:
  *
  *     <gmf-editfeature
+ *         gmf-editfeature-dirty="ctrl.dirty"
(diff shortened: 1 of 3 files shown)


adube:
+1 for me.  Thanks a lot @fredj for taking care of the issue I've been having with this.

fredj:
Thanks for the review

PR #1862, opened 2016-09-15
WIP - Fix permalink for LayerTree refactor

adube:
Work still in progress.


adube:
@ger-benjamin This is the very first step towards fixing the permalink service to make it work with the refactoring of the layer tree.  See that I watch the tree ctrl references array from the tree manager.  I intent to make it public, i.e. mark it as `@export`.

The event is called once with the array full of new items and not once per item added (like in the ol.Collection).  So, I individually check each one of them that has been added or removed and register or unregister it accordingly.

From there, I watch the `state` property, which is currently private too.  I think I'll need to mark it public as well to be able to access it directly and watch it.

I'm currently there in my work.  My next step is to handle each kind of treeCtrl individually, i.e. make a big chunk of if statements to determine what to do with the change.

Feel free to comment.

ger-benjamin:
That's not too important, but I don't understand why the state should be public. You can use the getState method no ?

adube:
>That's not too important, but I don't understand why the state should be public. You can use the getState method no ?

In Angular, can to watch the result of a called method?  I haven't tried, but I think you need to listen a property, right?

ger-benjamin:
You can pass two parameters, one value or function to watch and another function to handle the change. So yes I think that should work.
Like there https://github.com/camptocamp/ngeo/blob/master/contribs/gmf/src/directives/print.js#L359
That should be tried.

adube:
Oh, cool!  I'll try that, thanks.

adube:
Replaced by: #1867

PR #1868, opened 2016-09-19
Add public getTreeCtrlReferences method

adube:
This PR adds a method to the treemanager to return the Layertree controllers kept as reference.  The goal of this is to make possible to watch the array for newly added or removed elements.


adube:
@ger-benjamin Ready for review.

PR #1875, opened 2016-09-20
Handle more branches in gh-pages branches cleanup

pgiraud:
When pushing branches to github pages, the `gh-pages` Makefile target also cleans up local repo by removing branches that are not in the remote repo (for example when they've been deleted).
Unfortunately we rely on a request on the github api which returns only a fragment of the branches list because of pagination. This results in branches not being sent to gh-pages because the list is truncated.

With this pull request I make sure that at least 100 branches are supported.

Please review.

### DIFF
--- a/buildtools/cleanup-ghpages.py
+++ b/buildtools/cleanup-ghpages.py
@@ -13,7 +13,8 @@ urllib3.disable_warnings()
 
 def main():
     try:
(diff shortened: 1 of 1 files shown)


sbrunner:
Thanks merge when Travis is OK :-)

PR #1878, opened 2016-09-20
[Work in progress] EditFeature Snapping

adube:
Work still in progress.


adube:
The first step towards supporting snapping is to listen to the tree manager added/removed Layertree controllers.  Those that are 'snappable' and that have the 'on' state are added to an array.  This array will be shared with the `gmf-editfeature` directive.   From there, the "SnappingHandler" will be created.

This is done this way to allow snappable layers to be turned on/off while an other layer is being edited.

PR #1882, opened 2016-09-21
Add an option to bypass the local storage

fredj:
Add a new options to gmfx.PermalinkOptions to deactivate to storage into the local storage.
The desktop_alt now uses this option.

### DIFF
--- a/contribs/gmf/src/services/permalink.js
+++ b/contribs/gmf/src/services/permalink.js
@@ -19,6 +19,7 @@ goog.require('ngeo.format.FeatureHash');
 goog.require('ngeo.WfsPermalink');
 goog.require('goog.asserts');
 goog.require('ol.Feature');
+goog.require('ol.functions');
 goog.require('ol.geom.Point');
 goog.require('ol.proj');
 goog.require('ol.layer.Group');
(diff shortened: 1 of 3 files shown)


contribs/gmf/src/services/permalink.js
```diff
@@ -146,6 +146,10 @@ gmf.Permalink = function($timeout, ngeoBackgroundLayerMgr, ngeoDebounce,
    */
   this.ngeoStateManager_ = ngeoStateManager;
 
+  if (gmfPermalinkOptions.useLocalStorage === false) {
+    this.ngeoStateManager_.localStorage = null;
```
gberaudo: Maybe use the hash `{ 'isAvailable': function() {return false;}}`?

fredj: Or even `this.ngeoStateManager_.localStorage.isAvailable = ol.functions.FALSE;` ?

fredj: ping @gberaudo

gberaudo: Yes anyway is +1. Thanks for asking.

PR #1814, opened 2016-09-07
Layertree refactor

ger-benjamin:
Fix: #1752 

Main changes:
- There is a state in each node of the layertree.
- Set state of a layertree and ask to sync will sync the layers based on the tree.
- The creation of layers was totally refactored (a layer is now created / managed when the corresponding layertree is created).
- A syncLayertreeMap service create the layers and sync the tree.
- Externs are ready to reflect now the server response.
- Layertree is now well tested. 
- We easily can get a ngeo.Layeree based on it's node id (but it retrieve the first found !) (then active or deactive it's state, like that: https://github.com/camptocamp/ngeo/blob/layertree_refactor/contribs/gmf/src/services/treemanager.js#L357-L375)

# TODO for this PR

## Help needed
**Features that must be fixed before merge or in other issues**:
 - Permalink (strange behaviors with multiple groups, try with the edit theme)
 - ~~Edition (problem cause now, layers are created/updated one by one)~~ fixed :-)

# TODO After this PR

## Issues opened
 - #1885

## To check / improve

We probably can move some things in the syncLayertreeMap service. Like:
 - Dimension management (currently in gmf.Layertree)
 - Opacity (currently in gmf.Layertree)

Perhaps, things can be simplified in these files (files where there was gmfThemesNode object):
 - contribs/gmf/service/querymanager.js
 - contribs/gmf/service/themeservice.js
 - contribs/gmf/service/treemanager.js
 - contribs/gmf/service/permalink.js
 - contribs/gmf/directive/themeselector.js
 - contribs/gmf/directive/layertree.js
 - contribs/gmf/directive/editfeatueselector.js
 - contribs/gmf/directive/editfeatue.js

Others files to check: 
 - src/service/layerhelper.js
 - src/directive/layertree.js

And A lot of `gmfWmsUrl` contantes exist, it's not longer used like in layertree. Must be removed

PR #1883, opened 2016-09-21
Fix editfeature dir. layer getting

adube:
This PR fixes how the OpenLayers layer is obtained in the `gmf-editfeature` directive.  It used to only check the parent treeCtrl, but the layer may be in a higher level group (i.e. if we have groups inside groups).

To fix this, a new static method has been added to the gmf layertree controller.  It can be used to find a layer object from a ngeo treeCtrl that has a leaf as node (because I think it only makes sense to work from leaves when you want to know the associated layer, you don't really care about the ol.layer.Group).

### DIFF
--- a/contribs/gmf/src/directives/editfeature.js
+++ b/contribs/gmf/src/directives/editfeature.js
@@ -4,6 +4,7 @@ goog.provide('gmf.editfeatureDirective');
 
 goog.require('gmf');
 goog.require('gmf.EditFeature');
+goog.require('gmf.SyncLayertreeMap');
 goog.require('gmf.XSDAttributes');
 /** @suppress {extraRequire} */
 goog.require('ngeo.attributesDirective');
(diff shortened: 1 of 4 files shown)


adube:
Ready for review.  @pgiraud: if you'd like to take a look...

ger-benjamin:
Mhhh. That's a little bit awkward... I just remember that I've already create a function that get the layer.
See there: https://github.com/camptocamp/ngeo/pull/1814/files#diff-0a312b12ae20110929a33176555f1dd6R395

Even if the principle is different, in practice, the result will be always the same, no ?

adube:
Oh, you're right.  Nevermind this, then.  I'll simply use this function instead.

adube:
Oops, hold on.

adube:
Okay, ready for review.

ger-benjamin:
Okay for me. I merge

PR #1886, opened 2016-09-21
Make sure map is resized at the end of animation

pgiraud:
In https://github.com/Geoportail-Luxembourg/geoportailv3/pull/1444 it was reported that the resize of the map doesn't work on some computers.
The following patch ensures that the map is correctly resized at least at the end of the animation.

Please review.


fredj:
@gberaudo did you had the same issue with the epfl project ?

gberaudo:
Thanks @pgiraud, I will test it.

PR #1906, opened 2016-09-30
Change user icon when logged in

pgiraud:
A tiny modification which will help the user know that he/she's authenticated.
Demo
https://pgiraud.github.io/ngeo/user_icon/examples/contribs/gmf/apps/desktop


kalbermattenm:
That's cool !

But when "hoovering" the button you get "Se connecter"... Could it be the user name, or if too complicated, just rename the button translation in something more neutral, like "Paramètres de connexion" (or whatever...)

presse-papiers-1

pgiraud:
I totally agree with you @kalbermattenm 
I tried to solve this issue as well, but couldn't find any solution because of translation issues.

sbrunner:
@fredj should we merge it?

PR #1907, opened 2016-09-30
Layertree refactor permalink

sbrunner:


sbrunner:
Know issues:
 * click on group `couches` => select only `Cinéma`
 * the `initLayers_` method is called two times ...

PR #1930, opened 2016-10-06
Make having a blank background layer optional

gberaudo:


sbrunner:
Looks good :-)

gberaudo:
Thanks for the review.

PR #1933, opened 2016-10-07
Include the dimensions values in the query params

fredj:


src/services/query.js
```diff
@@ -539,6 +545,17 @@ ngeo.Query.prototype.doGetFeatureInfoRequests_ = function(
...
         goog.uri.utils.setParam(wmsGetFeatureInfoUrl, 'QUERY_LAYERS', lyrStr);
 
+    // add dimensions values
+    var dimensions = items[0].source.dimensions;
+    if (dimensions) {
+      for (var key in dimensions) {
+        var value = this.dimensions[key];
+        if (value !== undefined) {
+          wmsGetFeatureInfoUrl = goog.uri.utils.setParam(wmsGetFeatureInfoUrl, key, value);
+        }
```
sbrunner: if the value is undefined and it's defined in `items[0].source.dimensions` shouldn't we get the value from `items[0].source.dimensions`?

fredj: yes, I will fix this

fredj:
thanks for the review

PR #1948, opened 2016-10-14
Reorder delete

### LINKED ISSUE #1744: All groups are deleted after moving a group

pgiraud:
Fixes #1744 

This pull request aim is to fix the problem with layertree DOM elements being messed up after a reordering (drag&drop).
The main issue is due to the fact that the `ng-repeat` directive relies on HTML comments to work properly.

In order to get the comments and DOM elements being correctly ordered it is mandatory that the variable used as configuration for `ng-repeat` is updated after the reorder. The simplest solution is to use the same variable.
https://github.com/camptocamp/ngeo/commit/91e7b9a74956e1db99a3860bb3dfecf697849674

Once done, there's no issue with the deletion of groups anymore. However, the order of layers in the map is updated.

In order to do that, I watch for changes in the order of `root.children` (list of children config, used in the ng-repeat), then update the `rootCtrl.children` to apply the same order change. We can then loop into the first level groups, get the layers and update the gmfLayerTreeCtrl.layers which will eventually change the order of layers in the map.

https://github.com/camptocamp/ngeo/commit/a179cb77555390a9dde98716694ee80bf03cb9b1

Note : In my opinion, this should not be so complex.
Note 1 : I'm not sure that the permalink is updated when the order is changed.

### DIFF
--- a/contribs/gmf/src/directives/layertree.js
+++ b/contribs/gmf/src/directives/layertree.js
@@ -246,6 +246,32 @@ gmf.LayertreeController = function($http, $sce, $scope, ngeoCreatePopup,
     this.map.render();
   }.bind(this));
 
+  // watch any change on root children to update the layers on the map
+  $scope.$watchCollection(function() {
+    return this.gmfTreeManager_.root.children;
+  }.bind(this),
+  function(newValue, oldValue) {
+    // we want to watch for order change only
+    if (newValue != oldValue &&
+        newValue.length == oldValue.length &&
+        this.gmfTreeManager_.rootCtrl) {
+      // create a new children array with the new order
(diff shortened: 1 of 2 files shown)


ger-benjamin:
The order on the node is normal.

But your watch looks complex. Why don't use the "sync" method in the gmfSyncLayertreeMap ? I'll do some try

ger-benjamin:
That's not even _so simple_... the PR don't works with permalink and order ont he map is strange... I try to correct that...

PR #1954, opened 2016-10-17
Layertree order delete

ger-benjamin:
Replace: #1948 
Fix: #1744 
Example: https://ger-benjamin.github.io/ngeo/layertree_order_delete/examples/contribs/gmf/apps/desktop_alt

(More explanation in the PR of pgiraud)
- The order is now done on the same object that the ng-repeat build the layertree (Group node of the root treeCtrl)
- There is now a callback function that is called after the move end of the sortable directive
- Via that, after move end, the gmf layertree update the treeCtrls order and layers from the map

Also, there is a problem with the permalink but it's not bound with this issue I think.

### DIFF
--- a/contribs/gmf/src/directives/layertree.js
+++ b/contribs/gmf/src/directives/layertree.js
@@ -246,32 +246,6 @@ gmf.LayertreeController = function($http, $sce, $scope, ngeoCreatePopup,
     this.map.render();
   }.bind(this));
 
-  // watch any change on root children to update the layers on the map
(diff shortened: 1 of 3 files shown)


ger-benjamin:
Can I merge ?

ger-benjamin:
:octocat:

sbrunner:
@fredj ?

fredj:
looking at it now

ger-benjamin:
Thanks @fredj

Issue #1931, opened 2016-10-06
Missing documentation


sbrunner:
This file is not visible in the documentation ...
https://github.com/camptocamp/ngeo/blob/master/contribs/gmf/externs/gmf-themes.js

ger-benjamin:
It is. But there is no link clear in the left menu.

Look at the end of the "global" page: http://localhost:3000/.build/apidoc/global.html#GmfThemesResponse

sbrunner:
Is there a reason that we don't had a namespace in this file?

ger-benjamin:
Not especially. What would you like ? "gmx" as the others ?

sbrunner:
Yes (gmfx or gmf-themes) I suspect that with that it will directly be displayed in the menu :-)

PR #1955, opened 2016-10-17
Prefix gmfthemes

ger-benjamin:
Fix:  #1931 
Example: https://ger-benjamin.github.io/ngeo/prefix_gmfthemes/apidoc/

That's a little strange that almost all gmfThemes.* are _class_ and not _type_ . I've should leave theme defined as _class_ because _type_ doesn't support the `@extend` tag / function (see apidoc and commit 2 to understand)


sbrunner:
Thanks, for me the css of our docs isn't so good but that's an other story ...

You can merge when Travis tests pass :-)

PR #1958, opened 2016-10-18
Update tested themes.json

### LINKED ISSUE #1885: Some tests are based on an old 'Themes' server response object...
And don't really work anymore.

So, some tests:
 - contribs/gmf/test/spec/services/querymanager.spec.js
 - contribs/gmf/test/spec/services/treemanager.spec.js
 - contribs/gmf/test/spec/services/permalinkservice.spec.js
 - contribs/gmf/test/spec/services/treemanager.spec.js

(`git grep FIXME`)
are based on a old version of the server response 'themes' object. (A version without ogc_server among other things). Obviously some test don't pass with the current version of the code.

This old version is in the file `contribs/gmf/test/spec/data/old_themes.js`
It must be as least use the `contribs/gmf/test/spec/data/themes.js` which is up to date (21 septembre).

Then, if we want that tests are useful, we should use a real server response and not use a static file. Or find a method to update the fake server response when the database change...

ger-benjamin:
Fix  #1885 

Done:
 1. Tested themes.json is up to date (18 october 2016)
 1. The old_themes is no more used (is removed)
 1. A lot of now useless FIXME (in tests) have been removed
 1. Debug contribs/gmf/test/spec/services/querymanager.spec.js (on test is still in FIXME but that's out of this PR scope)

Question:
How can we do that ? _Then, if we want that tests are useful, we should use a real server response and not use a static file. Or find a method to update the fake server response when the database change..._
Just by update this themes.json file manually every month ?


sbrunner:
Thanks a lot :-)

> Question...

Good question...

PR #1963, opened 2016-10-21
Return parent node iconUrl if defined

fredj:
work in progress

apps to be tested


fredj:
thanks for the review

Issue #1966, opened 2016-10-24
Print crashes when printing a rotated text


ybolognini:
* Add a text (drawing)
 * Rotate it with slider
 * Print zone with the text
 * Print crashes

PR #1974, opened 2016-10-25
Fix printing rotated labels

tsauerwein:
Closes https://github.com/camptocamp/ngeo/issues/1966

With https://github.com/mapfish/mapfish-print/pull/468 this change is no longer required, but this avoids having to update MFP/GMF.


tsauerwein:
Travis is failing because the demo server is returning errors in the themes service:

```
JavaScript ERROR: The themes contain some errors:

The WMS layer name 'point', can't be two times in the same block (first level group).

The WMS layer name 'polygon', can't be two times in the same block (first level group).

The WMS layer name 'line', can't be two times in the same block (first level group).
```
https://travis-ci.org/camptocamp/ngeo/jobs/170431139#L2580

Could you please take a look, @sbrunner?

sbrunner:
I'm looking about that :-)

tsauerwein:
Thanks!

PR #1995, opened 2016-11-01
#2600 - Add ObjectEditing Hub example

adube:
This PR features a "Hub" example that will be the point of entry to open examples and applications that use the ObjectEditing tools.

**Note** For now, the example opens the authentication example.  This will be changed when we have an example featuring the ObjectEditing tools.

## Todo

 * [ ] Review

## Live example

 * https://adube.github.io/ngeo/oe-2600-hub-example/examples/contribs/gmf/objectediting-hub.html


adube:
@fredj Ready for review.

fredj:
looks good, thanks

PR #1999, opened 2016-11-02
ObjectEditing manager service

adube:
This PR introduces the ObjectEditing manager service, which is responsible of looking for parameters in the url, use them to build a request to the WFS Protocol to fetch a feature to use for the ObjectEditing tool.

The returned feature is also used within the Permalink service, i.e. if an ObjectEditing feature is returned, then the map is centered on it.  Otherwise, the default centering takes place.

## Todo

 * [x] Review

## Live demo

Once the hub page is completed loading, click on the "Launch" button.

 * https://adube.github.io/ngeo/oe-2593-oe-manager/examples/contribs/gmf/objectediting-hub.html

You can test with separate layers / features.  You can also try to modify the url of the opened page to see how it reacts to any parameter missing or if the query returns nothing.

### DIFF
--- a/contribs/gmf/examples/objectediting-hub.js
+++ b/contribs/gmf/examples/objectediting-hub.js
@@ -52,7 +52,7 @@ app.MainController = function($http, $q, $scope, gmfThemes) {
    * @export
    */
   this.urls = [
-    'authentication.html'
+    'objectediting.html'
   ];
 
   /**
@@ -179,7 +179,7 @@ app.MainController = function($http, $q, $scope, gmfThemes) {
 app.MainController.prototype.run = function() {
 
   var feature = this.selectedFeature;
-  var layer = this.selectedGmfLayerNode.name;
+  var layer = this.selectedGmfLayerNode.id;
   var property = 'name';
   var id = feature.get(property);
 
--- a/contribs/gmf/examples/objectediting-hub.js
+++ b/contribs/gmf/examples/objectediting-hub.js
@@ -4,6 +4,7 @@ goog.require('ngeo.proj.EPSG21781');
 goog.require('gmf');
 goog.require('gmf.ObjectEditingManager');
 goog.require('gmf.Themes');
+goog.require('gmf.XSDAttributes');
 goog.require('ol.format.WFS');
 
 
@@ -19,14 +20,19 @@ app.module.value('gmfTreeUrl',
     'https://geomapfish-demo.camptocamp.net/2.1/wsgi/themes?version=2&background=background');
 
 
+app.module.value('gmfLayersUrl',
+    'https://geomapfish-demo.camptocamp.net/2.1/wsgi/layers/');
+
+
 /**
  * @param {angular.$http} $http Angular $http service.
  * @param {angular.$q} $q Angular $q service.
  * @param {!angular.Scope} $scope Angular scope.
  * @param {gmf.Themes} gmfThemes The gmf themes service.
+ * @param {gmf.XSDAttributes} gmfXSDAttributes The gmf XSDAttributes service.
  * @constructor
  */
-app.MainController = function($http, $q, $scope, gmfThemes) {
+app.MainController = function($http, $q, $scope, gmfThemes, gmfXSDAttributes) {
 
   /**
    * @type {angular.$http}
@@ -45,6 +51,12 @@ app.MainController = function($http, $q, $scope, gmfThemes) {
    * @private
    */
   this.gmfThemes_ = gmfThemes;
+
+  /**
+   * @type {gmf.XSDAttributes}
+   * @private
+   */
+  this.gmfXSDAttributes_ = gmfXSDAttributes;
 
   /**
    * @type {Array.<string>} List of example and application urls that contain
@@ -103,6 +115,18 @@ app.MainController = function($http, $q, $scope, gmfThemes) {
    */
   this.selectedFeature = null;
 
+  /**
+   * @type {Object.<number, string>}
+   * @export
+   */
+  this.geomTypeCache_ = {};
+
+  /**
+   * @type {?string}
+   * @export
+   */
+  this.selectedGeomType = null;
+
   $scope.$watch(
     function() {
       return this.selectedGmfLayerNode;
@@ -113,6 +137,9 @@ app.MainController = function($http, $q, $scope, gmfThemes) {
       if (newVal) {
         this.getFeatures_(newVal).then(
           this.handleGetFeatures_.bind(this, newVal)
+        );
+        this.getGeometryType_(newVal).then(
+          this.handleGetGeometryType_.bind(this, newVal)
         );
       }
     }.bind(this)
@@ -178,12 +205,14 @@ app.MainController = function($http, $q, $scope, gmfThemes) {
  */
 app.MainController.prototype.run = function() {
 
+  var geomType = this.selectedGeomType;
   var feature = this.selectedFeature;
   var layer = this.selectedGmfLayerNode.id;
   var property = 'name';
   var id = feature.get(property);
 
   var params = {};
+  params[gmf.ObjectEditingManager.Param.GEOM_TYPE] = geomType;
   params[gmf.ObjectEditingManager.Param.ID] = id;
   params[gmf.ObjectEditingManager.Param.LAYER] = layer;
   params[gmf.ObjectEditingManager.Param.THEME] = this.themeName;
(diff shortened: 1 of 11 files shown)


adube:
@fredj about: https://github.com/camptocamp/ngeo/pull/1999/commits/127fdbbd40f3a573d374959094509d321293556d

The EditFeature service uses the layers url.  All methods that were there before require to have url set, so instead of doing a dummy promise for them, I added a `goog.asserts.assert` call.  For the new `getFeaturesWithComparisonFilters`, it's the one that's called by the ObjectEditing service, which is called by the permalink all the time.  That one I allow to have the url undefined and return a dummy promise as if a query had been made.

Ready for review.

adube:
There's currently an error with the tests that also occurs on the `master` branch:

```
PhantomJS 2.1.1 (Linux 0.0.0) gmfWMSTime service should format the time regarding the resolution and with a mode set on value FAILED
	Expected '2013' to be '2014'.
	contribs/gmf/test/spec/services/wmstime.spec.js:30:27
	Expected '2013-12' to be '2014-01'.
	contribs/gmf/test/spec/services/wmstime.spec.js:36:27
	Expected '2013-12-31' to be '2014-01-01'.
	contribs/gmf/test/spec/services/wmstime.spec.js:42:27
PhantomJS 2.1.1 (Linux 0.0.0) gmfWMSTime service should format the time regarding the resolution and with a mode set on range FAILED
	Expected '2013/2015' to be '2014/2015'.
	contribs/gmf/test/spec/services/wmstime.spec.js:60:27
	Expected '2013-12/2015-12' to be '2014-01/2015-12'.
	contribs/gmf/test/spec/services/wmstime.spec.js:67:27
	Expected '2013-12-31/2015-12-30' to be '2014-01-01/2015-12-31'.
	contribs/gmf/test/spec/services/wmstime.spec.js:74:27
PhantomJS 2.1.1 (Linux 0.0.0): Executed 212 of 212 (2 FAILED) (2.09 secs / 1.896 secs)
```

adube:
@sbrunner Would you please take care of reviewing this?

adube:
Update: I added the `geom_type` parameter in order to be able to avoid using the theme service in the gmf-objectediting directive.

contribs/gmf/src/services/editfeature.js
```diff
@@ -26,21 +27,36 @@ gmf.EditFeature = function($http, gmfLayersUrl) {
...
+  /**
+   * Url to the GeoMapFish layers service. Required in applications that use:
+   * - the editfeature tools
+   * - the objectediting tools
+   *
+   * @type {?string}
+   * @private
+   */
+  this.baseUrl_ = $injector.has('gmfLayersUrl') ?
+    $injector.get('gmfLayersUrl') : null;
```
sbrunner: The gmfLayerUrl is optional?

contribs/gmf/src/services/editfeature.js
```diff
@@ -26,21 +27,36 @@ gmf.EditFeature = function($http, gmfLayersUrl) {
...
+ * of layer ids inside a specified extent.
+ *
  * @param {Array.<number>} layerIds List of layer ids to get the features from.
  * @param {ol.Extent} extent The extent where to get the features from.
  * @return {angular.$q.Promise} Promise.
  * @export
  */
-gmf.EditFeature.prototype.getFeatures = function(layerIds, extent) {
+gmf.EditFeature.prototype.getFeaturesInExtent = function(layerIds, extent) {
+  goog.asserts.assert(this.baseUrl_, 'GMF layers url must be defined');
```
sbrunner: It seem not...

adube: I had previously discussed about this with @fredj.  The `gmf.EditFeature` service is now injected in the `gmf.ObjectEditingManager`, which is injected in the `gmf.Permalink`.  This would make the `gmfLayerUrl` required to be added in **all** examples.

Since we didn't want to do that, we made it optional.  You **have to** set it when you need to use methods such as `getFeaturesInExtent`, but you can leave it undefined otherwise.

sbrunner: Than it's not here that we have an optional injection, it's probably in the permalink with the `EditFeature` ...

adube: I kinda disagree.  The property to set is `gmfLayerUrl`, which is used in the `gmf.EditFeature`.  It seems simpler to manage where's it's needed instead of one level above, in my opinion.  What do you think?

contribs/gmf/examples/objectediting-hub.js
```diff
@@ -268,6 +297,74 @@ app.MainController.prototype.getFeaturesFromCache_ = function(gmfLayerNode) {
...
+  var geomType = /** @type {string} */ (
+    this.getGeometryTypeFromCache_(gmfLayerNode));
+  this.selectedGeomType = geomType;
+};
+
+
+/**
+ * @param {gmfThemes.GmfLayerWMS} gmfLayerNode Layer node.
+ * @return {?string} The type of geometry.
+ * @export
```
gberaudo: Why is this private method `@export`? Shouldn't it be `@private` instead?

contribs/gmf/src/services/permalink.js
```diff
@@ -72,6 +73,8 @@ gmf.module.value('gmfPermalinkOptions',
  * @param {ol.Collection.<ol.Feature>} ngeoFeatures Collection of features.
  * @param {ngeo.LayerHelper} ngeoLayerHelper Ngeo Layer Helper.
  * @param {ngeo.StateManager} ngeoStateManager The ngeo StateManager service.
+ * @param {gmf.ObjectEditingManager} gmfObjectEditingManager The gmf
```
gberaudo: This could be optional.

contribs/gmf/src/services/editfeature.js
gberaudo: See https://github.com/camptocamp/ngeo/pull/1999/files#r86995638 which depends on gmfEditFeature.

contribs/gmf/src/services/editfeature.js
```diff
@@ -52,6 +68,56 @@ gmf.EditFeature.prototype.getFeatures = function(layerIds, extent) {
...
+ * the permalink service, i.e. it's always called. Since we don't have to
+ * define the url to the GMF Protocol (layers) a dummy promise returns an
+ * empty array of features if the url is not defined.
+ *
+ * @param {Array.<number>} layerIds List of layer ids to get the features from.
+ * @param {Array.<gmfx.ComparisonFilter>} filters List of comparison filters
+ * @return {angular.$q.Promise} Promise.
+ * @export
+ */
+gmf.EditFeature.prototype.getFeaturesWithComparisonFilters = function(
```
gberaudo: Should this method be accessed from a partial? If not it should not be marked `@export`.

contribs/gmf/src/services/objecteditingmanager.js
```diff
@@ -1,5 +1,126 @@
...
+    var id = this.ngeoLocation_.getParam(
+      gmf.ObjectEditingManager.Param.ID);
+    var layer = this.ngeoLocation_.getParam(
+      gmf.ObjectEditingManager.Param.LAYER);
+    var property = this.ngeoLocation_.getParam(
+      gmf.ObjectEditingManager.Param.PROPERTY);
+    var theme = this.ngeoLocation_.getParam(
+      gmf.ObjectEditingManager.Param.THEME);
+
+    if (id && layer && property && theme) {
```
gberaudo: OK, these are strings.

gberaudo:
I added a few remarks. Please merge when addressed.

contribs/gmf/examples/objectediting-hub.js
```diff
@@ -268,6 +297,74 @@ app.MainController.prototype.getFeaturesFromCache_ = function(gmfLayerNode) {
...
+
+  return this.getGeometryTypeDeferred_.promise;
+};
+
+
+/**
+ * @param {gmfThemes.GmfLayerWMS} gmfLayerNode Layer node.
+ * @export
+ */
+app.MainController.prototype.issueGetAttributesRequest_ = function(
```
gberaudo: Should it be `@private` here?

contribs/gmf/examples/objectediting-hub.js
```diff
@@ -268,6 +297,74 @@ app.MainController.prototype.getFeaturesFromCache_ = function(gmfLayerNode) {
 
 
 /**
+ * @param {gmfThemes.GmfLayerWMS} gmfLayerNode Layer node.
+ * @return {angular.$q.Promise} The promise attached to the deferred object.
+ * @export
```
gberaudo: Idem, should it be @private here?

adube:
@gberaudo The build fails with other examples.  Do you have a hint how to fix this?  Thanks.

gberaudo:
@sbrunner is looking into it. The build tool is incorporating some unwanted files.

adube:
Understood.  Please, let me know when I can rebase.

Issue #1977, opened 2016-10-25
Layers in mapserver groups can't be queried


ger-benjamin:
Layers in a **MapServer** GROUP can't be queried.

Try with theme: _OSM_, first-level-group: _Group_ layer: _OSM_. At least the _bank_ layer should be queried, but nothing is queried.
I think (but i'm not comfortable with the ngeo query service) that's because query sources are only based on final layer (like bank) and this layer is not directly on the map.

(You can also try the bank layer directly to see that the query system works on this layer, try it in theme: _OSM_, first-level-group: _OSM function_, layer: _layer with a very very very very very long name_)

https://geomapfish-demo.camptocamp.net/2.1/theme/OSM

ger-benjamin:
@adube 
The problem come from here: https://github.com/camptocamp/ngeo/blob/391fe66920a765c6d4c02f24fe9591ea87b55f0f/src/services/query.js#L426-L452

Precisely, at line 447. In my case, the "cfgLayer" (queried source) value is `group` and the "layerLayers" (existing layer) value are `hotel,bank,place_of_worship`. Both doesn't match but one is part of the other and must match. (That's a Mapserver group)

I don't understand why we need that `item.source.validateLayerParams` check. Perhaps we can just removesit. Or should we get the group layer of `hotel,bank,place_of_worship`.
What should I do ?

adube:
@ger-benjamin This might help: https://github.com/camptocamp/ngeo/blob/master/externs/ngeox.js#L514-L531

Otherwise, you can contact me.

ger-benjamin:
Thanks adube

PR #2000, opened 2016-11-03
Can query map server groups

ger-benjamin:
Fix: #1977

Example: https://ger-benjamin.github.io/ngeo/query_mapserver_groups/examples/contribs/gmf/apps/desktop/

(Before it takes only the first childlayer and tests if it's on the map. Now it does this check on the layer)

(2000th PR :tada:  )

### DIFF
--- a/src/services/query.js
+++ b/src/services/query.js
@@ -441,12 +441,12 @@ ngeo.Query.prototype.getQueryableSources_ = function(map, wfsOnly) {
             'validateLayerParams option.'
         );
         var layerLayers = layerSource.getParams()['LAYERS'].split(',');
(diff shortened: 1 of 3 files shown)


src/services/query.js
```diff
@@ -441,12 +441,12 @@ ngeo.Query.prototype.getQueryableSources_ = function(map, wfsOnly) {
             'validateLayerParams option.'
         );
         var layerLayers = layerSource.getParams()['LAYERS'].split(',');
-        var cfgLayer = item.source.wmsSource.getParams()['LAYERS'].split(',');
+        var cfgLayer = item.source.layers.split(',');
```
ger-benjamin: Done

PR #2002, opened 2016-11-04
Encode values for fragment parameters

tsauerwein:
When using values like `6+,7a+` as fragment parameters, the values have to be url encoded/decoded.

### DIFF
--- a/src/services/location.js
+++ b/src/services/location.js
@@ -4,6 +4,7 @@ goog.provide('ngeo.MockLocationProvider');
 goog.require('goog.Uri');
 goog.require('goog.object');
 goog.require('ngeo');
+goog.require('ngeo.string');
 
 
 /**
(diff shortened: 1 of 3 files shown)


src/services/location.js
```diff
@@ -128,11 +128,12 @@ ngeo.Location.prototype.getParam = function(key) {
...
  * Get a param from the fragment of the location's URI.
  * @param {string} key Param key.
- * @return {string} Param value.
+ * @return {string|undefined} Param value.
  * @export
  */
 ngeo.Location.prototype.getFragmentParam = function(key) {
-  return /** @type {string} */ (this.getFragmentUri_().getQueryData().get(key));
+  var val = /** @type {string} */ (this.getFragmentUri_().getQueryData().get(key));
+  return val !== undefined ? goog.string.urlDecode(val) : undefined;
```
gberaudo: Can we get ride of the goog library?
See https://github.com/google/closure-library/blob/master/closure/goog/string/string.js#L527

tsauerwein: I added a new module `ngeo.string`. There are still other places where `goog.string` is used, but I don't want to handle it in this PR.

tsauerwein:
Thanks for the reviews!

PR #2009, opened 2016-11-10
ObjectEditing manager service

sbrunner:
Replace #1999 

Requires https://github.com/openlayers/ol3/pull/6098


ger-benjamin:
@sbrunner , so, can we close the PR #1999 ?

src/services/layerHelper.js
```diff
@@ -318,11 +321,12 @@ ngeo.LayerHelper.prototype.isLayerVisible = function(layer, map) {
  * @param {ol.layer.Image|ol.layer.Tile} layer Layer to refresh.
  */
 ngeo.LayerHelper.prototype.refreshWMSLayer = function(layer) {
-  var source = layer.getSource();
+  var source_ = layer.getSource();
```
ger-benjamin: Why do you do things like this ?

test/spec/directives/layertree.spec.js
```diff
@@ -41,7 +41,7 @@ describe('ngeo.layertreeDirective', function() {
       return new ol.layer.Image();
     };
 
-    inject(function($rootScope, $compile, $sce) {
+    inject(function($rootScope, $compile, $sce, $httpBackend) {
```
ger-benjamin: ... or this ? (It seems not used)

ger-benjamin:
That's very hard to review. Changes seem interrelated...

One question more: Now, ngeo example and gmf example provide "app.component" (instead of a ngeo.component and a gmf.component). Is this system functional if I provide, for example a component "test"  in ngeo AND in gmf ? ( My ngeo example file `test.js` will be provided as `app.test` AND my gmf example file will also be provided as `app.test`, That will create an error no ?)

sbrunner:
> My ngeo example file `test.js` will be provided as `app.test` AND my gmf example file will also be provided as `app.test`, That will create an error no ?

Yes!

Makefile
```diff
@@ -5,8 +5,8 @@ GMF_DIRECTIVES_PARTIALS_FILES := $(shell ls -1 contribs/gmf/src/directives/parti
 
 OS := $(shell uname)
 
-EXAMPLES_JS_FILES := $(shell find examples -maxdepth 1 -type f -name '*.js')
 EXAMPLES_HTML_FILES := $(shell find examples -maxdepth 1 -type f -name '*.html')
+EXAMPLES_JS_FILES := $(EXAMPLES_HTML_FILES:.html=.js)
```
sbrunner: Yes, I do that because if I don't do that the https.js will be compiled and failed...

src/services/layerHelper.js
sbrunner: To have a casted version of source bellow...

test/spec/directives/layertree.spec.js
sbrunner: This one looks to be en error ...

ger-benjamin:
Thanks merge if the tests pass

PR #2032, opened 2016-11-17
ngeox and gmfx are not real externs

### LINKED ISSUE #2026: ngeox.js and gmfx.js are not extern files ...
move `externs/ngeox.js` to `src/typedef.js`,
and `contribs/gmf/externs/gmfx.js` to `contribs/gmf/src/typedef.js`.

sbrunner:
should fix #2026


contribs/gmf/src/typedefs.js
```diff
@@ -1,11 +1,10 @@
 /**
- * Externs for GeoMapFish
- *
+ * This file contains the typedefs of the options of the methods.
+ * He can be includes as externs if you needs that not renamed.
```
gberaudo: It can be included as extern if you want to prevent renaming.

sbrunner:
> Can't we remove the @externs tag in option/ngeox and contribs/gmf/options/gmfx ?

No, if we remove it it's not possible to use it as extern...

PR #2035, opened 2016-11-17
ObjectEditing - show list of queryable layers only when needed

adube:
This PR ensures that the list of querable layers is only visible when a tool that requires it is activated.


adube:
Ready for review.

PR #2036, opened 2016-11-17
ObjectEditing - Check intersects b4 deleting

adube:
This PR adds the following validation in the ObjectEditing tools that use the 'DELETE' process type:

Before applying the `difference` operator using JSTS, check if the geometries intersect first.  If they don't, no need to apply the `difference`.  This avoids pushing changes of geometries that have no actual changes when deleting.


adube:
Ready for review.

PR #2042, opened 2016-11-21
Switch to glob2

gberaudo:
Python 2.7 glob does not support '**' which is used in the template cache mako file.
Switching to glob2 makes it work automagically.


sbrunner:
You should add something like this:
https://github.com/camptocamp/ngeo/blob/master/Makefile#L231-L233

gberaudo:
Thanks @sbrunner. I updated the PR.

gberaudo:
Thanks for the review.

PR #2053, opened 2016-11-25
Search padding

### LINKED ISSUE #2043: Weird alignment of search results
In both desktop and mobile applications, the results in the search are shifted to the right. Does someone a good reason for this? I can't easily find which commit(s) introduced this.

If there's no good reason to have this left space, I'd be in favor to remove it.

screenshot from 2016-11-22 14 10 09
screenshot from 2016-11-22 14 10 31

pgiraud:
Fixes #2043.

### DIFF
--- a/contribs/gmf/less/desktop.less
+++ b/contribs/gmf/less/desktop.less
@@ -169,7 +169,6 @@ gmf-search {
 
       .gmf-search-header {
         padding: @app-margin;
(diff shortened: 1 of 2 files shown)


pgiraud:
screenshot from 2016-11-25 10 06 15
screenshot from 2016-11-25 10 06 25

PR #2058, opened 2016-11-25
ObjectEditing - Add viewer application

adube:
This PR adds the ObjectEditing "Viewer" application, in which the WFSPermalink is used to display features using their ids (i.e. using the attribute defined in the `identifierAttributeField` in the metadata).

The application can be launched from the "hub" example.

## Todo

 * [ ] Review


adube:
Ready for review.

PR #2060, opened 2016-11-25
Do not write null & undefined values in FeatureHash format

adube:
This PR fixes the `ngeo.format.FeatureHash` when used with features that have `null` or `undefined` property values.

My use-case is this: I'm using the `ol.format.GPX`, which does the following:

    feature.set('extensionsNode_', undefined);

So, when attempting to use the featurehash format with features written by the gpx format, it fails because of this undefined property.

### DIFF
--- a/src/ol-ext/format/featurehash.js
+++ b/src/ol-ext/format/featurehash.js
@@ -1142,7 +1142,10 @@ ngeo.format.FeatureHash.prototype.writeFeatureText = function(feature, opt_optio
        * @param {string} key Key.
        */
       function(value, key) {
-        if (key !== feature.getGeometryName()) {
+        if (value !== undefined &&
+            value !== null &&
+            key !== feature.getGeometryName()
+        ) {
           if (encodedProperties.length !== 0) {
             encodedProperties.push('\'');
           }


adube:
Ready for review.

src/ol-ext/format/featurehash.js
```diff
@@ -1142,7 +1142,10 @@ ngeo.format.FeatureHash.prototype.writeFeatureText = function(feature, opt_optio
        * @param {string} key Key.
        */
       function(value, key) {
-        if (key !== feature.getGeometryName()) {
+        if (value !== undefined &&
+            value !== null &&
```
sbrunner: => `!!value`?

adube: that would reject values with `0` or `""`.  We want those.

fredj: or simply `if (value && key !== feature.getGeometryName())` ?

fredj: @adube good point

sbrunner: effectively thanks :-)

Issue #2073, opened 2016-11-30
We should be able to compile even if external services don't work


ger-benjamin:
In these examples:

 - examples/backgroundlayer.js
 - examples/backgroundlayerdropdown.js
 - examples/layerloading.js
 - examples/layerorder.js
 - examples/layervisibility.js

We use external server (http://demo.boundlessgeo.com/geoserver/wms and http://demo.opengeo.org/geoserver/wms)

If these external servers don't work, we can not longer compile ngeo...

gberaudo:
Done, thanks.

PR #2074, opened 2016-11-30
Use https for boundless services

ger-benjamin:
To fix the make check

Work done for : #2073


ger-benjamin:
I wait on a make check, I've got some errors i'll update this PR soon

ger-benjamin:
That works with boundlessgeo but not with opengeo...

I propose to merge so @adube can continue to work

PR #2076, opened 2016-12-01
Revert "use https for boundless and opengeo"

ger-benjamin:
Related to: #2073

This reverts commit 72dbe4a097e666491b2b46bebd817ec239df96e1.

Seems that the servers was just down (sorry, I was in a hurry, and I'm too suspicious about http url now...)

But do we really want http url in our examples ? Someone know example served with https ?


fredj:
I think it's fine to use http

ger-benjamin:
Ok I'll merge that but I'll keep open and rename the issue because compilation should work even if boudless or opengeo don't answer.

PR #2081, opened 2016-12-02
Add missing OL deps for JSTS

adube:
This PR adds the messing properties required to be exported for JSTS to work properly.


fredj:
thanks

PR #2086, opened 2016-12-05
Typo fix in OL deps for JSTS

adube:
Fixes a copy-paste typo in OL deps for JSTS.


adube:
Yeah, sorry about that.

PR #2100, opened 2016-12-09
ObjectEditing - Fix list of copyable layers with undefined values

adube:
This PR fixes the list of copyables layers in the ObjectEditing tools that have 'undefined' values when using minimized code.

## Live demo

 * https://adube.github.io/ngeo/oe-layers-list-undefined/examples/contribs/gmf/objecteditinghub.html

### DIFF
--- a/contribs/gmf/src/services/objecteditingquery.js
+++ b/contribs/gmf/src/services/objecteditingquery.js
@@ -214,13 +214,11 @@ gmf.ObjectEditingQuery.GetQueryableLayerNodesResponse;
 
 
 /**
- * @typedef {{
- *     ogcServer: (gmfThemes.GmfOgcServer),
- *     layerNode: (gmfThemes.GmfLayerWMS)
- * }}
+ * @constructor
+ * @struct
  * @export
  */
-gmf.ObjectEditingQuery.QueryableLayerInfo;
+gmf.ObjectEditingQuery.QueryableLayerInfo = function() {};
 
 
 /**
--- a/contribs/gmf/src/services/objecteditingquery.js
+++ b/contribs/gmf/src/services/objecteditingquery.js
@@ -220,3 +220,17 @@ gmf.ObjectEditingQuery.GetQueryableLayerNodesResponse;
  * }}
  */
 gmf.ObjectEditingQuery.QueryableLayerInfo;
+
+
+/**
+ * @type {gmfThemes.GmfOgcServer}
+ * @export
+ */
+gmf.ObjectEditingQuery.QueryableLayerInfo.prototype.ogcServer;
+
+
+/**
+ * @type {gmfThemes.GmfLayerWMS}
+ * @export
+ */
+gmf.ObjectEditingQuery.QueryableLayerInfo.prototype.layerNode;
--- a/contribs/gmf/src/services/objecteditingquery.js
+++ b/contribs/gmf/src/services/objecteditingquery.js
@@ -218,6 +218,7 @@ gmf.ObjectEditingQuery.GetQueryableLayerNodesResponse;
  *     ogcServer: (gmfThemes.GmfOgcServer),
  *     layerNode: (gmfThemes.GmfLayerWMS)
  * }}
+ * @export
  */
 gmf.ObjectEditingQuery.QueryableLayerInfo;
 


adube:
**Note** : I'm currently not able to do `make check`, therefore I can't test this patch against a minimized version.  I don't know if the suggested fix works or not.

@sbrunner or @fredj, looking a this issue: do you think my approach is correct?

sbrunner:
From where is it used (and needs the export)?

adube:
Here's a preview of the application:

list-undefined

See the list on the right?  That's the list of layers that can be copied / cut from.  Its creation uses the `ng-options` but something must be renamed by the compiler because it doesn't work in minimized version (it does in dev.).

See: https://github.com/camptocamp/ngeo/blob/master/contribs/gmf/src/directives/partials/objectediting.html#L19

Also, if you look on the left, you'll see that the layer tree has the same names.  It has to be cause by the compiler renaming, but I haven't been able to find how to debug this nor test this.  If you have any hint regarding this, that would be welcome.

sbrunner:
Thanks, @adube that looks good to me :-)

adube:
Thanks.

I'll repeat, just to make sure: I haven't tested this fix, because I couldn't.  I don't know if this fixes the issue or not.  Any hint to help me be able to test this is welcome.

sbrunner:
```
make examples-hosted
```
Will generate th built example in `.build/examples-hosted/contribs/gmf/apps/`
then you should be able to test.

contribs/gmf/src/services/objecteditingquery.js
```diff
@@ -220,3 +220,17 @@ gmf.ObjectEditingQuery.GetQueryableLayerNodesResponse;
  * }}
  */
 gmf.ObjectEditingQuery.QueryableLayerInfo;
+
```
sbrunner: I thing that it should ends with ` = function() {};`...

adube: `@typedef` --> don't need to.

`@struct / @consrtructor` requires it.

I think... Thoughts?

sbrunner: @gberaudo ?
I think that good to use `@struct` and `@consrtructor`

gberaudo: `@struct` and `@constructor` are not for typedefs.
Does it work to put an `@export` on a typedef?

adube: Well, as it turns out it does!  `make check` worked after I rebased today and I was able to put a live demo.  Look at: https://adube.github.io/ngeo/oe-layers-list-undefined/examples/contribs/gmf/objecteditinghub.html, then run the example or hosted app.  You'll see that the list is now working.

It exactly as I predicted: the compiler was renaming the inner properties of the typedef because they were not exported.  In the templates, the code is not minimized so it was using the original name, not the name 'renamed' by the compiler.  So yes: putting `@export` did work.

sbrunner:
@gberaudo can you merge it if it's ok for you?

contribs/gmf/src/services/objecteditingquery.js
gberaudo: I tested using `@export` in https://closure-compiler.appspot.com/home#code%3D%252F%252F%2520%253D%253DClosureCompiler%253D%253D%250A%252F%252F%2520%2540compilation_level%2520ADVANCED_OPTIMIZATIONS%250A%252F%252F%2520%2540output_file_name%2520default.js%250A%252F%252F%2520%253D%253D%252FClosureCompiler%253D%253D%250A%250A%252F%252F%2520ADD%2520YOUR%2520CODE%2520HERE%250Agoog.provide(%27toto%27)%253B%250A%250A%252F**%250A%2520*%2520%2540typedef%2520%257B%257B%250A%2520%2520%2520ramonus%253A%2520string%250A%2520%257D%257D%250A%2520*%252F%250Atoto%253B%250A%250A%252F**%250A%2520*%2520%2540type%2520%257Bstring%257D%250A%2520*%2520%2540export%250A%2520*%252F%250Atoto.prototype.ramonus%253B%250A%250A%252F**%250A%2520*%2520%2540param%2520%257Btoto%257D%2520name%250A%2520*%252F%250Afunction%2520hello(name)%2520%257B%250A%2520%2520alert(%27Hello%252C%2520%27%2520%252B%2520name)%253B%250A%257D%250Ahello(generate())%253B%250A%250Afunction%2520generate()%2520%257B%250A%2520%2520return%2520%257B%250A%2520%2520%2520%2520ramonus%253A%2520Math.random()%250A%2520%2520%257D%253B%250A%257D but as you can see it did not prevent the renaming of the attribute to `a`.

adube: Right.  But telling it to 'export' must ensure that both `a` and the original property exist.  The patch does fix the issue, right?

adube:
Any news regarding this?

sbrunner:
It has an error here: https://travis-ci.org/camptocamp/ngeo/jobs/183272792#L3333...

adube:
The current state of this patch is this: if fixes the issue; it works in minimized version; travis complains.

Any hint would be welcome.

sbrunner:
I do this change https://github.com/camptocamp/ngeo/commit/8df79fe40be5a2f50dbd953f254f2e84b64aaf08 and it seam working but I don't understand why ...

adube:
Here's a new attempt.  I tried it locally using `make examples-hosted` and it worked.  The only thing remaining would be to see if Travis passes.

adube:
Travis passes.  I confirm that the patch fixes the issue.  If you approve this PR, then we could merge.

PR #2101, opened 2016-12-09
ObjectEditing - Ensure multi-geometry on INSERT

adube:
This PR fixes the ObjectEditing directive when it attemps to insert a feature.  The geometry has to be MULTI.

### DIFF
--- a/contribs/gmf/src/directives/objectediting.js
+++ b/contribs/gmf/src/directives/objectediting.js
@@ -936,7 +936,8 @@ gmf.ObjecteditingController.prototype.handleSketchFeaturesAdd_ = function(evt) {
     }
 
   } else if (this.process === gmf.ObjecteditingtoolsController.ProcessType.ADD) {
-    this.feature.setGeometry(sketchGeom.clone());
+    this.feature.setGeometry(
+      gmf.ObjecteditingController.toMultiGeometry_(sketchGeom.clone()));
   }
 
   this.sketchFeatures.clear();


adube:
Ready for review.

contribs/gmf/src/directives/objectediting.js
```diff
@@ -936,7 +936,8 @@ gmf.ObjecteditingController.prototype.handleSketchFeaturesAdd_ = function(evt) {
     }
 
   } else if (this.process === gmf.ObjecteditingtoolsController.ProcessType.ADD) {
-    this.feature.setGeometry(sketchGeom.clone());
+    this.feature.setGeometry(
+      gmf.ObjecteditingController.toMultiGeometry_(sketchGeom.clone()));
```
sbrunner: Shouldn't be done only if the geometry type is a multi?

adube: First, the GMF server throws an error unless we give it a multi geometry.  That's the reason we must ensure that the sent geometry is always multi.

Then, if you check the `toMultiGeometry_`, it already takes care of checking the type of geom and return the same geom if it's already multi, else it creates a multi.

Therefore, I believe the fix is okay.  What do you think?

sbrunner: It's not that I speak about ...
It the result of e.g.: https://geomapfish-demo.camptocamp.net/2.1/wsgi/layers/111/md.xsd
Is a geometry type: `gml:LineStringPropertyType` instance of `gml:MultiLineStringPropertyType` we shouldn't use a multi geometry :-)

sbrunner: OK, thanks :-)

adube:
Ready for review / merge.

PR #2102, opened 2016-12-09
Use the struct and constructor like the other elements

sbrunner:


contribs/gmf/externs/gmf-themes.js
```diff
@@ -89,7 +105,7 @@ gmfThemes.GmfTheme.prototype.functionalities;
  * neither a WMS group.
  * This represent « first level group » (Block in the layer tree),
  * or all sub nodes that's not al leaf.
- * @constructor
+ * @record
  * @struct
  * @extends gmfThemes.GmfBaseNode
```
gberaudo: Do `@record` works together with `@extends`?

externs/mapfish-print-v3.js
```diff
@@ -366,9 +383,57 @@ var MapFishPrintSymbolizer = function() {};
 MapFishPrintSymbolizer.prototype.type;
 
 
+/**
+ * @type {string}
+ */
+MapFishPrintSymbolizer.prototype.fillColor;
```
gberaudo: Should add `|undefined` for optional attributes.

externs/mapfish-print-v3.js
```diff
@@ -530,7 +597,8 @@ MapFishPrintSymbolizerPolygon.prototype.strokeDashstyle;
 
 
 /**
- * @constructor
+ * @record
+  @struct
```
gberaudo: Missing a star.

contribs/gmf/externs/gmf-themes.js
sbrunner: It seem yes :-)

PR #2112, opened 2016-12-14
Don't use controllerAs

fredj:
Use `controller: 'FooController as ctrl'` instead.


fredj:
thanks for the review

PR #2122, opened 2016-12-15
Fix layertree parent state

gberaudo:
When the first child is checked and there exist another child which
is not checked, the parent was incorrectly checked instead of being
indeterminate.

To avoid this, the parent state is always refreshed.

The root cause is that state management is done too early, when the
layertree controller hierarchy is instanciated.

A more involved fix would be to:
- stop initializing layertrees in their constructors;
- have layertrees register a ready listener on the root layertree;
- have the root layertree triggers an onReady event in its postLink
  function.


pgiraud:
I don't see the wrong behaviour you describe on the gmf demo. Where can we see the issue?

gberaudo:
See http://plan.epfl.ch/
The "Informations EPFL" and "Guichet étudiants" are checked, not the other sibling children.
However, the whole group "Guichet étudiants " is marked checked (it appears in mes points d'intérêts above).

gberaudo:
Thanks for the review.

PR #2123, opened 2016-12-15
Add support for multi geometries

### LINKED ISSUE #2084: Not possible to add new feature in edit interface
It is not possible to safe a new feature in the edit interface (editing the geometry and attributes of an existing feature is possible)
Tested with:
- IE 11 Win 8.1
- Edge Win10
- Safari latest Mac El Cap.
- Chrome Samsung Galaxy Tab 4

pgiraud:
This fixes #2084.

The idea is to ensure that the drawn geometry is multi if the geometry type (as defined in the .xsd) is multi as well.


sbrunner:
This should be on the 2.1 branch...

PR #2129, opened 2016-12-16
Add Bloodhound service for GeoAdmin location search

tsauerwein:
This PR proposes to add a service for the [GeoAdmin location search](http://api3.geo.admin.ch/services/sdiservices.html#search) which creates a Bloodhound object that parses the response from the location service and creates features (similar to the `createGeoJSONBloodhound` service).

### DIFF
--- a/examples/locationsearch.js
+++ b/examples/locationsearch.js
@@ -134,13 +134,11 @@ app.SearchController.prototype.createAndInitBloodhound_ = function(ngeoCreateLoc
  * @private
  */
 app.SearchController.select_ = function(event, suggestion, dataset) {
-  var map = /** @type {ol.Map} */ (this.map);
   var feature = /** @type {ol.Feature} */ (suggestion);
-  var bbox = /** @type {ol.Extent} */
-      (feature.get('bbox'));
-  var mapSize = /** @type {ol.Size} */ (map.getSize());
-  map.getView().fit(bbox, mapSize,
-      /** @type {olx.view.FitOptions} */ ({maxZoom: 16}));
+  var bbox = /** @type {ol.Extent} */ (feature.get('bbox'));
+  var mapSize = this.map.getSize();
+  goog.asserts.assert(mapSize !== undefined);
+  this.map.getView().fit(bbox, mapSize, {maxZoom: 16});
 };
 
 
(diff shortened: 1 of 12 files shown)


examples/locationsearch.js
```diff
@@ -0,0 +1,174 @@
...
+
+/**
+ * @param {jQuery.Event} event Event.
+ * @param {Object} suggestion Suggestion.
+ * @param {TypeaheadDataset} dataset Dataset.
+ * @this {app.SearchController}
+ * @private
+ */
+app.SearchController.select_ = function(event, suggestion, dataset) {
+  var map = /** @type {ol.Map} */ (this.map);
```
gberaudo: Is it necessary to cast the map?

examples/locationsearch.js
```diff
@@ -0,0 +1,174 @@
...
+  var feature = /** @type {ol.Feature} */ (suggestion);
+  var bbox = /** @type {ol.Extent} */
+      (feature.get('bbox'));
+  var mapSize = /** @type {ol.Size} */ (map.getSize());
```
gberaudo: Have you considered using an assert instead of casting?

examples/locationsearch.js
```diff
@@ -0,0 +1,174 @@
...
+  map.getView().fit(bbox, mapSize,
+      /** @type {olx.view.FitOptions} */ ({maxZoom: 16}));
```
gberaudo: Is this cast necessary?

src/modules/search/createlocationsearchbloodhound.js
```diff
@@ -0,0 +1,154 @@
...
+        parseFloat(match[3]),
+        parseFloat(match[4])
+      ];
+    } else {
+      return null;
+    }
+  };
+
+  var removeHtmlTags = function(label) {
+    return label.replace(/<\/?[ib]>/g, '');
```
gberaudo: Shouldn't you remove all tags and not only b and i?

tsauerwein: `<b>` and `<i>` are the only tags used.

examples/locationsearch.js
tsauerwein: Thanks for noticing, there were quite some unneeded casts in this example.

tsauerwein:
Thanks for the review!

PR #2134, opened 2016-12-19
Add a query layer pointer

rbovard:
Add a query layer pointer in `mapQueryDirective`.

Edit: now only the queriable layers are used.

PS: I cannot publish the example to GitHub pages because of access rights of repository.

### DIFF
--- a/examples/mapquery.js
+++ b/examples/mapquery.js
@@ -103,7 +103,8 @@ app.MainController = function($scope, ngeoDataSources, ngeoToolActivateMgr) {
   const busStopLayer = new ol.layer.Image({
     'source': new ol.source.ImageWMS({
       'url': 'https://geomapfish-demo.camptocamp.net/1.6/wsgi/mapserv_proxy',
-      params: {'LAYERS': 'bus_stop'}
+      params: {'LAYERS': 'bus_stop'},
+      crossOrigin: 'anonymous'
     })
   });
 
@@ -120,7 +121,8 @@ app.MainController = function($scope, ngeoDataSources, ngeoToolActivateMgr) {
   const informationLayer = new ol.layer.Image({
     'source': new ol.source.ImageWMS({
       'url': 'https://geomapfish-demo.camptocamp.net/1.6/wsgi/mapserv_proxy',
-      params: {'LAYERS': 'information'}
+      params: {'LAYERS': 'information'},
+      crossOrigin: 'anonymous'
     })
   });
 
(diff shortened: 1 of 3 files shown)


rbovard:
To see it in action, don't add OSM layer to the map (comment those lines https://github.com/rbovard/ngeo/blob/1f3dfe7a47541d10a82f857d109a661a6105137b/examples/mapquery.js#L125-L127)

Example: http://localhost:3000/examples/mapquery.html

rbovard:
Ready for review.

See example http://localhost:3000/examples/mapquery.html

rbovard:
I fixed the conflict in `src/directives/mapquery.js` but now my example is broken... Did I miss something?

From:

```
    link: function(scope, elem, attrs) {
      var map = scope.$eval(attrs['ngeoMapQueryMap']);
      var clickEventKey_ = null;
```

To:

```
    link(scope, elem, attrs) {
      const map = scope.$eval(attrs['ngeoMapQueryMap']);
      let clickEventKey_ = null;
```

rbovard:
Okay I noticed that there was big changes introduced this morning by @adube (see https://github.com/camptocamp/ngeo/pull/2239) in the Query structure.

It's a a bit complicated for me to understand all those changes, I'll need help to make my PR working again :(

fredj:
If I understand correctly, the goal is to change the cursor when a query result is hovered?

(query result is the vector point)

rbovard:
Yes, exactly. Like you can see here: http://openlayers.org/en/latest/examples/getfeatureinfo-image.html

It was working before the mentionned PR.

fredj:
See https://github.com/camptocamp/ngeo/compare/pointer-wms

It used the `querySourceIds` property to determine if the layer is queryable. This property is set by `gmf.Themes` so it won't work with the mapquery example (thanks to @adube for the help !).

This system will generate a lot of false positive results, for example:

wrong_cursor

The cursor is changed because there's something under the mouse (the text) but if the user click nothing happens

rbovard:
Thanks for the explanations and for the adaptations of my code, it seems much simpler with the `querySourceIds`.

You're right about the false positives when hovering the labels, do you have an idea of how to handle this specific case? I still think this feature is very interessant (with no labeled layers) and I'm wondering how to implement it? Maybe @cedricmoullet can give us a feedback about how they did with geoadmin?

fredj:
They have the same issue:
https://map.geo.admin.ch/?topic=ech&lang=en&bgLayer=voidLayer&layers=ch.swisstopo.zeitreihen,ch.bfs.gebaeude_wohnungs_register,ch.swisstopo.lotabweichungen&layers_visibility=false,false,true&layers_timestamp=18641231,,&X=153157.10&Y=532629.62&zoom=7&catalogNodes=457,491

and hover the label: the cursor is changed but there's no query result

rbovard:
So can *we* live with it?

Maybe we can simply add an option to let the integrator the choice to use this feature or not?

sbrunner:
+1

fredj:
ok, let's do this

rbovard:
Cool, thanks!

fredj:
see #2304

PR #2147, opened 2016-12-23
ObjectEditing - Fix save requests after insert

adube:
After inserting a new feature in the ObjectEditing tool, the id returned in the JSON was not stored in the existing feature.  Therefore, further actions were not working because they require the id.

This fixes this issue.

### DIFF
--- a/contribs/gmf/src/directives/objectediting.js
+++ b/contribs/gmf/src/directives/objectediting.js
@@ -527,13 +527,22 @@ gmf.ObjecteditingController.prototype.handleDeleteFeature_ = function(resp) {
  * @private
  */
 gmf.ObjecteditingController.prototype.handleEditFeature_ = function(resp) {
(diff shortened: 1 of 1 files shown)


contribs/gmf/src/directives/objectediting.js
```diff
@@ -527,13 +527,22 @@ gmf.ObjecteditingController.prototype.handleDeleteFeature_ = function(resp) {
  * @private
  */
 gmf.ObjecteditingController.prototype.handleEditFeature_ = function(resp) {
+  // (1) Update the id
+  var features = new ol.format.GeoJSON().readFeatures(resp.data);
```
sbrunner: missing requires!

adube: Fixed.  Do you know why `make dist` didn't complain when we're missing a require?

PR #2152, opened 2017-01-04
Don't add unknown keys in the permalink

### LINKED ISSUE #1987: Strange things in the permalink
Like `...Bflex-direction%3Acolumn%3Bmin-height%3A0%7D.x-layout-fit.x-stretched%3E.x-layout-fit-item%7Bdisplay%3Aflex%3Bdisplay%3A-webkit-box%3Bdisplay%3A...`

related to: https://github.com/camptocamp/c2cgeoportal/issues/2610

sbrunner:
fix #1987

### DIFF
--- a/contribs/gmf/src/services/permalink.js
+++ b/contribs/gmf/src/services/permalink.js
@@ -1,6 +1,7 @@
 goog.provide('gmf.Permalink');
 
 goog.require('gmf');
+goog.require('ngeo');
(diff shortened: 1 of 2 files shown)


fredj:
Will it work when we have to add items in the permalink at application level ? (example `floor=2` for the epfl project)

sbrunner:
I hope yes, you should be able to override the `ngeoUsedKeyRegexp` value.

PR #2157, opened 2017-01-06
The graphic format shouldn't be hard coded to an arbitrary value, it …

### LINKED ISSUE #2149: Print - Can't print svg symbols
The ngeo print service use "image/png" for png like for others symbole type. The result is that SVG can't be printed. (should be "image/svg" for svg)

Issue in these lines: https://github.com/camptocamp/ngeo/blob/a859f97ed0e411fc78598b4ba44bdd3c679e27b8/src/services/print.js#L614-L618

sbrunner:
…should be automatically detected by the print

fix #2149


sbrunner:
@ger-benjamin As I see in mapfish-print it's possible that's already working, can you test it in the EPFL project?

ger-benjamin:
For some reason, that disappoint me. But yes, that work fine.

Thanks, please merge.

PR #2158, opened 2017-01-06
Add ES6 support

sbrunner:


buildtools/check-example.js
```diff
@@ -2,16 +2,19 @@
 // A PhantomJS script used to check that the hosted examples load
 // without errors. This script is executed by the Makefile's
 // check-examples target.
-//
-var args = require('system').args;
+
+// For ES6 support
+require('babel-polyfill/dist/polyfill');
```
gberaudo: Do you still need it?

contribs/gmf/src/directives/mobilenav.js
```diff
@@ -50,16 +50,9 @@ gmf.mobileNavDirective = function() {
...
-        /**
-         * @param {angular.Scope} scope Scope.
-         * @param {angular.JQLite} element Element.
-         * @param {angular.Attributes} attrs Atttributes.
-         * @param {gmf.MobileNavController} navCtrl Controller.
-         */
-        function(scope, element, attrs, navCtrl) {
-          navCtrl.init(element);
-        }
+    link(scope, element, attrs, navCtrl) {
```
gberaudo: You may want to keep the comments here and below. They simply need to be moved before `link` and their indentation reduced.

examples/asitvd.html
```diff
@@ -31,7 +31,7 @@
     <script src="default.js"></script>
     <script src="../utils/watchwatchers.js"></script>
     <script>
-      var ngeoModule = angular.module('ngeo');
+      let ngeoModule = angular.module('ngeo');
```
gberaudo: The examples must also be viewable with ES5 browsers so you can not use let here and in any html files.

src/modules/import/file.js
```diff
@@ -10,7 +10,7 @@ goog.provide('ngeo.File');
  * @ngInject
  */
 ngeo.File = function($q, $http, $window, gettext) {
-  var fileReader, canceler;
+  let fileReader, canceler;
```
gberaudo: The import module may be used independently of the rest of the code, with a different build system.
I see that babel that you used is able to convert a whole directory to ES5. Do you think it is OK?

https://babeljs.io/docs/usage/cli/

Compile the entire src directory and output it to the lib directory. You may use --out-dir or -d. This doesn’t overwrite any other files or directories in lib.
```
babel src/modules/import --out-dir es5_import
```
Compile the entire src directory and output it to the one concatenated file.
```
babel src/modules/import --out-file es5_import_one_file.js
```

sbrunner: I don't completely catch what you say...

fgravin: Great !
Very nice initiative and a nice step forward :)
Did you test to use the `Class` uppon a Controller for example ?
Maybe we could automate with lebab the refactoring from es5 to es6 classes, taking care of closure and doc annotations.

Maybe in a next step.

sbrunner: No, not yet :-)

package.json
```diff
@@ -28,42 +28,43 @@
...
+    "karma-chrome-launcher": "2.0.0",
     "karma-coverage": "1.1.1",
     "karma-jasmine": "1.1.0",
-    "karma-chrome-launcher": "2.0.0",
     "less": "2.7.2",
     "less-plugin-autoprefix": "1.5.1",
     "less-plugin-clean-css": "1.5.1",
     "nomnom": "1.8.1",
     "openlayers": "git://github.com/openlayers/ol3#8f6e07b",
+    "phantomjs-prebuilt": "^2.1.14",
```
gberaudo: remove the ^

karma-conf.js
```diff
@@ -116,8 +116,14 @@ module.exports = function(config) {
 
     // start these browsers
     // available browser launchers: https://npmjs.org/browse/keyword/karma-launcher
-    browsers: ['PhantomJS'],
+    browsers: ['optchrome'],
```
tsauerwein: PhantomJS 2.5 will support ES 6, the beta was released on Sunday:
https://groups.google.com/forum/#!topic/phantomjs/AefOuwkgBh0

Makefile
```diff
@@ -460,6 +461,10 @@ dist/gmf.js.map: dist/gmf.js
 	mkdir -p $(dir $@)
 	cp $< $@
 
+.build/examples-hosted/lib/transpile.js: node_modules/closure-util/.deps/library/87f29481f1e57fbba986e2304ac80d4173f94c9c/closure/goog/transpile.js
```
gberaudo: CLOSURE_LIBRARY_PATH := $(shell node -e 'process.stdout.write(require("closure-util").getLibraryPath())' 2> /dev/null)

karma-conf.js
gberaudo: Great news!

sbrunner: Yes :-) but I don't see how to use it :-(

src/modules/import/file.js
gberaudo: @sbrunner tested it: the closure comments were preserved and the ES6 constructs like `for of` where replaced.

PR #2168, opened 2017-01-10
Continue to use var in the application html

sbrunner:


tsauerwein:
Would it be possible to merge this PR? I would need it to test with non-ES6 browsers.

tsauerwein:
I also had to do the following changes:

```
diff --git a/examples/https.js b/examples/https.js
index 0f48faf..5ce31de 100644
--- a/examples/https.js
+++ b/examples/https.js
@@ -4,7 +4,7 @@
 if (window.location.protocol == 'http:' &&
     window.location.port != '3000' &&
     window.location.host != 'localhost') {
-  let restOfUrl = window.location.href.substr(5);
+  var restOfUrl = window.location.href.substr(5);
   /** @type {Location} */
   window.location = 'https:' + restOfUrl;
 }
diff --git a/utils/watchwatchers.js b/utils/watchwatchers.js
index 5963238..963e1d6 100644
--- a/utils/watchwatchers.js
+++ b/utils/watchwatchers.js
@@ -11,10 +11,10 @@
  */
 (function () {
 
-  let root = angular.element(document.getElementsByTagName('body'));
+  var root = angular.element(document.getElementsByTagName('body'));
 
-  let countWatchers_ = function(element, scopes, count) {
-    let scope;
+  var countWatchers_ = function(element, scopes, count) {
+    var scope;
     scope = element.data().$scope;
     if (scope && !(scope.$id in scopes)) {
         scopes[scope.$id] = true;
```

PR #2169, opened 2017-01-10
Fix examples build

sbrunner:
actually they don't works anymore ...
http://camptocamp.github.io/ngeo/master/examples/


sbrunner:
Can be tested here: http://camptocamp.github.io/ngeo/master-examples/examples/

Issue #2083, opened 2016-12-05
CSV export not working for Profiles


marionb:
It is not possible to make a CSV export from the profile.
This issue occurred for:
- IE11 Win8.1
- Edge Win10
- Safari latest Mac El Cap.
- Chrome Samsung Galaxy Tab 4 (tested with https://camptocamp.github.io/ngeo/master/examples/contribs/gmf/apps/desktop)

sbrunner:
It's working on Firefox 50 Ubuntu Xenial

tsauerwein:
I can't reproduce with IE11/Win8.1, Edge/Win10 and Chrome/Samsung Galaxy Tab 4, but I am looking into the issue with Safari (https://github.com/eligrey/FileSaver.js/issues/215).

ger-benjamin:
@ybolognini  I close and remove "Done" label.

PR #2177, opened 2017-01-11
Work-around for download on Safari

tsauerwein:
There are several issues with the file download on Safari, see https://github.com/eligrey/FileSaver.js/issues/12. The latest Safari Tech Preview fixes these issues, but as long as no official release is available, it would be good to have a work-around.

The first commit addresses the problem that Safari is not able to start a download for files with a mime-type other than `text/plain`. So, when using Safari, the mime-type is set to `text/plain` (instead of the mime-type for csv/kml/gpx). This will open the file in a new tab and the user can download with `File -> Save as`. Not optimal, but as least the user can get the data.

The second commit deals with the problem that on Safari a download can only be started within 1000ms in the event handler of a *trusted event* (an event generated by a user action, like clicking the export button). Currently, a request is made to get the profile as CSV and then when the request succeeds, the download is started. But this doesn't work for Safari, the file would be shown in the same tab.
Instead the CSV file is generated on the client-side with the already available data.

Closes https://github.com/camptocamp/ngeo/issues/2083

### DIFF
--- a/contribs/gmf/apps/desktop/index.html
+++ b/contribs/gmf/apps/desktop/index.html
@@ -281,7 +281,6 @@
         module.constant('authenticationBaseUrl', 'https://geomapfish-demo.camptocamp.net/2.1/wsgi');
         module.constant('fulltextsearchUrl', 'https://geomapfish-demo.camptocamp.net/2.1/wsgi/fulltextsearch?limit=30&partitionlimit=5&interface=desktop');
         module.constant('gmfRasterUrl', 'https://geomapfish-demo.camptocamp.net/2.1/wsgi/raster');
-        module.constant('gmfProfileCsvUrl', 'https://geomapfish-demo.camptocamp.net/2.1/wsgi/profile.csv');
         module.constant('gmfProfileJsonUrl', 'https://geomapfish-demo.camptocamp.net/2.1/wsgi/profile.json');
         module.constant('gmfPrintUrl', 'https://geomapfish-demo.camptocamp.net/2.1/wsgi/printproxy');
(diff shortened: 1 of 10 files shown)


sbrunner:
https://docs.angularjs.org/error/$injector/unpr?p0=ngeoCsvDownloadProvider%20%3C-%20ngeoCsvDownload%20%3C-%20GmfProfileController in the profile example (on Travis) :-(

tsauerwein:
> https://docs.angularjs.org/error/$injector/unpr?p0=ngeoCsvDownloadProvider%20%3C-%20ngeoCsvDownload%20%3C-%20GmfProfileController in the profile example (on Travis) :-(

Thanks, fixed.

fredj:
LGTM, @sbrunner do we want to keep the `profile.csv` service in c2cgeoportal ?

sbrunner:
Yes because it steel used by old CGXP applications :-)

tsauerwein:
Thanks for the review, @fredj!

stef74:
For me it's not work on iOS on chrome and safari. They show GPX file as a text but not usable (no way to export or open with dedicated app). On desktop I can save an reuse (so it usable).

fredj:
You need an app that can open gpx file.

stef74:
I have it .... If i open gpx file on my mail or google drive or other website it's work perfectly (it show open with ... and I can select my GPX app )

sbrunner:
Your right, the main issue is here: 
https://github.com/eligrey/FileSaver.js/issues/12

stef74:
But I have also the issue on chrome (on iOS).

stef74:
on opera mini ... it open google search with blob:http://*******
on firefox : impossible to open the page. 

It'seem more IOS related ...

PR #2182, opened 2017-01-11
Restore jsdoc to ES5

gberaudo:


fgravin:
Is it better to get back to `var` or `use strict` couldn't make the trick ?

Do you know why travis didn't complain and why the apidoc is well published on ghpages ?

gberaudo:
Maybe this part is run with node6 on travis.
I added "use strict" to all `.js` files but it was still failing. Maybe I should add something to the `.tmpl` files? Anyway I think it is fine to keep jsdoc directory as-is, we can convert to ES6 when we officially switch to node6 (in April?).

PR #2185, opened 2017-01-12
Fix translations

oterral:
This PR replace uses of `gettextCatalog` by `gettext` to fix the translation and also remove use of tranlsate filter when it's unneeded.

### DIFF
--- a/src/modules/import/partials/import-local.html
+++ b/src/modules/import/partials/import-local.html
@@ -14,6 +14,6 @@
           ng-class="{'btn-danger': loading}"
           ng-disabled="!isValid(file) || (loading && !fileReader)"
           ng-click="loading ? cancel() : handleFile()">
-    <span>{{userMessage | translate }}</span>
+    <span>{{userMessage | translate}}</span>
   </button>
 </div>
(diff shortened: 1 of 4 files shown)


src/modules/import/partials/import-local.html
```diff
@@ -13,7 +13,8 @@
   <button type="button" class="btn btn-default btn-sm ngeo-load"
           ng-class="{'btn-danger': loading}"
           ng-disabled="!isValid(file) || (loading && !fileReader)"
-          ng-click="loading ? cancel() : handleFile()">
-    <span>{{userMessage | translate }}</span>
+          ng-click="loading ? cancel() : handleFile()"
+          translate>
```
gberaudo: You need to keep using the `translate` filter. Why is this change needed?

src/modules/import/partials/import-online.html
```diff
@@ -8,7 +8,8 @@
   <button type="button" class="btn btn-default btn-sm"
           ng-class="{'btn-danger': canceler}"
           ng-disabled="!fileUrl || !isValid(fileUrl) || (loading && !canceler)"
-          ng-click="loading ? cancel() : handleFileUrl()">
-    {{userMessage | translate}}
+          ng-click="loading ? cancel() : handleFileUrl()"
+          translate>
```
gberaudo: Idem

oterral:
According to angular-gettext doc it should work. But ok if it's necessary.

oterral:
@gberaudo  filter re-added

gberaudo:
> According to angular-gettext doc it should work. But ok if it's necessary.
Really? I tried both and the behaviour were differents.

gberaudo:
Waiting for the build to finish. Will merge as soon as Travis is green.

gberaudo:
Thanks @oterral.

Issue #2199, opened 2017-01-12
make dist fails


fgravin:
Fresh ngeo after `npm install`

`make dist` command fails
seems still related with es6 changes.

```
mkdir -p dist/
node buildtools/build.js .build/ngeo.json dist/ngeo.js
ERR! Trouble generating info: /home/fgravin/dev/ngeo/buildtools/jsdoc/info/define-plugin.js:9
let DEFAULT_VALUE = /default\s+is\s+`?(true|false)`?/i;
^^^

SyntaxError: Block-scoped declarations (let, const, function, class) not yet supported outside strict mode
    at exports.runInThisContext (vm.js:53:16)
    at Module._compile (module.js:373:25)
    at Object.Module._extensions..js (module.js:416:10)
    at Module.load (module.js:343:32)
    at Object.load (/home/fgravin/dev/ngeo/node_modules/jsdoc/node_modules/requizzle/lib/loader.js:112:15)
    at requizzle (/home/fgravin/dev/ngeo/node_modules/jsdoc/node_modules/requizzle/lib/requizzle.js:90:24)
    at infectProxy (/home/fgravin/dev/ngeo/node_modules/jsdoc/node_modules/requizzle/lib/loader.js:82:28)
    at Module.targetModule.require (/home/fgravin/dev/ngeo/node_modules/jsdoc/node_modules/requizzle/lib/loader.js:101:11)
    at require (internal/module.js:12:17)
    at Object.exports.installPlugins (/home/fgravin/dev/ngeo/node_modules/jsdoc/lib/jsdoc/plugins.js:23:18)
    at Object.module.exports.cli.createParser (/home/fgravin/dev/ngeo/node_modules/jsdoc/cli.js:340:17)
    at module.exports.cli.main (/home/fgravin/dev/ngeo/node_modules/jsdoc/cli.js:225:20)
    at Object.module.exports.cli.runCommand (/home/fgravin/dev/ngeo/node_modules/jsdoc/cli.js:180:5)
    at /home/fgravin/dev/ngeo/node_modules/jsdoc/jsdoc.js:103:9
    at Object.<anonymous> (/home/fgravin/dev/ngeo/node_modules/jsdoc/jsdoc.js:104:3)
    at Module._compile (module.js:409:26)
 
make: *** [dist/ngeo.js] Error 1
```

sbrunner:
What's this `make dist` for?

fgravin:
Just build the dist.

[...]

PR #2200, opened 2017-01-12
Add use strict to buildtools/jsdoc files

gberaudo:


gberaudo:
@fgravin does this fixes your issue #2199  with `make dist`?

fgravin:
Yes thanks !

PR #2221, opened 2017-01-19
Filter - Add ngeo.DataSource & gmf.FilterSelector example and directive

adube:
This PR introduces the `ngeo.DataSource` object, which will be used as base for anything that wants to get information about specific data sources, such as: which ones are queryable, how can they be queried, etc.

This PR introduces the `gmf.DataSourcesManager` service, which is hooked to the theme and treeManager services in order to create and add the `ngeo.DataSource` objects from the themes.  The treeManager is used to synchronize the `visible` property, i.e. when the treeCtrl state becomes "on", the visibility of the data source is set.

Finally, this PR introduces an example in which the manager is injected for the data sources to be created.  It also includes the `gmf.FilterSelector` directive, which doesn't currently do anything.  It will do something in further pull requests.

## Todo

 * [x] Review

PR #2223, opened 2017-01-20
Filter - Add and synchronize ngeo.DataSource inRange property

adube:
This PR introduces the `ìnRange` dynamic property within the `ngeo.DataSource` object.

It is a dynamic property, i.e. its value can change and be watched.

The value is calculated every time the map view resolution changes and uses the `maxResolution` and `minResolution` properties.  This is not directly managed within the `ngeo.DataSource` object, but inside a `ngeo.SyncDataSourcesMap` service in order to listen only once for each change of resolution.  That service is injected in the `ngeo.Map` directive.

## Todo

 * [x] Wait for #2221 to be merged
 * [ ] Review ~~(you can look at the code in the last commit only)~~

### DIFF
--- a/src/directives/map.js
+++ b/src/directives/map.js
@@ -2,6 +2,7 @@ goog.provide('ngeo.mapDirective');
 
 goog.require('goog.asserts');
 goog.require('ngeo');
+goog.require('ngeo.SyncDataSourcesMap');
 goog.require('ol.Map');
 
 
@@ -18,12 +19,14 @@ goog.require('ol.Map');
  * [../examples/simple.html](../examples/simple.html)
  *
  * @htmlAttribute {ol.Map} ngeo-map The map.
+ * @param {ngeo.SyncDataSourcesMap} ngeoSyncDataSourcesMap Ngeo sync
+ *     data sources map service.
  * @return {angular.Directive} Directive Definition Object.
  * @ngInject
  * @ngdoc directive
  * @ngname ngeoMap
  */
-ngeo.mapDirective = function() {
+ngeo.mapDirective = function(ngeoSyncDataSourcesMap) {
   return {
     restrict: 'A',
     /**
@@ -39,6 +42,8 @@ ngeo.mapDirective = function() {
       goog.asserts.assertInstanceof(map, ol.Map);
 
       map.setTarget(element[0]);
+
+      ngeoSyncDataSourcesMap.map = map;
     }
   };
 };
--- a/src/services/syncDataSourcesMap.js
+++ b/src/services/syncDataSourcesMap.js
@@ -1,10 +1,12 @@
 goog.provide('ngeo.SyncDataSourcesMap');
 
 goog.require('ngeo');
+goog.require('ngeo.DataSource');
 /** @suppress {extraRequire} */
 goog.require('ngeo.DataSources');
 goog.require('ol.Collection');
 goog.require('ol.Observable');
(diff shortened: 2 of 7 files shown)


adube:
@fredj Can I rebase this onto #2221, which is up-to-date with master?

fredj:
yes, please rebase

adube:
Rebased.

adube:
(Reminder: you only need to review the last commit)

adube:
Ready for review.

src/directives/map.js
```diff
@@ -18,12 +19,14 @@ goog.require('ol.Map');
...
  * @htmlAttribute {ol.Map} ngeo-map The map.
+ * @param {ngeo.SyncDataSourcesMap} ngeoSyncDataSourcesMap Ngeo sync
+ *     data sources map service.
  * @return {angular.Directive} Directive Definition Object.
  * @ngInject
  * @ngdoc directive
  * @ngname ngeoMap
  */
-ngeo.mapDirective = function() {
+ngeo.mapDirective = function(ngeoSyncDataSourcesMap) {
```
fredj: I'm not a big fan of having `ngeo.SyncDataSourcesMap` here. The map initialization should be moved elsewhere (like the map initialization of `ngeo.FeatureOverlayMgr`)

adube: The `ngeo.FeatureOverlayMgr` seems have its map set at the "application" level, i.e. in the controllers or in the examples.

I think we always want the `ngeoSyncDataSourcesMap` to do its magic, i.e. synchronize the `inRange` property, otherwise you have to make sure you do not forget about it.  If we're okay with that, then fine.  I personnally think it would be a dangerous move.  If you forget about it, in an example (for example) it would become very hard for a third person to understand why it doesn't behave like it should (i.e. a data source that's supposed to be out of range would always be).  I think we should leave it there for that reason.

What do you think?

fredj: ok, make sense

adube:
Rebased onto master.

fredj:
code looks good to me; @sbrunner please merge if you're happy too ....

src/services/syncDataSourcesMap.js
```diff
@@ -0,0 +1,171 @@
...
+  }
+
+  /**
+   * Called when the resolution of the map view changes. Synchronize the
+   * datasources to current resolution of the view.
+   * @param {ol.ObjectEvent} evt Event.
+   * @private
+   */
+  handleViewResolutionChange_(evt) {
+    const view = /** @type {ol.View} */ (evt.target);
```
sbrunner: We should avoid to use type cast, it's preferable to have:
```js
const view = evt.target;
const view = goog.asserts.assetInstanceof(evt.target, ol.View);
```

src/services/syncDataSourcesMap.js
```diff
@@ -0,0 +1,171 @@
...
+  }
+
+  /**
+   * Called when a new data source is added to the ngeo collection. If there's
+   * map bound, update its `inRange` right away.
+   * @param {ol.Collection.Event} evt Event
+   * @private
+   */
+  handleDataSourcesAdd_(evt) {
+    const dataSource = /** @type {ngeo.DataSource} */ (evt.element);
```
sbrunner: same here

adube: I'm on it.

adube:
Fixed.

PR #2212, opened 2017-01-17
Update openlayers and closure compiler to latest version

gberaudo:
Replaces #2109.

PR #2234, opened 2017-01-24
Removes some dangerous typecast

sbrunner:
GCC allows to cast a `string` into a `number` that's in fact impossible than we shouldn't use it...

### DIFF
--- a/contribs/gmf/src/services/permalink.js
+++ b/contribs/gmf/src/services/permalink.js
@@ -350,7 +350,7 @@ gmf.Permalink = function($timeout, ngeoBackgroundLayerMgr, ngeoDebounce,
   });
   this.rootScope_.$on('ngeo-layertree-opacity', (event, treeCtrl) => {
     const newState = {};
-    const opacity = treeCtrl.layer.opacity;
+    const opacity = treeCtrl.layer.getOpacity();
     const stateName = (treeCtrl.parent.node.mixed ?
         gmf.PermalinkParamPrefix.TREE_OPACITY : gmf.PermalinkParamPrefix.TREE_GROUP_OPACITY
     ) + treeCtrl.node.name;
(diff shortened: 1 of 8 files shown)


contribs/gmf/src/services/permalink.js
```diff
@@ -453,12 +453,14 @@ gmf.Permalink.prototype.addListenerKey_ = function(uid, key, opt_isol) {
  */
 gmf.Permalink.prototype.getMapCenter = function() {
   let center = null;
-  const x = /** @type {number} */ (this.ngeoStateManager_.getInitialValue(
-    gmf.PermalinkParam.MAP_X));
-  const y = /** @type {number} */ (this.ngeoStateManager_.getInitialValue(
-    gmf.PermalinkParam.MAP_Y));
+  const x = this.ngeoStateManager_.getInitialValue(
```
fredj: minor but can you put all in the same line?
```js
const x = this.ngeoStateManager_.getInitialValue(gmf.PermalinkParam.MAP_X);
```

same for all the changes in this pr

contribs/gmf/src/services/permalink.js
```diff
@@ -453,12 +453,14 @@ gmf.Permalink.prototype.addListenerKey_ = function(uid, key, opt_isol) {
...
+    gmf.PermalinkParam.MAP_X);
+  const y = this.ngeoStateManager_.getInitialValue(
+    gmf.PermalinkParam.MAP_Y);
 
   if (x !== undefined && y !== undefined) {
+    goog.asserts.assertNumber(x);
```
gberaudo: I would instead introduce a `ngeoStateManager.getInitialNumberValue` returning a number.

sbrunner:
Requires #2212

PR #2236, opened 2017-01-24
For an unknown reason the value can be null

sbrunner:
see: https://travis-ci.org/camptocamp/c2cgeoportal/builds/194753416#L4777


sbrunner:
In fact we should remove nearly all the 400 type cast to make it possible to debug...

sbrunner:
Thanks :-)

PR #2239, opened 2017-01-24
Replace the 'query' service for the 'querent' and 'mapquerent'

adube:
**Work in progress**

This PR introduces replacement for the `ngeo.Query` service.  Their replacements are:

 * `ngeo.Querent` - a service that issues WMS GetFeatureInfo and/or WFS GetFeature queries through promises. It doesn't store the result.  It uses a list of `ngeo.DataSource` as point of entry to do its requests.  It supports combining requests when data sources use the same url.

 * `ngeo.MapQuerent` - the direct equivalent of the `ngeo.Query` service, but it uses the `ngeo.Querent` to issue queries.

All directives, services, examples and app controllers are updated in this PR to use these new components instead of the `ngeo.Query`.

The structure of the `ngeo.QueryResult` is the same as before.  It is defined in the `ngeo.MapQuerent` service. The only difference is that the sources are removed on each new query.  Before, they were created when sources were created.  This change doesn't affect how they are used in the end.

## Not included in this PR

The following features are not included in this PR.  They should come in in further PRs:

 * support of 'dimensions'
 * support WFS resultType = hits.
 * remove the query.js file

## Todo

 * [x] Wait for #2221 to be merged
 * [x] Wait for #2223 to be merged
 * [x] Review (you can look at the code in the last commit only)

### DIFF
--- a/contribs/gmf/examples/displayquerywindow.html
+++ b/contribs/gmf/examples/displayquerywindow.html
@@ -301,15 +301,15 @@
 
     <p id="desc">
       This example shows how to use the <code>ngeo-map-query</code>
-      directive in combination with the <code>gmf.QueryManager</code>.
-      The QueryManager fetches the themes returned by the theme service and
-      adds one source to the query service per layer definition found. In this
-      example, the layer tree is responsible of creating the layer. You can
-      switch theme and turn on/off layers to see the impact it has on results
-      returned by the query service.
-      To display results, this example use the <code>gmf-displayquerywindow
-      </code> component. Here, this last directive uses a custom style to
-      display all selected features.</p>
+      directive in combination with the <code>gmf.DataSourcesManager</code>.
+      The DataSourcesManager fetches the themes returned by the theme
+      service and adds one DataSource, which is used by the querent service
+      per layer definition found. In this example, the layer tree is
+      responsible of creating the layer. You can switch theme and turn
+      on/off layers to see the impact it has on results returned by the
+      query service.  To display results, this example use the
+      <code>gmf-displayquerywindow </code> component. Here, this last
+      directive uses a custom style to display all selected features.
     </p>
 
     <input type="checkbox"
--- a/src/services/datasourceshelper.js
+++ b/src/services/datasourceshelper.js
@@ -1,0 +1,95 @@
+goog.provide('ngeo.DataSourcesHelper');
+
+goog.require('ngeo');
+goog.require('ngeo.DataSource');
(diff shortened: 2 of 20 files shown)


adube:
@fredj You can review the last commit only.

contribs/gmf/examples/displayquerywindow.html
```diff
@@ -301,12 +301,12 @@
...
       This example shows how to use the <code>ngeo-map-query</code>
-      directive in combination with the <code>gmf.QueryManager</code>.
-      The QueryManager fetches the themes returned by the theme service and
-      adds one source to the query service per layer definition found. In this
-      example, the layer tree is responsible of creating the layer. You can
-      switch theme and turn on/off layers to see the impact it has on results
-      returned by the query service.
+      directive in combination with the <code>gmf.DataSourcesManager</code>.
+      The DataSourcesManager fetches the themes returned by the theme service
+      and adds one DataSource, which is used by the querent service per lay
```
fredj: `layer` word is cut

adube: fixed

adube:
Rebased.

src/services/datasourceshelper.js
```diff
@@ -0,0 +1,92 @@
...
+  }
+
+  /**
+   * Called when a new data source is added to the ngeo collection. Add it
+   * to the cache.
+   * @param {ol.Collection.Event} evt Event
+   * @private
+   */
+  handleDataSourcesAdd_(evt) {
+    const dataSource = /** @type {ngeo.DataSource} */ (evt.element);
```
sbrunner: We should avoid type-case ...

src/services/datasourceshelper.js
```diff
@@ -0,0 +1,92 @@
...
+  }
+
+  /**
+   * Called when a data source is removed from the ngeo collection. Remove it
+   * from the cache.
+   * @param {ol.Collection.Event} evt Event
+   * @private
+   */
+  handleDataSourcesRemove_(evt) {
+    const dataSource = /** @type {ngeo.DataSource} */ (evt.element);
```
sbrunner: type-case ...

sbrunner:
You have an unreported error here:
https://travis-ci.org/camptocamp/ngeo/jobs/195867987#L588
...

sbrunner:
@fredj if it's OK for you you can merge it :-)

fredj:
Thanks @adube

PR #2248, opened 2017-01-26
Number and strings shouldn't be set to null

sbrunner:


sbrunner:
requires #2234

contribs/gmf/src/services/datasourcesmanager.js
```diff
@@ -435,7 +435,7 @@ gmf.DataSourcesManager = class {
    * layertree controller being a 'leaf'.
    *
    * @param {ngeo.LayertreeController} treeCtrl The layer tree controller
-   * @param {?string} newVal New state value
+   * @param {string=} newVal New state value
```
gberaudo: Is it optional? If no it should be `|undefined`.

contribs/gmf/src/services/snapping.js
```diff
@@ -369,7 +369,7 @@ gmf.Snapping.prototype.getWFSConfig_ = function(treeCtrl) {
 
 /**
  * @param {ngeo.LayertreeController} treeCtrl The layer tree controller
- * @param {?string} newVal New state value
+ * @param {string=} newVal New state value
```
gberaudo: Idem, is it optional?

adube:
I just noticed that this change broke something.  In the `ngeo.SyncDataSourcesMap` service, we do something like this:

```
const minResolution = dataSource.minResolution;
const inMinRange = minResolution === null || resolution >= minResolution;
```

`dataSource.minResolution` used to be either a number or null.  Now, it's a number or undefined, so the above no longer works as expected.  The compiler was not smart enough to make us aware of this.

So, before I make any fix for this issue, I need to ask: why was this change made?  A `null` value is something defined, i.e. `dataSource.minResolution`, when it could be `null` was something that we were aware that existed.  Now it can be `undefined`.  `dataSource.foo` can also be undefined.  See where I'm getting at?  Both are technically the same but that's not the case at all.  Okay, we have a compiler that would make us aware that `dataSource.foo` is not a valid property, but `null` is a proper way to say "hey, that value doesn't have anything set yet".

So yeah, why?

@gberaudo @sbrunner ?

PR #2276, opened 2017-02-03
Remove goog.provide for controllers

fredj:


fredj:
Note that the `@export` is removed

sbrunner:
Should be rebased ...

PR #2300, opened 2017-02-08
Removes usage of `goog.storage.mechanism.HTML5LocalStorage`, fix master

sbrunner:


fredj:
From memory, the local storage is not available on safari in incognito mode (raises an error).
A test could be added to avoid this

src/services/statemanager.js
```diff
@@ -50,15 +61,14 @@ ngeo.StateManager = function(ngeoLocation, ngeoUsedKeyRegexp) {
 
   if (paramKeys.length === 0 ||
       (paramKeys.length === 1 && paramKeys[0] == 'debug')) {
-    if (this.localStorage.isAvailable()) {
-      for (const key in this.localStorage) {
-        goog.asserts.assert(key !== null);
-        goog.asserts.assert(key !== undefined);
+    if (this.useLocalStorage !== false) {
```
gberaudo: Should be simplified with `if (this.useLocalStorage)`

src/services/statemanager.js
```diff
@@ -142,14 +152,14 @@ ngeo.StateManager.prototype.getInitialBooleanValue = function(key) {
  */
 ngeo.StateManager.prototype.updateState = function(object) {
   this.ngeoLocation.updateParams(object);
-  if (this.localStorage.isAvailable()) {
+  if (this.useLocalStorage !== false) {
```
gberaudo: Idem: `if (this.useLocalStorage)`

src/services/statemanager.js
```diff
@@ -161,8 +171,8 @@ ngeo.StateManager.prototype.updateState = function(object) {
  */
 ngeo.StateManager.prototype.deleteParam = function(key) {
   this.ngeoLocation.deleteParam(key);
-  if (this.localStorage.isAvailable()) {
-    this.localStorage.remove(key);
+  if (this.useLocalStorage !== false) {
+    delete window.localStorage[key];
```
gberaudo: Idem if condition.
Shouldn't it be `window.localStorage.removeItem(key)`?

PR #2304, opened 2017-02-08
Add a query layer pointer if ngeoQueryOptions.hoverCursor is true

fredj:
replace #2134 

todo: test with an authenticated layer

### DIFF
--- a/src/services/layerHelper.js
+++ b/src/services/layerHelper.js
@@ -81,7 +81,8 @@ ngeo.LayerHelper.prototype.createBasicWMSLayer = function(sourceURL,
   const source = new ol.source.ImageWMS({
     url: sourceURL,
     params,
-    serverType: olServerType
+    serverType: olServerType,
+    crossOrigin: 'anonymous'
   });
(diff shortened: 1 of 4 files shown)


src/services/layerHelper.js
```diff
@@ -81,7 +81,8 @@ ngeo.LayerHelper.prototype.createBasicWMSLayer = function(sourceURL,
   const source = new ol.source.ImageWMS({
     url: sourceURL,
     params,
-    serverType: olServerType
+    serverType: olServerType,
+    crossOrigin: 'anonymous'
```
sbrunner: I think that we should add an option for that, and in a second time id should be configurable by the OGC server :-)

rbovard:
> todo: test with an authenticated layer

@fredj Did you managed to test it? Do you think it won't work as it?

rbovard:
Thanks @fredj!

rbovard:
@fredj It was merged in master but which version was it on march (`2.1` or `2.2`)?

PR #2307, opened 2017-02-09
Fix the WMTS layers initialisation

sbrunner:


gberaudo:
What was the issue? A compiler error?

sbrunner:
We use the name in place of layer and in the last deploy the layer and the name will be different => braking the application :-(

gberaudo:
OK

PR #2283, opened 2017-02-06
Filter - Add list of filtrable data sources to filterselector directive

adube:
This PR introduces a dropdown list of "filtrable" data sources within the `ngeo.FilterSelector` directive (component).

Important note: the GMF server is not yet ready and doesn't have any information about the filtrable layers yet.  For now, the queryable data sources are used instead.  I can't go much further for now...

PR #2309, opened 2017-02-09
Use gmfUser to populate filtrable data sources in gmf.FilterSelector

adube:
This PR is the follow-up of #2283.  It ensures that the data sources that are filtrables are added in the `gmf.FilterSelector` directive (component).  For that, the `gmf.User` service is used, from where the `filtrable_layers` are obtained in the `functionnalities`.  In other words, you need to be logged in to use the filter tool.

Once the tool is activated AND the list of filtrable layers are available, that's where the data sources are registered and added to the list.

## Todo

 * [x] Wait for #2283 to be merged, then rebase
 * [x] Review

### DIFF
--- a/contribs/gmf/examples/filterselector.js
+++ b/contribs/gmf/examples/filterselector.js
@@ -57,14 +57,13 @@ gmfapp.MainController = class {
    *     manager service.
    * @param {gmf.Themes} gmfThemes The gmf themes service.
    * @param {gmf.TreeManager} gmfTreeManager gmf Tree Manager service.
-   * @param {gmfx.User} gmfUser User.
    * @param {ngeo.DataSources} ngeoDataSources Ngeo collection of data sources
    *     objects.
    * @param {ngeo.ToolActivateMgr} ngeoToolActivateMgr Ngeo ToolActivate manager
    *     service.
    * @ngInject
    */
-  constructor($scope, gmfDataSourcesManager, gmfThemes, gmfTreeManager, gmfUser,
+  constructor($scope, gmfDataSourcesManager, gmfThemes, gmfTreeManager,
       ngeoDataSources, ngeoToolActivateMgr
   ) {
 
@@ -81,12 +80,6 @@ gmfapp.MainController = class {
      * @export
      */
     this.gmfTreeManager = gmfTreeManager;
-
-    /**
-     * @type {gmfx.User}
-     * @export
-     */
-    this.gmfUser = gmfUser;
 
     /**
      * @type {ol.Map}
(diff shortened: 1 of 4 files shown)


adube:
Travis passes.

adube:
Rebased

contribs/gmf/options/gmfx.js
```diff
@@ -476,6 +476,14 @@ gmfx.AuthenticationFunctionalities.prototype.default_basemap;
  */
 gmfx.AuthenticationFunctionalities.prototype.default_theme;
 
+
+/**
+ * A list of layer names that can be filtered.
+ * @type {Array.<string>|undefined}
+ */
+gmfx.AuthenticationFunctionalities.prototype.filtrable_layers;
```
fredj: should also be in the `gmfx.AuthenticationFunctionalities` typedef above

contribs/gmf/examples/filterselector.js
```diff
@@ -80,6 +83,12 @@ gmfapp.MainController = class {
     this.gmfTreeManager = gmfTreeManager;
 
     /**
+     * @type {gmfx.User}
+     * @export
+     */
+    this.gmfUser = gmfUser;
```
fredj: not used ?

adube:
Fixed.  Thanks for the review.

fredj:
Not related but we want to stop `goog.provide` the controllers if it's not needed.
Can you please remove it? (maybe you will have to add the `@private` tag to avoid compiler errors)

see #2276

adube:
Sure.  Is it okay if I do this in an other PR?

fredj:
> Sure. Is it okay if I do this in an other PR?

yes, no problem

contribs/gmf/options/gmfx.js
```diff
@@ -476,6 +477,14 @@ gmfx.AuthenticationFunctionalities.prototype.default_basemap;
  */
 gmfx.AuthenticationFunctionalities.prototype.default_theme;
 
+
+/**
+ * A list of layer names that can be filtered.
+ * @type {Array.<string>|undefined}
```
sbrunner: => `{!Array.<string>|undefined}` it can't be null

adube: I'm looking at all the other definitions and none have the `!`.  I think the compiler requires a `?` when a value can be null.  I think this definition is okay.  What do you think?

Otherwise, if all other properties can't be null, then they would all need `!`.

sbrunner: By default an object can be null, for coherence live it like it, we will change it later if necessary :-)

adube: Understood

adube:
Travis is happy

PR #2321, opened 2017-02-13
Filter - Allow anonymous filtering

adube:
This PR allows the list of filtrable data sources to be filled upon loading the page and if we are not connected.  The 'anonymous' user can contain functionnalities, therefore can contain filtrable layers.

**Note**: This works, but only when loading the page.  If you log in with a user, then log out you're not as the 'anonymous' user thereafter and no filtrable layers are there.  If this is an issue, this should be addressed in an other PR as this is out of scope of the Filter project.


adube:
Ready for review.

adube:
Travis is happy.

PR #2322, opened 2017-02-13
Use injector in the gmfPermalink service

### LINKED ISSUE #1962: Removes permalink required dependencies
On this file:
https://github.com/camptocamp/ngeo/blob/master/contribs/gmf/src/services/permalink.js

We should:
- Replace nearly all the ingection by the ingector.
- Get the "elements" from the ingector.
- Don't crash if an element isn't present.

ger-benjamin:
Fix #1962 

Question (from the issue):
> Don't crash if an element isn't present.

I can do that but it will crash later... What I'm supposed to do ? Adding if statements everywhere in this file ?

### DIFF
--- a/contribs/gmf/src/services/permalink.js
+++ b/contribs/gmf/src/services/permalink.js
@@ -13,7 +13,6 @@ goog.require('ngeo.FeatureHelper');
 goog.require('ngeo.Features');
 goog.require('ngeo.FeatureOverlay');
 goog.require('ngeo.FeatureOverlayMgr');
-goog.require('ngeo.LayerHelper');
 goog.require('ngeo.Popover');
 goog.require('ngeo.StateManager');
 goog.require('ngeo.format.FeatureHash');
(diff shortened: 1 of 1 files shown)


sbrunner:
I think that we should use the has method on the injector to know if an element is present?

Than if an element isn't present the reference will be null, and we should add some if to don't fill and use the related part...

ger-benjamin:
Done, thanks Sbrunner

contribs/gmf/src/services/permalink.js
```diff
@@ -122,115 +123,109 @@ gmf.Permalink = function($timeout, ngeoBackgroundLayerMgr, ngeoDebounce,
...
-  /**
-   * @type {string}
-   * @private
-   */
-  this.defaultTheme_ = defaultTheme;
+    $injector.get('gmfThemeManager') : null;
 
   /**
-   * @type {angular.Scope}
+   * @type {?string}
```
sbrunner: Prefer `{string|undefined}` for native types (string, number and boolean) :-)

contribs/gmf/src/services/permalink.js
```diff
@@ -73,36 +72,38 @@ gmf.module.value('gmfPermalinkOptions',
  * - whether to add a crosshair feature in the map or not
  * - the dimensions value
  *
+ * To avail the whole possibilities offer by the permalink, these services
```
fredj: `To have` instead of `To avail` ?

PR #2324, opened 2017-02-14
Don't use goog.Uri anymore

sbrunner:


src/services/location.js
```diff
@@ -47,48 +45,129 @@ ngeo.Location = function(location, history) {
...
-  } else {
-    extendedUri = this.uri_;
+ngeo.Location.prototype.getUriString = function() {
+  const out = [];
+
+  if (this.schema_) {
+    out.push(this.schema_, ':');
+  }
+
+  if (this.domain_ || this.schema_ == 'file') {
```
gberaudo: Should be `===`.

src/services/location.js
```diff
@@ -47,48 +45,129 @@ ngeo.Location = function(location, history) {
...
+
+    out.push(this.domain_);
+
+    if (this.port_ !== undefined) {
+      out.push(':', String(this.port_));
+    }
+  }
+
+  if (this.path_) {
+    if (this.domain_ && this.path_.charAt(0) != '/') {
```
gberaudo: Should be `!==`.

src/services/location.js
```diff
@@ -47,48 +45,129 @@ ngeo.Location = function(location, history) {
...
+      const value = queryData[key];
+      if (value) {
+        sb.push(`${encodeURIComponent(key)}=${encodeURIComponent(value)}`);
+      } else {
+        sb.push(encodeURIComponent(key));
+      }
+    }
+    return sb;
+  }
+  const eqs = getEncodedQuery(this.queryData_);
```
gberaudo: Could be more explicit like, at least `eQueryData`.

src/services/location.js
```diff
@@ -47,48 +45,129 @@ ngeo.Location = function(location, history) {
...
+  if (eqs.length > 0) {
+    out.push('?', eqs.join('&'));
+  }
+
+  const ef = getEncodedQuery(this.fragment_);
```
gberaudo: Idem, could be more explicit like, at least `eFragment`.
Also rename `getEncodedQuery` to `encode` since it is not specific to `query`.

src/services/location.js
```diff
@@ -221,37 +305,37 @@ ngeo.Location.prototype.getParamKeysWithPrefix = function(prefix) {
  * @export
  */
 ngeo.Location.prototype.getFragmentParamKeysWithPrefix = function(prefix) {
-  return this.getFragmentParamKeys().filter(key => key.indexOf(prefix) == 0);
+  const keys = [];
```
gberaudo: What about an helper function to create an array out of on object, optionaly passing a filter?

src/services/location.js
```diff
@@ -47,48 +45,129 @@ ngeo.Location = function(location, history) {
...
+    fillQueryData(this.queryData_, location.search.substring(1));
+  }
+
+  /**
+   * @type {!Object.<string, string>}
+   * @private
+   */
+  this.fragment_ = {};
+  if (location.hash) {
+    fillQueryData(this.fragment_, location.hash.substring(1));
```
gberaudo: Since it is local and not specific to query data. I suggest to rename the function from `fillQueryData` to `fill`.

src/services/location.js
sbrunner: For the second, I don't totally agree because we use the standard query syntax for the fragment and that's not standard...

gberaudo: OK, in this case you are right we should keep the naming.

src/services/location.js
sbrunner: We fill the fragment using the query data format than it's correct for me ...

src/services/location.js
sbrunner: Why not but it's not easy, we should be able to do generic typing to avoid casting.
We filter the key than we should do tow for ...

gberaudo: The code is small and readable so it is fine to me if you keep it as-is.

sbrunner:
Updated :-)

sbrunner:
Not related errors. merging

PR #2328, opened 2017-02-16
Remove unused import

fredj:


fredj:
demo.boundlessgeo.com is down, merging

PR #2330, opened 2017-02-16
Fix the WMTS layers initialisation

sbrunner:
Backport of #2307


sbrunner:
Only boundless errors, merging :-)

PR #2374, opened 2017-02-23
Filter - Create rule component, support 'text' and 'select' types

adube:
This PR introduces the `ngeo-rule` component, which is used to set the value of a rule.  It is used by the `ngeo-filter` directive.

This PR includes:

 * support of `text` rule type
 * support of `select` rule type
 * rename metadata `EnumeratedAttributes` to `enumeratedAttributes`
 * rename metadata `DirectedFilterAttributes` to `directedFilterAttributes`
 * support of removing custom rules

This PR doesn't include:

 * support of `date` rule type (will come in the next PR)
 * support of `geometry` rule type

Please note that I decided to go with always using "apply" and "cancel" buttons  and prevent the closing of the drop-down menu with click outs.  I did this to allow the `active` property of the `ngeo-rule` component to be toggled on while the drop-down menu is shown, which is also linked to a tool activate object (using the tool activate manager).  This allows the deactivation of the query tool while the menu is shown.  This will be especially useful when we introduce the vector features with the geometry filters.

## Todo

 * [ ] Review

### DIFF
--- a/contribs/gmf/examples/filterselector.html
+++ b/contribs/gmf/examples/filterselector.html
@@ -63,9 +63,14 @@
         opacity: 1;
       }
 
+      .ngeo-filter-rule-custom-rm-btn {
+        float: right;
+        margin: 0.4rem 0;
+      }
+
       ngeo-rule {
         display: block;
-        margin: 1rem 0;
+        margin: 1rem 2.5rem 1rem 0;
       }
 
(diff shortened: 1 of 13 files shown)


adube:
Travis is happy.  Ready for review.

fredj:
LGTM too, thanks @adube

adube:
Thanks for the review.

PR #2379, opened 2017-02-24
Only perform animation on elements that have a gmf-animatable css class

### LINKED ISSUE #2378: angular-animate performance issue
ngAnimate causes some performance issue, this can be seen in the debugger console (timeline tab in chrome).

Every time ngIf, ngClass, ngRepeat, ... directives are used, a [forced reflow](https://developers.google.com/web/fundamentals/performance/rendering/avoid-large-complex-layouts-and-layout-thrashing#avoid-forced-synchronous-layouts) is made.

fredj:
fixes #2378


sbrunner:
Shouldn't be on 2.1?

PR #2387, opened 2017-02-27
Transform the ObjectEditing directive into a component

sbrunner:


contribs/gmf/src/directives/objectediting.js
```diff
@@ -270,80 +268,58 @@ gmf.ObjecteditingController = function($scope, $timeout, gettextCatalog,
...
-  this.state_ = geometry ? gmf.ObjecteditingController.State.UPDATE :
-    gmf.ObjecteditingController.State.INSERT;
+  this.state_;
 
   /**
-   * @type {Array.<?ol.geom.Geometry>}
+   * @type {!Array.<!ol.geom.Geometry>}
    * @private
    */
   this.geometryChanges_ = [];
```
adube: This array is supposed to be one or null and geometry types.  This change broke the tool.

contribs/gmf/src/directives/objectediting.js
```diff
@@ -685,6 +714,7 @@ gmf.ObjecteditingController.prototype.resetGeometryChanges_ = function() {
   if (this.geometryChanges_.length === 0) {
     const geometry = this.feature.getGeometry();
     const clone = gmf.ObjecteditingController.cloneGeometry_(geometry);
+    goog.asserts.assert(clone);
```
adube: More specifically, this line broke the tool.  A `null` geometry is valid when we have a feature without any geom, i.e. a new feature.

PR #2391, opened 2017-02-27
Filter custom rules

adube:
This PR allows a distinction between the "directed" and "custom" rules. "Custom" rules can have their operator set using a drop down list.  Directed rules always use the same operator.

## Todo

 * [x] Wait for #2374 to be merged first.
 * [ ] Review

PR #2394, opened 2017-02-28
Filter - Support date rule

adube:
This PR adds the support of filter rules using a date.  For that purpose, the `ngeo-date-picker` directive is used.

## About the datepicker

The datepicker can't have its values set other by itself.  That's contrary to the other types of rules, i.e. a text rule can have its value input to something, then the user clicks 'cancel' and the menu is closed and value reverted.  That can't happen with the date picker.  However, it's possible to create a date picker giving it initial values.  For that reason, the date picker directive is created every time the menu of a rule is opened and destroyed on close.  This allow the control of the value to set.

Also, please note the following important things:

 * when using a date range, the "upperValue" is equal to "today" by default.
 * when using a date rante, the "lowerValue" is equal to "a week ago" by default.

## Todo

 * [x] Wait for #2391 to be merged
 * [x] Adjust the "grey" value being displayed (currently, the milliseconds are shown)
 * [x] Review

### DIFF
--- a/src/directives/rule.js
+++ b/src/directives/rule.js
@@ -1,6 +1,8 @@
 goog.provide('ngeo.ruleComponent');
 
 goog.require('ngeo');
+/** @suppress {extraRequire} */
+goog.require('ngeo.DatePickerDirective');
 goog.require('ngeo.RuleHelper');
 goog.require('ngeo.ToolActivate');
 goog.require('ngeo.ToolActivateMgr');
@@ -11,6 +13,7 @@ ngeo.RuleController = class {
 
   /**
    * @param {!angularGettext.Catalog} gettextCatalog Gettext service.
+   * @param {!angular.Scope} $scope Angular scope.
    * @param {!angular.$timeout} $timeout Angular timeout service.
    * @param {!ngeo.RuleHelper} ngeoRuleHelper Ngeo rule helper service.
    * @param {!ngeo.ToolActivateMgr} ngeoToolActivateMgr Ngeo ToolActivate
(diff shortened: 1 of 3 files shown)


adube:
Ready for review.

src/directives/rule.js
```diff
@@ -52,17 +57,24 @@ ngeo.RuleController = class {
...
+    /**
      * @type {!angular.$timeout}
      * @private
      */
     this.timeout_ = $timeout;
 
     /**
      * @type {!ngeo.RuleHelper}
-     * @private
+      * @private
```
fredj: extra space

src/directives/rule.js
```diff
@@ -120,6 +171,42 @@ ngeo.RuleController = class {
...
+    // This chunk of code ensures that the rule properties are synchronized
+    // with the TimeProperty objects required to build the datepickers.
+    if (this.clone.type === ngeo.AttributeType.DATE ||
+        this.clone.type === ngeo.AttributeType.DATETIME
+    ) {
+      // Watch 'expression'
+      this.unlisteners_.push(this.scope_.$watch(
+        () => this.clone.expression,
+        (newVal) => {
+          const value = (newVal === null) ? this.createDate_() : newVal;
```
fredj: unneeded parentheses

src/directives/rule.js
```diff
@@ -138,6 +225,10 @@ ngeo.RuleController = class {
     this.active = false;
     this.ngeoToolActivateMgr_.unregisterTool(
       this.toolGroup, this.toolActivate_);
+    for (const unlistener of this.unlisteners_) {
```
fredj: `this.unlisteners_.forEach(unlistener);` ? (untested)

adube:
Thanks for the review.  I'll make the fixes.

src/directives/rule.js
adube: That one is not working.  If it's okay, I'll leave it as it is.

fredj: ok

adube:
Okay, I applied all but the last required changes (see my comment).

adube:
I restarted the Travis build twice and it always fail.  Locally, it builds.  What is the issue, here?

fredj:
@adube the travis issue is not related to your PR, see #2408

adube:
@fredj Thanks for pointing this out.

PR #2402, opened 2017-03-03
Filter - Support spatial rule

adube:
This PR adds the support of spatial rules to the `ngeo-rule` directive.

It also changes the way a rule is activated. Now, an `active` property exists within the `ngeo.rule.Rule` class to manage the activation. This allows the parent directive, `ngeo-filter`, to deactivate a rule before removing it.

Upon activating a geometry rule, an operator is immediately selected and the draw tools are shown.  The user can change the operator using the list and the draw tools are updated accordingly.

When a draw tool is clicked, the operator list is deactivated.  The user can then draw on the map. When the drawing is completed, the tool is deactivated and the geometry can be immediately modified.

Even when deactivated, the feature stays on the map.

## Note about the compiler and the 'expression' property

Also, the `ngeo.rule.Geometry` object has been created.  Because of a current limitation with the compiler, ES6 getter/setter functions can't be extended by child classes.  In the `ngeo.rule.Geometry`, the expression is still used as the common way to have a value for the rule.  It's a GeoJSON string of the geometry.  The rule also has a `feature` property to hold the actual geometry.  The conversion to the expression is done within the rule object itself.  In order to be able to perfectly synchronize both properties, the `expression` setter had to be extended, but the compiler threw an error.  For that reason, the getter/setter for the `expression` were renamed `getExpression` and `setExpression`.  This could be re-changed in the future when the issue with the compiler's gone.

## Todo

 * [x] Wait for #2394 to be merged
 * [ ] Validate the workflow, which is slighly different than the specs
 * [ ] Review

## Live demo

 * https://adube.github.io/ngeo/filter-spatial/examples/contribs/gmf/filterselector.html

Please, disregard the fact that there are missing fonts on that example.

### DIFF
--- a/contribs/gmf/examples/filterselector.html
+++ b/contribs/gmf/examples/filterselector.html
@@ -37,6 +37,57 @@
         display: block;
       }
 
+      /* drawfeature */
+
+      .gmf-icon-circle:before {
+        content: "Circle";
+      }
+      .gmf-icon-line:after {
+        content: 'Line';
+      }
+      .gmf-icon-point:after {
+        content: 'Point';
+      }
+      .gmf-icon-polygon:after {
+        content: 'Polygon';
+      }
+      .gmf-icon-rectangle:after {
+        content: 'Rectangle';
+      }
+      .gmf-icon-text:after {
+        content: 'Text';
+      }
+
+      .tooltip {
+        position: relative;
+        background: rgba(0, 0, 0, 0.5);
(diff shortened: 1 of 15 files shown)


adube:
Ready for review.

Please, validate workflow as well.

adube:
Travis is happy.

fredj:
Code looks very good to me, let's merge. If there's a problem with the workflow we can change it later

PR #2411, opened 2017-03-07
Filter - Support saved filters

adube:
This PR introduces the saving of filters in the local storage.

The `gmf.SavedFilters` is a new service responsible of:

 * storing a list of saved filter items locally
 * storing a list of saved filters for a particular 'active' data source, i.e. it's built using the above list items that have an id equal to the current active data source.
 * saving the items in the local storage

The service is injected in the `gmf-filterselector` component to be able to have the list of items bound to a single data source.  The component has 2 new items added: a drop-down list of the items and a save button.

When saving, we check if there's an item that has the same name for the given active data source first.  If that's the case, an confirmation to overwrite the previous one is prompted.  Otherwise, the save occurs.

## Also introduced in this PR

A new `aRuleIsActive` property has been added to the `gmf-filterselector` and `ngeo-filter` components.  That property is set on when there's one rule currently active.  When that's the case, a bunch of components are disabled, such as: remove buttons, save button, apply button, etc.

When a rule is active, its edition must be completed first before being able to do something else.  This prevents a lot of unwanted behaviour.

## Todo

 * [x] Wait for #2402 to be merged
 * [ ] Review

## Live demo

 * https://adube.github.io/ngeo/filter-saved/examples/contribs/gmf/filterselector.html

### DIFF
--- a/contribs/gmf/examples/filterselector.html
+++ b/contribs/gmf/examples/filterselector.html
@@ -95,6 +95,16 @@
         margin: 1.5rem 0 0.5rem 0;
       }
 
+      .gmf-filterselector-savebtn {
+        display: block;
+        text-align: left;
+      }
+
+      .gmf-filterselector-savefilter-desc {
+        color: #999999;
+      }
+
+
       .ngeo-filter-condition-button,
       .ngeo-filter-condition-button:hover,
       .ngeo-filter-condition-button:focus {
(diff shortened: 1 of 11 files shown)


adube:
Travis complains, but locally it built fine for me.  Is there something wrong with Travis again?

fredj:
temporary error with geomapfish-demo.camptocamp.net. I've restarted the checks

adube:
Travis passes.  I'll rebase.

adube:
Ready for review.

fredj:
That was a big one ! LGTM, thanks

adube:
Thanks for the review.

Issue #2381, opened 2017-02-24
Default view


rbovard:
* GeoMapFish version: **2.1.2.dev1**

I'm not sure if it's global or specific to my project, but by default the `map_x`, `map_y`, `map_zoom` values are not set on loading with:

```
mapViewConfig: {
  center: [507000, 137850],
  zoom: 2,
  resolutions: [50, 20, 10, 5, 2.5, 2, 1, 0.5, 0.25, 0.1, 0.05, 0.02]
}
```

And if I try to pan, the values are set to `NaN` and `0`.

Is there something else to do?

ger-benjamin:
I can reproduce it on the demo; It's not specific to your project.

ger-benjamin:
Closed with #2415

PR #2415, opened 2017-03-08
Return null if x,y or z in NaN

ger-benjamin:
Fix https://github.com/camptocamp/ngeo/issues/2381


fredj:
`isNaN(undefined) === true` but `isNaN(null) === false` !

ger-benjamin:
Yes, but that's always to test a value after a parseFloat.
And a parseFloat can onkly return a number or `NaN` (never undefined, never null)

PR #2445, opened 2017-03-15
Filter - Fix saved list of filters

adube:
Fix the saved list of filters in the filter selector tool.

## Todo

 * [x] Wait for #2442 to be merged
 * [ ]  Review

PR #2449, opened 2017-03-16
Filter - Fix various CSS and other issues

adube:
This PR features various fixes for the filter tool:

 * several CSS fixes
 * prevent disabled elements to be clicked
 * deactivate draw/edit tools within the filter tool when we switch to an other tool

## Todo

 * [x] Review

## Live demo

 * https://adube.github.io/ngeo/filter-m-various-fixes/examples/contribs/gmf/apps/desktop_alt/

### DIFF
--- a/contribs/gmf/examples/filterselector.html
+++ b/contribs/gmf/examples/filterselector.html
@@ -95,11 +95,6 @@
         margin: 1.5rem 0 0.5rem 0;
       }
 
-      .gmf-filterselector-savebtn {
-        display: block;
-        text-align: left;
-      }
-
       .gmf-filterselector-savefilter-desc {
         color: #999999;
       }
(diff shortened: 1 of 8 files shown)


ybolognini:
Looks really good.

A tiny detail about separators: action "Add a new criteria" should be in the same block as advanced filters criteria:

 * Block 1: Layer selector
 * Block 2: Wheel + Saved filters, directed filters
 * Block 3: Advanced filters, "Add a new criteria" action
 * Block 4: Apply + Get data, "Save" action

adube:
@ybolognini The blocks are now fixed.  I have **not** updated the live demo.

adube:
@fredj, Let's forget #2445, which is included in here, and only review this PR instead.

PR #2455, opened 2017-03-20
Set theme selector max-height

### LINKED ISSUE #2441: Scroll in theme panel
In GMF 2.1.1, is the theme panel scrollable ?

If not, I think it would be good to set a max size and add a scroll bar if this size is reached.

fredj:
fixes #2441

### DIFF
--- a/contribs/gmf/less/desktop.less
+++ b/contribs/gmf/less/desktop.less
@@ -211,6 +211,9 @@ gmf-search {
 
 gmf-themeselector {
   width: 1.5 * @left-panel-width;
+  max-height: 1.5 * @left-panel-width;
+  overflow: hidden;
+  overflow-y: auto;
 }
 gmf-backgroundlayerselector {
   width: 25rem;


contribs/gmf/less/desktop.less
```diff
@@ -211,6 +211,9 @@ gmf-search {
 
 gmf-themeselector {
   width: 1.5 * @left-panel-width;
+  max-height: 1.5 * @left-panel-width;
```
ger-benjamin: @left-panel-width ? Is that correcton a small height screen ?

fredj: it's just to have (at max) a square div with a scrollbar. this size is arbitrary (I've checked with @ybolognini )

ger-benjamin: OK

kalbermattenm:
Thanks a lot @fredj

Issue #2443, opened 2017-03-15
App freezing


kalbermattenm:
When changing theme to one which has (a lot of) groups with not visible layers and moving the map around, it looks like a request to the mapserver is made (even if there are not activated) for each group and these requests persist.

This freezes the app and even the server. I have to stop Apache the hard way (which in Windows means kill the task...) and restart it to have it back on track...

This really worries me (and I am not sure if this is not related to #2377)

![image](https://cloud.githubusercontent.com/assets/1681332/23950168/55edfb3a-098a-11e7-97d2-96477b1338d3.png)

fredj:
Even if no layers are activated, the GetLegendGraphic requests are made (on zoom change)

fredj:
@kalbermattenm #2456 should have fixes this issue, can you please test again?

kalbermattenm:
Yes, it is much better... Sorry for the lack of support on this and thanks a lot @fredj

PR #2456, opened 2017-03-20
Don't update the legend image is the div is not visible

fredj:
related to #2443


fredj:
forget to update the title, thanks for the review

arnaud-morvan:
@fredj : How about backporting this into 2.1

PR #2467, opened 2017-03-22
Fix some namings

gberaudo:


gberaudo:
Thanks for the review @fredj.

src/proj/epsg2056.js
```diff
@@ -1,8 +1,14 @@
-goog.provide('ngeo.proj.EPSG2056');
+goog.module('ngeo.proj.EPSG2056');
+goog.module.declareLegacyNamespace();
```
sbrunner: What's that?

PR #2496, opened 2017-04-03
Add dimensions in WFS GetFeature requests

arnaud-morvan:


arnaud-morvan:
@fredj : Could you take a look here ?

Note that I do not understand why to not dispatch all global dimensions to each node. In my project I will have to add FLOOR dimension on approximately all layers.

arnaud-morvan:
Note that when I put one dimension on one layer (gmf admin), it is applied on layer parent group (ol.Layer object), can't-we get this applied on all layer tree nodes, regardless of the dimensions set in admin interface ?
For me it does not make sense to put different dimensions on layers, the dimensions values should be only set on the main controller.
For WMS layers, it is passed as a GET parameter, server will use or ignore it.
For WMTS layers, if is is present in url template, server whould use it, if not it will be ignored.

contribs/gmf/src/directives/layertree.js
```diff
@@ -290,6 +290,7 @@ gmf.LayertreeController.prototype.updateLayerDimensions_ = function(layer, node)
       var value = this.dimensions[key];
       if (value !== undefined) {
         dimensions[key] = value;
+        node.dimensions[key] = value;
       }
```
sbrunner: This if looks strange for me, for me it should be like this:
```
       var value = this.dimensions[key];
       dimensions[key] = value !== undefined ? value : node.dimensions[key];
```
@fredj what do you think?

src/services/query.js
```diff
@@ -621,6 +621,22 @@ ngeo.Query.prototype.doGetFeatureRequests_ = function(
...
+      if (dimensions) {
+        for (var key in dimensions) {
+          // get the value from the global dimensions
+          var value = this.dimensions[key];
+          if (value === undefined) {
+            // get the value from the layer default value
+            value = dimensions[key];
+          }
+          if (value !== undefined) {
+            url = goog.uri.utils.setParam(url, key, value);
```
sbrunner: Do not use `goog`!!!

src/services/query.js
```diff
@@ -621,6 +621,22 @@ ngeo.Query.prototype.doGetFeatureRequests_ = function(
...
+      // add dimensions values
+      var dimensions = items[0].source.dimensions;
+      if (dimensions) {
+        for (var key in dimensions) {
+          // get the value from the global dimensions
+          var value = this.dimensions[key];
+          if (value === undefined) {
+            // get the value from the layer default value
+            value = dimensions[key];
+          }
```
sbrunner: Looks duplicate code from layer tree...

contribs/gmf/src/directives/layertree.js
fredj: `node.dimensions` are the default dimension values for a node; it should not be modified

arnaud-morvan: Without this I cannot apply current dimensions on WFS GetFeature requests

src/services/query.js
arnaud-morvan: Ok, can you explain, why do not use goog, what should I use instead ? Note that this comes from doGetFeatureInfoRequests_ existing code.
I see ``goog`` everywhere in this file.

sbrunner: We try to don't use goog anymore to be able to build the project without google library. 
In the master branch almost all the goog (Except goog.[asserts, requires, provide, module]) will be removes, in this case there a method in openlayers that can be used.

arnaud-morvan:
I've changed my desktop.js file with this : 
    this.dimensions['FLOOR'] = '+0';
instead of : 
    this.dimensions = {'FLOOR': '+0'};
so I get the same global dimensions object in main controller and query service (no need to overwrite the node dimensions anymore).

I've passed the parameters directly to $http, so I do not use goog anymore.

What do you think about this updated version, is this good for merging.

arnaud-morvan:
Note that I still have an issue on the server side, where I find really boring to add same dimensions on all layers.

sbrunner:
For the server part you can write a script like this one: https://github.com/camptocamp/epfl_authgeoportail/blob/master/epfl_authgeoportail/scripts/manage_layers.py

arnaud-morvan:
@fredj : Are you OK for merging this PR ?

PR #2499, opened 2017-04-04
Fix CreateFeature - revert to being a directive

adube:
In the [Upgrades for Angular 1.6](784c025fe04c956c5d1df5c0571f5060013fd3bb) task, a bunch of changes were made, such as switching from directives to components.  For one of them, the `ngeo-createfeature`, the change was incompatible.  A component only support being declared as the element level, i.e. `<ngeo-createfeature>`, i.e. it doesn't support being declared using the attribute of an existing element`<div ngeo-createfeature>`.

The tool was made so that it supported being declared using attributes, making it not compatible with components.  The change that was made broke the tool.

This PR reverts that specific change for that tool only.


adube:
Ready for review.

fredj:
Thanks @adube

PR #2508, opened 2017-04-07
Fix the 2.1 merge

sbrunner:


sbrunner:
Have we a solution for that? 
https://travis-ci.org/camptocamp/ngeo/jobs/219669076#L945

PR #2511, opened 2017-04-11
Don't stop tree traversal after first non mixed group

### LINKED ISSUE #2470: NGEO mobile - permalink does not work with WMTS layers
Version 2.1.2

To reproduce : 
1°/ Generate a permalink with CGXP application, here is one : https://preprod.cartoriviera.ch/v2/s/es2I 
2°/ Open the link in a new tab > you see a WMTS overlay layer (historic color map of 2000)
3°/ Add mobile/ after V2 in the URL > NGEO mobile opens, but the WMTS layer is not activated...

Tested with several WMTS layers of this theme, does not work.

fredj:
fixes #2470

### DIFF
--- a/contribs/gmf/src/services/permalink.js
+++ b/contribs/gmf/src/services/permalink.js
@@ -887,7 +887,6 @@ gmf.Permalink.prototype.initLayers_ = function() {
                 treeCtrl.setState(enable ? 'on' : 'off', false);
               }
             });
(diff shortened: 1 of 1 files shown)


sbrunner:
Looks good, but should be on branch 2.1 ...

sbrunner:
It's already on branch 2.1 :-)

PR #2513, opened 2017-04-11
Change value before setting the value to the params obj

fredj:


fredj:
`olServerType != opt_serverType` :-)

PR #2525, opened 2017-04-19
Add number support in editing

sbrunner:
Add missing number support, should we separate them?

What about boolean support?


sbrunner:
string with maxlength should also be supported ...

PR #2555, opened 2017-04-26
Use the message id provided by the parsing result

oterral:
import local and online uses the result object differently. With this PR they works the same .


fredj:
importOnline code:
https://github.com/camptocamp/ngeo/blob/master/src/modules/import/importonline.js#L147-L148

Issue #2530, opened 2017-04-20
edit: no error message displayed is a required field is not set


fredj:
https://camptocamp.github.io/ngeo/master/examples/contribs/gmf/apps/desktop/

PR #2559, opened 2017-04-26
Support more attribute types in editing

### LINKED ISSUE #2527: Missing types support in editing
The following types should be supported:
* integer
* double
* boolean
* string with limit

First micro step: https://github.com/camptocamp/ngeo/pull/2525

Can be tested on layer point

adube:
Fixes #2527 

This PR introduces:

 * a new type of `ngeox.Attribute`: `boolean`
 * a new property for the `ngeox.Attribute`: `numType`, which distincts how to render a number attribute.  The two current values are: `integer` and `float`.
 * a new property for the `ngeox.Attribute`: `maxLength`

These changes are used in the `ngeo.Attributes` component, i.e. in the html template.  Only the text attribute support the "maxLength".  The boolean is rendered as a checkbox.  The "numType" is used to distinguish integers from floats accordingly, and are also both renderered as number input field.

**Note** This PR only changes the attributes that are used in the editing tools.  The ones used in the filter tools remain unchanged as they use WFS DescribeFeatureType requests to get the attributes.

## Todo

 * [ ] Review

### DIFF
--- a/options/ngeox.js
+++ b/options/ngeox.js
@@ -27,6 +27,13 @@ ngeox.AttributeBase.prototype.geomType;
 
 
 /**
+ * Set only if the attribute is a number type. Determines the type of number.
+ * @type {string|undefined}
+ */
+ngeox.AttributeBase.prototype.numType;
+
+
+/**
  * The attribute type, which determines how to render it.
  * @type {string|undefined}
  */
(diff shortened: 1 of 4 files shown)


sbrunner:
Looks good, but the validations errors messages isn't display anymore :-(

fredj:
@sbrunner the validation display is an other issue, see #2530

sbrunner:
OK, then @fredj merge if is't OK for you :-)

PR #2564, opened 2017-04-27
Fix several regressions in ObjectEditing

adube:
There are changes that were made to several components in ngeo and gmf that affected the ObjectEditing project.  This patch fixes some of those issues.

Please, review.


fredj:
Thanks @adube

sbrunner:
Why it isn't possible to use a component?

adube:
https://docs.angularjs.org/guide/component

See under: Comparison between Directive definition and Component definition

**restrict**

Directive: yes
Component: No (restricted to elements only)

sbrunner:
OK, but we should one time to convert it into a component ...

adube:
There's no need to.  First, back when these kind of directives were created, it was decided that having the possibility to define them at the attribute level was nice, because you let more customization.  I.e. one could define any type of node for that directive.  That would not be possible with components.

So, second, if we change them to components, we woudl prevent a lot of customization.

In the end, it's a matter of choice.  A component is something that should be very specific.  Such buttons are okay to be customized and it's okay to still use directives.

What do you think?

PR #2576, opened 2017-05-01
Stop referencing ngeo symbols in ngeox.js

gberaudo:
Referencing ngeo symbols adds a hard dependency on the referenced symbol, forcing applications using the library to pass extra js code to their invocation of the closure compiler.

### DIFF
--- a/options/ngeox.js
+++ b/options/ngeox.js
@@ -377,8 +377,59 @@ ngeox.DataSourceOptions.prototype.wmtsUrl;
 
 /**
  * @interface
+ * @struct
  */
-ngeox.DataSource;
+ngeox.DataSource = function() {};
+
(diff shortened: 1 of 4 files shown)


gberaudo:
There was an issue with the proposed patch: GCC was silently failing to check the accesses to the datasource properties.
With this new commit, I now use an empty function and the `@struct` keyword. With these changes GCC is now actively checking the code, and bails out on bad accesses.
Could you please review again? Thanks

fredj:
rebased on master

gberaudo:
Thanks @fredj for the rebase. Saves me 1 hour! ^^

gberaudo:
Thanks for the reviews.

PR #2578, opened 2017-05-01
FeatureHash - Use same accuracy as in cgxp

### LINKED ISSUE #2573: Drawing: Circles are corrupted when reloading page
Check on Desktop github.io example:
 * Draw a circle
 * Reload app

The more you zoom, the more the issue is visible.

circle

ger-benjamin:
FIX #2573 

Use accuracy of 0.1m instead of 1m. Like in cgxp: https://github.com/camptocamp/cgxp/blob/95ee7b5dbe9a49119d141e302047850787e623ae/core/src/script/CGXP/widgets/RedLiningPanel.js#L75

### DIFF
--- a/src/ol-ext/format/featurehash.js
+++ b/src/ol-ext/format/featurehash.js
@@ -194,7 +194,7 @@ ngeo.format.FeatureHash.CHAR64_ =
  * @const
  * @private
  */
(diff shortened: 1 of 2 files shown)


ger-benjamin:
(to port to 2.2)

PR #2579, opened 2017-05-01
ObjectEditing - Geometry changes can contain null values

adube:
This patch fixes the regression introduced in https://github.com/camptocamp/ngeo/pull/2387/commits/114e05d2dba44bbfe8d7c957085f76cdf0cac1a9

The array `this.geometryChanges_` can have null values.  A null value occurs when the feature has no geometry, i.e. when a feature has no geometry yet in the ObjectEditing project.


adube:
Ready for review. @sbrunner or @fredj .

fredj:
Look good to me, @sbrunner please merge if you agree

PR #2580, opened 2017-05-01
Filter - 'Not' condition should include an OR, not an AND

### LINKED ISSUE #2574: Filters: does "None" option work?
I tried to select "None" option and result looks wrong to me.

Looks like boolean expression is NOT (A AND B) when I was expecting NOT (A OR B).

aucun

adube:
Fixes #2574.

Instead of `NOT (A AND B)`, we now have `NOT (A OR B)`.


adube:
Ready for review.

adube:
Rebased onto master.

PR #2581, opened 2017-05-01
StreetView - do not reposition if change comes from SV navigation

### LINKED ISSUE #2571: StreetView: some points are unreachable
Sometimes, when two StreetView points are close to each other, clicking on an arrow in StreetView makes it start to move to another point and then it comes back to previous one:

 * Open Desktop github.io example
 * Search for "532817 152244"
 * Open StreetView
 * Click on the center
 * Try to go west twice while using StreetView's white arrows

adube:
Fixes #2571.

With this fix, changing the position of the panorama using the controls within it do will not attempt to re-calculate the nearest location thereafter.

### DIFF
--- a/src/directives/googlestreetview.js
+++ b/src/directives/googlestreetview.js
@@ -143,7 +143,7 @@ ngeo.GooglestreetviewController = class {
      * @type {boolean}
      * @private
      */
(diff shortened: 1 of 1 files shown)


adube:
Ready for review.

PR #2582, opened 2017-05-01
StreetView - make 100% height in gmf desktop apps

### LINKED ISSUE #2570: StreetView: variable height
Currently StreetView width AND height are fixed. Specifications weren't really clear about height, but testing now shows that variable height would be much better.

adube:
Fixes #2570.

Variable height for StreetView in desktop and desktop_alt applications.


adube:
Ready for review.

PR #2584, opened 2017-05-02
Add new on-init callback to gmf.searchDirective

fredj:
Same as #2557 but for the master branch.

The first pull request was wrongly for the 2.1 branch


fredj:
fixed with 0312598d486389a514815aaa863923697c9eac7d

PR #2543, opened 2017-04-25
Add directive to import layers from a WMTS server

Jenselme:
I have a working solution that looks good to me. However, I have an error I don't know how to correctly fix when I run `make check`:

```
node buildtools/build.js .build/examples/interactionbtngroup.json .build/examples/interactionbtngroup.js
info ol Parsing dependencies
ERR! Unsatisfied dependency "ol.source.WMTS.optionsFromCapabilities" in script: /home/jenselme/Work/forks/ngeo/src/modules/import/wmtsgetcap.js 
Makefile:346: recipe for target '.build/examples/interactionbtngroup.js' failed
make: *** [.build/examples/interactionbtngroup.js] Error 1
```

I require `ol.source.WMTS.optionsFromCapabilities` in [wmtsgetcap.js](https://github.com/Jenselme/ngeo/blob/wmts-import/src/modules/import/wmtsgetcap.js#L5) because I use it [here](https://github.com/Jenselme/ngeo/blob/wmts-import/src/modules/import/wmtsgetcap.js#L42). How can I require it without causing any issues during the build?

Close #2537

PR #2588, opened 2017-05-04
Correct return type of some get template URL functions

Jenselme:
They used boolean instead of string.

See [this comment](https://github.com/camptocamp/ngeo/pull/2543#discussion_r114507611) in #2543 for the original report.


fredj:
thanks @Jenselme

PR #2590, opened 2017-05-04
Removes stange blur on search bar

sbrunner:


fredj:
?

sbrunner:
Remove that blur
image4198

PR #2591, opened 2017-05-04
ObjectEditing - Ensure 2D geometry

adube:
In the ObjectEditing tool, when using the 'copy' tool, WMS GetFeatureInfo requests are made to fetch the geometry of the features to copy.  The geometry read may contains "XYZ" coordinates.  When those are used to INSERT features using GMF, it fails.  GMF doesn't seem to like having "XYZ" coordinates.

This only occurs on INSERT for features that are copied, because otherwise there would only be "XY" coordinates.

This patch makes sure to convert any "XYZ" coordinates to "XY" just before we save.


adube:
Ready for review.

fredj:
@adube please squash your commits and it will be good to merge

PR #2592, opened 2017-05-04
(WIP) - Support Time in query

adube:
This PR introduces the support of TIME in the querent service.  On WMS, the `TIME` parameter is used. On WFS requests, it is added to any existing filter on the data source, which gets converted to WFS filter in the end.

Note that the time filter works differently than the filters defined in the Filter tool.  A time filter is set directly on the data source.  It can be changed using the widgets that are drawn in the layer tree.  Those used to set the property directly to the layer.  Now, they set the time value in the data source instead.

A data source that has such a time filter can't be combined to other data source when a query occurs.

## Moment.js

This PR also introduces the Moment.js library, which is used to calculate the range of a DURING filter when we only have 1 time value.  MapServer understands it when we give it as `TIME` property in WMS, but in WFS we need to create it manually.  When we have a 2 time values (start, end), those are used "as-is".

The library has been added to all apps, plus all examples that feature either a map query tool or filter tool.

## Todo

 * [x] Support WMS requests properly, i.e. add `TIME` param in WMS GetFeatureInfo requests
 * [x] Support WFS requests properly, i.e. add them to the filter
 * [ ] Review
 * [ ] Use an official version of OL (currently using @fredj 's branch)


adube:
Not officially ready for review, but you can begin one if you want.  You can also try it (locally).

adube:
Ready for review!

adube:
Rebased onto master.

fredj:
@adube I can open a PR in OpenLayers if you thing that the during filter is ready

adube:
From what I've seen so far, I'd say that this is ready.  I would feel safer if someone else could also test this.

adube:
I fixed some errors reported by Travis.  How about what's remaining?  Would someone tell me what's wrong?  Thanks.

sbrunner:
I think it's clear (when we found where it is :-) ):
https://travis-ci.org/camptocamp/ngeo/jobs/229993021#L363
https://travis-ci.org/camptocamp/ngeo/jobs/229993025#L1557

fredj:
@adube to search the cause of an error in travis, I search (using the browser search) for `with 2` (or `with 1` or `error 1`)

adube:
Thanks, but that's not what I meant.  I'm sorry, I should have made it more clear: I don't understand the errors.  How can I fix them?

.eslintrc.yaml
```diff
@@ -21,3 +21,4 @@ globals:
     gmfapp: false
     inject: false
     jsts: false
+    moment: false
```
rbovard: Missing new line (see https://travis-ci.org/camptocamp/ngeo/jobs/229993021#L363)

sbrunner:
Search for `jsts.min.js` in the `Makefile` and you will see what's missing for `moment.min.js` :-)

adube:
Thank you very much.

adube:
Rebased onto master.

sbrunner:
The new version of OpenLayers don't works anymore with node 4 than you should remove this:
https://github.com/camptocamp/ngeo/blob/master/.travis.yml#L26-L30

sbrunner:
And the `desktop` and `desktop_alt` will have an assertion, can you reproduce that on debug mode?

adube:
Thanks for pointing those out.  I made 2 new fixes.

fredj:
thanks @adube

adube:
Thanks @fredj.  I'm surprised to see this merged with the use of you ol3 branch.  Would you please validate that's what you intended to do?

Many thanks,
Alexandre

fredj:
yes, that's intended; so we can test it more easily. I'm creating an issue to not forget to switch to openlayers in package.json once the pull request is merged

adube:
Understood, thanks.

PR #2593, opened 2017-05-05
UI adjustments to gmf oeedit app based on user feedback

pfirpfel:
- Removed panel when in query mode, because there is no content to display in there and it uses up valuable space.
- Closing edit mode panel switches state to the only other option, query mode.
- Query details window didn't handle long content well. Now it allows horizontal scrolling and is 50% wider, since there is more space available because the query mode doesn't have a panel any more.


fredj:
Thanks @pfirpfel , please merge

PR #2609, opened 2017-05-10
Allow ngeo.interaction.ModifyRectangle to be used with other interactions

fredj:
can be tested here:
http://camptocamp.github.io/ngeo/modify_rectangle_you_are_not_alone/examples/contribs/gmf/apps/desktop


src/ol-ext/interaction/modifyrectangle.js
```diff
@@ -95,6 +95,17 @@ ol.inherits(ngeo.interaction.ModifyRectangle, ol.interaction.Pointer);
 
 
 /**
+ * @param {boolean} active Active.
+ * @private
```
pfirpfel: compiles when adding @override

fredj: added, thanks

PR #2629, opened 2017-05-15
Use the good GetMap url from the GetCapabilities

oterral:
The good WMS url to use is written in the Capabilities section not in Service

WMS links to test:

https://geoweb.so.ch/wms/isboden?SERVICE=WMS&REQUEST=GetCapabilities
http://services.sentinel-hub.com/v1/wms/570197e0-4b40-4783-b261-30a054d715bb

### DIFF
--- a/src/modules/import/wmsgetcap.js
+++ b/src/modules/import/wmsgetcap.js
@@ -67,7 +67,7 @@ exports = function($window, gettext, gettextCatalog, ngeoWmsGetCapTemplateUrl) {
     }
 
     if (!layer['isInvalid']) {
-      layer['wmsUrl'] = getCap['Service']['OnlineResource'];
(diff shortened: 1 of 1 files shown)


oterral:
any opinions?

ponceta:
http://www.sigip.ch/main/wsgi/mapserv_proxy

Other possible wms link to test. (GMF base WMS url)

gberaudo:
@sbrunner, do you have opinion on this?

sbrunner:
Actually I don't have any opinion about it, it's seem better like it but we will have probably an other opinion the day we have an issue with a server who has not the same URL :-(

oterral:
According to the spec the good url is in the Capability tag.

sbrunner:
Yes be we also use it to do GetFeature Info, than id the URLs are defendant we should store all of them. For me we can merge it and when we had an issue about that, we will store all of them ...

gberaudo:
OK,then I am merging it, thanks all.

PR #2639, opened 2017-05-17
Fix the i18n string collection

### LINKED ISSUE #2635: 2.1 - Coordinates not translated
"Coordinates" is not translated (lower-right of image):

image

It is strange, because it is called through the `gettext` function: https://github.com/camptocamp/ngeo/blob/2.1/contribs/gmf/src/directives/mouseposition.js#L88

(By the way... I made a suggestion on Transifex... `Coordonnées` should be in its plural form...)

### LINKED ISSUE #2636: Translation question...
There is a string here:
https://github.com/camptocamp/ngeo/blob/2.1/contribs/gmf/src/directives/objectediting.js#L418
As I understand it, this one should be translated.

But there is nothing about it on Transifex: https://www.transifex.com/camptocamp/ngeo/translate/#fr/gmf-2_1/97419280

How is that working ???

Is there some mechanism which adds string to translate to Transifex ?

Should I do that that manually ?

I am willing to help, but I do not get how the whole thing works...

### LINKED ISSUE #2638: Geometry type in measure should be translated
image

The types which should be translated: Polygon, Circle, LineString, Point (I know... in French it would be OK, but not in German...), Rectangle

### LINKED ISSUE #2640: Translate elevation data sources
We should be able to translate them : 

image

(even in english...)

sbrunner:
fix #2635
fix #2636
fix #2640
fix #2638

### DIFF
--- a/contribs/gmf/src/directives/authenticationdirective.js
+++ b/contribs/gmf/src/directives/authenticationdirective.js
@@ -170,6 +170,7 @@ gmf.AuthenticationController = function(gettextCatalog, $scope,
  * @export
  */
 gmf.AuthenticationController.prototype.changePassword = function() {
+  var gettextCatalog = this.gettextCatalog;
 
   var oldPwd = this.oldPwdVal;
   var newPwd = this.newPwdVal;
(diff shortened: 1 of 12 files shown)


kalbermattenm:
You are faster than me being able to write issues... ;-)

Thanks

kalbermattenm:
Thanks a lot !

PR #2648, opened 2017-05-18
Put some order in easting/northing coordinates filter

### LINKED ISSUE #2641: Contextual data - latitude / longitude
I do not know if this is done on purpose, but for the WGS84 coordinates, once it shows longitude / latitude (decimal version) and once latitude / longitude (DMS version).

Would it not be more consistent to use the same order for both of them ? (or if not, what is the point in doing so...?)

image

sbrunner:
fix #2641


kalbermattenm:
Thanks !

PR #2652, opened 2017-05-22
Make translation targets configurable

pvalsecc:
GeoMapfish uses .tx/config.mako as is for the project it generates and those
projects must support more than de and fr.
Now, the languages are passed as a variable to the MAKO file.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -927,7 +927,7 @@ $(HOME)/.transifexrc:
 
 .tx/config: .tx/config.mako .build/python-venv/bin/mako-render
 	PYTHONIOENCODING=UTF-8 .build/python-venv/bin/mako-render \
-		--var "tx_version=$(TX_VERSION)" $< > $@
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -927,7 +927,7 @@ $(HOME)/.transifexrc:
 
 .tx/config: .tx/config.mako .build/python-venv/bin/mako-render
 	PYTHONIOENCODING=UTF-8 .build/python-venv/bin/mako-render \
-		--var "tx_version=$(TX_VERSION)" $< > $@
+		--var "tx_version=$(TX_VERSION)" --var "languages=de fr" $< > $@
```
fredj: Use the `L10N_LANGUAGES` variable?

pvalsecc: Good point. Done.

sbrunner:
Thanks :-)

PR #2653, opened 2017-05-23
Escape string before using it as a regex

oterral:
Avoid a js error see https://github.com/geoadmin/mf-geoadmin3/issues/3785


oterral:
The failure doesn't seem to be related to the code.

sbrunner:
Why using a regexp for that?
We can't simply use `=` (or includes)?

oterral:
No particular reason. We could use an other method too.

sbrunner:
+1 for not using regex :-)

oterral:
Removed

gberaudo:
Thanks

PR #2656, opened 2017-05-26
Switch to WMS layer in print

### LINKED ISSUE #2463: Use WMS instead of WMTS for print
I'm trying to use WMS instead of WMTS for printing background layers.

According to @eleu, it should be possible by adding the properties "wmsUrl" and "wmsLayers" to the background layer's metadata, but this has no effect. Is there some configuration missing? A test by @eleu on the GMF 2.1 demo application didn't work either.

I'm using GeoMapFish version 2.1.1 and the resulting themes response looks like this:
```
{
  "background_layers": [
    {
      "layer": "grundkarte_farbig", 
      "dimensions": {}, 
      "name": "grundkarte_farbig", 
      "url": "http://tile.geoview.bl.ch/1.0.0/WMTSCapabilities.xml", 
      "type": "WMTS", 
      "id": 9999998, 
      "imageType": "image/png", 
      "metadata": {
        "wmsUrl": "http://geoview.bl.ch/main/wsgi/mapserv_proxy", 
        "wmsLayers": "grundkarte_farbig_group", 
        "thumbnail": "http://geoview.bl.ch/main/wsgi/sfr/static/images/baselayer_farbig.png"
      }
    }, 
    ...	
  ]
  ...
}
```

fredj:
work in progress

fixes #2463

### DIFF
--- a/contribs/gmf/src/directives/print.js
+++ b/contribs/gmf/src/directives/print.js
@@ -8,6 +8,8 @@ goog.require('ngeo.CreatePrint');
 goog.require('ngeo.FeatureOverlayMgr');
 goog.require('ngeo.LayerHelper');
 goog.require('ngeo.PrintUtils');
+goog.require('ol.Map');
+goog.require('ol.layer.Group');
 
 
 /**
@@ -133,6 +135,7 @@ gmf.module.directive('gmfPrint', gmf.printDirective);
  * @param {ngeox.QueryResult} ngeoQueryResult ngeo query result.
  * @param {ngeo.FeatureHelper} ngeoFeatureHelper the ngeo FeatureHelper service.
  * @param {angular.$filter} $filter Angular $filter service.
+ * @param {gmf.Themes} gmfThemes The gmf Themes service.
  * @constructor
  * @export
  * @ngInject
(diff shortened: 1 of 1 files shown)


contribs/gmf/src/directives/print.js
```diff
@@ -665,7 +681,34 @@ gmf.PrintController.prototype.print = function(format) {
...
+  var map = new ol.Map({});
+  map.setView(this.map.getView());
+  var ol_layers = this.ngeoLayerHelper_.getFlatLayers(this.map.getLayerGroup());
+  var new_ol_layers = [];
+  for (var i = 0, ii = ol_layers.length; i < ii; i++) {
+    var layer = ol_layers[i];
+    var metadata = layer.get('metadata');
+    if (metadata) {
+      var server_name = metadata.ogcServer;
+      var layer_names = metadata.layers || metadata.printLayers;
```
sbrunner: layers and print layers should be inverted :-)

contribs/gmf/src/directives/print.js
```diff
@@ -665,7 +681,34 @@ gmf.PrintController.prototype.print = function(format) {
...
+      if (server_name && layer_names) {
+        var server = this.ogcServers_[server_name];
+        goog.asserts.assert(server);
```
sbrunner: In this case we should log an error with production code, the server name isn't verified before ...

contribs/gmf/src/directives/print.js
fredj: done

contribs/gmf/src/directives/print.js
fredj: done

PR #2659, opened 2017-05-29
ObjectEditing - Disable save button when feature has no geom

adube:
This PR fixes an issue that occured in the ObjectEditing tools.  When a user manually deletes the geometry of an existing feature, whether by using the "minus" or "scisors" buttons, clicking on "save" caused the server to fail because it was attempting to save with an empty geometry.

To prevent this, we add an additionnal condition to the save button.  It gets disabled automatically if the feature has no geometry at any point.  The user can still use the delete button to remove the feature, which doesn't use the empty geometry if the user had deleted it manually, and the request to delete it works.

## Todo

 * [x] Review


adube:
@fredj Travis is happy.

PR #2665, opened 2017-06-02
Be able to have custom print attribute

sbrunner:


contribs/gmf/examples/print.html
```diff
@@ -180,7 +180,7 @@
 
     <p id="desc">
       This example shows how to use the <code>gmf.printDirective</code>.
-      Directive that generates fields and manages tools to print with a
+      Directive that generates layout informations and manages tools to print with a
```
gberaudo: No "s" to information.

contribs/gmf/src/directives/print.js
```diff
@@ -69,16 +69,31 @@ gmf.module.value('gmfPrintTemplateUrl',
...
- * Example:
+ * Simple example:
  *
  *      <gmf-print
  *        gmf-print-map="mainCtrl.map"
  *        gmf-print-active="printActive"
  *        gmf-print-rotatemask="true">
  *      </gmf-print>
  *
+ * Example with u ser defined attribute:
```
gberaudo: Extra space

contribs/gmf/src/directives/print.js
```diff
@@ -344,6 +363,12 @@ gmf.PrintController = function($rootScope, $scope, $timeout, $q, $injector,
   this.rotation = 0;
 
   /**
+   * @type {Array.<string>}
+   * @export
+   */
+  this.hiddenAttributes;
```
gberaudo: Might be better to name it `hiddenAttributeNames` to distinguish with `attributesOut` which are not an array of strings.

contribs/gmf/src/directives/print.js
```diff
@@ -461,7 +486,7 @@ gmf.PrintController.prototype.getCapabilities_ = function(opt_roleId) {
 
 /**
  * Create the list of layouts, get the formats, get the first layout in
- * gmf print v3 capabilities and then update the print panel fields.
+ * gmf print v3 capabilities and then update the print panel layout informations.
```
gberaudo: No s.

contribs/gmf/src/directives/print.js
```diff
@@ -500,46 +525,48 @@ gmf.PrintController.prototype.updateFields_ = function() {
...
+  this.attributesOut = this.layoutInfo['simpleAttributes'];
+
   // Force the update of the mask
   this.map.render();
 };
 
 
 /**
- * Update customs fields with gmfx.Customfield to be able to generate a form
+ * Update simple attributes informations with gmfx.Customfield to be able to generate a form
```
gberaudo: No s here and in other occurences below.

contribs/gmf/src/directives/print.js
```diff
@@ -109,11 +125,14 @@ gmf.printDirective = function(gmfPrintTemplateUrl) {
...
     replace: true,
     restrict: 'E',
+    transclude: true,
     scope: {
       'map': '=gmfPrintMap',
       'active': '=gmfPrintActive',
       'rotateMask': '&?gmfPrintRotatemask',
-      'fieldValues': '&?gmfPrintFieldvalues'
+      'fieldValues': '&?gmfPrintFieldvalues',
+      'hiddenAttributeNames': '=gmfPrintHiddenattributes',
```
fgravin: Are the 2 new fields mandatory ? If not use `?`, otherwise it breaks all code that currently use the directive.

Also, is the 2 ways binding usefull ? if not use the `<`, it seems not as you pass expression in the attribute value like `gmf-print-hiddenattributes="['title', 'comments']"`.

@sbrunner @fredj @gberaudo @ger-benjamin 
Please try to more strict with your bindings.

kalbermattenm:
@sbrunner, this went into master, right ?

Thus I will have to wait 2.3 to be able to use it ?

sbrunner:
No it was done before create the 2.2 we branch :-), it should be available on 2.2.0 :-)

kalbermattenm:
OK... then I have some questions.... because I cannot make it work...

1) `gmf-print-attributes-out="attributes"`, `attributes` corresponds to the content of the print `config.yaml` file:

```yaml
throwErrorOnExtraParameters: true

templates:
    1 A4 portrait: !template
        reportTemplate: A4_Portrait.jrxml
        attributes:
            title: !string &title
                default: ""
            comments: !string &comments
                default: ""
            debug: !boolean &debug
                default: false
            legend: !legend &legend {}
            map: !map &map
                maxDpi: 254
                dpiSuggestions: [254]
                zoomLevels: !zoomLevels
                    scales: [100, 250, 500, 2500, 5000, 10000, 25000, 50000, 100000, 500000]
width: 555
...
```
And this is all retrieved by making a request to capabilities.json

Moreover, I simply tried:

```html
<gmf-print
  gmf-print-map="mainCtrl.map"
  gmf-print-active="mainCtrl.printActive"
  gmf-print-attributes-out="attributes"
>
  <div ng-repeat="attribute in attributes" class="gmf-print-layoutinfo">
    <div ng-if="attribute.name == 'img_commune'" class="form-group" style="margin-top: 1rem;">
      <label class="control-label col-md-5">COMMUNE</label>
    </div>
  </div>
</gmf-print>
```
But the result:
image

It looks like the stuff inside the div containing `ng-repeat` isn't taken into account. Moreover, the default stuff is still showed (you can see img_commune rendered as some kind of default text box).

Am I missing something ?

sbrunner:
Deeper investigations needed ...

kalbermattenm:
Just for info, I copied-pasted the wrong config.yaml content... I updated it just now...

PR #2668, opened 2017-06-02
Fix decimal / precision mixage and unworking

### LINKED ISSUE #2664: gmf-mobile-measurelength-decimals is not used
the mobileMeasureLengthDirective api
http://camptocamp.github.io/ngeo/master/apidoc/gmf.mobileMeasureLengthDirective.html
say we can use a gmf-mobile-measurelength-decimals option.
but that option value is never used,
see:

https://github.com/camptocamp/ngeo/blob/master/src/ol-ext/interaction/measure.js#L275-L285

since the decimal, according to @sbrunner has no meaning (see with him directly), remove that unused gmf-mobile-measurelength-decimals option

sbrunner:
fix #2664

### DIFF
--- a/contribs/gmf/src/directives/mobilemeasurelength.js
+++ b/contribs/gmf/src/directives/mobilemeasurelength.js
@@ -37,8 +37,7 @@ gmf.module.value('gmfMobileMeasureLengthTemplateUrl',
  *
  * @htmlAttribute {boolean} gmf-mobile-measurelength-active Used to active
  * or deactivate the component.
- * @htmlAttribute {number=} gmf-mobile-measurelength-decimals number of decimal
- *     to display.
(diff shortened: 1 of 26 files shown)


contribs/gmf/src/directives/mobilemeasurelength.js
```diff
@@ -113,6 +112,12 @@ gmf.MobileMeasureLengthController = function($scope, ngeoDecorateInteraction, $f
    * @type {number|undefined}
    * @export
    */
+  this.precision;
+
+  /**
+   * @type {number|undefined}
+   * @export
+   */
   this.decimals;
```
fredj: `decimals` is removed from the scope, is `this.decimals` still useful?

sbrunner: Just removed :-)

PR #2669, opened 2017-06-06
OSRM routing

pfirpfel:
**Current status:** seems to work, needs more testing

## Open tasks

- [x] move from gmf to ngeo
- [x] check documentation in source code if it needs more explanation
- [ ] add documentation in geoportal admin/integrator guide -> [PR](https://github.com/camptocamp/c2cgeoportal/pull/3291)
- [x] in source code
- [x] add an (ngeo) example
- [x] change close behaviour (don't delete route), add clear button

### done:
- [x] Create specs -> see [c2cgeoportal](https://github.com/camptocamp/c2cgeoportal/wiki/Spec-%23%3F%3F%3F-Routing)
- [x] Add modify-interaction for start/end
- [x] Add icons for start/end
- [x] Useful formatting for ~~duration~~ / distance (12km instead of 12345m)
- [x] Replace inputs by search fields (using Nominatim)
- [x] Properly shut down panel on closing (don't keep displaying route)
- [x] Recenter map on route
- [x] Fetch nearest useful location from [nearest-service](https://github.com/Project-OSRM/osrm-backend/blob/master/docs/http.md#nearest-service) and replace user input with it
- [x] Via-function
- [x] Allow multiple profiles
- [x] fix bugs in reverse route feature
- [x] add filter for duration

## Open questions

- ~What should happen, if the routing tab gets closed?~
- ~Should I move code from contrib/gmf to ngeo?~
- ~Should we keep the nominatim-service (forward search)? Or replace it with the integrated full-text-search? Note: the routing cannot handle all types of features as results. It currently requires point geometries. The reverse-lookup will always be done through the connected OSRM-engine.~
- ~Where could I run a demo?~


fredj:
Please use an angular component instead of a directive

contribs/gmf/src/directives/routing.js
```diff
@@ -0,0 +1,213 @@
...
+   * @type {ol.Feature}
+   * @private
+   */
+  this.startFeature_ = null;
+
+  /**
+   * @type {string}
+   * @private
+   */
+  this.startFeatureLabel_ = '';
```
fredj: This variable is used in the template, it should be rename to `this.startFeatureLabel` and marked as `@export` instead of `@private`

pfirpfel:
Screenshots of the current state:

Enable drawing to set start point
1 - enable drawing to set start point

Set start point on map
2 - set start point on map

Enable drawing of end point
3 - enable drawing of end point

Set end point, route gets calculated
4 - set end point - route gets calculated


Difference between user-defined and waypoint used by routing
user point vs waypoint

gjn:
This is a really nice addition to ngeo. We are looking forward to use it in map.geo.admin.ch

Do you plan to also support isochrone functions, to determine reachable places from a given point with distance/time?

See https://disaster.openrouteservice.org/reach?a=11.11128,7.7227&b=0&i=1&j1=30&j2=15&k1=en-US&k2=km for an example to this.

pfirpfel:
Our [supported routing back-end](https://github.com/Project-OSRM/osrm-backend/) doesn't not have an isochrones feature like [openrouteservice](https://developers.openrouteservice.org/portal/features).

This front-end currently only works with an OSRM back-end API, which you can host and setup yourself unlike the openrouteservice.org API (could only find the code for their front-end). But in theory, we could add support for different kinds of routing APIs.

The isochrones feature has quite different UI requirements. In my opinion this should be done as a separate component.

gjn:
Thanks alot for the fast reply!

contribs/gmf/apps/desktop/index.html
```diff
@@ -333,6 +349,17 @@
...
+        module.constant('gmfRoutingProfiles', [
+          {label : 'Car', profile: 'routed-car'},
+          {label : 'Bike (City)', profile: 'routed-bike'},
+          {label : 'Bike (Touring)', profile: 'routed-bike-touring'},
+          {label : 'Foot', profile: 'routed-foot'},
+          {label : 'Hiking', profile: 'routed-hiking'}
+        ]);
+        module.constant('gmfRoutingSearchDefaultParams', {
+          'countrycodes': 'CH'
+        });
```
fredj: Instead of having multiple options why not have a single object `gmfRoutingOptions` with everything inside?

contribs/gmf/src/directives/partials/routing.html
```diff
@@ -0,0 +1,65 @@
...
+      <option ng-repeat="profile in routeCtrl.routingProfiles" ng-value="profile">{{profile.label}}</option>
+    </select>
+  </div>
+
+  <div id="routingError" ng-hide="routeCtrl.errorMessage === ''">
+    {{routeCtrl.errorMessage}}
+  </div>
+
+  <div id="routeDuration" ng-hide="routeCtrl.routeDuration === null">
+    <strong>{{'Duration' | translate}}:</strong> {{routeCtrl.routeDuration}} {{'minutes' | translate}}
```
fredj: instead of having:
```
<strong>{{'Duration' | translate}}:</strong>
```
you can have the simpler:
```
<strong translate>Duration:</strong>
```

contribs/gmf/src/directives/routingfeature.js
```diff
@@ -0,0 +1,312 @@
...
+  const onSuccess = (function(resp) {
+    const lon = parseFloat(resp['data']['lon']);
+    const lat = parseFloat(resp['data']['lat']);
+    const coordinate = [lon, lat];
+    const label = resp['data']['display_name'];
+    this.setFeature_(coordinate, label);
+  }).bind(this);
+
+  const onError = (function(resp) {
+    this.errorMessage = 'Error: nominatim server not responding.';
```
fredj: `this.errorMessage` should be in the constructor

contribs/gmf/src/directives/routingfeature.js
```diff
@@ -0,0 +1,312 @@
...
+ * function of the routing service. Replaces the feature.
+ * @param {ol.Feature} feature Feature to snap
+ * @private
+ */
+gmf.GmfRoutingFeatureController.prototype.snapFeature_ = function(feature) {
+  const coord = this.getLonLatFromPoint_(feature);
+  const config = {};
+
+  const onSuccess = (function(resp) {
+    const lon = parseFloat(resp['data']['lon']);
```
fredj: `this.errorMessage` should be set to `''`

contribs/gmf/src/directives/partials/routing.html
```diff
@@ -0,0 +1,65 @@
...
+  </button>
+  <button type="button" id="gmfRoutingAddViaButton" class="btn btn-default" ng-click="routeCtrl.addVia()">
+    <span class="fa fa-plus"></span> {{'Add via' | translate}}
+  </button><br><br>
+
+  <div id="routingProfile" ng-if="routeCtrl.routingProfiles.length > 1">
+    <label for="gmfRoutingProfileSelect">{{'Profile' | translate}}:</label>
+    <select id="gmfRoutingProfileSelect"
+            name="gmfRoutingProfileSelect"
+            ng-model="routeCtrl.selectedRoutingProfile">
```
fredj: `selectedRoutingProfile` is not defined in the controller

pfirpfel: It should be defined in directives/routing.js

pfirpfel:
Redesign:
selection_034

fredj:
One last issue: when switching from the routing panel to an other one (for example from routing to print) the start and destination points remains on the map

contribs/gmf/src/directives/routingfeature.js
fredj: (fixed)

contribs/gmf/src/directives/partials/routing.html
```diff
@@ -0,0 +1,97 @@
...
+        {{routeCtrl.routeDuration}} {{'minutes' | translate}}
+      </div>
+    </div>
+
+    <div class="row" ng-hide="routeCtrl.routeDistance === ''">
+      <div class="col-md-4 text-right" translate>
+        Distance
+      </div>
+      <div class="col-md-8">
+        {{routeCtrl.routeDistance}}
```
sbrunner: `routeDistance` should be a number in 'm' and the ngeoUnitPrefix filter should be called with: `{{routeCtrl.routeDistance | ngeoUnitPrefix}}`

sbrunner:
I don't know it its a good idea to put the routing in gmf, for me it should be in ngeo, and in the gmf application it should use the fulltextsearch, not nominatim ...
And finally for me it should be added in the desktop_alt, not in the standard desktop :-)

sbrunner:
Is there a demo somewhere?

contribs/gmf/src/directives/partials/routing.html
```diff
@@ -0,0 +1,97 @@
...
+      <div class="col-md-12">
+        <strong translate>Route statistics</strong>
+      </div>
+    </div>
+    <div class="row" ng-hide="routeCtrl.routeDuration === null">
+      <div class="col-md-4 text-right" translate>
+        Duration
+      </div>
+      <div class="col-md-8">
+        {{routeCtrl.routeDuration}} {{'minutes' | translate}}
```
sbrunner: You should create a filter for durations ...

src/services/filters.js
```diff
@@ -317,3 +317,107 @@ ngeo.trustHtmlFilter = function($sce) {
...
+      case TimeUnits.HOURS:
+        formattedUnit = amount > 1 ? 'hours' : 'hour';
+        break;
+      case TimeUnits.DAYS:
+        formattedUnit = amount > 1 ? 'days' : 'day';
+        break;
+      default:
+        break;
+    }
+    return `${amount} ${gettextCatalog_.getString(formattedUnit)}`;
```
sbrunner: You should put the gettextCatalog_.getString directly on each string to make the string collection working ...

and it should probably be named gettextCatalog without ending _

sbrunner:
Is there any reason to format the distance in the controller instance of in the partial?

contribs/gmf/src/directives/partials/routing.html
```diff
@@ -85,12 +85,12 @@
...
     </div>
 
-    <div class="row" ng-hide="routeCtrl.routeDistance === ''">
+    <div class="row" ng-hide="routeCtrl.routeDistance <= 0">
       <div class="col-md-4 text-right" translate>
         Distance
       </div>
       <div class="col-md-8">
-        {{routeCtrl.routeDistance}}
+        {{routeCtrl.routeDistance | ngeoUnitPrefix:'m'}}
```
sbrunner: Thanks :-)

fredj:
See https://github.com/camptocamp/ngeo/compare/gmf-routing

fredj:
see #3719

PR #2681, opened 2017-06-12
Fix the coordinates filters

### LINKED ISSUE #2680: context menu broken
* https://camptocamp.github.io/ngeo/master/examples/contribs/gmf/apps/desktop
 * right click

```
desktop.js:140 Error: [$parse:syntax] http://errors.angularjs.org/1.6.4/$parse/syntax?p0=%7B&p1=is%20an%20unexpec…n&p2=40&p3=coord_4326%7CngeoNumberCoordinates%3A2%3A%7By%7D%20%7Bx&p4=%7Bx
    at desktop.js:23
    at s.throwError (desktop.js:253)
    at s.ast (desktop.js:245)
    at uc.parse (desktop.js:280)
    at desktop.js:149
    at k (desktop.js:127)
    at la (desktop.js:98)
    at jc (desktop.js:82)
    at Ma (desktop.js:79)
    at Ma (desktop.js:79)
```

sbrunner:
fix #2680

### DIFF
--- a/contribs/gmf/apps/desktop/contextualdata.html
+++ b/contribs/gmf/apps/desktop/contextualdata.html
@@ -12,7 +12,7 @@
       Wgs Coord.
     </td>
     <td>
-      {{coord_4326|ngeoNumberCoordinates:2:{y} {x}}
+      {{coord_4326|ngeoNumberCoordinates:2:'{y} {x}'}}
     </td>
   </tr>
   <tr>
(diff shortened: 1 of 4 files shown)


fredj:
The single quotes are not necessary, the only thing that need to be added is the missing `}` in:
https://github.com/camptocamp/ngeo/blob/master/contribs/gmf/apps/desktop/contextualdata.html#L15

sbrunner:
Required because we have come {} in the text :-)

ger-benjamin:
I'm not sur that's beacause we have some bracket in the texte. But this parameter of the filter function is a text (template), we expect some texte, so, the quote are required.

PR #2682, opened 2017-06-12
#2679 - Fix date filters

### LINKED ISSUE #2679: Date filters crash
When setting a date in a filter, the result map is empty (0 byte response). The problem is for all dates, not only for Time layers.

Tested on desktop_alt with "Filtres" layers.

adube:
Fixes #2679 

When using WFS, date filters need to use the proper type of operator.  The standard ones wont work, i.e. those with `property equals to`, `between`, etc. do not work.  We need to use temporal ones instead.

MapServer currently supports `During` only, so that's the one we'll use.

The fix is applied to the filter tool, when we're dealing with a date attribute.  We removed the `>`, `<` and `not` operators as they are not required / make less sense than just using `=`, `>=`, `<=`.

Since During always have a beginning and an end, we need to calculate their value when dealing with an operator other than "between".  For min value, we use `1970-01-01`.  For the max value, we add **30 years**.

## Todo

 * [x] Fix issues reported by Travis
 * [x] Show a live demo
 * [ ] Review

## Live demo

https://adube.github.io/ngeo/2679-fix-date-filters/examples/contribs/gmf/filterselector.html?tree_enable_osm_open=true&tree_enable_ch.swisstopo.geologie-geotechnik-gk500-gesteinsklassierung&tree_enable_OSM%2520map&tree_enable_osm_time_r_dp_2&tree_groups=filters

* Select `osm_open` as layer to filter
 * Add criteria / timestamp
 * Choose November 9th 2010 as value, keep 'equals to' as operator
 * Apply / Apply

You can try other possibilities.

### DIFF
--- a/options/ngeox.js
+++ b/options/ngeox.js
@@ -2120,6 +2120,14 @@ ngeox.rule.Rule = function() {};
  * @struct
  * @extends ngeox.rule.RuleOptions
  */
+ngeox.rule.DateOptions = function() {};
+
+
+/**
+ * @record
+ * @struct
+ * @extends ngeox.rule.RuleOptions
+ */
 ngeox.rule.GeometryOptions = function() {};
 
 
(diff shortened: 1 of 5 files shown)


adube:
Ready for review.

fredj:
LGTM, @ybolognini can you take a look?

ybolognini:
It works! thanks, please merge

PR #2718, opened 2017-07-03
Handle dimensions in ngeoMapQuery and ngeoBboxQuery

arnaud-morvan:
Add dimensions attributes to ngeo.mapQueryDirective and
ngeo.bboxQueryDirective and bind them with mainCtrl.dimensions.

Pass query directives dimensions to MapQuerent and Querent services throught
ngeox.IssueGetFeaturesOptions.

In Querent service:
- combine data sources regarding dimensions
- combine datasources default dimensions with ones from options
- pass dimensions in getFeature and getFeatureInfo requests as query parameters.

### DIFF
--- a/contribs/gmf/apps/desktop/index.html
+++ b/contribs/gmf/apps/desktop/index.html
@@ -237,10 +237,12 @@
           ngeo-map-query-map="::mainCtrl.map"
           ngeo-map-query-active="mainCtrl.queryActive"
           ngeo-map-query-autoclear="mainCtrl.queryAutoClear"
+          ngeo-map-query-dimensions="mainCtrl.dimensions"
           ngeo-bbox-query=""
           ngeo-bbox-query-map="::mainCtrl.map"
           ngeo-bbox-query-active="mainCtrl.queryActive"
-          ngeo-bbox-query-autoclear="mainCtrl.queryAutoClear">
+          ngeo-bbox-query-autoclear="mainCtrl.queryAutoClear"
+          ngeo-bbox-query-dimensions="mainCtrl.dimensions">
         </gmf-map>
 
         <!--infobar-->
--- a/contribs/gmf/apps/mobile/index.html
+++ b/contribs/gmf/apps/mobile/index.html
@@ -24,7 +24,8 @@
       <gmf-map gmf-map-map="mainCtrl.map"
         ngeo-map-query=""
         ngeo-map-query-map="::mainCtrl.map"
-        ngeo-map-query-active="mainCtrl.queryActive"></gmf-map>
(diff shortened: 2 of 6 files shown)


adube:
Originally, the dimensions were introduced in the data sources for them to be independant.  I'm not sure if they should be removed and added to the quey directly instead.

The dimensions may be used for other purpose, such as: filter, layers.  By only giving them to the querent service, we would need to do so for the filters and layers as well and thus duplicate a lot of code.  I think it should have stayed as it was and, instead, make sure that when the global dimensions property changes then update the active dimensions in the data sources.

Does that make sense?

arnaud-morvan:
For me it make sense that each data source keep track of its own active dimensions and manage related filter with underlying data provider.
In my usecase, dimensions are used for floors. When using filter-selector directive to filter visible data, I would like to keep active dimensions applied, but when getting tabular data, I would like to be able to get data from multiple floors.
Maybe we should open a specific issue for discussing this ?

adube:
@arnaud-morvan: @gberaudo and I discussed about this and came up with an alternate solution.  I'll work on a new PR featuring it, plus we're going to have an other one in the near future.  Let me explain.

### First PR - organize the dimensions within the data sources

I'm going to work on a first new PR.  I'm going to pick some of the elements you have implemented in yours (thank you for that, btw).

First, we're going to inject the `gmf.DataSourcesManager` service in the Abstact controller and give it a direct reference to the global `dimensions` property.

In the `gmf.DataSourceManager`, data sources are going to be created with a reference to this `dimensions` property.  The name of the property within the data source will be `dimensions`, thus the current `dimensions` property will have to be renamed to: `dimensionsConfig`, i.e. the dimensions that are used by this data source.

Then, the `activeDimensions` property will be changed into a `getter method` instead and will return the "key:value" in an object for the current "active dimensions" of the data source.  That's going to be used in the querent service.

For the combinaison, the logic of your `dimensionsOverlap_` will be converted into one withing the `ngeo.DataSource` class, where 2 data sources will be compared and checked for the exact same dimensions for combinaison.

The querent will then use the dimensions as params (like you introduced).  That would be the end of this first PR.

### Second PR - alternate dimensions

This would be regarding the questions you raised, i.e.

> When using filter-selector directive to filter visible data, I would like to keep active dimensions applied, but when getting tabular data, I would like to be able to get data from multiple floors.

Yes, we should discuss about this first.  @gberaudo and I talked about it a little, but it would be best to clarify what the use-cases would be to focus on how we're going to implement this.  Let's talk about this when you return.

adube:
@arnaud-morvan The fix introduced #2769 will be used instead.  I'll close this one.

Let's discuss the what should be next when you return.

PR #2719, opened 2017-07-03
Search-parameter

pfirpfel:
## New feature
Introduces two new parameters:

- **search**: (string) recenters map on first result of fulltext search
- **search-select-index**: (number) use nth result from search instead

If a feature is found, it gets added to the search input field. Clicking the X-button there deletes the feature from the feature-overlay.

## Example:
demo-search-param

`[APP-URL]/desktop/?search=Lausanne`
In the gmf demo app, this would center the map on "District de Lausanne" (district layer)

`[APP-URL]/desktop/?search=Lausanne&search-select-index=4`
In the gmf demo app, this would center the map on "Lausanne" (osm layer), the fourth result in the response of the fulltextsearch call.

## Open questions / issues:

- [technical] Is the search directive the right place to put this?
- ~[usability] Different name/behaviour for _search-offset_?~ Changed to **search-select-index**
- [usability / technical] The result order does not always match the drop down menu of the search bar. This is due some re-arranging. Is this a problem?

### DIFF
--- a/contribs/gmf/src/directives/search.js
+++ b/contribs/gmf/src/directives/search.js
@@ -460,11 +460,11 @@ gmf.SearchController.prototype.$onInit = function() {
   if (this.ngeoLocation_) {
     const searchQuery = this.ngeoLocation_.getParam('search');
     if (searchQuery) {
-      let resultOffset = 0;
-      if (this.ngeoLocation_.getParam('search-offset')) {
-        resultOffset = parseInt(this.ngeoLocation_.getParam('search-offset'), 10);
+      let resultIndex = 1;
+      if (this.ngeoLocation_.getParam('search-select-index')) {
+        resultIndex = parseInt(this.ngeoLocation_.getParam('search-select-index'), 10);
       }
-      this.fulltextsearch_(searchQuery, resultOffset);
+      this.fulltextsearch_(searchQuery, resultIndex);
     }
   }
 };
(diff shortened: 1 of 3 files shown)


pfirpfel:
If a feature is found, I add it to the FeatureOverlay. I need to find a useful user interaction that would make sense to get rid of the feature once it is displayed. Any ideas?

pfirpfel:
Demo:
demo-search-param

yjacolin:
ping @brossardm

fredj:
> [usability] Different name/behavior for search-offset?

I found this notion of offset a bit confusing because of the n + 1.
What about a `search-select-index` that tells the index of the element to select ? the default value will be `1` (first result) and the user will set it to `3` to select the third result.

contribs/gmf/src/directives/search.js
```diff
@@ -205,6 +208,17 @@ gmf.SearchController = function($scope, $compile, $timeout, gettextCatalog,
   this.timeout_ = $timeout;
 
   /**
+   * @type {ngeo.Location}
+   * @private
+   */
+  this.ngeoLocation_ = $injector.get('ngeoLocation');
```
fredj: I think we should not put a strong dependency between the search and the ngeoLocation, the search should work even if no ngeoLocation is present.

Please change to something like:
```js

/**
 * @type {ngeo.Location|undefined}
 * @private
 */
this.ngeoLocation_;

if ( $injector.has('ngeoLocation')) {
  this.ngeoLocation_ = $injector.get('ngeoLocation');
}
```

contribs/gmf/src/directives/search.js
```diff
@@ -205,6 +208,17 @@ gmf.SearchController = function($scope, $compile, $timeout, gettextCatalog,
...
+
+  /**
+   * @private
+   */
+  this.fullTextSearch_ = $injector.get('gmfFulltextSearchService');
```
fredj: Now because this is a "strong" dependency (can't work without), please don't use the injection but add in the constructor parameter (just like `gmfTreeManager`)

pfirpfel:
@fredj: I made changes according to your suggestions.

fredj:
@pfirpfel looks very good to me, thanks !

fredj:
@pfirpfel please rebase into master and merge

PR #2728, opened 2017-07-04
Fix default theme resolution

arnaud-morvan:
Default theme from user functionalities was only applied when the user log in.

* Move gmfUser.functionalities.default_theme check in permalink service.
* Get the default theme by priority order from url, local storage, user functionalities or defaultTheme constant.

### DIFF
--- a/contribs/gmf/src/controllers/abstract.js
+++ b/contribs/gmf/src/controllers/abstract.js
@@ -103,16 +103,17 @@ gmf.AbstractController = function(config, $scope, $injector) {
   this.gmfThemes_ = $injector.get('gmfThemes');
 
   /**
+   * Permalink service
+   * @type {gmf.Permalink}
+   * @private
+   */
+  this.permalink_ = $injector.get('gmfPermalink');
+
+  /**
    * Authentication service
    * @type {gmf.Authentication}
    */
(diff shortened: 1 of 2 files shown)


arnaud-morvan:
@ger-benjamin : After some git grep, there is already a default theme loaded by the permalink that comes from the main template, see https://github.com/camptocamp/ngeo/blob/a135fdebaab1860cf53c61782e3aef701913325d/contribs/gmf/apps/desktop/index.html#L328.

It seems that we have 2 differents ways of configuring / loading the default theme.
Through defaultTheme constant and gmfUser.functionalities.

arnaud-morvan:
To load the default theme with correct priorities, which should be : 
* url
* local storage
* user functionalities
* defaultTheme constant

It seems necessary to have all the code in one place. I will propose a PR to add a permalink.getDefaultThemeName() method that evaluate each one in proper order and take care of user.functionalitites.default_theme.

arnaud-morvan:
@ger-benjamin : I moved the gmfUser.functionalities.default_theme check to permalink service. Can you have a look ?

arnaud-morvan:
Bad destination branch, I will open I new one on 2.2.

PR #2735, opened 2017-07-10
Fix angular gettext extern

gberaudo:
From `node_modules/angular-gettext/dist/angular-gettext.js`.

### DIFF
--- a/externs/angular-gettext.js
+++ b/externs/angular-gettext.js
@@ -38,24 +38,26 @@ angularGettext.Catalog.prototype.debug;
  * @param {number} n
  * @param {string} string
  * @param {string} stringPlural
- * @param {Object.<string, string>=} opt_context
+ * @param {string=} opt_context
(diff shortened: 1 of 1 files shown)


externs/angular-gettext.js
```diff
@@ -38,24 +38,26 @@ angularGettext.Catalog.prototype.debug;
...
 angularGettext.Catalog.prototype.getPlural = function(n, string, stringPlural, opt_context) {};
 
 
 /**
- * @param {string} string
+ * @param {string} language
+ * @param {string} key
  * @param {number} n
- * @return {string}
+ * @param {String=} opt_context translation key context
```
apastuhov: string object? it seems that it must be `string`

gberaudo: True, fixing it.

gberaudo:
Thanks for the reviews.

PR #2737, opened 2017-07-11
Remove trailing slash from authenticationBaseUrl

fredj:
fixes https://github.com/camptocamp/c2cgeoportal/issues/2991


gberaudo:
LGTM, though I am wondering why there is a slash at the end of the authentication URL?

PR #2744, opened 2017-07-12
Enforce ngeo.rule.Rule and ngeo.DataSource are not used in ngeox.js

gberaudo:
Depending on actual types in ngeox.js is bad since it forces users of
ngeo to depend on the js files providing these types, even if they are
not used in their project.

To avoid this, use ngeox.rule.Rule instead of ngeo.rule.Rule and
ngeox.DataSource instead of ngeo.DataSource.

See c6d6e8fd9cef1e308073fd3c65441163ddb6e0b4.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -201,6 +201,7 @@ dist: dist/ngeo.js dist/ngeo-debug.js dist/gmf.js
 
 .PHONY: check
 check: git-attributes eof-newline lint check-examples test dist build-gmf-apps
+	grep -nE "ngeo.rule.Rule|ngeo.DataSource" options/ngeox.js && (echo "Only use ngeox.rule.Rule and ngeox.DataSource in options/ngeox.js" && exit 1)
 
 .PHONY: build-gmf-apps
 build-gmf-apps: $(foreach APP,$(GMF_APPS),$(addprefix contribs/gmf/build/$(APP),.js .css)) \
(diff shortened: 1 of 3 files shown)


gberaudo:
Thanks for the review.

Makefile
```diff
@@ -201,6 +201,7 @@ dist: dist/ngeo.js dist/ngeo-debug.js dist/gmf.js
 
 .PHONY: check
 check: git-attributes eof-newline lint check-examples test dist build-gmf-apps
+	grep -nE "ngeo.rule.Rule|ngeo.DataSource" options/ngeox.js && (echo "Only use ngeox.rule.Rule and ngeox.DataSource in options/ngeox.js" && exit 1)
```
sbrunner: Shouldn't be in a separate rule and tested by Travis?

src/services/rulehelper.js
```diff
@@ -379,7 +379,7 @@ ngeo.RuleHelper = class {
    */
   createFilter(options) {
 
-    const dataSource = options.dataSource;
+    const dataSource = /** @type {ngeo.DataSource} */ (options.dataSource);
```
sbrunner: Shouldn't be an asserts?

Makefile
gberaudo: I expected it would be checked by Travis. Will create a PR fixing that.

src/services/rulehelper.js
gberaudo: Unfortunately there is no way to check that: we can not do an assert on a typedef.

sbrunner: ok :-(

PR #2750, opened 2017-07-14
Fix theme name in theme selector for IE11

### LINKED ISSUE #2745: Themeselector no-flush: IE11 doesn't display theme name below theme images
On GMF 2.1 & 2.2, desktop interface. Only on IE11 and with the flush mode disabled for theme selection (I saw no issue with flush mode active on IE11):
image

It should be displayed:
image

It seems that the issue is related to the CSS flex attribute used in the `<span>`.

llienher:
Fix #2745


sbrunner:
Merged without any review !!!

fredj:
@sbrunner it was #2746 but for the 2.1 branch

PR #2758, opened 2017-07-17
Fix legends loading for mobile

### LINKED ISSUE #2755: mobile legends aren't displayed directly
Only on 2.2, doesn't happens on 2.1.

Mobile interface legends are not displayed directly after clicking button "show legend". Instead  we have this small pixel bar (which is the border of the legend window, when empty):
image

The legend only loads directly if you click "open legend", then activate the layer. It doesn't work the other way around. Or if you have the layer active, only of it is hidden then it gonna show up or if the menu is closed then re-opened.

Looking at 2.1 it just opens when asked.

llienher:
Follow up from #1055 (this fixed some performance issues with legends but not the loading)
Fix #2755


sbrunner:
Merged without any review?

ger-benjamin:
We have done this modification together. So, I approve

sbrunner:
:-), good to put a comment event it it's posted with the author user :-)

PR #2763, opened 2017-07-20
Move ngeox check to a dedicated rule

gberaudo:
Follow up of #2744.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -200,8 +200,11 @@ apidoc: .build/apidoc
 dist: dist/ngeo.js dist/ngeo-debug.js dist/gmf.js
 
 .PHONY: check
-check: git-attributes eof-newline lint check-examples test dist build-gmf-apps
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -223,7 +226,7 @@ eof-newline:
 	buildtools/test-eof-newline
 
 .PHONY: test
-test: .build/ol-deps.js .build/ngeo-deps.js .build/gmf-deps.js .build/templatecache.js .build/gmftemplatecache.js .build/node_modules.timestamp .build/examples-hosted/lib/proj4.js
+test: .build/ol-deps.js .build/ngeo-deps.js .build/gmf-deps.js .build/templatecache.js .build/gmftemplatecache.js .build/node_modules.timestamp .build/examples-hosted/lib/proj4.js check-ngeox
```
sbrunner: Why adding that here, for me it should be added in `check`, and this
```
- if [ "${DO_LINT}" = TRUE ]; then make check-ngeox; fi
```
should be added in `.travis.yaml` https://github.com/camptocamp/ngeo/blob/master/.travis.yml#L105

gberaudo: OK, done

sbrunner:
UP...

gberaudo:
... and click DOWN. ^^

PR #2764, opened 2017-07-20
Update map size on window resize - use in mobile app

### LINKED ISSUE #2741: Problem when rotating mobile NGEO with iOS 10
When rotating device twice with mobile NGEO (tested on 2.1 and 2.2), the map is streched. It has been reproduced on several latest iOS devices (only with iOS 10 on Safari AND Chrome) and on Cartoriviera application (https://map.cartoriviera.ch) AND official demo application (https://geomapfish-demo.camptocamp.net/2.2/mobile/). 

This is a relative important issue that has to be fixed ASAP. 

image

adube:
Fixes #2741.

Add options and method to both gmf and ngeo map directives to allow the listening of the browser window resizing and update the size of the map accordingly.

The gmf mobile application is the only one that actually have the option set (with this PR).

The options are:

 * `ngeo-map-manage-resize` - defaults to undefined, if set will register the event listener
 * `ngeo-map-resize-transition` - defaults to undefined, if set it defines the duration in ms to apply the resize (useful if animations occur while resizing)

The same are in gmf.

## Live example

 * https://adube.github.io/ngeo/2741-fix-map-strech-on-window-resize/examples/contribs/gmf/apps/mobile/index.html


adube:
@gberaudo I did not test on iOS yet, where the issue was originally reported.  I'll try to setup a live demo to be able to test from there.  You can still start a review.

src/directives/map.js
```diff
@@ -44,6 +55,49 @@ ngeo.mapDirective = function(ngeoSyncDataSourcesMap) {
...
+      const manageResizeProp = attrs[manageResizeAttr];
+      const manageResize = scope.$eval(manageResizeProp);
+
+      if (manageResize) {
+        const resizeTransitionAttr = 'ngeoMapResizeTransition';
+        const resizeTransitionProp = attrs[resizeTransitionAttr];
+        const resizeTransition = scope.$eval(resizeTransitionProp);
+
+        ol.events.listen(
+          window,
```
gberaudo: You may want to use `$window`.

src/directives/map.js
```diff
@@ -44,6 +55,49 @@ ngeo.mapDirective = function(ngeoSyncDataSourcesMap) {
...
+                  loop = false;
+                }
+              };
+              adjustSize();
+            } else {
+              // A single plain resize
+              map.updateSize();
+            }
+          },
+          this
```
gberaudo: This `this` is unnecessary.

src/directives/map.js
```diff
@@ -75,7 +75,7 @@ ngeo.mapDirective = function(ngeoSyncDataSourcesMap) {
           window,
           'resize',
           () => {
-            if (resizeTransition) {
+            if (resizeTransition && Number.isInteger(resizeTransition)) {
```
gberaudo: In which cases could it not be a number?

contribs/gmf/src/directives/map.js
```diff
@@ -31,45 +37,81 @@ goog.require('ol.Map');
...
+  this.gmfSnapping_ = gmfSnapping;
 };
 
 gmf.module.controller('GmfMapController', gmf.MapController);
+
+
+/**
+ * Called on initialization of the controller.
+ */
+gmf.MapController.prototype.$onInit = function() {
```
gberaudo: The map is a directive, not a component. Are you sure you can use `$onInit`?

src/directives/map.js
adube: The compiler complains because we get `resizeTransition` without knowing its type.  You're right, it can't.  Any suggestion?

src/directives/map.js
adube: okay

contribs/gmf/src/directives/map.js
adube: There are other directives that use controllers with this `$onInit` method.  You're right, it's odd since it seems to be documented in the components only.  But, it's a controller's method.

Should we leave this as it is?

src/directives/map.js
```diff
@@ -44,6 +55,49 @@ ngeo.mapDirective = function(ngeoSyncDataSourcesMap) {
...
+      // animation that may occur on the div of the map to smootly resize the
+      // map while in progress.
+      const manageResizeAttr = 'ngeoMapManageResize';
+      const manageResizeProp = attrs[manageResizeAttr];
+      const manageResize = scope.$eval(manageResizeProp);
+
+      if (manageResize) {
+        const resizeTransitionAttr = 'ngeoMapResizeTransition';
+        const resizeTransitionProp = attrs[resizeTransitionAttr];
+        const resizeTransition = scope.$eval(resizeTransitionProp);
```
adube: Would it be okay if I did this?

`const resizeTransition = /**  {number|undefined} */ (scope.$eval(resizeTransitionProp));`

contribs/gmf/src/directives/map.js
gberaudo: Fine to me to keep the code your way if it works in practice and like you said it is a common pattern in other places.

src/directives/map.js
gberaudo: Yes, looks good.
Have you also tried this?
```js
 /** {number|undefined} */ 
const resizeTransition = scope.$eval(resizeTransitionProp);
```

adube:
Fixes made.  Rebased onto master.

adube:
With the transition, it fixes the issue for me (tested on an old iPhone with iOS 9.3.3, on Safari).

llienher:
I tested on iphone 7 + iOS10.3 and Safari, it was working fine.

PR #2765, opened 2017-07-20
Fix rectangle resize with rotation

### LINKED ISSUE #2742: Drawing: rotation lost when resizing a rectangle
I don't know if it is a current limitation or a new issue. 
When drawing a rectangle, then rotate it (right click action), the resize action will reset the rotation.

alexbrault:
Fixes #2742

When starting to drag a rectangle, the sides vectors are precalculated and used for the duration of the resize operation instead of as the rectangle is resized. If one of the vectors is invalid (i.e. the situation that caused the rectangles to turn into triangles), we can replace it with a vector perpendicular to the other, valid vector. 
In the very unlikely event of both vectors being invalid (which requires all 4 corners of the rectangle to be at the exact same point when the resize starts), the algorithm still maintains a rectangle shape.

### DIFF
--- a/src/ol-ext/interaction/modifyrectangle.js
+++ b/src/ol-ext/interaction/modifyrectangle.js
@@ -304,23 +304,26 @@ ngeo.interaction.ModifyRectangle.prototype.handleDrag_ = function(evt) {
     const siblingYPixel = this.getMap().getPixelFromCoordinate(siblingYCoordinate);
 
     // 4 - The point opposite of the handle we dragged
-    const origin = [
-      siblingXCoordinate[0], siblingYCoordinate[1]
-    ];
-    const originPixel = [
-      siblingXPixel[0], siblingYPixel[1]
-    ];
+    let opposite = siblingY.get('siblingY');
+    goog.asserts.assertInstanceof(opposite, ol.Feature);
+    if (ol.getUid(feature) == ol.getUid(opposite)) {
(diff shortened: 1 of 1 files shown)


gberaudo:
Hi @alexbrault, could you please rebase?

src/ol-ext/interaction/modifyrectangle.js
```diff
@@ -191,11 +191,81 @@ ngeo.interaction.ModifyRectangle.prototype.willModifyFeatures_ = function(evt) {
     this.modified_ = true;
     this.dispatchEvent(new ol.interaction.Modify.Event(
       ol.interaction.ModifyEventType.MODIFYSTART, this.features_, evt));
+    this.params_ = this.initializeParams_();
```
gberaudo: Why don't you put the initialization in the `handleDown_` listener?

alexbrault: The idea was to delay the vector calculation until the resizing actually starts so we don't calculate the vectors for nothing each time the user clicks on a feature.

gberaudo: ok

gberaudo:
@alexbrault, I ran the test and it looked to work well. There are some compiler errors in Travis. Can you come up with a fix? Then I think we can merge.

sbrunner:
Thanks @alexbrault :-)

PR #2766, opened 2017-07-20
Set default decimals and precision values for drawing

### LINKED ISSUE #2756: Drawing: azimut angles are too precise
This is the case on 2.2 demo and master but not in 2.1 demo
drawing_angle

llienher:
Fix #2756 

The current display will be 0 decimals (set in abstract controller) as default value. Then it can be overridden in project html or js files. I added the precision value as well as it wasn't set previously. Preview:
image


sbrunner:
UP ?

PR #2769, opened 2017-07-21
Use 'dimensions' properly in querent service

adube:
This is an alternative to #2718, i.e. a fix for the use of the `dimensions` property within the querent service.

### Technical details

The dimensions are globally defined in the GMF abstract controller.  From there, the `gmf.DataSourcesManager` is injected and given the global dimensions, allowing it to give them to the created data sources.

Within a data source, we have:

 * `globalDimensions`: property, is a reference to the global dimensions that come from 
 * `innerDimensions`: property, the static dimensions that are unique for the data source
 * `activeDimensions`: method, returns the "current" dimensions to use with the data source.  If a inner dimensions is defined with a `null` value, then the global dimensions is used.

Two data sources can be **combined** for a WMS or WFS query if they both have the exact same **active dimensions**.

Then, the ngeo querent service use these active dimensions for both WMS and WFS types of queries.  Note that we only pick the first data source active dimensions, since we always issue queries with combinable data sources and they all share the same active dimensions.

### DIFF
--- a/src/datasource.js
+++ b/src/datasource.js
@@ -34,6 +34,13 @@ ngeo.DataSource = class {
     // === DYNAMIC properties (i.e. that can change / be watched ===
 
     /**
+     * The dimenions configuration for the data source.
+     * @type {?ngeox.Dimensions}
+     * @private
+     */
+    this.dimensionsConfig_ = options.dimensionsConfig || null;
+
+    /**
      * The filter condition to apply to the filter rules (if any).
      * @type {string}
      * @private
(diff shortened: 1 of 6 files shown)


src/datasource.js
```diff
@@ -775,6 +767,28 @@ ngeo.DataSource = class {
   // ===================================
 
   /**
+   * @return {!ngeox.DimensionsActive} Active dimensions
+   * @export
+   */
+  get activeDimensions() {
+    const active = {};
+    const global = this.globalDimensions_ || {};
+    const inner = this.innerDimensions || {};
```
gberaudo: Here you mix two things:
- the `globalDimensions` objet wich are the current values for the dimensions, from the application;
- the `innerDimensions` object which is the configuration of which dimensions to use.

It is confusing. I think `innerDimensions` should be renamed to `innerDimensionsConfig` or simply `dimensionsConfig`.
Idem for the type: `ngeox.Dimensions` could be `ngeox.DimensionsConfig`.

adube: I agree.  I'll work on this very soon.

adube:
Fixed.  I'll merge once Travis is happy.

I prefer keeping the name `ngeox.Dimensions`, because the same definition is used by the dimensions and dimensions config.

PR #2783, opened 2017-07-31
Provide vertical swipe

apastuhov:
https://jira.camptocamp.com/browse/GSGMF-12


contribs/gmf/src/directives/swipe.js
```diff
@@ -0,0 +1,264 @@
...
+              swipeHandler(scope, {$event: event});
+            });
+          }
+        }
+      }, pointerTypes);
+    };
+  }]);
+}
+
+// Down is negative X-coordinate, up is positive.
```
romanzoller: I guess this should be Y-coordinate?

PR #2795, opened 2017-08-04
resizeMap when switching from opened panels to street-view panel and vice-versa

### LINKED ISSUE #2786: StreetView: squeezed map
Map is squeezed (see screenshot) in the following case: 
 * Edit tool
 * Click on StreetView tool

squeezed

llienher:
Fixes #2786 

This fix add a CSS transition on width changes used by street-view panel. The transition is then listened by the resizeMap directive that will run the resize actions at the 'transitionend' event: https://github.com/camptocamp/ngeo/blob/f6c897e3ada535f9f483ba18f290b056d9387703/src/directives/resizemap.js#L62-L65

The current duration used is 0.01s because we don't want a animation to start and be visible, we just want to run the transition to make the resizeMap directive do his job on width changes.

### DIFF
--- a/contribs/gmf/less/desktop.less
+++ b/contribs/gmf/less/desktop.less
@@ -264,7 +264,7 @@ gmf-backgroundlayerselector {
   .gmf-app-tools-content {
     width: @right-panel-width;
     margin-right: -@right-panel-width;
-    transition: margin-right 0.2s ease;
+    transition: margin-right 0.2s ease, width 0.001s ease;
     float: right;
     height: 100%;
     overflow: auto;


llienher:
Browser tests:
- Chrome: OK
- Firefox: OK
- Edge: not tested
- IE11: not tested (ES6 is not supported with the source code)

Another ressource: 
http://caniuse.com/css-transitions/embed/
IE 10 and above, Edge, are supported.

fredj:
@sbrunner looks good to you ?

sbrunner:
+1

PR #2797, opened 2017-08-04
Fix move/resize issues and improve styling

romanzoller:
- Fix issue with "share map" modal that was not resizeable
- Make query results window draggable and resizable
- Adapt styling of modals in gmf

See https://jira.camptocamp.com/browse/GSGMF-8


romanzoller:
Thanks for the reviews!

PR #2806, opened 2017-08-08
Add missing 'strokeLinecap' in mapfish-print externs

fredj:


sbrunner:
Shouldn't be in the 2.2 branch?

PR #2808, opened 2017-08-09
Support user functionalities to toggle the filter tool on upon init

adube:
This PR introduces the support of the two following GMF user "functionalities":

`"open_layer_filter_panel": ["true"]` - when defined, the filter panel is open upon loading a desktop application.

`"preset_layer_filter": ["inventaire_equipements"]` - when defined, the name of the layer (data source) automatically: 1) open the filter panel, 2) select the data source in the filter selector tool.

### DIFF
--- a/contribs/gmf/externs/gmf-themes.js
+++ b/contribs/gmf/externs/gmf-themes.js
@@ -375,6 +375,28 @@ gmfThemes.GmfFunctionalities = function() {};
  */
 gmfThemes.GmfFunctionalities.prototype.default_basemap;
 
+
+/**
+ * When set, contains the name of the panel to open upon loading an application.
+ *
+ * Note: altough this is a list, only one can be defined.
+ *
+ * @type {Array.<!string>|undefined}
+ */
+gmfThemes.GmfFunctionalities.prototype.open_panel;
+
+
+/**
+ * Name of the layer (data source) that should be toggled in the filter tool
+ * upon loading an application.
+ *
+ * Note: altough this is a list, only one can be defined.
+ *
+ * @type {Array.<!string>|undefined}
+ */
+gmfThemes.GmfFunctionalities.prototype.preset_layer_filter;
+
+
 /**
  * @constructor
  * @struct
--- a/contribs/gmf/src/controllers/abstract.js
+++ b/contribs/gmf/src/controllers/abstract.js
@@ -133,6 +133,18 @@ gmf.AbstractController = function(config, $scope, $injector) {
 
   const userChange = function(evt) {
     const roleId = (evt.user.username !== null) ? evt.user.role_id : undefined;
+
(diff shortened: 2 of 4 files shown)


contribs/gmf/externs/gmf-themes.js
```diff
@@ -375,6 +375,29 @@ gmfThemes.GmfFunctionalities = function() {};
...
+
+/**
+ * Whether to open the filter tool panel upon loading an application.
+ *
+ * Note: altough this is a list, it should have been boolean. The value
+ * required to have this set is actually: `['true']`.
+ *
+ * @type {Array.<!string>|undefined}
+ */
+gmfThemes.GmfFunctionalities.prototype.open_layer_filter_panel;
```
gberaudo: The functionality should be `open_panel` and contain `['layer_filter']` to open that panel.

contribs/gmf/src/controllers/abstract.js
```diff
@@ -604,10 +621,12 @@ gmf.AbstractController.prototype.updateCurrentTheme_ = function() {
...
-    if (functionalities && 'default_theme' in functionalities) {
-      const defaultTheme = functionalities.default_theme;
-      if (defaultTheme.length > 0) {
-        themeName = defaultTheme[0];
+    if (functionalities) {
+      if ('default_theme' in functionalities) {
+        const defaultTheme = functionalities.default_theme;
+        if (defaultTheme.length > 0) {
+          themeName = defaultTheme[0];
+        }
```
gberaudo: Why change this part?

contribs/gmf/src/directives/filterselector.js
```diff
@@ -266,6 +274,14 @@ gmf.FilterselectorController = class {
...
       this.filtrableLayerNodeNames_ = null;
     }
+    if (usrFunc &&
+        usrFunc['preset_layer_filter'] &&
+        usrFunc['preset_layer_filter'][0]
+    ) {
+      this.defaultFiltrableDataSourceName_ = usrFunc['preset_layer_filter'][0];
+    } else {
+      this.defaultFiltrableDataSourceName_ = undefined;
+    }
```
gberaudo: Could be simplified to `this.defaultFiltrableDataSourceName_ = usrFunc && usrFunc['preset_layer_filter'] && usrFunc['preset_layer_filter'][0]`.

adube: That would set a `false` value.  I want an `undefined` value.  Anyway, having a simplified version of this is not that useful since it already gets minimized and simplified automatically when building the minimized version.  I'll leave this as it is.

adube:
Fixed.

contribs/gmf/src/directives/filterselector.js
gberaudo: You can try it, that would set an `undefined` value. But fine to me to keep as you did.

PR #2810, opened 2017-08-10
Fix theme selector title when not in flush mode.

arnaud-morvan:
When gmfThemeManager is not in flushMode, themeName remains equal to null.

So when not in flush mode replace "Theme:Loading..." by "Themes".


arnaud-morvan:
Bad destination branch, I need it in 2.2.

sbrunner:
Now it's possible to change the destination branch :-)

PR #2811, opened 2017-08-10
Fix theme selector title when not in flush mode.

arnaud-morvan:
When gmfThemeManager is not in flushMode, themeName remains equal to null.

So when not in flush mode replace "Theme:Loading..." by "Themes".


sbrunner:
The push Travis check is due to a Travis issue, and the other build error is because https://docs.angularjs.org/ than don't wary about it :-)

PR #2812, opened 2017-08-10
Fix default theme resolution.

arnaud-morvan:
* Move gmfUser.functionalities.default_theme check in permalink service.

* Get the default theme by priority order from url, local storage, user
functionalities or defaultTheme constant.

### DIFF
--- a/contribs/gmf/src/services/permalink.js
+++ b/contribs/gmf/src/services/permalink.js
@@ -234,6 +234,12 @@ gmf.Permalink = function($timeout, $rootScope, $injector, ngeoDebounce,
     $injector.get('ngeoWfsPermalink') : null;
 
   /**
+   * @type {gmfx.User}
+   * @export
+   */
+  this.gmfUser_ = $injector.get('gmfUser');
+
+  /**
    * @type {?ol.Map}
    * @private
    */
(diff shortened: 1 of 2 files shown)


ger-benjamin:
> Get the default theme by priority order from url, local storage, user
functionalities or defaultTheme constant.

The local storage is always set, no ? (except the very first time you arrive on GMF).
So if you login (and logout ?), the priority change. You want the theme change for your user default theme.

I need a link to to test this change before you merge.

contribs/gmf/src/services/permalink.js
```diff
@@ -803,31 +809,53 @@ gmf.Permalink.prototype.setThemeInUrl_ = function() {
...
-    }
-
-    if (!themeName) {
-      // check if we have a theme in the local storage
-      const tn = this.ngeoStateManager_.getInitialStringValue('theme');
-      if (tn) {
-        themeName = tn;
-      }
-    }
+    let themeName = this.defaultThemeName();
```
ger-benjamin: +100 :-)

arnaud-morvan:
Note : The localStorage seems to be set only in flush mode (only one theme at a time).

IMHO, when I log in, if I have a theme in permalink, I do not want to skip it, like it is for now.
Imagine I click on a link with a theme in url, I need to log in to see the theme, I log in, I do not want to my default theme, but the url one, for sure.

Concerning the localStorage, we may add a flag indicating that we are logging in.
If we log in we get the functionalities first, else, we get the one from local storage first.
The one from localStorage is valid until the end of the session.

Does everybody agree with this ? @ger-benjamin  @ybolognini

arnaud-morvan:
@ger-benjamin @ybolognini : What about this ? 
I can give an url to aeroportdelyon, but you will need to install extended support release of firefox, install and configure iceteaweb to connect a tunnel through a java applet ...

ger-benjamin:
I've put your my instance on the test server (and start to debug the branche 2.2 of the demo project. I'ts not fully completed know, but it's enough for this example).

So you can see the result of your change here: https://testgmf.sig.cloud.camptocamp.net/bge (very temporarily demo)

If you login, you'll never get the "user theme" you've defined in the admin interface. Because the local storage is always set and the defaultThemeName method in the permalink return too soon: https://github.com/arnaud-morvan/ngeo/commit/1c17ea1c5206cf21f5eebddfc489c7982251e274#diff-8e58e57e4aec6dad1c9d6bad99cf30bfR822

sbrunner:
Looks good :-)

arnaud-morvan:
I've updated my PR, now I do not change behavior when the user log in from ngeo interface. I've only added the default theme from functionalities when he open the main page as already authenticated user.

PR #2816, opened 2017-08-14
Use OpenLayers v4.3.1

fredj:


sbrunner:
Should we also use this version in the 2.2 branch?

fredj:
@sbrunner so in 2.2 instead of master ?

fredj:
@sbrunner ok, changed to the 2.2 branch

PR #2819, opened 2017-08-16
Fix activeDimensions precedence order

arnaud-morvan:
Regarding layertree, global dimensions take precedence over node ones.
https://github.com/camptocamp/ngeo/blob/5aede6eda19fe7bbcd07b76be77c93cb42e1caba/contribs/gmf/src/directives/layertree.js#L318:L321

With this, the selected features are the visible ones.

### DIFF
--- a/src/datasource.js
+++ b/src/datasource.js
@@ -776,10 +776,8 @@ ngeo.DataSource = class {
     const config = this.dimensionsConfig || {};
 
     for (const key in config) {
-      if (config[key] === null) {
-        if (dimensions[key] !== undefined && dimensions[key] !== null) {
-          active[key] = dimensions[key];
-        }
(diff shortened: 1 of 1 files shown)


sbrunner:
It seem that the code you reference who is buggy...

arnaud-morvan:
Ok, but I do not found how to set NULL value in dimension value in admin interface.
With empty string, I get errors.
```
The themes contain some errors:
The layer 'lampadaires' has an unsupported dimension value '' ('FLOOR').
The layer 'baies' has an unsupported dimension value '' ('FLOOR').
The layer 'murs_images' has an unsupported dimension value '' ('FLOOR')
```

sbrunner:
:-/ should be fixed with: https://github.com/camptocamp/c2cgeoportal/pull/3042

PR #2821, opened 2017-08-16
Fix layertree dimensions precedence

arnaud-morvan:
Apply global dimensions only if layer config value is null.
To be consitent with ngeo.datasource.

Replace #2819


sbrunner:
The build error is because https://docs.angularjs.org/ than don't wary about it :-)

PR #2832, opened 2017-08-21
Adds some missing requires for annotations 

sbrunner:
Required by c2cgeoportal build on Docker

### DIFF
--- a/contribs/gmf/src/services/permalink.js
+++ b/contribs/gmf/src/services/permalink.js
@@ -3,6 +3,8 @@ goog.provide('gmf.Permalink');
 goog.require('gmf');
 goog.require('ngeo');
 goog.require('ngeo.AutoProjection');
+/** @suppress {extraRequire} */
+goog.require('gmf.ObjectEditingManager');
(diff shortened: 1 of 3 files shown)


sbrunner:
Yes ...

contribs/gmf/src/services/permalink.js
```diff
@@ -3,6 +3,8 @@ goog.provide('gmf.Permalink');
 goog.require('gmf');
 goog.require('ngeo');
 goog.require('ngeo.AutoProjection');
+/** @suppress {extraRequire} */
+goog.require('gmf.ObjectEditingManager');
```
gberaudo: Ideally you should not require `gmf.ObjectEditingManager` here since it is an optionnal dependency.

sbrunner: :-) then where, If I add it in the gmf.js, it's a requirement for all the module...

sbrunner: For me it's an issue of closure-utils ...

PR #2834, opened 2017-08-24
Always give the template cache to Google closure compiler

sbrunner:
This is needed for the build into Docker


sbrunner:
Thanks :-)

PR #2835, opened 2017-08-24
Allow overwriting of map_tooltip and map_crosshair

pfirpfel:


contribs/gmf/src/services/permalink.js
```diff
@@ -511,6 +522,35 @@ gmf.Permalink.prototype.getMapCrosshair = function() {
 };
 
 
+/**
+ * Sets the map crosshair to the center (or the map center if nothing provided).
+ * Overwrites an existing map crosshair.
+ * @param {?ol.Coordinate|undefined} center Optional center coordinate.
```
fredj: Instead of having `{?ol.Coordinate|undefined}` use `{?ol.Coordinate=}` an rename the parameter to `opt_center`.
This is the standard way for optional parameter with the closure compiler.

contribs/gmf/src/services/permalink.js
```diff
@@ -523,6 +563,38 @@ gmf.Permalink.prototype.getMapTooltip = function() {
   return this.ngeoStateManager_.getInitialStringValue(gmf.PermalinkParam.MAP_TOOLTIP);
 };
 
+/**
+ * Sets the map tooltip to the center (or the map center if nothing provided).
+ * Overwrites an existing map tooltip.
+ * @param {string} tooltipText Text to display in tooltip.
+ * @param {?ol.Coordinate|undefined} center Optional center coordinate.
```
fredj: Same remark as above

contribs/gmf/src/services/permalink.js
```diff
@@ -511,6 +522,35 @@ gmf.Permalink.prototype.getMapCrosshair = function() {
 };
 
 
+/**
+ * Sets the map crosshair to the center (or the map center if nothing provided).
+ * Overwrites an existing map crosshair.
+ * @param {?ol.Coordinate|undefined} center Optional center coordinate.
+ * @export
```
fredj: If this function is not used in a html template, the `@export` is not needed

fredj:
please merge if you think it's ready

PR #2838, opened 2017-08-25
Allow passing a custom limit to the bbox query directive

gberaudo:


gberaudo:
Thanks for the review @adube.

PR #2851, opened 2017-08-30
Test if OperationMetadata tag exists

### LINKED ISSUE #2850: Impossible to import WMTS
Impossible to import this WMTS[1] because it doesn't contain the optional tag OperationMetadata [2].

If the OperationMetadata  tag doesn't exist the requestEncoding should be REST by default.

[1] http://www.ogc.stadt-zuerich.ch/mapproxy/wmts/1.0.0/WMTSCapabilities.xml
[2] https://github.com/camptocamp/ngeo/blob/master/src/modules/import/wmtsgetcap.js#L52

oterral:
Fix #2850

### DIFF
--- a/src/modules/import/wmtsgetcap.js
+++ b/src/modules/import/wmtsgetcap.js
@@ -56,7 +56,7 @@ exports = function(gettext, gettextCatalog, ngeoWmtsGetCapTemplateUrl) {
         }
         const layerOptions = {
           'layer': layer['Identifier'],
-          'requestEnconding': requestEncoding
+          'requestEncoding': requestEncoding
         };
         layer['sourceConfig'] = ol.source.WMTS.optionsFromCapabilities(getCap, layerOptions);
         if ('ServiceProvider' in getCap) {
--- a/src/modules/import/wmtsgetcap.js
+++ b/src/modules/import/wmtsgetcap.js
@@ -49,8 +49,11 @@ exports = function(gettext, gettextCatalog, ngeoWmtsGetCapTemplateUrl) {
       }
 
       if (!layer['isInvalid']) {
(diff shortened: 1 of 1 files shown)


src/modules/import/wmtsgetcap.js
```diff
@@ -49,8 +49,11 @@ exports = function(gettext, gettextCatalog, ngeoWmtsGetCapTemplateUrl) {
...
-        const getTileMetadata = getCap['OperationsMetadata']['GetTile']['DCP']['HTTP']['Get'][0];
-        const requestEncoding = getTileMetadata['Constraint'][0]['AllowedValues']['Value'][0];
+        let requestEncoding;
+        if ('OperationsMetadata' in getCap && 'GetTile' in getCap['OperationsMetadata']) {
+          const getTileMetadata = getCap['OperationsMetadata']['GetTile']['DCP']['HTTP']['Get'][0];
+          requestEncoding = getTileMetadata['Constraint'][0]['AllowedValues']['Value'][0];
+        }
         const layerOptions = {
           'layer': layer['Identifier'],
           'requestEnconding': requestEncoding
```
fredj: oups, the key name is `requestEnconding` but it should be `requestEncoding`

oterral: Nice catch.

oterral:
Maybe the typo fix will break some WMTS. Let me try before merging this.

oterral:
In fact this these 2 lines are useless because we do exactly the same in `ol.source.WMTS.optionsFromCapabilities`

sbrunner:
The Travis job 5 failed because https://docs.angularjs.org/ is down ...

PR #2853, opened 2017-08-30
Support user functionalities to toggle the filter tool on upon init

gberaudo:


gberaudo:
@sbrunner, this is a backport from the ngeo master branch. See #2808

PR #2864, opened 2017-09-01
To avoid an issue that GeoTools do an uppercase

sbrunner:


src/services/print.js
```diff
@@ -280,12 +280,18 @@ ngeo.Print.prototype.encodeImageWmsLayer_ = function(arr, layer) {
  * @private
  */
 ngeo.Print.prototype.encodeWmsLayer_ = function(arr, opacity, url, params) {
+  const url_url = new URL(url);
   const customParams = {'TRANSPARENT': true};
-  // remove empty params
+  if (url_url.searchParams) {
+    for (const element of url_url.searchParams) {
+      customParams[element[0]] = element[1];
```
ger-benjamin: I don't understand what you are doing here

sbrunner: I convert the base URL query string in customParams (The first one is uppercased but not the second one ...)

Issue #2848, opened 2017-08-29
Type is not describe in the doc for directives/components


ger-benjamin:
Type is not described in the documentation for directives/components (but the type is described in the doc)
Example: https://camptocamp.github.io/ngeo/master/apidoc/gmf.elevationDirective.html

ger-benjamin:
Closed with #2872

PR #2872, opened 2017-09-04
Type for directives and components

ger-benjamin:
Fix: #2848

The documentation is already better with that (type is now defined in directive and components).
But I still don't have links on documentation on the types.

To do that, I need to copy all the related content of "helper.registerLink" (like https://github.com/camptocamp/ngeo/blob/master/jsdoc/template/publish.js#L450-L461) in the "plugin" file.
It isn't a better way to do that ? Can't I have a "common" file ? Do you have any clue @sbrunner @gberaudo  ?

### DIFF
--- a/jsdoc/template/tmpl/params.tmpl
+++ b/jsdoc/template/tmpl/params.tmpl
@@ -84,9 +84,10 @@
             <?js } ?>
 
             <td class="type">
-
             <?js if (param.typeDefinitionUrl) {?>
                 <?js= param.typeDefinitionUrl ?>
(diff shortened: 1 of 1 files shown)


gberaudo:
@ger-benjamin can you add a screen capture of the result of your change?

ger-benjamin:
selection_029

Here (the type column)

ger-benjamin:
Done, thanks Stéphane (second commit)

ger-benjamin:
Errored only because of my fork. I merge

Issue #2844, opened 2017-08-29
Non working gmf example


fredj:
https://camptocamp.github.io/ngeo/master/examples/contribs/gmf/elevation.html
https://camptocamp.github.io/ngeo/master/examples/contribs/gmf/drawfeature.html

Error is:
```
Error: [$injector:unpr] http://errors.angularjs.org/1.6.6/$injector/unpr?p0=gmfLayersUrlProvider%20%3C-%20gmfLayersUrl%20%3C-%20gmfEditFeature%20%3C-%20gmfObjectEditingManager%20%3C-%20gmfPermalink
```

fredj:
The `gmfLayersUrl` constant is missing

ger-benjamin:
Please always add a label 'ready' and a milestone for our dashboard

pfirpfel:
Still both broken for me with current master, but they work with 2.2.

Error message:
```
angular.js:14700 Error: [$injector:unpr] Unknown provider: gmfLayersUrlProvider <- gmfLayersUrl <- gmfEditFeature <- gmfObjectEditingManager <- gmfPermalink
http://errors.angularjs.org/1.6.6/$injector/unpr?p0=gmfLayersUrlProvider%20%3C-%20gmfLayersUrl%20%3C-%20gmfEditFeature%20%3C-%20gmfObjectEditingManager%20%3C-%20gmfPermalink
    at angular.js:116
    at angular.js:4826
    at Object.getService [as get] (angular.js:4981)
    at angular.js:4831
    at getService (angular.js:4981)
    at injectionArgs (angular.js:5006)
    at Object.instantiate (angular.js:5052)
    at Object.<anonymous> (angular.js:4889)
    at Object.invoke (angular.js:5040)
    at Object.enforcedReturnValue [as $get] (angular.js:4873)
```

pfirpfel:
Fixed with #3157

sbrunner:
Create :-)

PR #2846, opened 2017-08-29
Don't include gmf.ObjectEditingManager in permalink

fredj:
fixes #2844 

was added in #2832

PR #2873, opened 2017-09-04
Retrieve gmfLayersUrl via injector only if it is defined

pfirpfel:
I think this fixes: #2844 

The examples work again after this change.

### DIFF
--- a/contribs/gmf/src/services/editfeature.js
+++ b/contribs/gmf/src/services/editfeature.js
@@ -12,11 +12,11 @@ goog.require('ol.format.GeoJSON');
  *
  * @constructor
  * @struct
+ * @param {angular.$injector} $injector Main injector.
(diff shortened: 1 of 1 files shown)


sbrunner:
Looks strange for me, for me the optional dependency isn't as the right place ...

pfirpfel:
It is an alternative to remove the ObjectEditingManager from permalink.js (#2846).
As long as gmfLayersUrl is in the parameters of the constructor, angular will try to resolve it and break all the apps and examples were it isn't defined.

sbrunner:
Then the optional injection should be in the permalink ...

sbrunner:
Or we should fix the closure_util just before abandon it :-/

pfirpfel:
Apparently fixed by #3025

PR #2874, opened 2017-09-05
Export modeFlush for layertree title

ger-benjamin:
Commit 1: needed by the layertree title (otherwise, it doesn't exist in the HTML so the value is always false)

Questions:
 - It is useful in the mobile app ? https://github.com/camptocamp/ngeo/blob/2.2/contribs/gmf/apps/mobile/index.html#L177

- This line was switched, is there a reason for that ? (that make one change more in the ngeo.diff during an upgrade: https://github.com/camptocamp/ngeo/blob/2.2/contribs/gmf/apps/desktop/index.html#L303 (was at line 299)


sbrunner:
1) The build system don't allows to have deferent list, then it's better that the debug list reflect that.

2) Just to be coherent with the others files :-)

PR #2876, opened 2017-09-05
Switch to protocol-agnostic GitHub shortcut url [2.2]

pfirpfel:
Using the git-protocol explicitly there has caused issues on multiple instances. This configuration change allows an automatic fall-back to https.

Same issue as #2875


fredj:
@pfirpfel codacy check is stuck, please merge anyway

PR #2881, opened 2017-09-06
What's authCtrl.translate?

### LINKED ISSUE #2880: "Forgotten password" window is empty
"Forgotten password" feature is broken. A tiny window is displayed and it is empty.

May be linked to https://github.com/camptocamp/ngeo/issues/2879

sbrunner:
fix #2880

### DIFF
--- a/contribs/gmf/src/directives/partials/authentication.html
+++ b/contribs/gmf/src/directives/partials/authentication.html
@@ -152,9 +152,8 @@
         {{'Password forgotten?' | translate}}
       </h4>
     </div>
-    <div class="modal-body">
-      {{ ::authCtrl.translate(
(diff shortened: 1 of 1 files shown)


contribs/gmf/src/directives/partials/authentication.html
```diff
@@ -153,8 +153,7 @@ <h4 class="modal-title">
       </h4>
     </div>
     <div class="modal-body">
-      {{ ::authCtrl.translate(
-      'A new password has just been sent to you by e-mail.') }}
+      {{'A new password has just been sent to you by e-mail.' | translate}}
```
gberaudo: What about putting the translate as an attribute of the div?

PR #2890, opened 2017-09-11
measure tools tooltip is moving when panning the map

### LINKED ISSUE #2885: Mobile: measure tools tooltip is moving when panning the map
GMF 2.2, mobile interface, tested and reproduced on both mobile and tablet devices, and browserstack.

Activate the measure tool (either coordinate or length), then pan the map in any direction. The quicker the pan is, the more descentered the tooltop is away from the initial point.

Seems to be related to dynamic calculation of the tooltip position while we ideally just want the content to be dynamically generated.

What we want is a fixed position for the tooltip (similar to the cross-hair at the center).

fredj:
fixes #2885

### DIFF
--- a/src/ol-ext/interaction/measure.js
+++ b/src/ol-ext/interaction/measure.js
@@ -398,16 +398,14 @@ ngeo.interaction.Measure.prototype.onDrawStart_ = function(evt) {
   const geometry = this.sketchFeature.getGeometry();
 
   goog.asserts.assert(geometry !== undefined);
-  this.changeEventKey_ = ol.events.listen(geometry,
-    ol.events.EventType.CHANGE,
-    function() {
-      this.handleMeasure((measure, coord) => {
-        if (coord !== null) {
-          this.measureTooltipElement_.innerHTML = measure;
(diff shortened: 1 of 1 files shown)


src/ol-ext/interaction/measure.js
```diff
@@ -420,9 +439,10 @@ ngeo.interaction.Measure.prototype.onDrawEnd_ = function(evt) {
   this.dispatchEvent(new ngeo.MeasureEvent(ngeo.MeasureEventType.MEASUREEND,
     this.sketchFeature));
   this.sketchFeature = null;
-  if (this.changeEventKey_ !== null) {
-    ol.events.unlistenByKey(this.changeEventKey_);
-  }
+  goog.asserts.assert(this.changeEventKey_ !== null);
+  goog.asserts.assert(this.postcomposeEventKey_ !== null);
+  ol.events.unlistenByKey(this.changeEventKey_);
+  ol.events.unlistenByKey(this.postcomposeEventKey_);
```
gberaudo: Shouldn' you nullify the event keys?

fredj: yes, fixed.

PR #2892, opened 2017-09-11
searchtext icon overlaps placeholder on ipad

### LINKED ISSUE #2884: Mobile: searchtext icon overlaps placeholder on ipad
On gmf 2.2, only noticable on the ipad tablet (tested safari and chrome).
Appears when on portrait orientation, apparently just a css issue.

fredj:
fixes #2884


fredj:
Thanks for the review

PR #2894, opened 2017-09-12
Enable eslint-plugin-googshift in some places

gberaudo:
In addition:
- src/modules is removed, see the .eslintrc.yaml in search and utils directories;
- ngeo.module is introduced (currently as an alias to the ngeo.module property of the namespace);
- the base template for modules is kept, it can be transformed to handle all templates in the code.

To check more directories, copy the .eslintrc.yaml file there.


src/ngeo.js
```diff
@@ -3,15 +3,17 @@
  * - a "virtual" angular module root used to automatically register finely included ngeo dependencies;
  * - a JS namespace for constants and types;
  * - a list of requires (for olx, ol3) to please GCC (using hide_warnings_for GCC parameter might help here);
- * - a GCC entry point with requires on all parts of ngeo to produce the dist/ngeo.js file (badly broken).
+ * - a GCC entry point with requires on all parts of ngeo to produce the dist/exports.js file (badly broken).
```
fredj: should not be changed

src/ngeo.js
```diff
@@ -44,9 +46,9 @@ goog.require('ol.style.AtlasManager');
 
 
 /** @type {!angular.Module} */
-ngeo.module = angular.module('ngeo', [
+exports.module = angular.module('ngeo', [
   'gettext', 'ui.date', 'floatThead'
-  // src/modules/* were added for producing the dist/ngeo.js file, which is badly broken.
+  // src/modules/* were added for producing the dist/exports.js file, which is badly broken.
```
fredj: should not be changed

Makefile
```diff
@@ -21,7 +21,7 @@ GMF_EXAMPLES_HTML_FILES := $(shell find contribs/gmf/examples -maxdepth 1 -type
 GMF_EXAMPLES_JS_FILES := $(GMF_EXAMPLES_HTML_FILES:.html=.js)
 GMF_APPS += mobile desktop desktop_alt mobile_alt oeedit oeview
 GMF_APPS_JS_FILES := $(shell find contribs/gmf/apps/ -type f -name '*.js')
-GMF_APPS_LESS_FILES := $(shell find contribs/gmf/less src/modules -type f -name '*.less')
+GMF_APPS_LESS_FILES := $(shell find contribs/gmf/less -type f -name '*.less')
```
sbrunner: Your sure about this?

PR #2923, opened 2017-09-21
Support changing opacity and showing legend in mobile app

adube:
This PR introduces a "mobile" version of the tooltip that contains the opacity slider and "show/hide legend" that are in the desktop apps.

Instead of a tooltip, a plain div is shown underneath the node in the tree.  A "cog" button has been added and is used to toggle on/off this menu.

## Todo

 * [x] review
 * [x] use the proper icon for show/hide legend (see image in comment below)
 * [x] slider style
 * [x] move methods to gmf


adube:
preview

adube:
@ybolognini Would you please tell me the name of the font-awesome icon to use instead?

ybolognini:
@adube 
The open/close legend icon is fa-th-list

Please note that in the mockup, the two separator lines split completely the layer tree (whole column). The idea here was that the menu splits the layer tree and then comes back when the user closes the menu. We could even add an animation where the layer tree is being split (if easy to do)

adube:
>The open/close legend icon is fa-th-list
Thanks, I'll make the fix.

>Please note that in the mockup, the two separator lines split completely the layer tree (whole column). The idea here was that the menu splits the layer tree and then comes back when the user closes the menu. We could even add an animation where the layer tree is being split (if easy to do)

I'm not sure that's currently possible to do that.  Please, anyone correct me if I'm wrong: the current UI is a tree, i.e. elements within elements with indentation.  The "menu" (i.e. what's shown when the "cog" button is clicked" is part of the tree and therefore follows the identation.

If we want the menu to be 100% width of the tree, then we would need to remove all the indentation of the tree.  Please, confirm if that's what you want.

src/directives/layertree.js
```diff
@@ -439,6 +439,60 @@ ngeo.LayertreeController.prototype.setDataSource = function(dataSource) {
...
+/**
+ * @return {boolean} Whether the layer tree controller supports having its
+ *     layer opacity being changed or not.
+ * @export
+ */
+ngeo.LayertreeController.prototype.supportsOpacityChange = function() {
+  return this.layer &&
+    (
+      (
+        this.depth === 1 && !this.node['mixed']
```
fredj: Instead of using the quoted notation, can you try to change `this.node` type from `Object` to `gmfThemes.GmfGroup` ?

sbrunner: +1

sbrunner: with an assert

adube: We can't.  This is a file from "ngeo".  We can't put code from "gmf" there.

adube: @fredj @sbrunner let me know if this is okay.

fredj: @sbrunner can we do an exception for the externs ?

fredj: instead of changing the `@type` here, these two function must be moved from `ngeo` to `gmf`

adube:
Here's what's remaining to do:

 * apply the same style to the slider that's in the desktop version
 * move the methods from ngeo to gmf, while avoiding the use of `['']`

I'll work on this soon.

sbrunner:
Probably just a missing asserts : https://travis-ci.org/camptocamp/ngeo/jobs/279666721#L1028-L1045

fredj:
@adube thanks for your work on this; please merge

adube:
Done.

PR #2928, opened 2017-09-22
GetData on filtered data source only

### LINKED ISSUE #2927: Filters: all visible layers are queried
When using action "Get Data", data from all visible layers is returned. There should be data from the filtered layer only.

To test:
 * https://camptocamp.github.io/ngeo/master/examples/contribs/gmf/apps/desktop/
 * Layer group Filters
 * Ouverture OSM layer activated
 * Create Ouverture OSM filter
 * Get Data
 * Results include data from Carte OSM, it shouldn't

adube:
Fixes #2927.

In the ngeo filter component, when using `getData`, only get data from the data source that is currently being filtered, not all currently visible data sources.


gberaudo:
@adube, to be able to use the fix in EPFL you must merge it into 2.2 (or cherry-pick it there).

sbrunner:
@adube  as @gberaudo say it's required on the branch 2.2, can you re-base on it (I just change the base in the PullRequest)?

PR #2945, opened 2017-10-02
Remove unused import of $compile

pfirpfel:
Was looking for usages of $compile to have an example. $compile gets injected but not used here.
I checked the history of the file, $compile was never used. Probably a leftover from development.

### DIFF
--- a/src/directives/drawfeature.js
+++ b/src/directives/drawfeature.js
@@ -120,7 +120,7 @@ ngeo.module.directive('ngeoDrawfeature', ngeo.drawfeatureDirective);
  * @ngdoc controller
  * @ngname ngeoDrawfeatureController
  */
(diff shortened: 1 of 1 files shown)


src/directives/drawfeature.js
```diff
@@ -121,8 +120,8 @@ ngeo.module.directive('ngeoDrawfeature', ngeo.drawfeatureDirective);
  * @ngdoc controller
  * @ngname ngeoDrawfeatureController
  */
-ngeo.DrawfeatureController = function($scope, $compile, $sce,
-  gettextCatalog, ngeoDecorateInteraction, ngeoFeatureHelper, ngeoFeatures) {
+ngeo.DrawfeatureController = function($scope, $sce, gettextCatalog
```
fredj: missing comma

PR #2949, opened 2017-10-03
Fix definitions leaking outside of the provide

gberaudo:
All definitions should be local to the provided symbol.
As such, a symbol ngeo.search.toto cannot define Toto in ngeo.search.


gberaudo:
Thanks

PR #2952, opened 2017-10-03
Only update the scale value at the end of the zoom animation

fredj:
Otherwise `$applyAsync` is called on every frame change (~60 / second)

### DIFF
--- a/src/directives/scaleselector.js
+++ b/src/directives/scaleselector.js
@@ -225,9 +225,11 @@ ngeo.ScaleselectorController.prototype.handleResolutionChange_ = function(e) {
   //
   // For that reason we use $applyAsync instead of $apply here.
 
-  this.$scope_.$applyAsync(() => {
-    this.currentScale = currentScale;
-  });
+  if (currentScale !== undefined) {
+    this.$scope_.$applyAsync(() => {
+      this.currentScale = currentScale;
+    });
+  }
 };
 
 


adube:
@fredj I'm interested in this.  Here's why.

Since OpenLayers 3.20.0 (if I recall correctly) the `change:resolution` event (on the resolution property of the `ol.View`) is fired on every frame change.  Before that, it used to be fired once, and it had the value of the 'end' resolution right away.

That change caused this kind of problem that you raised and addressed with this fix.  Oh, btw I agree with your fix.

The thing is: natively, within OpenLayers, it would be nice to be able to listen to an event that would only be fired when the resolution changes to one of the resolution we have in the constraint.  Hear me out.  I would suggest the following enhancement within OL.

Introduce a new `ol.ViewProperty.CONSTRAINT_RESOLUTION` that would only be set when the value is one among the resolution that the user specified as constraint when the view was created.  When setting the resolution, the view would be responsible of checking if it's one of the constraint resolution and set it as well if it does.  With this, one could register an event on that property, which would avoid the management that needs to be repeated otherwise.

I would also add: it would make sense to set the property to `null` (only once) if the current property has a value, but we're currently with a resolution between those defined as constrains.  This would allow a listener to act on `null` value as well.  For example:

 * while zooming (not in constraint, value is null) - hide something
 * at the end (constrain value is set) - show something

What do you think?

fredj:
Yes; it definitely can be improved in openlayers; maybe using a combination of `change:resolution` and `moveend`

PR #2951, opened 2017-10-03
Import DataSource Restructuring (1)

adube:
This PR introduces a restructuring of the data source definition within ngeo and gmf to allow the upcoming new "file" data source type.

In summary:

 * we now have `ngeo.datasource` and `gmf.datasource` namespaces (directories)
 * all classes directly related to data sources were moved in there
 * `ngeo.Datasource` has been replaced by:
  * `ngeo.datasource.DataSource` - the abstract class for a data source
  * `ngeo.datasource.OGC` - a data source that uses OGC services to fetch/display its data
 * `gmf.DataSource` has been replaced by `gmf.datasource.OGC`

The 'externs' definitions have been updated as well.

PR #2956, opened 2017-10-05
Import DataSource Component (2)

adube:
This PR introduces the `gmf-importdatasource` component:

 * contains 2 modes:
  * local (currently does nothing as part of this PR.  Will come in an other PR)
  * online (currently shows a plain text input instead of a list as part of this PR.  The list will come in an other PR)

The `ngeo.Querent` service is now able to issue WMS GetCapatilities requests, and even cache the result in case we're using the OnlineResource url as unique key for the cache.

A `gmf.datasource.ExternalDataSourcesManager` service is introduced in the PR to take care of managing the creation of the data sources that are created from the WMS Capabilities (WMTS and files will come in other tickets).

The component has also been added to the `desktop_alt` app, in order to be able to test the logging in and out, i.e. while doing so, external data sources should be kept (this is the case in this PR).

## Change to the gmf.datasource.DataSourcesManager service

The manager used to always clear the collection of ngeo data sources upon changing theme, i.e. when logging in/out we cleared all of the ngeo data sources.  This is no longer the case as part of this PR, as external data sources are added in there as well, and changing theme should not remove them.

To fix this, only the data sources that are in the manager cache are removed, thus leaving the external data sources in there, untouched.

## Id of the external data sources

The concept of data source id is critical for the manager, queries, etc. to do their job accordingly.  Each data source must have a unique numeric id.  This is the case already for the data source that are created from the theme JSON.

For external data sources, we do not have an id, only a name.  Also, a name is not enough as id, since multiple servers could have layers of the same name.  As part of this PR, the id issue is addressed like this:

 * generate an id on the capabilities object using `ol.getUid`
 * add a million

This solution works as part of GMF, as long as the ids returned by the theme remain under the million.  Please, validate this.

## Todo

 * [ ] Rebase onto master once #2951 is merged
 * [ ] Review
 * [x] Manage changing theme
 * [x] Manage login/logout
 * [ ] Approve the strategy involving the external data source ids generation

## Example

You can use the following url to test: http://wms.geo.admin.ch/?lang=fr

You can test either in the GMF example, or in the desktop_alt app.

As part of this PR, no layer is actually added.  The data source is simply pushed in the `ngeoDataSources` collection, nothing else.  The creation of an OpenLayers layer will come in a separate PR.


adube:
@fredj Once #2951 gets merged and this is rebased on top of master, then it would be ready for review.

adube:
The revision will take place in #2988.

PR #2961, opened 2017-10-09
Adding missing datasource attribute to mapfish-print externs

pfirpfel:
see:
https://mapfish.github.io/mapfish-print-doc/attributes.html#!datasource

### DIFF
--- a/externs/mapfish-print-v3.js
+++ b/externs/mapfish-print-v3.js
@@ -45,7 +45,7 @@ MapFishPrintAttributes.prototype.map;
 
 
 /**
- * @type {Array}
+ * @type {Array.<Object>}
  */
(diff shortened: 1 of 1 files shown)


fredj:
@sbrunner is `{Array}` correct? or can we have an other type for the `datasource` ?

fredj:
if the type is free, we should use `{?}` meaning "Indicates that the variable can take on any type, and the compiler should not type-check any uses of it."

see https://github.com/google/closure-compiler/wiki/Type-Expressions#the-unknown-type

fredj:
@pfirpfel According to @sbrunner ; the type should be `Array.<Object>`

fredj:
@pfirpfel thanks, please merge

PR #2962, opened 2017-10-09
[gmf] Create a promise for each enumerated attribute

arnaud-morvan:
When getting enumerations for enumerated attributes,
we need a different promise for each attribute.


arnaud-morvan:
Fixed by 0863a4caf5647c732c33962a80c905ad85444c05

PR #2964, opened 2017-10-10
Set correct print labelAlign

### LINKED ISSUE #2963: print: label align is ignored
To reproduce the issue:
 * https://camptocamp.github.io/ngeo/master/examples/contribs/gmf/apps/desktop/
 * draw a label
 * print the page

-> the label on the map is centered but the printed label is aligned to the left

fredj:
Default values in OpenLayers are 'center' and 'middle'.

fixes #2963


fredj:
Thanks for the review

Issue #2965, opened 2017-10-10
[GMF] Login/Logout causes angular error


pfirpfel:
Happens on current master.

## Steps to reproduce
1. Open GMF desktop app
2. Login
3. Logout

## Expected
No errors

## Observed
Errors in console:
![selection_055](https://user-images.githubusercontent.com/621371/31378109-8078d8e2-adaa-11e7-959d-74f71e216c7f.png)

PR #2966, opened 2017-10-10
Move modal out of ng-if-block to avoid linking it again

pfirpfel:
Fixes: https://github.com/camptocamp/ngeo/issues/2965


pfirpfel:
Alternative would be to change the ng-if blocks into ng-show (with ng-cloak).

PR #2974, opened 2017-10-11
Update closure-util to the latest version

fredj:


fredj:
I've an issue (locally) with the `transpile.js` (file not found)

package.json
```diff
@@ -33,7 +33,7 @@
     "async": "2.5.0",
     "bootstrap": "3.3.7",
     "clean-css-cli": "4.1.10",
-    "closure-util": "camptocamp/closure-util#487fac6",
+    "@camptocamp/closure-util": "camptocamp/closure-util#f1152e1",
```
sbrunner: Can't we do a release on npm?

fredj: You're right, version 1.24.0 is already published on npm

sbrunner: :-)

fredj:
:tada:

PR #2976, opened 2017-10-11
WIP - Import DataSource - Manage layer and layer tree (3)

adube:
This PR introduces 2 mains things:

 * the addition of an OpenLayers layer for the external data sources that are added as part of a same server
 * the addition of nodes for these external data sources in the layer tree.

More details below.

## ngeo.datasource.WMSGroup

This is a new class that has been introduced to manage the collection of multiple OGC data sources that come from the same server.  What it does:

 * creates the `ol.layer.Image` object
 * manages the `LAYERS` param property of the WMS source depending on the `visible` property of the data sources within the group
 * provide a `visibilityState` property (on, off, indeterminate) that is automatically calculated and can be watched nicely from a template

## gmf-wmsgrouptree

This new component is bound to a single `ngeo.datasource.WMSGroup`. It displays the data source within exactly like a layer tree would.

It "borrows" the same CSS class names and the same type of HTML structure.

We had to do this, because the point of entry are data sources, and not `node` objects.  Nodes are what the current layer tree uses as point of entry.  It wouldn't have been possible to manage nodes.

The UI of a WMS Group tree is, therefore, very similar to the gmf-layertree, with minor exceptions:

 * it is possible to remove a group OR a data source
 * no opacity slider
 * no legend
 * no filter

## gmf-layertree

The GMF layertree directive now injects the `gmf.datasource.ExternalDataSourcesManager` service.  Within its template, the collection of WMS groups is watched to build the list of `gmf-wmsgrouptree` components that are immediately shown after the "normal" nodes.  Only enabled at the "root" level, i.e. they are added at the first 'depth' level.

## Todo

 * [ ] Rebase onto master once #2956 is merged
 * [ ] Review
 * [ ] There seems to be an error thrown by OpenLayers for the images. See error below:
 * [ ] Small UI issue: the block for the external data sources in the layer tree is following the others.

**Error**:
```
Uncaught DOMException: Failed to execute 'getImageData' on 'CanvasRenderingContext2D':
The canvas has been tainted by cross-origin data.
```

## Example

You can use the following url to test: http://wms.geo.admin.ch/?lang=fr

You can test either in the GMF example, or in the desktop_alt app.


adube:
The revision will take place in #2988.

PR #2979, opened 2017-10-12
Document nodejs installation

arnaud-morvan:


docs/developer-guide.md
```diff
@@ -7,6 +7,15 @@ This document provides information for developers working on ngeo.
...
 Check it with `java -version`.
 
+NodeJS >= 6
+
+```shell
+sudo curl -sL https://deb.nodesource.com/setup_6.x | bash -
+apt-get install -y nodejs
+```
+
+source : https://nodejs.org/en/download/package-manager/
```
romanzoller: Minor: `:` should not be preceded by a space in English, i.e. `Source:` instead of `Source :`

docs/developer-guide.md
```diff
@@ -7,6 +7,15 @@ This document provides information for developers working on ngeo.
 The Closure Compiler requires Java 7 or higher. You can use [OpenJDK](http://openjdk.java.net/install/).
 Check it with `java -version`.
 
+NodeJS >= 6
+
+```shell
+sudo curl -sL https://deb.nodesource.com/setup_6.x | bash -
+apt-get install -y nodejs
+```
```
sbrunner: Why copy the external documentation, I think adding the link is enough...

ger-benjamin:
@arnaud-morvan Probably not your priority, but what about this PR ?

arnaud-morvan:
Seems that every one has it's own method of installing nodejs, so as sbrunner said, a link would be sufficient

gberaudo:
Nowadays Node < 6.x is no more supported; see https://github.com/nodejs/Release#release-schedule.
So I am closing the PR.

PR #2987, opened 2017-10-13
Change ngeo.Scalify filter to be stateful

### LINKED ISSUE #2983: Scaleselector does not get updated after language change
Due to a race condition, this is easier to reproduce on the compiled version.

How to reproduce (on current master):
`make build-gmf-apps examples-hosted-apps serve`
Got to: http://localhost:3000/.build/examples-hosted/contribs/gmf/apps/desktop_alt/?lang=fr

(I'm using a browser that request German as language)

The scale selector should show "1 : 20,000" (the English default, I guess).

Zoom into the map.

The scale selector shows "1 : 20 000" (the french formatting)

Open the scale selector. All values should still use the old, English formatting.

I assume, because the model of the scaleselector does not change, the template does not get refreshed.

fredj:
To reevaluate the value when the locale change

fixes #2983

### DIFF
--- a/src/directives/partials/scaleselector.html
+++ b/src/directives/partials/scaleselector.html
@@ -5,7 +5,7 @@
   <ul class="dropdown-menu btn-block" role="menu">
     <li ng-repeat="zoomLevel in ::scaleselectorCtrl.zoomLevels">
       <a href ng-click="scaleselectorCtrl.changeZoom(zoomLevel)"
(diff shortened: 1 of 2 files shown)


pfirpfel:
@sbrunner Since 2.1 is also affected, can we have the fix there as well?

fredj:
@pfirpfel See #2991

PR #2988, opened 2017-10-13
Import DataSource - Component (2), Layer and tree (3), WMTS (4), Permalink (5)

adube:
This PR introduces the following:

 * 02 - component to display the UI to import/connect external data sources
 * 03 - the addition of the OpenLayers layer, and addition of the node in the layer tree, with WMS
 * 04 - WMTS support
 * 05 - Added data sources are added in the Permalink service

## Todo

 * [x] wait for #2951 to be merged and this to be rebased onto master
 * ~~should the "clear all" button in the layer tree clear the external data sources too?~~
 * ~~validate the UI of the external data sources in the layer tree~~
 * [x] review

We'll leave these 2 comments to be discussed in other issues.

## How to test

Use the `desktop_alt` application, with any of these urls:

WMS:
 * swisstopo: http://wms.geo.admin.ch/?lang=fr

WMTS:
 * ASIT VD: https://ows.asitvd.ch/wmts/1.0.0/WMTSCapabilities.xml
 * swisstopo: https://wmts.geo.admin.ch/1.0.0/WMTSCapabilities.xml?lang=fr

# Details

Here's details for each part.

## 02 - Component

This commit introduces the `gmf-importdatasource` component:

 * contains 2 modes:
  * local (currently does nothing as part of this PR.  Will come in an other PR)
  * online (currently shows a plain text input instead of a list as part of this PR.  The list will come in an other PR)

The `ngeo.Querent` service is now able to issue WMS GetCapatilities requests, and even cache the result in case we're using the OnlineResource url as unique key for the cache.

A `gmf.datasource.ExternalDataSourcesManager` service is introduced in the PR to take care of managing the creation of the data sources that are created from the WMS Capabilities (WMTS and files will come in other tickets).

The component has also been added to the `desktop_alt` app, in order to be able to test the logging in and out, i.e. while doing so, external data sources should be kept (this is the case in this PR).

### Change to the gmf.datasource.DataSourcesManager service

The manager used to always clear the collection of ngeo data sources upon changing theme, i.e. when logging in/out we cleared all of the ngeo data sources.  This is no longer the case as part of this PR, as external data sources are added in there as well, and changing theme should not remove them.

To fix this, only the data sources that are in the manager cache are removed, thus leaving the external data sources in there, untouched.

### Id of the external data sources

The concept of data source id is critical for the manager, queries, etc. to do their job accordingly.  Each data source must have a unique numeric id.  This is the case already for the data source that are created from the theme JSON.

For external data sources, we do not have an id, only a name.  Also, a name is not enough as id, since multiple servers could have layers of the same name.  As part of this PR, the id issue is addressed like this:

 * generate an id on the capabilities object using `ol.getUid`
 * add a million

This solution works as part of GMF, as long as the ids returned by the theme remain under the million.  Please, validate this.

## 03 - Layer and layer tree support

This commit introduces 2 mains things:

 * the addition of an OpenLayers layer for the external data sources that are added as part of a same server
 * the addition of nodes for these external data sources in the layer tree.

More details below.

### ngeo.datasource.WMSGroup

This is a new class that has been introduced to manage the collection of multiple OGC data sources that come from the same server.  What it does:

 * creates the `ol.layer.Image` object
 * manages the `LAYERS` param property of the WMS source depending on the `visible` property of the data sources within the group
 * provide a `visibilityState` property (on, off, indeterminate) that is automatically calculated and can be watched nicely from a template

### gmf-wmsgrouptree

This new component is bound to a single `ngeo.datasource.WMSGroup`. It displays the data source within exactly like a layer tree would.

It "borrows" the same CSS class names and the same type of HTML structure.

We had to do this, because the point of entry are data sources, and not `node` objects.  Nodes are what the current layer tree uses as point of entry.  It wouldn't have been possible to manage nodes.

The UI of a WMS Group tree is, therefore, very similar to the gmf-layertree, with minor exceptions:

 * it is possible to remove a group OR a data source
 * no opacity slider
 * no legend
 * no filter

### gmf-layertree

The GMF layertree directive now injects the `gmf.datasource.ExternalDataSourcesManager` service.  Within its template, the collection of WMS groups is watched to build the list of `gmf-wmsgrouptree` components that are immediately shown after the "normal" nodes.  Only enabled at the "root" level, i.e. they are added at the first 'depth' level.

## 04 - WMTS support

This commit introduces the possibility to import a WMTS data source.

### ngeo.datasource.Group

The previous `ngeo.datasource.WMSGroup` has been split:

 * ngeo.datasource.Group --> a collection of data sources, provide the `visibilityState` watchable property
 * ngeo.datasource.WMSGroup --> extends Group, adds the concept of the `ol.layer.Image` creation and watchers for the data source `visible` property that affects the layer

### gmf-wmsgrouptree --> gmf-datasourcegrouptree

This directive was renamed.  It makes more sense that way, and can be safely re-used for the list of WMS data sources as well as the WMTS data sources to be shown in the layer tree.

## 05 - Permalink

This commit introduces the support of the External data sources within the Permalink service.

The parameters that are saved are:

 * `eds_u` - stands for External Data Sources Urls.  This is a comma-separated list of the urls that were used to connect the data sources.  There's one url for each "group" of data source connected.
 * `eds_n` - stands for External Data Sources Names. This is a comma-separated list of the data source names (i.e. layer names) for each url.  For each url, the names are themselves separated by `;`, so a group can have multiple names.

The order is important here, as the first url is associated with the first "group" of names.

Currently, nothing happens if a name doesn't exist, or if an url is invalid.


adube:
@fredj - Ready for review.

adube:
Rebased onto master and fixed the 'make test' issue.

adube:
@fredj I don't understand what's wrong: I made fixes in the `import-ds-05-permalink` branch, but the PR doesn't update itself.  See the branch: https://github.com/camptocamp/ngeo/tree/import-ds-05-permalink

It even says it's this PR.  What's going on?

adube:
I tried resetting the last commit with "--amend", tried renaming the commit message, resetting the last commit, making a new one, nothing...

adube:
Oh... right... see: https://status.github.com/messages

adube:
> October 17, 2017
> 
> 11:27 EDTWe continue to investigate a large queue backlog. Commits will take longer than usual to appear in pull requests.
> 10:46 EDTWe continue to investigate a queue backlog. You may experience that commits take longer than usual to appear in pull requests.
> 10:23 EDTWe are investigating a queue backup. You may experience that commits take longer than usual to appear in pull requests.
> 10:12 EDTWe are investigating reports of elevated error rates.

adube:
Rebased onto master.

adube:
Rebased onto master.

adube:
Travis happy.  @fredj Please, review.

PR #2991, opened 2017-10-16
Change ngeo.Scalify filter to be stateful

fredj:
Port of #2987 for the 2.1 branch


fredj:
Note: @pfirpfel is testing it

PR #2992, opened 2017-10-16
Add quotes arround $stateful property

fredj:
To avoid renaming by the closure-compiler


pfirpfel:
Tested compiled version, works for me.

fredj:
Thanks for the tests @pfirpfel

PR #2998, opened 2017-10-17
Fix display query window minimum size

adube:
Fix the min size the display query window.  Also, make sure to show the proper pointer cursor when hovered to tell that it can be dragged and show an other cursor while being dragged.


adube:
@fredj Ready for review.

PR #3000, opened 2017-10-17
TEST - Import ds 05 permalink clone

adube:
Do not merge.


adube:
@fredj  Something seems to be wrong with Github... I had to create this PR manually, i.e. when visiting the home page did not propose to create the PR.  Related?

PR #3003, opened 2017-10-18
search parameter/map_zoom: Test for null to allow 0 as value

pfirpfel:
Addresses @fredj [feedback](https://github.com/camptocamp/ngeo/pull/3001#discussion_r145355559).

### DIFF
--- a/contribs/gmf/src/directives/search.js
+++ b/contribs/gmf/src/directives/search.js
@@ -965,7 +965,7 @@ gmf.SearchController.prototype.fulltextsearch_ = function(query, resultIndex, op
         const feature = format.readFeature(data.features[resultIndex - 1]);
         this.featureOverlay_.addFeature(feature);
         const fitOptions = /** @type {olx.view.FitOptions} */ ({});
-        if (opt_zoom) {
+        if (opt_zoom !== null) {
           fitOptions.maxZoom = opt_zoom;
         }
         this.map_.getView().fit(feature.getGeometry().getExtent(), fitOptions);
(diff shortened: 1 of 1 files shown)


contribs/gmf/src/directives/search.js
```diff
@@ -965,7 +965,7 @@ gmf.SearchController.prototype.fulltextsearch_ = function(query, resultIndex, op
         const feature = format.readFeature(data.features[resultIndex - 1]);
         this.featureOverlay_.addFeature(feature);
         const fitOptions = /** @type {olx.view.FitOptions} */ ({});
-        if (opt_zoom) {
+        if (opt_zoom !== null) {
```
fredj: if `opt_zoom` is not passed to the function, it will be `undefined` not `null`

fredj: And the `@param` for `opt_zoom` should be changed to:
```
@param {number=} opt_zoom Optional zoom level.
```

See https://github.com/google/closure-compiler/wiki/Annotating-JavaScript-for-the-Closure-Compiler#optional-parameter-in-a-param-annotation

contribs/gmf/src/directives/search.js
```diff
@@ -965,7 +965,7 @@ gmf.SearchController.prototype.fulltextsearch_ = function(query, resultIndex, op
         const feature = format.readFeature(data.features[resultIndex - 1]);
         this.featureOverlay_.addFeature(feature);
         const fitOptions = /** @type {olx.view.FitOptions} */ ({});
-        if (opt_zoom) {
+        if (typeof opt_zoom !== 'undefined') {
```
fredj: can be simplified to `if (opt_zoom !== undefined) {`

sbrunner: +1

fredj:
travis fails (probably) because the branch is from your fork, not the camptocamp one

PR #3005, opened 2017-10-18
Add resizable option to ngeo.Modal

adube:
This PR adds the `ngeo-modal-resizable` option to the modal directive. Default value is: `false`.

This changes the default behaviour of the modals, which could all be resized before.  However, it makes little sense to do so, as most of the times modals are used to show content from an "controlled" interface, i.e. content from other directives / components.  The content container of the modal (bootstrap) already manages to be bigger in height if it has lots of content, and show a scroll bar if necessary.

Making the modal resizable doesn't help anyway, as the whole container of the modal is resized, unnaffecting the header, footer and content. This results in having more "blank" space and nothing more.

### DIFF
--- a/src/directives/modal.js
+++ b/src/directives/modal.js
@@ -33,8 +33,10 @@ goog.require('ngeo');
  *
  * @param {angular.$parse} $parse Angular parse service.
  * @return {angular.Directive} The directive specs.
- * @htmlAttribute {boolean} ngeo-modal-destroy-content-on-hide Destroy the content
- * when the modal is hidden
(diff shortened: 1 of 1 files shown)


adube:
@fredj Ready for review.  I already discussed about this with @ybolognini.  You can check with him, if required.

adube:
Rebased onto master.

adube:
@fredj The error reported by Travis doesn't seem related to this PR.  Could I merge, regardless?

fredj:
The issue reported by travis has been fixed in #3002, merging

PR #3010, opened 2017-10-20
Set correct print labelAlign

fredj:
Default values in OpenLayers are 'center' and 'middle'.

This is a backport of #2964 for the 2.2 branch


fredj:
ignoring codacy failure

Issue #3011, opened 2017-10-20
Layertree popover style inconsistent on Internet Explorer


pfirpfel:
Affected version: master (and older)

The layertree popover (opacity, show/hide legend, filter) has an inconsistent look on Internet Explorer (reproduced on IE11).

Use the [demo as an example](https://geomapfish-demo.camptocamp.net/2.2/theme/Filters).

The first click on the cog opens this popover:
![opacity_1](https://user-images.githubusercontent.com/621371/31809760-6a33af68-b579-11e7-9ad7-f7097fdec710.png)

On the second click it looks like this:
![opacity_2](https://user-images.githubusercontent.com/621371/31809805-9fa95bfc-b579-11e7-9235-8f045a734be3.png)

On the third and any successive click after that, it looks like this:
![opacity_3](https://user-images.githubusercontent.com/621371/31809826-af87f024-b579-11e7-8171-a836b4fabdf2.png)

In comparison on Firefox, it always looks like this:
![selection_060](https://user-images.githubusercontent.com/621371/31809866-d71b45d2-b579-11e7-91d5-34fd4832defc.png)

PR #3016, opened 2017-10-20
IE11 doesnt consistently size the popover box.

ochriste:
So we need to set a min width rule to force it.

fix  https://jira.camptocamp.com/browse/GEO-609

### DIFF
--- a/contribs/gmf/less/popover.less
+++ b/contribs/gmf/less/popover.less
@@ -9,6 +9,8 @@
   .popover-content {
     ul {
       margin-bottom: 0;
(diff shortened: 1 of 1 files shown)


pfirpfel:
Fixes #3011

pfirpfel:
The look doesn't change now, it stays like this:
selection_063

Do you think, you could also fix the height of the box?

On other browsers, there is much less empty space:
selection_064

ochriste:
the space is because of the show/hide legend item
I can try to fix that

sbrunner:
Not present on 2.2?

PR #3023, opened 2017-10-24
Fix displayquerywindow content height

adube:
This PR makes the height content of the DisplayQueryWindow relative to its parent, i.e. the height is calculated instead of being fixed.

This allow the size to be dynamic.  For example, when resizing the window, the content height follows.

Note: the width is already working properly.


fredj:
codacy check can be ignored

adube:
Rebased onto master.

PR #3034, opened 2017-10-27
Fix (again) angular-gettext externs

gberaudo:


gberaudo:
Travis is unhappy for an unrelated reason: https://travis-ci.org/camptocamp/ngeo/jobs/293611860.
So merging for not being blocked uselessly.

PR #3047, opened 2017-11-02
Minor cleanup in ngeo externs

fredj:


fredj:
> Would be cool to also have it in the 2.2 branch.

done

PR #3050, opened 2017-11-02
Better coordinates formating in contextual menu

### LINKED ISSUE #3049: contextual data coordinates in EPSG:21781
In french, the coordinates in the contextual menu looks like `631 285 152 450`; it's hard to read.
what about adding a `/` ? something like:  `631 285 / 152 450`

fredj:
fixes #3049

### DIFF
--- a/contribs/gmf/apps/desktop/contextualdata.html
+++ b/contribs/gmf/apps/desktop/contextualdata.html
@@ -4,7 +4,7 @@
       Swiss grid (LV03)
     </td>
     <td>
-      {{coord_21781|ngeoNumberCoordinates}}
+      {{coord_21781|ngeoNumberCoordinates:0:'{x} / {y}'}}
     </td>
   </tr>
   <tr>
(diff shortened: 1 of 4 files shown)


contribs/gmf/apps/desktop/contextualdata.html
```diff
@@ -4,15 +4,15 @@
       Swiss grid (LV03)
     </td>
     <td>
-      {{coord_21781|ngeoNumberCoordinates}}
+      {{coord_21781|ngeoNumberCoordinates:0:'{x} / {y}'}}
```
sbrunner: Finally we want a /?

fredj: The problem with the comma is that in english we will have: `631,285, 152,450` which is not very readable.

sbrunner: And what about making it translatable?

fredj:
let's merge this pull request as it

PR #3053, opened 2017-11-03
Fix ImportDataSource description position (CSS, flex)

adube:
Fixes the position (CSS) of the description in the ImportDataSource tool.


adube:
Travis fails due to issues unrelated to this PR.  Since it's a simple patch that doesn't affect any actual JS code, I'll merge right away anyway.

PR #3054, opened 2017-11-03
Remove legacy ImportLayer from desktop_alt

adube:
This PR removes the legacy 'Import Layer' tool from the `desktop_alt` app.

Note that its components are not removed by this PR, as the new ImportDataSource tool replacing it doesn't currently support the "typeahead" feature, and code might still be borrowed from the legacy version.

Once the "typeahead" features is supported in the new tool, then we'll be able to do the last "cleanup" of the files that were used by the legacy tool.

### DIFF
--- a/contribs/gmf/apps/desktop_alt/index.html
+++ b/contribs/gmf/apps/desktop_alt/index.html
@@ -23,11 +23,6 @@
       </div>
     </header>
     <main>
-      <div
-         ngeo-import-dnd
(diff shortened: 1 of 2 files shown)


adube:
> LGTM, can you create an issue for the final cleanup so we don't forget it ? thanks.

Done: #3055

adube:
Error reported by Travis is not related to this PR.  Merging...

PR #3063, opened 2017-11-06
ImportDataSource - Add twitter typeahead to list servers

adube:
This PR introduces Twitter's typeahead in the ImportDataSource component.

The url is displayed in the suggestions, and used in the search.  This could be enhanced in the future (i.e. we could use the name instead / too).

## Todo

 * [x] review
 * [x] ~~validate CSS style - I "borrowed" the code from https://github.com/bassjobsen/typeahead.js-bootstrap-css/, is that okay?~~ Fixed by fred, thanks.
 * [x] find a solution about the fact that Typeahead only matches from the beginning of the searched string...
 * [x] requires master to build properly, see: #3071

### DIFF
--- a/contribs/gmf/src/directives/importdatasource.js
+++ b/contribs/gmf/src/directives/importdatasource.js
@@ -19,6 +19,7 @@ gmf.ImportdatasourceController = class {
   /**
    * @param {!angular.JQLite} $element Element.
    * @param {!angular.$filter} $filter Angular filter.
+   * @param {!angular.$injector} $injector Main injector.
    * @param {!angular.Scope} $scope Angular scope.
    * @param {!angular.$timeout} $timeout Angular timeout service.
    * @param {!gmf.datasource.ExternalDataSourcesManager}
@@ -31,7 +32,7 @@ gmf.ImportdatasourceController = class {
    * @ngdoc controller
    * @ngname GmfImportdatasourceController
    */
-  constructor($element, $filter, $scope, $timeout,
+  constructor($element, $filter, $injector, $scope, $timeout,
     gmfExternalDataSourcesManager, ngeoQuerent) {
 
     // Binding properties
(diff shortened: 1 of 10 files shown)


contribs/gmf/src/directives/importdatasource.js
```diff
@@ -167,6 +188,25 @@ gmf.ImportdatasourceController = class {
    */
   $onInit() {
     this.gmfExternalDataSourcesManager_.map = this.map;
+
+
+    if (this.serversEngine_) {
+      // Timeout to let Angular render the placeholder of the input properly,
+      // otherwise typeahead would copy the string with {{}} in it...
+      this.timeout_(() => {
+        const engine = goog.asserts.assert(this.serversEngine_);
```
fredj: I find this a bit complicated and uneasy to read; what about simply have:
```
goog.asserts.assert(this.serversEngine_);
```
and use `this.serversEngine_.ttAdapter()` bellow

contribs/gmf/src/directives/importdatasource.js
```diff
@@ -167,6 +188,25 @@ gmf.ImportdatasourceController = class {
...
+        const $urlInput = this.element_.find(
```
fredj: can probably be simplified to `this.element_.find('input[name=url]');`

fredj:
It doesn't seems to work for me; I've searched for `geo` or `swiss` in the input and had not response from typeahead

adube:
>It doesn't seems to work for me; I've searched for geo or swiss in the input and had not response from typeahead

Typeahead only matches from the beginning of the string and doesn't seem to be configurable otherwise.

"That's by design. Since typeahead.js isn't a search engine, it should be okay if it uses a somewhat naive algorithm to get suggestions."

Any suggestion?

adube:
Found a solution, trying it now...

https://stackoverflow.com/questions/22059933/twitter-typeahead-js-how-to-return-all-matched-elements-within-a-string

adube:
@fredj Alright, I applied the solution it it worked.  Please, review...

fredj:
Looks good to me. To have the same look between the search and the import data source (typeahead style), I've grouped the common css into a file. New commit to come

fredj:
@adube see a91821cd61de3acf6d58704b3ad5d1c8e0bc9ba2

Does it look good to you?

adube:
Code looks clean, however it changed the UI of the input in the ImportDS tool.  I've cherry-picked your commit and will make a quick fix right away.  Thank you very much for this.

adube:
Okay, this looks perfect, now.

adube:
@fredj Would you please take one last look at the last commit I made?  Just to make sure you agree with it.  If so, then I'll merge.

contribs/gmf/src/directives/importdatasource.js
```diff
@@ -176,10 +176,11 @@ gmf.ImportdatasourceController = class {
          * Borrowed from:
          * https://stackoverflow.com/questions/22059933/twitter-typeahead-js-how-to-return-all-matched-elements-within-a-string
          *
-         * @param {string} url Url.
-         * @return {!Array.<string>} List of datum tokenizers.
+         * @param {BloodhoundDatum} url Url.
```
fredj: the param name could be renamed to `datum` instead of `url`

adube:
Build fails.  See: #3071

fredj:
rebased

adube:
Travis passes.  Merging now...

PR #3067, opened 2017-11-07
Wait on layer construction before displaying disclaimer

### LINKED ISSUE #2912: Disclaimers of Layers within LayerGroups are not displayed
Using 2.2 branch, instance: https://geomapfish-demo.camptocamp.net/2.2

This seems to be just a symptom. I have not found the cause yet. 

I use the [2.2-demo](https://geomapfish-demo.camptocamp.net/2.2/theme/Demo) as an example. The Demo-Theme includes multiple LayerGroups which contain multiple layers.

Yet in in registerLayer_, the [recursive registering](https://github.com/camptocamp/ngeo/blob/f322efaad5b107d159b5ede6f88c2893105eb5b4/contribs/gmf/src/directives/disclaimer.js#L239) never happens, because the LayerGroups thrown at it are all empty or not LayerGroups.

ger-benjamin:
Fix #2912

From where come the issue:

- The node is detected as a not-mixed-group
- So an ol3 layer is created and added to the map
- A add_layer event is launched by ol3
- the disclaimer directive read immediately the layer and display the disclaimer information from the layer
- The js process continue and THEN the ol3 layer is filled with the name of the layers inside. (that correspond to add string inside the layer. Like when we check/un-check a layer. There is no special event to detect for that)

So, I've added a timeout to wait until the layer is completely fulfilled.

### DIFF
--- a/contribs/gmf/src/directives/disclaimer.js
+++ b/contribs/gmf/src/directives/disclaimer.js
@@ -77,6 +77,7 @@ gmf.module.component('gmfDisclaimer', gmf.disclaimerComponent);
  * @param {!angular.JQLite} $element Element.
  * @param {!angular.Scope} $scope Angular scope.
  * @param {!angular.$sce} $sce Angular sce service.
+ * @param {!angular.$timeout} $timeout Angular timeout service.
  * @param {!ngeo.CreatePopup} ngeoCreatePopup Popup service.
  * @param {!ngeo.Disclaimer} ngeoDisclaimer Ngeo Disclaimer service.
  * @param {!ngeo.EventHelper} ngeoEventHelper Ngeo Event Helper.
@@ -86,8 +87,8 @@ gmf.module.component('gmfDisclaimer', gmf.disclaimerComponent);
  * @ngdoc controller
  * @ngname GmfDisclaimerController
  */
-gmf.DisclaimerController = function($element, $scope, $sce, ngeoCreatePopup,
-  ngeoDisclaimer, ngeoEventHelper, ngeoLayerHelper) {
+gmf.DisclaimerController = function($element, $scope, $sce, $timeout,
+  ngeoCreatePopup, ngeoDisclaimer, ngeoEventHelper, ngeoLayerHelper) {
 
   /**
    * @type {?ol.Map}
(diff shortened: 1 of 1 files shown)


pfirpfel:
Fixes the issue for me, but does not compile:

```
node buildtools/build.js .build/app-mobile.json .build/mobile.js
info ol Parsing dependencies
info ol Compiling 604 sources
ERR! compile /home/mkuenzli/dev/ngeo/contribs/gmf/src/directives/disclaimer.js:198: ERROR - actual parameter 1 of gmf.DisclaimerController.prototype.registerLayer_ does not match formal parameter
ERR! compile found   : *
ERR! compile required: (null|ol.layer.Base)
ERR! compile     this.registerLayer_(layer);
ERR! compile                         ^^^^^
ERR! compile 
ERR! compile 
ERR! compile 1 error(s), 0 warning(s), 96.8% typed
ERR! compile 
```

contribs/gmf/src/directives/disclaimer.js
```diff
@@ -189,7 +194,9 @@ gmf.DisclaimerController.prototype.$onInit = function() {
 gmf.DisclaimerController.prototype.handleLayersAdd_ = function(evt) {
   const layer = evt.element;
```
pfirpfel: `const layer = /** @type {ol.layer.Base} */ (evt.element);`

With that the code compiles for me.

fredj: strange that the `goog.asserts.assertInstanceof` bellow doesn't work

ger-benjamin: Yes, that's because it's outside the timeout. I've moved it in. It should compile now.
(The @pfirpfel solution should work too)

ger-benjamin:
I think travis get only "esource error: 302, https://b.tile.openstreetmap.org/2/3/1.png" errors. Can someone confirm ? can I merge ? (I'll retry tomorrow to be sure)

fredj:
> I'll retry tomorrow to be sure

I've restarted the jobs

ger-benjamin:
I relaunch it but no way, It never passes....

PR #3069, opened 2017-11-08
Ignore ressource error from openstreetmap

fredj:


ger-benjamin:
It looks that for this travis build, you get any openstreetmap errors :sweat_smile:

PR #3072, opened 2017-11-09
Be able to correctly translate in french

sbrunner:
=> `Vous utilisez l'application mobile. Allez sur l'` + `application standard`


gberaudo:
@sbrunner, see https://rocketeer.be/blog/2016/06/translate-params/.

sbrunner:
=> ```html
<span translate 
  translate-params-url="http://camptocamp.github.io/ngeo/master/examples/contribs/gmf/apps/desktop/?no_redirect">You're using the mobile application. Check out the <a href="{{url}}">standard application</a>.</span>
```
?

**Comment by gberaudo** [comment]
@sbrunner, I think the URL should be quoted like this:
```html
  translate-params-url="'http://camptocamp.github.io/ngeo/master/examples/contribs/gmf/apps/desktop/?no_redirect'"
```

sbrunner:
To be rested but the URL should already be quoted in `<a href="{{url}}">`...

gberaudo:
No, I mean you need to quote the URL so that it is evalutated correctly by angular. Otherwise I guess it will consider your URL to be a variable.

sbrunner:
OK, thanks :-)

PR #3073, opened 2017-11-09
Add basics ol-cesium to ngeo

gberaudo:
image


gberaudo:
@sbrunner, what is the correct way to handle the dependency on the less file, especially in examples?

sbrunner:
Superman is back :-)
Shouldn-t the images be in SVG or if there is an issue in hiDPI?
For me the structure is OK :-)

fredj:
here's a bigger version:
cyclopsclassic

gberaudo:
Right @sbrunner, and I used the wrong icons.
Will see what I can find.

sbrunner:
It can be changed latter :-)

gberaudo:
Thanks @sbrunner. I modified the PR with pngs created from geoadmin version of the controls.
Unfortunately these controls are based on a font; I don't know if there are SVG versions of the resources.

Maybe the user group will want to change the default resources anyway when 3d is officially added to ngeo. Ping @ybolognini.

fgravin:
I know dist versions are already broken, but this kind of integration might cause issues on what we want to provide as ngeo dists.
Will it grab all `ol-cesium` and `cesium` sources ?

Might postpone this question when we'll work with webpack.

gberaudo:
My understanding is that ngeo dist is supposed to provide all ngeo.
There are custom builds and other optmization strategies if you want a trimed down version.

Like said in https://github.com/camptocamp/ngeo/pull/3073/commits/a6b9edc549d7b5b5009126fdb0dd90202998929c, the ol-cesium code is an hard dependency; Cesium is optional.

gberaudo:
Thanks for the review @fredj.

PR #3076, opened 2017-11-10
Make lint target reliable

gberaudo:
Previously only the files newer than the timestamp were passed to the linter.
Changing the linter config or version did not trigger a lint of all the files.

With this commit, all files are linted on changes.


gberaudo:
@sbrunner, a full lint is around 5s. I expect this change will have no impact on our daily work.
We can optimize later if needed (at the cost of a more complex target).

PR #3086, opened 2017-11-13
 Fix WMTS Capabilities file detection

### LINKED ISSUE #3082: WMTS GetCap is not recognized
Step to reproduce:
 - Import https://maps.zg.ch/basisplan_sw_tiled_lv03/ExtendedCapabilities.xml
 
->  [`isWmtsGetCap`](https://github.com/camptocamp/ngeo/blob/master/src/modules/import/file.js#L24)  returns false

Expected:
   `isWmtsGetCap ` returns true

fredj:
fixes #3082

ping @oterral

### DIFF
--- a/src/modules/import/file.js
+++ b/src/modules/import/file.js
@@ -21,7 +21,7 @@ ngeo.File = function($q, $http, gettext) {
   };
 
   this.isWmtsGetCap = function(fileContent) {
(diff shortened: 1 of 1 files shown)


oterral:
Nice!!! Thanks @fredj

Issue #3065, opened 2017-11-07
Text from Drawing Tools is Rendered with a too large Buffer when Printed


marionb:
Project:
* GitHub URL: https://github.com/camptocamp/ngeo/
* Host: geomapfish-demo.camptocamp.net
* GeoMapFish version: 2.1, 2.2 and master

Permalink directly where we can reproduce the issue: 
 - https://geomapfish-demo.camptocamp.net/2.2/s/PWfg
 - https://geomapfish-demo.camptocamp.net/2.1/s/nliS
 - https://camptocamp.github.io/ngeo/2.1/examples/contribs/gmf/apps/desktop/theme/Demo?lang=en&tree_group_layers_Transport=fuel&baselayer_ref=OSM&rl_features=Fl(ye61Hncnn!15XzhOh5d-18T~n*LineString%25201%27c*%2523DB4436%27a*0%27o*0.2%27m*false%27s*10%27k*1)a(ywjzG5vap!zsUz21_..hx6_1U~n*Polygon%25202%27c*%2523DB4436%27a*0%27o*0.2%27m*false%27s*10%27k*1)p(fsh1Hn8bp!~t*true%27n*Text%25203%27c*%2523000000%27a*0%27o*0.2%27m*false%27s*10%27k*1)&map_x=630724&map_y=188277&map_zoom=3&tree_groups=Transport
[...]

PR #3089, opened 2017-11-14
Better size for text stroke

ger-benjamin:
For #3065 (but don't fix it)

I set the default stroke to 2, that's more visible than 1. (And 3 is too much in the MapFishPrint report, see my comment in the issue: https://github.com/camptocamp/ngeo/issues/3065#issuecomment-344189374)

I add the stroke value to the createTextStyle_ function. Before, it was always using the default value (stroke size was 3)

### DIFF
--- a/buildtools/check-example.js
+++ b/buildtools/check-example.js
@@ -36,8 +36,12 @@ page.onAlert = function(msg) {
   exitCode = 2;
 };
 page.onResourceError = function(resourceError) {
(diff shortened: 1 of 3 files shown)


ger-benjamin:
I got the standard OSM error on layertree.js, I retry once.

ger-benjamin:
Same error... I've cherry-picked on 2.2 the fix of fredj for OSM

PR #3090, opened 2017-11-14
Use 'one' function from jQuery to close the ngeo-popover

fredj:
And use ngeo.EventHelper in gmf.Permalink


fredj:
travis error is related to `contribs/gmf/src/directives/contextualdata.js` but it will be fixed in an other PR

PR #3092, opened 2017-11-14
profile.js: Remove usage of goog.events

pfirpfel:


src/directives/profile.js
```diff
@@ -111,9 +110,7 @@ ngeo.profileDirective = function(ngeoDebounce) {
           }
         });
 
-      goog.events.listen(window, 'resize',
-        ngeoDebounce(refreshData, 50, true),
-        false, this);
+      window.addEventListener('resize', ngeoDebounce(refreshData, 50, true));
```
fredj: minor but I think we should to stick to `ol.events.*` and jQuery for the events. Could be changed to:
```
ol.events.listen(window, 'resize', geoDebounce(refreshData, 50, true));
```

PR #3095, opened 2017-11-14
Allow injection of gmfObjectEditingManager to fail in Permalink

### LINKED ISSUE #2844: Non working gmf example

pfirpfel:
Fixes #2844 

Not sure if try/catch is a good practice.

### DIFF
--- a/contribs/gmf/src/services/permalink.js
+++ b/contribs/gmf/src/services/permalink.js
@@ -239,12 +239,14 @@ gmf.Permalink = function($q, $timeout, $rootScope, $injector, ngeoDebounce, ngeo
    * @private
    */
   this.gmfObjectEditingManager_;
-  if ($injector.has('gmfObjectEditingManager')) {
-    try {
-      this.gmfObjectEditingManager_ = $injector.get('gmfObjectEditingManager');
-    } catch (e) {
-      this.gmfObjectEditingManager_ = null;
-    }
+  try {
+    // Attempt to inject gmfObjectEditingManager.
+    // This fails if the GMF application is not completely configured, like
+    // in the examples, where there is no gmfLayersUrl defined. Usually in
+    // those cases the gmfObjectEditingManager is not being used.
+    this.gmfObjectEditingManager_ = $injector.get('gmfObjectEditingManager');
+  } catch (e) {
+    this.gmfObjectEditingManager_ = null;
   }
 
   /**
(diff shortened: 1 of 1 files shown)


ger-benjamin:
Mhh the try catch is useless. You use already a "if" statement on the "$injector.has" function.
If it doesn't have the object, you can assign a "null" value in the else.

pfirpfel:
That's not what I observed. The $injector.has returns true regardless of whether the $injector.get will succeed or not.
The injection does not fail directly at the gmf.ObjectEditingManager, it fails upwards in the "injection-chain" in gmf.EditFeature.

pfirpfel:
These are the previous attempts to fix this issue:
#2846
#2873

ger-benjamin:
It's probably not easy, but I think that's not good to **catch** the issue, we should take time to **resolve** the issue (If I understand correctly, in the gmf.EditFeature)

pfirpfel:
I tried that in #2873. But @sbrunner thinks it shoudl be the other way around, it should be resolved from Permalink.

ger-benjamin:
Okay... so I don't know what to do. @fredj ?

fredj:
I've also tried in #2846 but @sbrunner said that it broke the build in docker

sbrunner:
In the permalink the `gmf.ObjectEditingManager` requirement should be removes and replaced by an interface, to be able to call the `getFeature`...

ger-benjamin:
It seems that the removal of "goog.require('gmf.ObjectEditingManager');" in the permalink is enough.

pfirpfel:
This apparently [breaks a docker build](https://github.com/camptocamp/ngeo/pull/2846#issuecomment-325644123).

sbrunner:
Probably because he in't replaced by an interface...

pfirpfel:
#3157 removes the gmfObjectEditingManager require

pfirpfel:
#3157 fixes this issue

PR #3098, opened 2017-11-15
Add ngeo.CustomEvent to rule them all

fredj:
work in progress


romanzoller:
I think this PR adds the following error for `make dist`, but I have no idea why...

```
info ol Parsing dependencies
info ol Compiling 545 sources
ERR! compile /tmp/exports1171020-27155-aii70r.7wseomvx6r.js:12395: ERROR - Property feature never defined on ol.source.Vector.Event.prototype
ERR! compile     ol.source.Vector.Event.prototype.feature);
ERR! compile                                      ^^^^^^^
ERR! compile 
ERR! compile 1 error(s), 0 warning(s), 97.3% typed
ERR! compile 
```

PR #3099, opened 2017-11-15
Change eslint object-shorthand check

fredj:
Don't force the shorthand syntax but ensures that either all shorthand or all longform will be used in an object literal.


pfirpfel:
This breaks the linting of a lot of files.

PR #3101, opened 2017-11-15
Replace goog.fx in sortable directive

pfirpfel:
~contribs/gmf/src/controllers/abstractmobile.js~ (I'm going to do this in a separate branch/pr)
 - [x] src/directives/sortable.js


pfirpfel:
To test sortable.js:
- layerorder example
- desktop app, use the demo theme

I tested debug and compile on Chrome and Firefox (desktop and mobile). Needs more testing on IE, Edge and Safari (especially mobile)

third-party/jquery-ui/jquery.ui.touch-punch.js
```diff
@@ -0,0 +1,180 @@
+/*!
```
sbrunner: From where does this file comes from?

pfirpfel: http://touchpunch.furf.com/
Jquery-ui does not support touch events. Without this patch it won't work on mobile devices.

sbrunner: Any reason to not using the npm package?
https://www.npmjs.com/package/jquery-ui-touch-punch

pfirpfel: My reasoning was to put it to the same place as jquery-ui, since they belong together. We don't use an npm package for jquery-ui either.
The touchpunch code hasn't changed in the last four years, so I don't expect updates.

sbrunner: The reason to don't use the `jquery-ui` npm it to have a custom build that's not possible with the npm package...

pfirpfel: Changed it.

Makefile
```diff
@@ -544,8 +550,8 @@ node_modules/angular/angular.min.js: .build/node_modules.timestamp
 		-e 's|\.\./node_modules/font-awesome/css/font-awesome.css|lib/font-awesome.min.css|' \
 		-e 's|\.\./node_modules/jquery/dist/jquery.js|lib/jquery.min.js|' \
 		-e 's|\.\./third-party/jquery-ui/jquery-ui.min\.js|lib/jquery-ui.min.js|' \
+		-e 's|\.\./node_modules/jquery-ui-touch-punch/jquery.ui.touch-punch.min.js|lib/jquery.ui.touch-punch.min.js|' \
 		-e 's|\.\./third-party/jquery-ui/jquery-ui.min\.css|lib/jquery-ui.min.css|' \
-		-e 's|\.\./node_modules/bootstrap/dist/js/bootstrap.js|lib/bootstrap.min.js|' \
```
fredj: is the removal of `bootstrap.js`wanted?

pfirpfel: Oops, good catch.

fredj:
In comparaison to https://camptocamp.github.io/ngeo/master/examples/contribs/gmf/apps/desktop:
* When dragging an open layertree item, it must be (temporary) closed.
* When moving an item, there must be a "shadow" box (dotted box)

src/directives/sortable.js
```diff
@@ -71,117 +69,69 @@ ngeo.sortableDirective = function($timeout) {
...
+        // underlying data.
         const children = element.children();
         for (let i = 0; i < children.length; ++i) {
           angular.element(children[i]).data('idx', i);
         }
 
-        if (dragListGroup !== null) {
-          dragListGroup.dispose();
-        }
+        const sortableElement = $(element[0]);
```
fredj: minor: index not need, it can be `const sortableElement = $(element);`

pfirpfel:
I'm looking into this.

>     When dragging an open layertree item, it must be (temporary) closed.
I never noticed that. I'm going to check where this happened before.


>     When moving an item, there must be a "shadow" box (dotted box)

Should be simple, see: http://jqueryui.com/sortable/#placeholder

fredj:
Please also add this change:
```diff
diff --git a/contribs/gmf/less/desktoplayertree.less b/contribs/gmf/less/desktoplayertree.less
index 85077d1..da6741e 100644
--- a/contribs/gmf/less/desktoplayertree.less
+++ b/contribs/gmf/less/desktoplayertree.less
@@ -8,7 +8,6 @@ gmf-layertree {
   width: 100%;
   ul {
     margin-bottom: 0;
-    height: 100%;
   }
   > :first-child {
     margin-right: @half-app-margin;
```
It was a workaround for a limitation in goog.fx

fredj:
And in addition to limit the axis to `y`, what about adding the `containment: 'parent'` option?

pfirpfel:
Axis and containment don't really complement each other. They do, in regards to the layertree, very similar things. 

Containment has the issue that the layertree parent element is very "tight" and leaves no tolerance. This can make replacing the top layer impossible if you grab the layer handle at the top:

containment-issue

Axis has the issue, that you can drag the layer (but not drop) to useless areas above and below the tree:

axis-issue


But using neither is worse, because you can get the layertree tab to scroll:

neither-issue



I suggest we use the "axis: 'y'"-option. Replacing the top layer comes up often enough.

fredj:
yes, go ahead with `axis: 'y'` only

pfirpfel:
I'm stuck with a CSS issue.

My goal is to collapse the layer group when it gets dragged, like it does with goog.fx.

The rules from bootstrap for the regular collapsing(-button) look like this:

```css
.collapse {
  display: none;

  &.in      { display: block; }
  tr&.in    { display: table-row; }
  tbody&.in { display: table-row-group; }
}
```

This class gets added to the ul-element inside the li of the layer group.
To be more specfic than this rule, I'm currently on this:

```css
.ui-sortable .gmf-layertree-node.gmf-layertree-dragger {
  ul.collapse.in {
    display: none;
  }
}
```

see: https://github.com/pfirpfel/ngeo/blob/replace-goog-fx-collapse/contribs/gmf/less/desktoplayertree.less#L112

In Chrome and Firefox it wins against the collapse.in-class, but the effect is different.
The display:none of collapse hides the element, the ul.collapse.in of mine leaves the empty space there.

I think the issue is, that the drag and drop API (HTML5) creates a picture that gets dragged around. So it isn't the actual DOM element. Or at least a ill-timed snapshot of it.

As an experiment I removed the .in class (would set display:block) from the element during drag via the drag-event-handlers. This had the same effect as the CSS stuff above.

I think I'll leave it as it is. The collapsing is only really necessary on a long list of layergroups.

gberaudo:
Yes @pfirpfel, we will improve in a second step if it is too disturbing with the new way.

src/directives/sortable.js
```diff
@@ -1,15 +1,14 @@
 goog.provide('ngeo.SortableOptions');
```
fredj: this line can be removed

PR #3103, opened 2017-11-16
Fix autorotate

llienher:
Use native javascript to fetch gyroscope data from mobile devices instead of ol.DeviceOrientation

TODO:
- [x] Fetch data from native APIs
- [x] Adapt the code
- [x] Remove old code (use of ol.DeviceOrientation)

### DIFF
--- a/src/directives/mobilegeolocation.js
+++ b/src/directives/mobilegeolocation.js
@@ -315,9 +315,8 @@ ngeo.MobileGeolocationController.prototype.autorotateListener = function() {
  * @private
  */
 ngeo.MobileGeolocationController.prototype.handleRotate_ = function(eventAlpha, currentAlpha) {
-  const alpha = eventAlpha;
-  if (Math.abs(alpha - currentAlpha) > 0.2) {
-    currentAlpha = alpha;
+  if (Math.abs(eventAlpha - currentAlpha) > 0.2) {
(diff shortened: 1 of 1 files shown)


src/directives/mobilegeolocation.js
```diff
@@ -334,11 +334,10 @@ ngeo.MobileGeolocationController.prototype.autorotateListener = function() {
...
  * @param {number} currentAlpha .
- * @return {number} currentAlpha .
+ * @return {number|undefined} .
  * @private
  */
 ngeo.MobileGeolocationController.prototype.handleRotate_ = function(eventAlpha, currentAlpha) {
-  console.log(eventAlpha, currentAlpha);
   const alpha = eventAlpha;
 
   if (alpha === undefined || alpha === null) {
```
sbrunner: apha is declared as a number than he can't be null or undefined ...

src/directives/mobilegeolocation.js
```diff
@@ -334,11 +334,10 @@ ngeo.MobileGeolocationController.prototype.autorotateListener = function() {
  * Handle rotation.
  * @param {number} eventAlpha .
  * @param {number} currentAlpha .
- * @return {number} currentAlpha .
+ * @return {number|undefined} .
```
sbrunner: For me it should be just {number}

PR #3105, opened 2017-11-16
Move scale selector options to the options file

ger-benjamin:
For GSGMF-141

### DIFF
--- a/options/ngeox.js
+++ b/options/ngeox.js
@@ -834,6 +834,22 @@ ngeox.QuerySource.prototype.wfsQuery;
 
 
 /**
+ * Options to configure the scale selector
+ * @typedef {{
+ *     dropup: (boolean|undefined)
+ * }}
+ */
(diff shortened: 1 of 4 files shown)


options/ngeox.js
```diff
@@ -834,6 +834,22 @@ ngeox.QuerySource.prototype.wfsQuery;
...
+ * }}
+ */
+ngeox.ScaleselectorOptions;
+
+
+/**
+ * True to get a drop menu that opens imself to the top, instead of the bottom.
+ * @type {boolean|undefined}
+ */
+ngeox.ScaleselectorOptions.prototype.dropup;
```
fredj: not sure that this line is needed (ping @gberaudo)

ger-benjamin: That allow to have a documentation on each options, like https://camptocamp.github.io/ngeo/master/apidoc/ngeox.html#QueryableDataSources

PR #3112, opened 2017-11-16
Simplify goog.provide in gmf apps

romanzoller:
And adapt entry_point value in Makefile.
See https://jira.camptocamp.com/browse/GSGMF-144


sbrunner:
This change should also be on c2cgeoportal, then I'n nor sure that's a simplification...

sbrunner:
Why not renaming `app.DesktopController` to `app.Controllerdesktop` to don't requires the `GMF_APP_ENTRY_POINT_*` variables?

fredj:
Instead of providing a namespace to `entry_point`, a file can be used:
https://github.com/openlayers/closure-util/blob/f0255279454c3965c6e66bd2f70afb4b1df0e34c/compiler-options.txt#L144-L152

(untested)

romanzoller:
Yes, it seems like @fredj's suggestion works, thx!

romanzoller:
Hmm, unfortunately doesn't work. I didn't find any documentation about it, but I suspect that entry_point only accepts a filename if goog.provide/require is not used.

sbrunner:
:-(

gberaudo:
Superseded by #3233.

PR #3116, opened 2017-11-17
Allow text in gmf-displayquerywindow to be selectable

adube:
Allow text in the gmf-displayquerywindow directive to be selectable.

The jQuery UI draggable widget was preventing text to be selected using a drag of the mouse.  Instead, it dragged the window.  To fix this, we use the `cancel` option of the widget, see: http://api.jqueryui.com/draggable/#option-cancel.  This prevents the dragging to occur for any defined selector.

Since the gmf-displayquerywindow always shows its content in a table, the `tr` is the chosen selector to prevent the selection. The rest of the selector are the default values that I thought made sense to keep.

While mouse hovering a `tr`, the CSS cursor changes to `text`.

### DIFF
--- a/contribs/gmf/less/displayquerywindow.less
+++ b/contribs/gmf/less/displayquerywindow.less
@@ -172,6 +172,11 @@
       font-family: FontAwesome;
     }
   }
+  &.ui-draggable {
+    tr {
+      cursor: text;
+    }
+  }
 }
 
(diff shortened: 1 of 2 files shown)


ybolognini:
If this looks good to @fredj then I think we can merge (so I can test and validate)

fredj:
@ybolognini merged

Issue #3085, opened 2017-11-13
Print a wms using IE does not work


rmichaelis:
In our projects we have to support IE11 and higher.
But ngeo uses unsupported javascripts : here : https://github.com/camptocamp/ngeo/blob/master/src/services/print.js#L283

Is ngeo supposed to still support Internet Explorer ?

gberaudo:
Hi @rmichaelis, the closure compiler should handle the transpilation.
Are you experimenting a specific issue in production mode?

For debug mode you are right, it is not supported, one should rely on sourcemaps.

rmichaelis:
Hi @gberaudo, Yes it happens in our production environment.  
If you try to print the following map : http://g-o.lu/3/NL4I you get the following error : `TypeError: Cet objet ne gère pas cette action
   at yB (http://map.geoportail.lu/main/wsgi/proj/728c826edeb44bdeb1054284e4e45399/build/build.js:2131:22)
   at xB (http://map.geoportail.lu/main/wsgi/proj/728c826edeb44bdeb1054284e4e45399/build/build.js:2124:100)
   at Anonymous function (http://map.geoportail.lu/main/wsgi/proj/728c826edeb44bdeb1054284e4e45399/build/build.js:2123:245)
   at wB (http://map.geoportail.lu/main/wsgi/proj/728c826edeb44bdeb1054284e4e45399/build/build.js:2123:216)
   at vB.prototype.rh (http://map.geoportail.lu/main/wsgi/proj/728c826edeb44bdeb1054284e4e45399/build/build.js:2122:467)
   at Anonymous function (http://map.geoportail.lu/main/wsgi/proj/728c826edeb44bdeb1054284e4e45399/build/build.js:2148:493)
   at Anonymous function (http://map.geoportail.lu/main/wsgi/proj/728c826edeb44bdeb1054284e4e45399/build/build.js:181:450)
   at m.prototype.$digest (http://map.geoportail.lu/main/wsgi/proj/728c826edeb44bdeb1054284e4e45399/build/build.js:192:417)
   at m.prototype.$apply (http://map.geoportail.lu/main/wsgi/proj/728c826edeb44bdeb1054284e4e45399/build/build.js:196:111)
[...]

PR #3123, opened 2017-11-20
Prepend protocol to WMS url

gberaudo:
The URL constructor does not support protocol-relative URLs.


gberaudo:
Related to #3085. @sbrunner, maybe it should be backported in 2.2?

PR #3124, opened 2017-11-20
stop using ol3 enums for events

fredj:
https://jira.camptocamp.com/browse/GSGMF-149

### DIFF
--- a/contribs/gmf/examples/profile.js
+++ b/contribs/gmf/examples/profile.js
@@ -103,11 +103,10 @@ gmfapp.MainController = function($scope, ngeoFeatureOverlayMgr) {
    * @type {ol.interaction.Draw}
    * @export
    */
-  this.drawLine = new ol.interaction.Draw(
-    /** @type {olx.interaction.DrawOptions} */ ({
-      type: 'LineString',
-      features: features
-    }));
+  this.drawLine = new ol.interaction.Draw({
+    type: /** @type {ol.geom.GeometryType} */ ('LineString'),
+    features: features
+  });
 
   this.drawLine.setActive(false);
   this.map.addInteraction(this.drawLine);
--- a/contribs/gmf/src/datasource/datasourcesmanager.js
+++ b/contribs/gmf/src/datasource/datasourcesmanager.js
@@ -156,8 +156,7 @@ gmf.datasource.DataSourcesManager = class {
       this.handleNgeoBackgroundLayerChange_,
       this
     );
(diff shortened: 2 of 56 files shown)


adube:
For my own knowledge: why is this being changed?  Isn't it lighter to use the constants, which are renamed / minimized in when compiling, resulting in having a lighter minimized file?

gberaudo:
@adube, we are trying to simplify the library and get closer to standard ecmascript.
People are used to have plain strings as event names.

Regarding the generated size, it is a good question. @fredj do you have figures before and after?
If there is a significant difference we could get in contact with the compiler guys to see if the compiler can be improved.

adube:
>@adube, we are trying to simplify the library and get closer to standard ecmascript.
People are used to have plain strings as event names.

It's somewhat the opposite for me.  When I'm looking for an event, I know how to easilly find it using the constant.  With the string, you have to know exactly what its name is.  Also, for OpenLayers, you have to know the `change:resolution` syntax, or is it `resolution:change`?  I don't know, but I know that those are pretty long and that in the end it gets minimized a lot instead of taking lots of characters, multiple times.

For the record, also, there are lots of event names that you are changing that are not standard to ecmascript, i.e. they have names unique to OpenLayers and also have some special syntax (`change:x`).  I'm not against this change... my mood is more like: `Aww... really?`

>Regarding the generated size, it is a good question. @fredj do you have figures before and after?
If there is a significant difference we could get in contact with the compiler guys to see if the compiler can be improved.

Every byte counts, in my opinion. :)

fredj:
One of the reason of using strings is that we don't have to change the code in ngeo when the internal implementation of the constants changes in OpenLayers (it happened at least twice)

adube:
>One of the reason of using strings is that we don't have to change the code in ngeo when the internal implementation of the constants changes in OpenLayers (it happened at least twice)

Good point.  But, it's fairly easy to search/replace constants.  Imagine if they decided to change the name of the events themselves.  That could happen too and it would be much harder to find/replace them all.  Something to think about...

fredj:
>  Imagine if they decided to change the name of the events themselves.

not on my watch :-)

adube:
>not on my watch :-)

That's great!  (and that made me smile a lot).  I'm not longer anxious about this change, then :)  Thanks a lot.

fredj:
I should have said: "not on my `$scope.$watch`" but it too late now ...

adube:
> I should have said: "not on my $scope.$watch" but it too late now ...

Priceless!! That made me laugh, even cry XD  .... so no, it wasn't too late.

PR #3126, opened 2017-11-20
Fix typeahead selection in ImportDataSource component

adube:
This PR fixes the following issue: in the DataSourceBrowser component, the "connect" button was sometimes disabled after the selection of an item.

This was due to the fact that the field itself is of `url` type and there's an angular rule applied on the connect button that automatically disables it if the form is invalid.  The selection using the drop-down list uses typeahead. Now, I don't know exactly why, but the selected value was not always properly copied in the model, i.e. it was pasted in the input field, but not the model.

This fix ensures that, after the selection of an item among the typeahead selections, the value is properly copied.  At the same time, I do a "focus" on the connect button.

This solution keeps the `url` type on the field, because I think it's very useful to ensure that any url either selected or manually typed are valid urls.

### DIFF
--- a/contribs/gmf/src/directives/importdatasource.js
+++ b/contribs/gmf/src/directives/importdatasource.js
@@ -226,6 +226,8 @@ gmf.ImportdatasourceController = class {
       this.timeout_(() => {
         goog.asserts.assert(this.serversEngine_);
         const $urlInput = this.element_.find('input[name=url]');
+        const $connectBtn = this.element_.find(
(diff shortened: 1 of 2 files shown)


fredj:
I've fixed the remaining travis issue by using `jQuery` type instead of `angular.JQLite`

fredj:
And because we have jquery loaded,  `angular.JQLite` is just an alias to `jQuery`:
https://docs.angularjs.org/api/ng/function/angular.element#overview

PR #1375, opened 2016-06-10
Drop dashed zone displayed even if dragging far down

pgiraud:
Fixes #1347.
Pure CSS alternative to #1367.

PR #3127, opened 2017-11-20
Show imported layers first in layertree, remove 100% height on ul

adube:
This PR changes how the external layers are added in the layer tree:

 * they are shown first (were last)
 * they are shown "as-if" they were part of the other layers, i.e. the gap between the 2 has been removed *

* Important note: the `height: 100%;` was removed to allow this. It was introduced in #1375.  I played with drag-dropping items and it didn't seem to affect the UI.  Please, make sure this PR didn't break #1375.

### DIFF
--- a/contribs/gmf/less/desktoplayertree.less
+++ b/contribs/gmf/less/desktoplayertree.less
@@ -8,7 +8,6 @@ gmf-layertree {
   width: 100%;
   ul {
     margin-bottom: 0;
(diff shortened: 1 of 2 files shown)


fredj:
The `height: 100%;` removal was suggested in https://github.com/camptocamp/ngeo/pull/3101#issuecomment-345189293 so it's fine to remove it in this PR

PR #3134, opened 2017-11-21
Don't change all the twitter-typeahead element in the page

fredj:
But only the children of the current search node

https://jira.camptocamp.com/browse/GSGMF-158


adube:
That makes a lot of sense.

+1

PR #3136, opened 2017-11-21
Directive to component for examples

apastuhow:
https://jira.camptocamp.com/browse/GSGMF-151

### DIFF
--- a/examples/backgroundlayer.js
+++ b/examples/backgroundlayer.js
@@ -18,33 +18,29 @@ app.module = angular.module('app', ['ngeo']);
 
 
 /**
- * The application-specific background layer directive.
+ * The application-specific background layer component.
  *
- * The directive is based on Angular's select, ngOptions, ngModel, and
- * ngChange directives. ngChange is used to avoid adding a watcher on
+ * The component is based on Angular's select, ngOptions, ngModel, and
+ * ngChange components. ngChange is used to avoid adding a watcher on
  * the ngModel expression.
  *
  * Note: we don't need two-way binding for ngModel here, but using ::
  * for the ngModel expression doesn't actually make a difference. This
  * is because ngModel doesn't actually watch the ngModel expression.
  *
- * @return {angular.Directive} Directive Definition Object.
- * @ngInject
+ * @type {!angular.Component}
  */
-app.backgroundlayerDirective = function() {
-  return {
-    restrict: 'E',
-    scope: {
-      'map': '=appBackgroundlayerMap'
-    },
-    templateUrl: 'partials/backgroundlayer.html',
-    bindToController: true,
-    controller: 'AppBackgroundlayerController as ctrl'
-  };
+app.backgroundlayerComponent = {
+  bindings: {
+    'map': '=appBackgroundlayerMap'
+  },
+  templateUrl: 'partials/backgroundlayer.html',
+  controller: 'AppBackgroundlayerController',
+  controllerAs: 'ctrl'
 };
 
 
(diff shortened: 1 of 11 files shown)


examples/backgroundlayer.js
```diff
@@ -18,33 +18,29 @@ app.module = angular.module('app', ['ngeo']);
...
-    bindToController: true,
-    controller: 'AppBackgroundlayerController as ctrl'
-  };
+app.backgroundlayerComponent = {
+  bindings: {
+    'map': '=appBackgroundlayerMap'
+  },
+  templateUrl: 'partials/backgroundlayer.html',
+  controller: 'AppBackgroundlayerController',
+  controllerAs: 'ctrl'
```
ger-benjamin: Can't we remove that ? Then use "$ctrl" in the partial ?

examples/backgroundlayerdropdown.js
```diff
@@ -18,28 +18,24 @@ app.module = angular.module('app', ['ngeo']);
...
-    bindToController: true,
-    controller: 'AppBackgroundlayerController as ctrl'
-  };
+app.backgroundlayerComponent = {
+  bindings: {
+    'map': '=appBackgroundlayerMap'
+  },
+  templateUrl: 'partials/backgroundlayerdropdown.html',
+  controller: 'AppBackgroundlayerController',
+  controllerAs: 'ctrl'
```
ger-benjamin: Same

examples/bboxquery.js
```diff
@@ -28,22 +28,17 @@ app.module.value('ngeoQueryOptions', {
...
-  return {
-    restrict: 'E',
-    scope: {},
-    controller: 'AppQueryresultController as qrCtrl',
-    bindToController: true,
-    templateUrl: 'partials/queryresult.html'
-  };
+app.queryresultComponent = {
+  controller: 'AppQueryresultController',
+  controllerAs: 'qrCtrl',
```
ger-benjamin: same

examples/search.js
```diff
@@ -20,49 +20,41 @@ app.module = angular.module('app', [ngeo.module.name]);
...
  *     create GeoJSON Bloodhound service.
  * @ngInject
  */
-app.SearchController = function($rootScope, $compile, ngeoSearchCreateGeoJSONBloodhound) {
+app.SearchController = function($element, $rootScope, $compile, ngeoSearchCreateGeoJSONBloodhound) {
+  /**
+   * @private
+   * @type {angular.JQLite}
+   */
+  this.$element = $element;
```
ger-benjamin: add an underscore if it's private:
this.$element_ = $element;

examples/backgroundlayer.js
apastuhov: yes, but then we need to update partials, I am not sure that it is critical and/or should be done in that PR

examples/search.js
apastuhov: sure, I will fix it

examples/backgroundlayer.js
ger-benjamin: That's not critical. That's why we'll probably never do that if we don't do that right now.
But it's not really important. If you think it takes too much time to do the change (and test) leave it like that :-)

fredj: it's fine to leave as it for the examples but it would be nice to do it the all the ngeo and gmf components (in the others pull requests).

apastuhow: I will do it as another PR on monday

PR #3144, opened 2017-11-22
Change seperator in csv download

julsbreakdown:
In order for the csv file to be opened directly in both excel & open/libre office. 
Changing the separator from ',' to ';' makes sense, especially in Europe

### DIFF
--- a/src/services/csvdownload.js
+++ b/src/services/csvdownload.js
@@ -62,7 +62,7 @@ ngeo.CsvDownload = function($injector, gettextCatalog) {
    * @private
    */
   this.separator_ = $injector.has('ngeoCsvSeparator') ?
-    $injector.get('ngeoCsvSeparator') : ',';
+    $injector.get('ngeoCsvSeparator') : ';';
 
   /**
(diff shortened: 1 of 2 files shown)


src/services/csvdownload.js
```diff
@@ -62,7 +62,7 @@ ngeo.CsvDownload = function($injector, gettextCatalog) {
    * @private
    */
   this.separator_ = $injector.has('ngeoCsvSeparator') ?
-    $injector.get('ngeoCsvSeparator') : ',';
+    $injector.get('ngeoCsvSeparator') : ';';
```
sbrunner: We shouldn't change the default value...

sbrunner:
The coma will open well with LibreOffice ...

julsbreakdown:
But not in excel

sbrunner:
By default we will support the real standards, and if the administrator want to change it it should be possible, actually it seem to be possible.

julsbreakdown:
Ok thanks for the info, i will close =)

PR #3146, opened 2017-11-22
Enable to configurate the default display of measurements

julsbreakdown:
Adds a showMeasure variable that can be configurated in the template to check by default the display measurement check box.

`                  gmf-drawfeature-showmeasure="true"
`


src/directives/drawfeature.js
```diff
@@ -97,7 +97,8 @@ ngeo.drawfeatureDirective = function() {
     bindToController: {
       'active': '=ngeoDrawfeatureActive',
       'features': '=?ngeoDrawfeatureFeatures',
-      'map': '=ngeoDrawfeatureMap'
+      'map': '=ngeoDrawfeatureMap',
+      'showMeasure': '=?ngeoDrawfeatureShowmeasure'
```
fredj: please also add a small doc for this option above (`@htmlAttribute`)

PR #3154, opened 2017-11-22
Move modal to component and ES6

apastuhov:


src/directives/modal.js
```diff
@@ -31,91 +31,159 @@ goog.require('ngeo');
...
-            modal.find('.modal-content').append(clone);
-          }
-        });
+  /**
+   * @private
+   */
+  onShow() {
+    this.childScope = this.$scope.$new();
+    // * @param {!function(!angular.Scope=, !function(Element)=)=} transcludeFn
+    // function((angular.Scope|null)=, function(angular.Scope, Element): ?=, (Element|null)=, string=): ?
```
pfirpfel: `// function((angular.Scope|null)=, function(angular.Scope, Element): ?=, (Element|null)=, string=): ?`

I'm a bit confused by this annotation.

pfirpfel:
There are a bunch of files that still reference **ngeo.modalDirective**:

./examples/modal.html
./examples/modal.js
./contribs/gmf/src/directives/filterselector.js
./contribs/gmf/src/directives/editfeature.js
./contribs/gmf/src/directives/authenticationdirective.js
./contribs/gmf/examples/share.js

pfirpfel:
The example is broken:
./examples/modal.html
./examples/modal.js

It just doesn't work, there is no error. I have not look at the example too thoroughly.

src/directives/modal.js
apastuhov: It was an old annotation, I just forgot to remove it. Sorry for that)

apastuhov:
@pfirpfel you are right.. I also did not found high-level problem. I will continue debugging on monday

src/directives/modal.js
```diff
@@ -31,91 +31,161 @@ goog.require('ngeo');
...
-    template: '<div class="modal fade" tabindex="-1" role="dialog">' +
-        '<div class="modal-dialog">' +
-        '<div class="modal-content">' +
-        '</div>' +
-        '</div>' +
-        '</div>',
-    restrict: 'E',
-    require: 'ngModel',
-    transclude: true,
+ngeo.modalDirective;
```
fredj: namespace creation not needed

apastuhow: I keep it to have a doc on top

PR #3155, opened 2017-11-23
GMF Print: Move from directive to component

ger-benjamin:
For GSGMF-155

Work well on example, and apps

### DIFF
--- a/contribs/gmf/apps/desktop_alt/index.html
+++ b/contribs/gmf/apps/desktop_alt/index.html
@@ -110,9 +110,9 @@
                 <a class="btn close" ng-click="mainCtrl.printPanelActive = false">&times;</a>
               </div>
               <gmf-print
-                gmf-print-map="mainCtrl.map"
+                gmf-print-map="::mainCtrl.map"
                 gmf-print-active="mainCtrl.printActive"
-                gmf-print-rotatemask="true">
+                gmf-print-rotatemask="::true">
               </gmf-print>
             </div>
           </div>
--- a/contribs/gmf/src/directives/print.js
+++ b/contribs/gmf/src/directives/print.js
@@ -1,4 +1,4 @@
-goog.provide('gmf.printDirective');
+goog.provide('gmf.printComponent');
 
 goog.require('gmf');
 /** @suppress {extraRequire} */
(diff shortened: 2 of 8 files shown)


contribs/gmf/apps/desktop_alt/index.html
```diff
@@ -110,9 +110,9 @@
                 <a class="btn close" ng-click="mainCtrl.printPanelActive = false">&times;</a>
               </div>
               <gmf-print
-                gmf-print-map="mainCtrl.map"
+                gmf-print-map="::mainCtrl.map"
```
apastuhow: no real need.. `gmf-print` should use one-way binding

contribs/gmf/src/directives/print.js
```diff
@@ -365,84 +377,95 @@ gmf.PrintController = function($rootScope, $scope, $timeout, $q, $injector,
...
    */
-  this.capabilities_;
+  this.currentThemes_;
+};
 
+
+/**
+ * Init the controller
+ */
+gmf.PrintController.prototype.$onInit = function() {
```
apastuhow: you can use classes

contribs/gmf/src/directives/print.js
```diff
@@ -365,84 +377,95 @@ gmf.PrintController = function($rootScope, $scope, $timeout, $q, $injector,
...
+  const getSizeFn = () => this.paperSize_;
 
   /**
-   * @type {Array.<gmfThemes.GmfTheme>}
-   * @private
+   * @param {olx.FrameState} frameState Frame state.
+   * @return {number} Scale of the map to print.
    */
-  this.currentThemes_;
+  const getScaleFn = (frameState) => {
```
apastuhow: why not a method? it seems a bit complex arrow-function

contribs/gmf/apps/desktop_alt/index.html
ger-benjamin: That's not the same.
'<' stand for "one-WAY" bindings.
`&` (in js) or `::` (in html) stand for "one-TIME" binding.
I think the more important is the one-time binding for the map and with FredJ, we find that it's more practical to use the html notation.

contribs/gmf/src/directives/print.js
ger-benjamin: I think I should do that in another PR (that already a big one and other component still use this notation)

contribs/gmf/src/directives/print.js
ger-benjamin: I agree, but I'm not here tomorrow, if I don't merge now, I'll merge Monday :-/

contribs/gmf/apps/desktop_alt/index.html
apastuhow: I told to use `one-way` because you are passing a map object, so profit is not so big.

As for me it makes sense only for some data that will never change like captions and(or) calculated expressions

https://docs.angularjs.org/guide/expression#reasons-for-using-one-time-binding

PR #3157, opened 2017-11-23
Don't need to add requires for type declaration

sbrunner:


gberaudo:
@sbrunner, I think there are other places with conditional import. See
```
git grep '$injector.has'
```
Are you going to also handle these?

Some time ago we set up some interfaces for datasource / rules which were also to workaround this behaviour of GCC. Are you willing to get rid of this too?

buildtools/mako_build.json
```diff
@@ -74,6 +74,18 @@ _ngeo_folder = '' if ngeo_folder is UNDEFINED else ngeo_folder
...
+    "${s}",
+% endfor
+% endif
+    "${_ngeo_folder}src/**/*.js",
+% if src_set == 'contribs_gmf':
+    "${_ngeo_folder}contribs/gmf/src/**/*.js",
+% endif
+    "node_modules/openlayers/src/**/*.js",
+    "node_modules/openlayers/build/ol.ext/*.js",
+    "node_modules/ol-cesium/src/**/*.js",
```
gberaudo: It is a copy of the src block, right? Can we use an intermediate variable to avoid the copy?

PR #3151, opened 2017-11-22
GMF Search: move from directive to component

ger-benjamin:
For GSGMF-155

Work well on example, and apps

PR #3172, opened 2017-11-27
Improve search and print

ger-benjamin:
Answer to comment in PR #3151 from @apastuhow 
Also for GSGMF-155

- Better bindings
- Use ES6 classes
- Clean

Interesting fact: With ES6 classes, a lot of undetected type errors is now detected.


fredj:
In the controllers, all the properties that are in the component bindings must be `@export` (map, active, ...)

ger-benjamin:
Done in the last commit @fredj

PR #3176, opened 2017-11-28
[WIP] Add 3d permalink

fgravin:
First shot that show a working example of 3d permalink.
Here the params used 
```
3d_enabled=true
3d_lon=23.49043
3d_lat=-40.49055
3d_elevation=993992
3d_heading=360.000
3d_pitch=-42.408
```

- [x] Load camera from permalink
- [x] Enable 3d if `3d_enabled` set in permalink
- [x] Update permalink from camera move
- [x] Add an example
- [x] Remove 3d permalink params when back to 2d
- [x] Update permalink when switch to 3d
- [x] Use ngeoStateManager
- [ ] Some configuration ?? debounce time...

### DIFF
--- a/src/olcs/Service.js
+++ b/src/olcs/Service.js
@@ -1,8 +1,5 @@
 goog.module('ngeo.olcs.Service');
 
-goog.require('ngeo.Debounce');
-goog.require('ngeo.Location');
-
 const Service = class {
 
   /**
@@ -10,8 +7,9 @@ const Service = class {
    * @param {angular.Scope} $rootScope Angular root scope.
    * @param {!ngeo.Debounce} ngeoDebounce ngeo debounce service.
    * @param {ngeo.Location} ngeoLocation ngeo location service.
+   * @param {ngeo.StateManager} ngeoStateManager The ngeo StateManager service.
    */
-  constructor($rootScope, ngeoDebounce, ngeoLocation) {
+  constructor($rootScope, ngeoDebounce, ngeoLocation, ngeoStateManager) {
     /**
      * @private
      * @type {olcs.contrib.Manager|undefined}
@@ -26,15 +24,21 @@ const Service = class {
 
     /**
      * @private
-     * @type {ngeo.Location}
+     * @type {ngeo.Debounce}
      */
     this.ngeoDebounce_ = ngeoDebounce;
 
     /**
      * @private
-     * @type {ngeo.Debounce}
+     * @type {ngeo.Location}
      */
     this.ngeoLocation_ = ngeoLocation;
+
+    /**
+     * @private
+     * @type {ngeo.StateManager}
+     */
+    this.ngeoStateManager_ = ngeoStateManager;
 
   }
 
@@ -49,7 +53,7 @@ const Service = class {
       () => this.manager_.getOl3d(),
       (ol3d) => {
         if (ol3d) {
-          if(this.ngeoLocation_.hasParam('3d_enabled')) {
+          if (this.ngeoStateManager_.getInitialBooleanValue('3d_enabled')) {
             this.cameraFromPermalink_();
           }
           this.setupPermalink_();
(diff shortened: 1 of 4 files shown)


examples/permalink3d.html
```diff
@@ -0,0 +1,47 @@
+<!DOCTYPE html>
```
gberaudo: IMO, there are already too many examples in ngeo. Instead of creating a new one, could you add permalink support to the simple3d example?

src/olcs/Service.js
```diff
@@ -41,6 +81,81 @@ const Service = class {
   getMaxTilt() {
     return 7 * Math.PI / 16;
   }
+
+  /**
+   * @private
+   */
+  cameraFromPermalink_() {
```
gberaudo: I discussed with @sbrunner about how to implement and where to put the code for this new feature.
The whole picture is:

A  `mapState` service should be created with the current code for 2d.
The code for handling 3d should be added in the same file.
The `mapState` deals with reading and writting the 2d view and 3d camera (not other parameters like the layers).
The `mapState` should not know about local storage or permalink.

src/olcs/Service.js
```diff
@@ -15,6 +44,17 @@ const Service = class {
    */
   initialize(manager) {
     this.manager_ = manager;
+
+    const unWatchFn = this.$rootScope.$watch(
+      () => this.manager_.getOl3d(),
```
gberaudo: You should use the promise returned by `manager.load()`.

src/olcs/Service.js
```diff
@@ -41,6 +81,81 @@ const Service = class {
...
+          '3d_lat': Cesium.Math.toDegrees(position.latitude).toFixed(5),
+          '3d_elevation': position.height.toFixed(0),
+          '3d_heading': Cesium.Math.toDegrees(camera.heading).toFixed(3),
+          '3d_pitch': Cesium.Math.toDegrees(camera.pitch).toFixed(3)
+        });
+      }
+    }, 1000, true));
+
+    this.$rootScope.$watch(
+      () => this.manager_.getOl3d().getEnabled(),
```
gberaudo: It looks like we are adding more and more watchers to track the manager state which worries me.
Could you make OLCesium and the Manager send events (OL events) instead?

examples/permalink3d.js
```diff
@@ -0,0 +1,69 @@
...
+   * @type {olcs.contrib.Manager}
+   */
+  this.manager = new ngeo.olcs.Manager(cesiumUrl, {
+    map: this.map
+  });
+  this.manager.setRootScope($rootScope);
+
+  ngeoOlcsService.initialize(this.manager);
+
+  if (ngeoLocation.hasParam('3d_enabled')) {
```
gberaudo: Can this be part of the library?

src/olcs/Service.js
fgravin: If i do this, cesium will be loaded on application bootstrap, cause calling `load()` function from `intialize` will trigger cesium loading, doesn't it ?

gberaudo: Sure, you need to check for the presence of the 3d parameters in the URL first.

examples/permalink3d.html
fgravin: Ok i removed it

examples/permalink3d.js
fgravin: I put it in `olcs.Service.initialize`

src/olcs/Service.js
fgravin: We can discuss it later, but the way you managed the load promise cannot work, we have to use a $watcher for now.

src/olcs/Service.js
fgravin: I pushed a change using `ngeoStateManager` instead of `ngeoLocation`. I think this cover the need for this project.
Shall we start the big refactoring under Luxembourg fundings ?
I think it's a big work and i'm not sure about the whole approach.

> The mapState should not know about local storage or permalink.

What do you mean ?

src/olcs/Service.js
fgravin: Yes we discussed about that as well with @tonio and other things to change, like ol3d load promise.

PR #3178, opened 2017-11-28
Rework olcs Manager

gberaudo:
- simplify service and make it optional;
- simplify example.


gberaudo:
Thanks for the review.

PR #3179, opened 2017-11-28
Move gmf search to it's own module

ger-benjamin:
For GSGMF-150
Do this: https://github.com/camptocamp/ngeo/wiki/ES6-modules-migration


contribs/gmf/src/search/component.html
```diff
@@ -6,7 +6,7 @@
     ngeo-search-datasets="$ctrl.datasets"
     ngeo-search-listeners="$ctrl.listeners">
   <span class="gmf-clear-button ng-hide"
-    ng-hide="!$ctrl.clearButton || ctrl.inputValue == ''"
+    ng-hide="!$ctrl.clearButton || $ctrl.inputValue == ''"
```
gberaudo: Good catch!

contribs/gmf/src/search/component.js
```diff
@@ -226,10 +239,10 @@ gmf.SearchController = class {
     this.gmfTreeManager_ = gmfTreeManager;
 
     /**
-     * @type {gmf.FulltextSearchService}
+     * @type {gmf.search.fulltextSearch.fulltextSearch_}
```
gberaudo: That is not good, you can not depend on a private type defined in a different file. See comment in the service file.

contribs/gmf/src/search/component.js
```diff
@@ -868,7 +881,7 @@ gmf.SearchController = class {
    * @param {jQuery.Event} event Event.
    * @param {Object|ol.Feature} suggestion Suggestion.
    * @param {TypeaheadDataset} dataset Dataset.
-   * @this {gmf.SearchController}
+   * @this {gmf.search.component.SearchController_}
```
gberaudo: Have you tried removing the `@this` annotation and the `static`? Would be good to fix this if that does not take too much time.

contribs/gmf/src/search/fulltextsearch.js
```diff
@@ -1,8 +1,15 @@
...
+goog.provide('gmf.search.fulltextSearch');
 
 goog.require('gmf');
 goog.require('ngeo.utils');
 
+
+/**
+ * @type {!angular.Module}
+ */
+gmf.search.fulltextSearch = angular.module('gmfSearchFulltextSearch', []);
```
gberaudo: See `src/utils/File.js` for how to do with services.

PR #3182, opened 2017-11-29
Add url polyfill

### LINKED ISSUE #3085: Print a wms using IE does not work

llienher:
Fixes #3085 

Should allow Url polyfill to be used in the projects.


Makefile
```diff
@@ -69,6 +70,7 @@ GMF_APPS_LIBS_JS_FILES += \
 	node_modules/corejs-typeahead/dist/typeahead.bundle.min.js \
 	node_modules/jsts/dist/jsts.min.js \
 	node_modules/moment/min/moment.min.js \
+	node_modules/url-polyfill/url-polyfill.js \
```
sbrunner: use min file ...

llienher:
@sbrunner I made the change for minified file, is everything ok now ?

PR #3189, opened 2017-11-30
gmf-displaywindow - drag/resize on same div

adube:
When the `gmf-displaywindow` was made smaller using the jQueryUI resizable widget, the parent div was keeping the original size.  That parent div had the jQueryUI draggable widget on it, which allows the window to be dragged even when seemed to be dragging outside the window.

To fix this, both widgets now interact with the same div: the first div within the widget (at the root).


adube:
@fredj After trying to change the displayquerywindow as well, I was going with the 'parent div' route and it did not work.  It was working for the displaywindow.

So, I changed to your solution (used the container instead) and it worked for both.

Ready for review.

adube:
@fredj I won't be able to unless you accept the PR :)

PR #3190, opened 2017-11-30
Text constraint filter expression

### LINKED ISSUE #3186: Filters: 'contains' rule give incomplete results
The 'contains' rule, in filters, in example for a name attribute on any filterable layer will work for the searched word and the following ones, but not what is before the said word. This issue is not present in 1.6, but in 2.2. In 1.6, the query is done on \*word\*.

Example: "Hôtel de ville"
Search "Hôtel": 
- The query will be done on hôtel*
- That means "Hôtel de ville" is going to be displayed. 

Search "ville":
- The query is done on ville*
- That means nothing is displayed as "ville" is not the first word.

llienher:
Fixes #3186

This PR allow the 'contains' filter to fetch result with pattern that are not only starting a attribute value, but also be in the middle of a word, or string, or at the end.

Example:
"med" will not give us only with result "medbase" but now the following results are valid as well:
"Bergrestaurant Jatzmeder", "Dr. med. dent. V. Nielsen", etc

Escaping works partially. Character '.' are working while the characters '!', '*' is not giving us the wanted result, but the issue is probably from MapServer does interpret it anyways as part of the expression.

### DIFF
--- a/src/services/rulehelper.js
+++ b/src/services/rulehelper.js
@@ -622,10 +622,13 @@ ngeo.RuleHelper = class {
         expression
       );
     } else if (operator === rot.LIKE) {
-      const stringExpression = String(expression);
+      const stringExpression = String(expression)
+        .replace(/!/g, '!!')
+        .replace(/\./g, '!.')
+        .replace(/\*/g, '!*');
       filter = ol.format.filter.like(
         propertyName,
-        stringExpression
+        `*${stringExpression}*`
       );
     } else if (operator === rot.NOT_EQUAL_TO) {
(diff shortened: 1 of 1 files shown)


llienher:
@adube Could you take a look at the code ? Do you see any inconvenient to implement this fix this way ? Or any issue that would occurs somewhere else in the filters from this fix ?

adube:
@llienher 

 * Code is clean
 * I don't see any inconvenience
 * It should not affect anything else in the filters

llienher:
@adube thanks for the quick answer!

PR #3191, opened 2017-11-30
Support of WFS attributes aliases

asaunier:
Some WFS backends such as QGIS Server may provide ``alias`` attributes for results of WFS requests. They are made available through a WFS DescribeFeatureType request. For instance:
```
...
<element alias="Vitesse" type="integer" name="Speed"/>
...
```

This PR aims to display those aliases if any in the query results window. See  https://jira.camptocamp.com/browse/GEORD-21

Still to do:
* [x] get the aliases if any when showing the query results
* ~~support layers added using the WMS browser~~
* [ ] figure out if the code in the ``gmf.WFSAliases`` service really deserves a distinct service or if it can be moved for instance to ``gmf.datasource.DataSourcesManager``
* ~~make sure WMS GFI are supported~~
* [ ] tests?

### DIFF
--- a/contribs/gmf/src/datasource/datasourcesmanager.js
+++ b/contribs/gmf/src/datasource/datasourcesmanager.js
@@ -4,6 +4,7 @@ goog.require('gmf');
 goog.require('gmf.datasource.OGC');
 goog.require('gmf.SyncLayertreeMap');
 goog.require('gmf.TreeManager');
+goog.require('gmf.WFSAliases');
 goog.require('ngeo.map.BackgroundLayerMgr');
 /** @suppress {extraRequire} */
 goog.require('ngeo.datasource.DataSources');
@@ -39,13 +40,14 @@ gmf.datasource.DataSourcesManager = class {
    * @param {!ngeo.map.LayerHelper} ngeoLayerHelper Ngeo Layer Helper.
    * @param {!ngeo.RuleHelper} ngeoRuleHelper Ngeo rule helper service.
    * @param {!ngeo.WMSTime} ngeoWMSTime wms time service.
+   * @param {!gmf.WFSAliases} gmfWFSAliases Gmf WFS aliases service.
    * @ngInject
    * @ngdoc service
    * @ngname gmfDataSourcesManager
    */
   constructor($q, $rootScope, $timeout, gmfThemes, gmfTreeManager,
     ngeoBackgroundLayerMgr, ngeoDataSources, ngeoLayerHelper, ngeoRuleHelper,
-    ngeoWMSTime
+    ngeoWMSTime, gmfWFSAliases
   ) {
 
     // === Injected properties ===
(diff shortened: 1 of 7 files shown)


asaunier:
d93b471 adds attributes aliases to the query results. Problem: original attributes are not removed:
capture du 2017-12-01 18-21-39
(another problem: the attributes order is not preserved)

How could we remove the original attributes? By creating new features instead of editing the existing one?

Another approach: directly filtering when passing the features to the querywindow partial:
https://github.com/camptocamp/ngeo/blob/9da48f8260059705ca6c540af0ff2dfce28bd2fd/contribs/gmf/src/directives/displayquerywindow.js#L450-L455
like in https://github.com/camptocamp/ngeo/blob/7076508ffd335d8f8903939af822998f1804663e/src/services/featurehelper.js#L612-L618
Problem: the list of alias is stored in the ``dataSource`` objects and not sure it is available here.

asaunier:
Thanks to @llienher I have been able to fix the issue reported at https://github.com/camptocamp/ngeo/pull/3191#issuecomment-348556968 => aliases are now correctly handled when performing a WFS GetFeature request on a layer available in the application's themes.

I am now investigating the case of external layers added through the "import layer" tool (available in the "desktop_alt" app). As for now I have figured out that:
* only WMS GFI request are supported for external layers, not WFS GetFeature requests (on the other hand it seems "internal" layers are only requested in GMF using WFS GetFeature, right?)
* data sources created by the external datasource manager do not provide ``ogcServerType`` info: https://github.com/camptocamp/ngeo/blob/c0f412ee9a196c42b74922f117be9ad18ca71a69/contribs/gmf/src/datasource/externaldatasourcesmanager.js#L351-L363 => the default ``mapserver`` type is then always used, including for the "qgisserver" test layer I am using. This is a problem because the aliases are only retrieved for "qgisserver" layers. I have not found how to get the real servertype in this part yet.
* If I force the servertype to "qgisserver", the WFS DescribeFeatureType (required to get the aliases) is blocked saying the WFS requests are not supported for this WMS layer, see https://github.com/camptocamp/ngeo/blob/master/src/services/querent.js#L191-L195 (``dataSource.supportsAttributes`` is false).

asaunier:
After more investigations about supporting aliases in external layers, I see several problems:
* the ``ogcServerType`` is not provided when creating a ``dataSource`` for external layers in the ``gmf.datasource.ExternalDataSourcesManager``. Not sure it is easy to have this information (for internal layers, the server type is provided by the "themes" service I think).
* external layers are imported as WMS layers, with no WFS support by default: when trying to make the WFS DescribeFeatureType request to get the list of attributes and aliases, the querent fails because the data source does not support WFS (no ``wfsUrl``), see
https://github.com/camptocamp/ngeo/blob/c0f412ee9a196c42b74922f117be9ad18ca71a69/src/services/querent.js#L189-L195 and
https://github.com/camptocamp/ngeo/blob/c0f412ee9a196c42b74922f117be9ad18ca71a69/src/datasource/ogc.js#L714-L737

Perhaps we could give up the support of aliases for external layers? At least in a first version of the tool.

PR #3204, opened 2017-12-04
Modularize recenter

ger-benjamin:
For GSGMF https://jira.camptocamp.com/browse/GSGMF-174

### DIFF
--- a/examples/recenter.js
+++ b/examples/recenter.js
@@ -1,7 +1,5 @@
-goog.provide('app.locationchooser');
+goog.provide('app.recenter');
 
-/** @suppress {extraRequire} */
-goog.require('ngeo.recenterDirective');
 goog.require('ol.Map');
 goog.require('ol.View');
 goog.require('ol.layer.Tile');
--- a/examples/recenter.html
+++ b/examples/recenter.html
@@ -1,7 +1,7 @@
 <!DOCTYPE html>
 <html ng-app='app'>
   <head>
-    <title>Location chooser example</title>
+    <title>Recenter example</title>
     <meta charset="utf-8">
     <meta name="viewport"
           content="initial-scale=1.0, user-scalable=no, width=device-width">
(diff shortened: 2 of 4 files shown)


src/recenter/directive.js
```diff
@@ -1,7 +1,15 @@
...
+goog.provide('ngeo.recenter.directive');
 
 goog.require('ngeo');
 
+/**
+ * @type {!angular.Module}
+ */
+ngeo.recenter.directive = angular.module('ngeoRecenter', []);
+
+const ngeoModule = ngeo.module;
```
ger-benjamin: I understand why I have to create this const, but not why it was not required in the gmf.search.component...

src/recenter/directive.js
```diff
@@ -1,7 +1,15 @@
-goog.provide('ngeo.recenterDirective');
+goog.provide('ngeo.recenter.directive');
```
gberaudo: Should be in `src/map/recenter.js`. @sbrunner, do you want to add a note in https://jira.camptocamp.com/browse/GSGMF-172 explaining the files need to be renamed (not moved in a `directive` subdirectory)?

Then the symbol should be `ngeo.map.recenter`. (I would drop the `directive` suffix).

src/recenter/directive.js
gberaudo: @ger-benjamin, no it is not correct to use this const here: you can not create a local constant, in a `goog.provide` file.
Do you have an error when using `ngeo.module` directly?

src/recenter/module.js
```diff
@@ -0,0 +1,13 @@
+/**
+ * @module ngeo recenter namespace
+ */
+goog.provide('ngeo.recenter.module');
```
gberaudo: This module is not necessary (there is no recenter top level module). You can recycle it by renaming it `ngeo.map.module` and moving it there if it does not already exist.

examples/recenter.js
```diff
@@ -1,17 +1,20 @@
...
 /** @suppress {extraRequire} */
 goog.require('ngeo.mapDirective');
-/** @suppress {extraRequire} */
-goog.require('ngeo.recenterDirective');
 goog.require('ol.Map');
 goog.require('ol.View');
 goog.require('ol.layer.Tile');
 goog.require('ol.source.OSM');
 
+goog.require('ngeo.recenter.module');
```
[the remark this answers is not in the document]
sbrunner: In the applications we will use in priority the tom level module, I think that's should be the same in the examples :-)

src/recenter/directive.js
sbrunner: +1

ger-benjamin:
Thanks @sbrunner @gberaudo . I'll wait to merge #3206 then address your comment

PR #3206, opened 2017-12-05
Ngeo map own module

ger-benjamin:
For https://jira.camptocamp.com/browse/GSGMF-174

### DIFF
--- a/examples/asitvd.js
+++ b/examples/asitvd.js
@@ -1,15 +1,18 @@
 goog.provide('app.asitvd');
 
-/** @suppress {extraRequire} */
-goog.require('ngeo.mapDirective');
 goog.require('ngeo.source.AsitVD');
 goog.require('ol.Map');
 goog.require('ol.View');
 goog.require('ol.layer.Tile');
 
+goog.require('ngeo.map.module');
 
-/** @type {!angular.Module} **/
-app.module = angular.module('app', ['ngeo']);
+
+/** @type {!angular.Module} */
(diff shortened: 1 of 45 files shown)


src/map/directive.js
```diff
@@ -1,12 +1,22 @@
...
 goog.require('ol.events');
 goog.require('ol.Map');
 
 
+/**
+ * @type {!angular.Module}
+ */
+ngeo.map.directive = angular.module('ngeoMap', []);
+
+const ngeoModule = ngeo.module;
```
gberaudo: This only works for `goog.module` (because with `goog.provide` it pollutes the window object (with window.ngeoModule property)).

gberaudo: So please use `ngeo.module` directly and disable the check (put it as warning).

examples/asitvd.js
```diff
@@ -1,15 +1,18 @@
...
 goog.require('ol.View');
 goog.require('ol.layer.Tile');
 
+goog.require('ngeo.map.module');
 
-/** @type {!angular.Module} **/
-app.module = angular.module('app', ['ngeo']);
+
+/** @type {!angular.Module} */
+app.module = angular.module('app', [
```
gberaudo: Create a Jira card so that we do not forget to transform:
-app.module => const appModule

before publishing as es6 modules.

ger-benjamin: Done: https://jira.camptocamp.com/browse/GSGMF-179

PR #3216, opened 2017-12-05
Adapt angular architecture: gmf.authentication

pfirpfel:


contribs/gmf/src/directives/print.js
```diff
@@ -2,7 +2,7 @@ goog.provide('gmf.printComponent');
 
 goog.require('gmf');
 /** @suppress {extraRequire} */
-goog.require('gmf.authenticationDirective');
+goog.require('gmf.authentication.module');
```
ger-benjamin: Why do we requested the directive ?
So what you have done is correct, but can't we require only the service ?

contribs/gmf/src/authentication/service.js
```diff
@@ -0,0 +1,242 @@
+goog.provide('gmf.authentication.service');
```
ger-benjamin: File name must be wrote with a uppercase s : Service.js. Cause it's a java-script object.
(Unlike the component, that is a description. The controller must also have an uppercase but anyway it's always private in it's own file)

Also. provide / require must use this uppercase S.
And also the Object name itself

contribs/gmf/src/authentication/service.js
```diff
@@ -0,0 +1,242 @@
...
+gmf.authentication.AuthenticationRouteSuffix = {
+  CHANGE_PASSWORD: 'loginchange',
+  IS_LOGGED_IN: 'loginuser',
+  LOGIN: 'login',
+  LOGOUT: 'logout',
+  RESET_PASSWORD: 'loginresetpassword'
+};
+
+// todo?
+gmf.module.value('gmfUser', {
```
ger-benjamin: Looks not suitable.
@gberaudo We move that in another file ?

contribs/gmf/src/authentication/service.js
```diff
@@ -0,0 +1,242 @@
...
+ * The possible API requests it supports, which are all self-explanatory, are:
+ *
+ * - changePassword
+ * - login
+ * - logout
+ * - resetPassword
+ *
+ * @extends {ol.events.EventTarget}
+ */
+gmf.authentication.service = class {
```
ger-benjamin: uppercase

contribs/gmf/src/directives/filterselector.js
```diff
@@ -2,7 +2,7 @@ goog.provide('gmf.filterselectorComponent');
 
 goog.require('gmf');
 /** @suppress {extraRequire} */
-goog.require('gmf.Authentication');
+goog.require('gmf.authentication.module');
```
ger-benjamin: Oh, I see, it's for the gmfUser. Definitely, it should be moved in another file. perhaps it's own file like the https://github.com/camptocamp/ngeo/blob/master/src/map/FeatureOverlay.js Object.

And then we require only the user, not the whole module.
(Also in another files with the same situation)

contribs/gmf/src/authentication/service.js
```diff
@@ -0,0 +1,242 @@
...
+/**
+ * @typedef {{
+ *     functionalities: (gmfx.AuthenticationFunctionalities|undefined),
+ *     is_password_changed: (boolean|undefined),
+ *     role_id: (number|undefined),
+ *     role_name: (string|undefined),
+ *     username: (string|undefined)
+ * }}
+ */
+gmf.authentication.AuthenticationLoginResponse;
```
ger-benjamin: Please, move this kind of descriptions in the contribs/gmf/options/gmfx.js file

contribs/gmf/src/authentication/service.js
```diff
@@ -0,0 +1,242 @@
...
+ *     success: boolean
+ * }}
+ */
+gmf.authentication.AuthenticationDefaultResponse;
+
+
+/**
+ * @enum {string}
+ */
+gmf.authentication.AuthenticationRouteSuffix = {
```
ger-benjamin: If it's in the gmf.authentification.Service file (and used by it), it should be prefixed like that I think:
`gmf.authentication.service.AuthenticationRouteSuffix = {...}`
Or even `gmf.authentication.service.RouteSuffix = {...}`

contribs/gmf/src/authentication/service.js
gberaudo: I do not understand why `gmfUser` is a standalone value instead of being a property of the authentication service.
To keep things simple I would let the value in this file (but register it into the authentication module).

Alternatives are:
- move it to its dedicated file, with its own Angular module and depend on it everywhere.
- make it a property of the authentication service.

contribs/gmf/src/authentication/component.js
```diff
@@ -0,0 +1,373 @@
+goog.provide('gmf.authentication.component');
+
+goog.require('gmf');
+/** @suppress {extraRequire} */
```
gberaudo: You can remove this suppress line.

contribs/gmf/src/authentication/service.js
```diff
@@ -0,0 +1,223 @@
...
+    for (const key in this.user_) {
+      this.user_[key] = null;
+    }
+    /** @type {gmfx.AuthenticationEvent} */
+    const event = new ngeo.CustomEvent('logout', {user: this.user_});
+    this.dispatchEvent(event);
+  }
+};
+
+ol.inherits(gmf.authentication.Service, ol.events.EventTarget);
```
gberaudo: This line is no more used.

pfirpfel:
@ger-benjamin Does it look ok to you now?

pfirpfel:
History is now preserved, see:
https://github.com/pfirpfel/ngeo/blame/gmf.authentication/contribs/gmf/src/authentication/component.js

PR #3226, opened 2017-12-06
Move last ngeo map services

ger-benjamin:
For https://jira.camptocamp.com/browse/GSGMF-183

Lasts module to move to the ngeo Map module:
 - Service ngeo LayerHelper (as extra of the module)
 - Service ngeo FeatureOverlayMgr
 - Object ngeo FeatureOverlay 

@gberaudo I'll like to have confirmation on some things I choose to do about module requirements:
 1.  ngeo.map.FeatureOverlay not in the ngeo.map.module , cause it's already in required by the FeatureOverlayMgr. Work but that's not like the gmf.search.Fulltextsearch service that is also put in the module).
 1.  Requirement of ngeo.map.FeatureOverlayMgr in gmf directive that use this service. I think that's right, but I can also require the whole ngeo.map.module

And the test doesn't pass. I need help to fix them

With this PR, we will have a whole top-level module :smiley:

### DIFF
--- a/contribs/gmf/src/directives/displayquerygrid.js
+++ b/contribs/gmf/src/directives/displayquerygrid.js
@@ -5,8 +5,6 @@ goog.require('ngeo.CsvDownload');
 goog.require('ngeo.GridConfig');
 /** @suppress {extraRequire} */
 goog.require('ngeo.gridComponent');
-goog.require('ngeo.FeatureOverlay');
-goog.require('ngeo.FeatureOverlayMgr');
 /** @suppress {extraRequire} - required for `ngeoQueryResult` */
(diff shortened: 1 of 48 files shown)


contribs/gmf/src/directives/displayquerygrid.js
```diff
@@ -15,6 +13,12 @@ goog.require('ol.style.Fill');
 goog.require('ol.style.Stroke');
 goog.require('ol.style.Style');
 
+goog.require('ngeo.map.FeatureOverlayMgr');
+
+
+// In futur module declaration, don't forget to require:
```
gberaudo: Typo: In the future.
No need to update the PR for this.

PR #3228, opened 2017-12-07
chore(package): update openlayers to version 4.6.2

sbrunner:
Closes #3225


fredj:
we need to wait for a new ol-cesium release

fredj:
#3232

PR #3248, opened 2017-12-11
Fix UI overlapping on iOS 10 (Safari)

### LINKED ISSUE #2695: iOS 10 safari UI overlaps with gmf 2.1
The problem appears on Safari for iOS 10. All devices were using iOS 10.3.2 (last stable version). I saw that this issue appears on Iphone 5S and Iphone 5C by clients as well in my own tests on Browserstack with more recent devices using this version of iOS. So it seems it is more about the OS version that the devices itself.

Reproduced on GMF 2.1.4 on landscape orientation.

Screen from the client:
img_0669

llienher:
Fixes #2695

This PR introduce a iOS less file that take specific classes. Then in the interface, the ios user-agent is checked to know if it is an iOS device and its version. If true and egal to iOS 10, it applies the classes to the HTML elements.

TODO:
- [x] (maybe) exlude iOS 9 and below
- [x] Test on real devices
- [x] Fix build (IE load the file on any interfaces)

Known issues:
- Overlap on the botton of the screen (it is possible to get the screen scrolled with a focus on the search field and drag at the bottom)
- ~Code is not minified/transpiled in the build (so may break on IE ?)~

Demo:
https://camptocamp.github.io/ngeo/ios-overlap-fix/examples/contribs/gmf/apps/mobile/

### DIFF
--- a/contribs/gmf/apps/mobile/index.html
+++ b/contribs/gmf/apps/mobile/index.html
@@ -190,7 +190,7 @@
     <script src="../../../../node_modules/url-polyfill/url-polyfill.js"></script>
     <script src="/@?main=mobile/js/controller.js"></script>
     <script src="../default.js"></script>
-    <script src="../ios-overlap-fix.js"></script>
+    <script src="../../../../utils/ios-overlap-fix.js"></script>
     <script src="../../../../utils/watchwatchers.js"></script>
     <script>
       (function() {
--- a/contribs/gmf/apps/mobile/index.html
+++ b/contribs/gmf/apps/mobile/index.html
@@ -194,15 +194,16 @@
     <script>
       // Hack for ios Safari browser (UI overlapping on iOS 10)
       var userAgent = window.navigator.userAgent;
-      var iOS = !!userAgent.match(/iP(hone|od|ad)/i);
-      var webkit = !!userAgent.match(/WebKit/i);
+      var regEx = /iP(hone|od|ad)/;
+      var iOS = regEx.test(userAgent);
+      var webkit = /WebKit/i.test(userAgent);
+      var criOS = /CriOS/i.test(userAgent);
+      var platform = regEx.test(window.navigator.platform);
 
-      if(iOS && webkit && !userAgent.match(/CriOS/i)) {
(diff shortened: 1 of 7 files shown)


contribs/gmf/less/iphone.less
```diff
@@ -0,0 +1,20 @@
+// Hack for ios Safari browser (UI overlapping on iOS 11)
+@media all and (orientation: landscape) {
+  @ios-margin: 5rem;
+
+  .ios-margin-top {
+    margin-top: @ios-margin;
+  }
+
+  .ios-zoom-btn {
+    top: (@app-margin + @map-tools-size + @app-margin) + @ios-margin !important;
```
sbrunner: You shouldn't use !important...

llienher: not working without it :-(

contribs/gmf/apps/mobile/index.html
```diff
@@ -194,15 +194,16 @@
...
-      var iOS = !!userAgent.match(/iP(hone|od|ad)/i);
-      var webkit = !!userAgent.match(/WebKit/i);
+      var regEx = /iP(hone|od|ad)/;
+      var iOS = regEx.test(userAgent);
+      var webkit = /WebKit/i.test(userAgent);
+      var criOS = /CriOS/i.test(userAgent);
+      var platform = regEx.test(window.navigator.platform);
 
-      if(iOS && webkit && !userAgent.match(/CriOS/i)) {
+      if(iOS && webkit && platform && !criOS && !window.MSStream) {
```
sbrunner: When I see this if I suspect that we will have an issue in near future ...

contribs/gmf/less/iphone.less
sbrunner: your selector isn't enough specific

llienher: My specific selectors are located in the javascript file (I need to do it as it's from the javascript that I activate or not the UI tweak). In the less I only set the properties I want to override with a simple name that I can retrieve easily in the javascript code.

The classes are added using $('selector').addClass('className').

llienher:
@fredj It's the final code, can you review ? Thanks.

PR #3264, opened 2017-12-13
Don't query the map on double click

### LINKED ISSUE #3261: Zooming in with double tap/click should not fire a query
Project:
* Host: https://geomapfish-demo.camptocamp.net/2.2/mobile
* GeoMapFish version: 2.2

Permalink directly where we can reproduce the issue:
* https://geomapfish-demo.camptocamp.net/2.2/mobile

Steps to reproduce the issue:

* Open above link with a mobile device
* Double tap to zoom in

Result:
* The map is zoomed in (OK)
* A query is fired (KO)

Expected result:
* No query should be fired

In fact this issue is also present in desktop interface with a double click, so it's a ngeo issue (it was fine in cgxp).

fredj:
fixes #3261

### DIFF
--- a/src/directives/mapquery.js
+++ b/src/directives/mapquery.js
@@ -78,7 +78,7 @@ ngeo.mapQueryDirective = function(ngeoMapQuerent, $injector) {
        * Listen to the map events.
        */
       const activate_ = function() {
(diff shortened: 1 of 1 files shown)


rbovard:
Thanks @fredj!

PR #3269, opened 2017-12-13
Display label drawing

julsbreakdown:
Enable the user to display the feature label. 
As it is done for the measurement label.
cf 
capture du 2017-12-13 14-43-17

### DIFF
--- a/src/services/featurehelper.js
+++ b/src/services/featurehelper.js
@@ -201,7 +201,6 @@ ngeo.FeatureHelper.prototype.getStyle = function(feature) {
  * @private
  */
 ngeo.FeatureHelper.prototype.getLineStringStyle_ = function(feature) {
-
   const strokeWidth = this.getStrokeProperty(feature);
   const showLabel = this.getShowLabelProperty(feature);
   const showMeasure = this.getShowMeasureProperty(feature);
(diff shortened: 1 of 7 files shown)


src/services/featurehelper.js
```diff
@@ -197,61 +197,88 @@ ngeo.FeatureHelper.prototype.getStyle = function(feature) {
...
+    }));
+  }
+  if (showLabel) {
+    offset += 25;
+    styles.push(new ol.style.Style({
+      text: this.createTextStyle_({
+        text: this.getNameProperty(feature),
+        offsetY: -(offset / 2 + 4)
+      })
+    }));
```
sbrunner: @fredj is this offset calculation correct?

src/services/featurehelper.js
```diff
@@ -197,61 +197,88 @@ ngeo.FeatureHelper.prototype.getStyle = function(feature) {
...
```
sbrunner: Same

src/services/featurehelper.js
```diff
@@ -355,13 +383,25 @@ ngeo.FeatureHelper.prototype.getPolygonStyle_ = function(feature) {
...
   }
+  if (showLabel) {
+    offset += 25;
+    styles.push(new ol.style.Style({
+      text: this.createTextStyle_({
+        text: this.getNameProperty(feature),
+        size: 10,
+        offsetY: -(offset / 2 + 4)
+      })
+    }));
```
sbrunner: again :-)

src/services/featurehelper.js
fredj: no idea

sbrunner: This 25 looks to be a magic number, shouln't we use multiline from OpenLayers: https://github.com/openlayers/openlayers/pull/4512/?

julsbreakdown:
In this case we only have the one style for the label, and use the multiline ('/n' option) if both (measure & label are checked)?

sbrunner:
Yes :-)

src/services/featurehelper.js
```diff
@@ -197,61 +197,88 @@ ngeo.FeatureHelper.prototype.getStyle = function(feature) {
...
+	  if (showMeasure){
+	    textLabelValue = textMeasure;
+	  }
+	  if (showLabel){
+	    textLabelValue = textProperty;
+	  }
+	  if (showLabel && showMeasure){
+		// display both label using  \n
+		textLabelValue =  textProperty + '\n' +textMeasure;
+	  }
```
sbrunner: We usely do like this:
```js
cost textLabelValues = [];
textLabelValues.push(this.getNameProperty(feature));
textLabelValues.push(this.getMeasure(feature));
cost textLabelValue = textLabelValues.join('\n');
```
=> shorter and no if :-)

PR #3274, opened 2017-12-14
Adapt angular: ngeo.filter

pfirpfel:


src/filter/component.js
```diff
@@ -294,7 +325,7 @@ ngeo.FilterController = class {
 
 
   /**
-   * @param {!ngeo.FilterController.Condition} condition Condition to set.
+   * @param {!ngeo.filter.component.Condition} condition Condition to set.
```
ger-benjamin: It's not better to have a ngeox.filterCondition ?

pfirpfel: Moved it.

PR #3285, opened 2017-12-18
Handle jsCodeShift errors

sbrunner:


buildtools/webpack-migration
```diff
@@ -1,11 +1,21 @@
 #!/bin/bash -ex
 
+function jscodeshift {
+    if node_modules/.bin/jscodeshift --dry $*  | grep '^0 errors$'
```
gberaudo: To speed up things you may want to use `tee` instead (untested):
```
node_modules/.bin/jscodeshift $* | tee jscodeshift_out.log
grep -q '^0 errors$'  jscodeshift_out.log && false
```

sbrunner: Done :-)

PR #3287, opened 2017-12-18
Stop providing ngeo.layertree.DecorateLayer as a service

fredj:
Use simple function instead


fredj:
I've added the units tests back

PR #3288, opened 2017-12-18
Create ngeo profile module

ger-benjamin:
For: https://jira.camptocamp.com/browse/GSGMF-206
 
It's right what I've done with the d3Elevation function ? 

I've renamed it to profile.elevation because we will get a "lidar" profile in the future.

### DIFF
--- a/contribs/gmf/src/directives/profile.js
+++ b/contribs/gmf/src/directives/profile.js
@@ -1,8 +1,6 @@
 goog.provide('gmf.profileComponent');
 
 goog.require('gmf');
-/** @suppress {extraRequire} */
-goog.require('ngeo.profileDirective');
 goog.require('ol.events');
 goog.require('ol.Feature');
 goog.require('ol.Overlay');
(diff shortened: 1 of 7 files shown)


contribs/gmf/src/directives/profile.js
```diff
@@ -16,11 +14,14 @@ goog.require('ol.style.Style');
 /** @suppress {extraRequire} */
 goog.require('ngeo.download.Csv');
 goog.require('ngeo.map.FeatureOverlayMgr');
+/** @suppress {extraRequire} */
+goog.require('ngeo.profile.elevation');
```
gberaudo: Should the component be renamed to `gmf.elevationProfileComponent`?

examples/profileElevation.js
```diff
@@ -1,7 +1,5 @@
-goog.provide('app.profile');
+goog.provide('app.profileElevation');
```
gberaudo: In English, the order should be reversed to `elevationProfile`.

src/profile/d3Elevation.js
```diff
@@ -58,7 +59,7 @@ goog.require('ol.obj');
  * @param {ngeox.profile.ProfileOptions} options Profile options.
  * @export
  */
-ngeo.profile = function(options) {
+ngeo.profile.d3Elevation = function(options) {
```
gberaudo: It is a constructor so it should start uppercase.

ger-benjamin: It is ? or it is a big function, but not a constructor ?

examples/profileElevation.js
ger-benjamin: I've named it like this because I think that you search for "profile" in example, not for elevation. But I can change

src/profile/d3Elevation.js
gberaudo: There is the annotation in the function comment.

pfirpfel: I think the constructor (never used with new) and struct (doesn't have any properties on itself) annotations are incorrect here.

ger-benjamin: I've removed the - * (at)constructor  - * (at)struct tags

PR #3290, opened 2017-12-18
Adapt angular: ngeo.query

pfirpfel:


src/query/Querent.js
```diff
@@ -1,17 +1,23 @@
-goog.provide('ngeo.Querent');
+/**
+ * @module ngeo query namespace
```
ger-benjamin: what's that ? Do you know ?

src/query/bboxquerydirective.js
```diff
@@ -1,8 +1,13 @@
-goog.provide('ngeo.bboxQueryDirective');
+/**
+ * @module ngeo query namespace
+ */
+goog.provide('ngeo.query.bboxQueryDirective');
```
ger-benjamin: Event if that's always an angular directive, can you names it "component ?".
That's a more generic term and In the future, we will change it to a component. So we (and our customers in their custom code) will have to update again the require.

contribs/gmf/src/services/querymanager.js
```diff
@@ -3,7 +3,7 @@ goog.provide('gmf.QueryManager');
 goog.require('ol.events');
 goog.require('gmf');
 goog.require('gmf.Themes');
-goog.require('ngeo.Query');
+goog.require('ngeo.query.Query');
```
ger-benjamin: ngeo.query.Service ?  query Query looks weird (In my own opinion)

gberaudo: Yes, should be Service.

src/query/Querent.js
gberaudo: We should not put `@module` annotations; they will be generated automatically later (in a specific format).

src/query/bboxquerydirective.js
gberaudo: +1

src/query/mapquerydirective.js
```diff
@@ -119,4 +123,11 @@ ngeo.mapQueryDirective = function(ngeoMapQuerent, $injector) {
...
 
-ngeo.module.directive('ngeoMapQuery', ngeo.mapQueryDirective);
+
+ngeo.query.mapQueryDirective.module = angular.module('ngeoMapQuery', [
+  ngeo.query.MapQuerent.module.name,
+]);
+
+ngeo.module.requires.push(ngeo.query.mapQueryDirective.module.name);
+
+ngeo.query.mapQueryDirective.module.directive('ngeoMapQuery', ngeo.query.mapQueryDirective);
```
gberaudo: I think we said to export the Angular module for directives and components (not the directive function).
```
ngeo.query.mapQueryDirective = angular.module('ngeoMapQuery', [
  ngeo.query.MapQuerent.name,
]);
```

ger-benjamin: +1
This file is strange. That should be:
`ngeo.query.mapQueryDirective = angular.module...`
`ngeo.module.requires....`
...
`ngeo.query.mapQueryDirective.directive_ = function...`
...
`ngeo.query.mapQueryDirective.directive('ngeoMapQuery', ngeo.query.mapQueryDirective.directive_);`

no ?

PR #3303, opened 2017-12-20
Move some services to misc module (second pack)

ger-benjamin:
For GSGMF-214

Then, we still need to move that in the extra "misc" module (in another PR):

- src/directives/btngroup.js
- src/directives/control.js
- src/directives/filereader.js
- src/directives/sortable.js 
- contribs/gmf/src/directives/swipe.js
- src/directives/colorpicker.js 
- src/directives/datepicker.js


src/misc/decorate.js
```diff
@@ -1,7 +1,8 @@
-goog.provide('ngeo.DecorateInteraction');
+goog.provide('ngeo.misc.decorate');
```
fredj: nice! we should move the layer decorator here (#3309)

ger-benjamin: Yes !

src/misc/decorate.js
```diff
@@ -15,17 +16,9 @@ goog.require('ngeo');
  *
  * See our live example: [../examples/interactiontoggle.html](../examples/interactiontoggle.html)
```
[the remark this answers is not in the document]
ger-benjamin: I remove it in the third (and last) "misc" pack

PR #3311, opened 2017-12-21
Don't update the view rotation if the geolocation is not active

fredj:


sbrunner:
It don't affect the version2.2?

fredj:
> It don't affect the version2.2?

checking now

sbrunner:
OK, thanks for the check :-)

PR #3329, opened 2018-01-03
Move filter template

pfirpfel:
Forgot to move the filter template in #3274


pfirpfel:
Changed it

pfirpfel:
Renamed the other one in this module as well.

PR #3332, opened 2018-01-04
Adapt angular ngeo.message

pfirpfel:


examples/modal.js
```diff
@@ -1,11 +1,14 @@
...
 
 /** @suppress {extraRequire} */
-goog.require('ngeo.modalDirective');
+goog.require('ngeo.message.modalComponent');
 
 
 /** @type {!angular.Module} **/
-app.module = angular.module('app', ['ngeo']);
+app.module = angular.module('app', [
+  'ngeo',
```
ger-benjamin: `ngeo.module.name` is better no ?
Look also other ngeo example (`git grep \'ngeo\'`)

src/message/Message.js
```diff
@@ -153,3 +125,31 @@ ngeo.Message.prototype.getMessageObjects = function(object, opt_type) {
 
   return msgObjects;
 };
+
+
+/**
+ * @enum {string}
+ * @export
+ */
+ngeo.message.Message.MessageType = {
```
ger-benjamin: That's a little bit repetitive for a name...
Can you still change it to `ngeo.message.Message.type` ?
(And the last name should start without uppercase no ?)

pfirpfel: Should start with a uppercase character. Google does it [this way](https://google.github.io/styleguide/javascriptguide.xml?showone=Naming#Naming).

How about `ngeo.message.Message.Type`?

ger-benjamin: okay

pfirpfel: Changed it

examples/modal.js
pfirpfel: This should be all of them.

ger-benjamin:
I've restarted the build, it was on error after a timeout.

PR #3343, opened 2018-01-08
Allow bloodhound rateLimitWait params to be settable in gmf

llienher:
This PR allow the rateLimitWait parameter, that is used with the Bloodhound engine for debouncing fulltext-search requests, to be set up. That allow as well to set different values for each interfaces. Like mobile could get a higher value in order to get less request / bandwidth used with the backend for fulltext-search.

If the value is not set from the directive html, or not existing in the interface controller, the search component is going to set it to the default value: 50 ms. This was the value we previously hard-coded. There should be no issue for project that have outdated html for the search component as the component is going to initialize the value correctly in any cases.

I updated every interface, only desktop_alt have a different value that the default one.
Fixed the gmf example as well.

### DIFF
--- a/contribs/gmf/src/directives/search.js
+++ b/contribs/gmf/src/directives/search.js
@@ -59,6 +59,7 @@ gmf.module.value('gmfSearchTemplateUrl',
  *        gmf-search-styles="ctrl.searchStyles"
  *        gmf-search-datasources="ctrl.searchDatasources"
  *        gmf-search-coordinatesprojections="ctrl.searchCoordinatesProjections"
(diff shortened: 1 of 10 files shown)


contribs/gmf/src/directives/search.js
```diff
@@ -298,6 +302,12 @@ gmf.SearchController = function($scope, $compile, $timeout, $injector, gettextCa
    */
   this.placeholder;
 
+  /**
+   * @type {string}
```
pfirpfel: `@type {string}` shouldn't this be `@type {number}`

PR #3351, opened 2018-01-09
Escape sourceLabel ids when invalid characters are present

### LINKED ISSUE #3273: Display query grid: No titles in tabs of second row
Project:
* GitHub URL: https://github.com/camptocamp/baselstadt_mapbs
* GeoMapFish version: 2.2.1

Steps to reproduce the issue:
* do a query over a surface
* open a tab in the second row
* the titles of the tabs in the second row are broken

Result:
display_query_grid

As soon as there are two rows of tabs in the display query grid the titles are broken.

llienher:
Fixes #3273.

With this fix, we escape the IDs (generated from the value in ctrl.mergeTabs for merged tabs or OL datasource). If an invalid value would result in an invalid ID selector (css), then the value is stripped from unauthorized characters. Other change is the use of the datasource IDs for the HTML ids instead of the label value that contained spaces, etc.

Ideally ids defined in the ctrl.mergeTabs should have no spaces, or any specific characters such as parentheses, brackets, etc. This fix allow all project to work without changing ids manually, but if possible having a correct value is just best practice...

In that context I changed the value in our desktop example, while the fix was tested on the previous value, 'OSM time merged'.

### DIFF
--- a/contribs/gmf/src/directives/displayquerygrid.js
+++ b/contribs/gmf/src/directives/displayquerygrid.js
@@ -348,6 +348,7 @@ gmf.DisplayquerygridController.prototype.updateData_ = function() {
     if (source.tooManyResults) {
       this.makeGrid_(null, source);
     } else {
+      source.id = this.escapeValue_(source.id);
       const features = source.features;
       if (features.length > 0) {
         this.collectData_(source);
@@ -355,7 +356,7 @@ gmf.DisplayquerygridController.prototype.updateData_ = function() {
     }
   });
 
-  if (this.loadedGridSources.length == 0) {
+  if (this.loadedGridSources.length === 0) {
     // if no grids were created, do not show
     this.active = false;
     return;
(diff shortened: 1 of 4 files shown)


llienher:
@ger-benjamin I refactored the code. Now it is escaped at data generation (before the function that going to fill the grid). I did the change to use Ids again in the HTML instead of the labels. Labels are still used for the tab labels, so nothing change for the user.

contribs/gmf/src/directives/displayquerygrid.js
```diff
@@ -381,6 +382,26 @@ gmf.DisplayquerygridController.prototype.hasOneWithTooManyResults_ = function()
   return this.ngeoQueryResult.sources.some(source => source.tooManyResults);
 };
 
+/**
+ * Returns an escaped value.
+ * @param {string|number} value A value to escape.
+ * @returns {string|number} value An escaped valjue.
```
ger-benjamin: typo

llienher:
@ger-benjamin I added two merged tabs in desktop_alt to make sure we have correct testing setup.

PR #3354, opened 2018-01-09
Adapt angular gmf.contextualdata

pfirpfel:


contribs/gmf/src/directives/drawprofileline.js
```diff
@@ -54,7 +54,7 @@ gmf.drawprofilelineDirective = function() {
      * @param {angular.Scope} scope Scope.
      * @param {angular.JQLite} element Element.
      * @param {angular.Attributes} attrs Attributes.
-     * @param {gmf.ContextualdataController} controller Controller.
+     * @param {gmf.DrawprofilelineController} controller Controller.
```
pfirpfel: I think this was a copy and paste mistake that was yet undiscovered.

PR #3360, opened 2018-01-09
Mouse position translation

### LINKED ISSUE #3344: Mouseposition: Translation of placeholder (Coordinates)
Within the mouseposition component the translation for the placeholder (markup for undefined coordinates) seems not to work correctly. There is always the English word "Coordinates" shown.

llienher:
Fixes #3344 

This fixes the translation issue with the mouse position OL control. Now we wait for gettextCatalog to fire an event that the language is loaded/changed to create the component with the correct language string.

### DIFF
--- a/contribs/gmf/src/directives/mouseposition.js
+++ b/contribs/gmf/src/directives/mouseposition.js
@@ -37,7 +37,8 @@ gmf.module.component('gmfMouseposition', gmf.mousepositionComponent);
 
 /**
  * @param {!angular.JQLite} $element Element.
- * @param {!angular.$filter} $filter Angular filter
+ * @param {!angular.$filter} $filter Angular filter.
+ * @param {!angular.Scope} $scope Angular scope.
  * @param {!angularGettext.Catalog} gettextCatalog Gettext catalog.
  * @constructor
  * @private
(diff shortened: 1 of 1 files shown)


contribs/gmf/src/directives/mouseposition.js
```diff
@@ -64,32 +65,64 @@ gmf.MousepositionController = function($element, $filter, gettextCatalog) {
    */
   this.projection;
 
+  /**
+   * @type {angular.Scope}
+   *
```
fredj: missing `@private`

contribs/gmf/src/directives/mouseposition.js
```diff
@@ -64,32 +65,64 @@ gmf.MousepositionController = function($element, $filter, gettextCatalog) {
...
+   */
+  this.$scope_ = $scope;
+
+  /**
+   * @type {angularGettext.Catalog}
+   * @private
+   */
+  this.gettextCatalog_ = gettextCatalog;
```
gberaudo: FWIW, you can mark it protected and remove the trailing underscore. Then there is no need to alias it in the code below.

PR #3361, opened 2018-01-10
Create gmf layertree module

ger-benjamin:
For: https://jira.camptocamp.com/browse/GSGMF-241


docs/guidelines.md
```diff
@@ -552,7 +552,7 @@ include the `ngeo-layertree` directive.
  * @return {angular.Directive} The directive Definition Object.
  * @ngInject
  */
-gmf.layertreeComponent = function() {
```
pfirpfel: Oh wow, I did not have the doc directory on my radar. Nice catch!

ger-benjamin: I think that's not really useful, because the doc is probably outdated. But Anyway, It's better like this.

PR #3362, opened 2018-01-10
Adapt angular gmf.import

pfirpfel:


pfirpfel:
@ger-benjamin I just tried contribs/gmf/examples/importdatasource.js`, it works fine.

PR #3363, opened 2018-01-10
Add line pcap support to vector encoder

gberaudo:


gberaudo:
Thanks @fredj for the quick review.

PR #3371, opened 2018-01-11
Use packages from npm

sbrunner:


package.json
```diff
@@ -59,8 +59,8 @@
     "less-plugin-clean-css": "1.5.1",
     "moment": "2.18.1",
     "nomnom": "1.8.1",
-    "openlayers": "openlayers/openlayers#4a6317d",
-    "ol-cesium": "openlayers/ol-cesium#3014d02de4",
+    "openlayers": "4.5.0",
```
fredj: do we want to have the latest stable version ? 4.6.4

sbrunner: Here we are on branch 2.2, and I get the first version after the pervius commit :-)

PR #3373, opened 2018-01-11
Adapt angular gmf.mobilemeasure

pfirpfel:


src/interaction/mobiledraw.js
```diff
@@ -12,16 +12,6 @@ goog.require('ol.layer.Vector');
 goog.require('ol.source.Vector');
 
 
-/**
- * @enum {string}
- */
-ngeo.interaction.MobileDrawProperty = {
```
ger-benjamin: Why do you remove that ?

pfirpfel: @gberaudo told me to remove the enum and use strings instead, see HipChat discussion on January 11th.

ger-benjamin: Ok, thanks !

PR #3378, opened 2018-01-12
Allow print symbols rotation in gmf projects

llienher:
This was the old behavior for 2.1 (icons were rotating with the map).
Adding parameter 'useNativeAngle' to 'false' fixes the issue in 2.2.

### DIFF
--- a/src/services/print.js
+++ b/src/services/print.js
@@ -216,19 +216,11 @@ ngeo.Print.prototype.encodeMap_ = function(map, scale, object) {
   object.layers = [];
 
   const mapLayerGroup = map.getLayerGroup();
-  goog.asserts.assert(mapLayerGroup !== null);
+  goog.asserts.assert(mapLayerGroup);
+  this.printNativeAngle_ = !(mapLayerGroup.get('printNativeAngle') !== false);
   let layers = this.ngeoLayerHelper_.getFlatLayers(mapLayerGroup);
   layers = layers.slice().reverse();
 
-  // (1.) We loop on every layers and if one of them doesn't use native angle,
-  // this set the global value to use for the print (default is true)
-  layers.forEach((layer) => {
-    if (layer.get('printNativeAngle') === false) {
-      this.printNativeAngle_ = false;
-    }
-  });
-
-  // (2.) We encode the layers
   layers.forEach((layer) => {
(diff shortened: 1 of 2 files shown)


src/services/print.js
```diff
@@ -214,6 +220,15 @@ ngeo.Print.prototype.encodeMap_ = function(map, scale, object) {
...
   layers = layers.slice().reverse();
 
+  // (1.) We loop on every layers and if one of them doesn't use native angle,
+  // this set the global value to use for the print (default is true)
+  layers.forEach((layer) => {
+    if (layer.getProperties()['useNativeAngle'] === false) {
+      this.useNativeAngle_ = false;
+    }
+  });
+
```
llienher: If one layer need it, it's active for every print request (layer groups).

I did this because I had an issue if I was mixing true and false between various layer groups (the layer was not visible once printed)... EPFL doesn't need to support this use case anyways but I don't want the print to break if some project just apply the layer.setProperties('useNativeAngle': false) to a single layer group.

src/services/print.js
```diff
@@ -214,6 +220,15 @@ ngeo.Print.prototype.encodeMap_ = function(map, scale, object) {
   let layers = this.ngeoLayerHelper_.getFlatLayers(mapLayerGroup);
   layers = layers.slice().reverse();
 
+  // (1.) We loop on every layers and if one of them doesn't use native angle,
+  // this set the global value to use for the print (default is true)
+  layers.forEach((layer) => {
+    if (layer.getProperties()['useNativeAngle'] === false) {
```
sbrunner: Shouldn't be 'printNativeAngle' To be clear that it's used only by the print?

llienher: @sbrunner You're right about that. I renamed it.

sbrunner:
The tests should be updated :-)

test/spec/services/print.spec.js
```diff
@@ -113,7 +113,8 @@ describe('ngeo.CreatePrint', () => {
                 type: 'wms',
                 opacity: 1,
                 serverType: undefined,
-                version: undefined
+                version: undefined,
+                useNativeAngle: false
```
sbrunner: It should be true by default ...

PR #3382, opened 2018-01-12
Set the background layer from the them at page load

fredj:
Correctly set the background layer at page load.

Change: at page load, the theme name is directly set to the url; it was not the case previously.


contribs/gmf/src/services/permalink.js
```diff
@@ -922,8 +922,7 @@ gmf.Permalink.prototype.initLayers_ = function() {
     const themeName = this.defaultThemeName();
 
     if (this.gmfThemeManager_ && this.gmfThemeManager_.modeFlush) {
-      // Set theme in stealth mode, we just want to set the theme name, not more.
-      this.gmfThemeManager_.setThemeName(`${themeName}`, true);
+      this.gmfThemeManager_.setThemeName(themeName);
```
sbrunner: There a missing `goog.asserts.assert` ...

PR #3393, opened 2018-01-16
Move gmf displaywindow to ngeo message

ger-benjamin:
For: https://jira.camptocamp.com/browse/GSGMF-245

Constat: displaywindow is in gmf but must be in ngeo.
Problem: displaywindow use display**query**window

I move the component and change the css to be able to use the same css but without duplicating the css rule.

Now, the displaywindow component is in the ngeo.messaage module. Should I do an example in ngeo ?


pfirpfel:
> Now, the displaywindow component is in the ngeo.messaage module. **Should I do an example in ngeo ?**

Yes, that would be nice. Every ngeo component should have one.

ger-benjamin:
Ok, I merge and I ask ybolognini to know what I must do. (Because that's not part of the ES6 tasks/budget)

PR #3394, opened 2018-01-16
Create gmf datasource module

ger-benjamin:
For: https://jira.camptocamp.com/browse/GSGMF-247


contribs/gmf/src/datasource/datasourcebeingfiltered.js
```diff
@@ -1,23 +0,0 @@
-goog.provide('gmf.datasource.DataSourceBeingFiltered');
```
pfirpfel: The renaming for this file got lost. 
contribs/gmf/src/datasource/DatasourceBeingFiltered.js

ger-benjamin: It's because it's nearly empty and I've changed everything... I think I can't proceed differently.

pfirpfel: To make it work, you would have to split the move and the change of content into two commits. It is annoying...

ger-benjamin: Done, but I'm not sure if that have changed something (except that I have good clear conscience).

PR #3396, opened 2018-01-16
Create gmf map module

ger-benjamin:
For: https://jira.camptocamp.com/browse/GSGMF-249


ger-benjamin:
The second commit is a fix for the mouseposition. I'll do the same fix in the 2.2

PR #3407, opened 2018-01-18
Fix draw issues

### LINKED ISSUE #3401: Labels issues and improvements in draw&measure
Bug with circles: Display label on a circle geometry doesn't do anything... The circle use "Display azimuth and radius" instead of "Display measure" as do other geometries. The current implementation doesn't take account of the specifics of circle label to make this use case works. Plus clicking only on display labels 
 just show the azimut (it should display nothing).

Possible improvement with polygons: long labels in polygons, rectangles, etc., doesn't exceed the available length regarding the current view zoom and the geometry size. This means the label disappears if there is no space left to entirely show it.

It could be possible to set an additional parameter to labels as possible here in the OL example (the 'exceed length' dropdown in the example): 
https://openlayers.org/en/latest/examples/vector-labels.html

Possible implementation is documented here:
http://openlayers.org/en/latest/apidoc/ol.style.Text.html
(overflow + position attributes)

sbrunner:
Fix #3401

### DIFF
--- a/contribs/gmf/src/directives/partials/featurestyle.html
+++ b/contribs/gmf/src/directives/partials/featurestyle.html
@@ -7,7 +7,7 @@
         class="form-control"/>
   </div>
   <div
-      ng-if="fsCtrl.type !== 'Text'"
+      ng-if="fsCtrl.type !== 'Text' && fsCtrl.type !== 'Circle'"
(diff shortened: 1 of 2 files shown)


sbrunner:
Effectively, thanks for the notice :-)

PR #3422, opened 2018-01-19
Fix ogc server encoding for IE11/Edge

llienher:
It seems that the ogcserver sent by IE11 and Edge is the following:
"ogcserver": "Main+PNG"

while the correct format sent by Chrome and Firefox is:
"ogcserver: "Main PNG"

The wrong format causes the plus sign to be encoded into %2B, and breaks the print request.
The condition I add is not doing anything in the cases when Chrome and Firefox are sending the right value.

### DIFF
--- a/src/services/print.js
+++ b/src/services/print.js
@@ -131,6 +131,12 @@ ngeo.Print = function(url, $http, ngeoLayerHelper) {
    * @private
    */
   this.printNativeAngle_ = true;
+
+  /**
+   * @type {boolean}
+   * @private
+   */
+  this.isMs = new URL('http://_/?a=+').searchParams.get('a') == '+';
 };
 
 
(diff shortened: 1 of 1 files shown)


src/services/print.js
```diff
@@ -291,7 +291,8 @@ ngeo.Print.prototype.encodeWmsLayer_ = function(arr, opacity, url, params) {
   const customParams = {'TRANSPARENT': true};
   if (url_url.searchParams) {
     for (const element of url_url.searchParams) {
-      customParams[element[0]] = element[1];
+      // work-around to normalize IE11 & Edge ogcserver encoding issue
+      element[0] === 'ogcserver' ? customParams[element[0]] = element[1].replace('+', ' ') : customParams[element[0]] = element[1];
```
ger-benjamin: `Main PNG` is something we define in the Admin interface. What if the name og the ogcserver is: `Mein schönes OGC server für "PNG" & nicht für "JPEG" !`. (Worst name I can find).

I means, that perhaps we should escape more. Can't you have a try ?

PR #3437, opened 2018-01-22
Add some ol requirements for Webpack (src)

ger-benjamin:
For ... I don't know. GSGMF-162 ?
It's only things in /src


ger-benjamin:
Useless work after #3454

PR #3441, opened 2018-01-23
Use the right assets file in production mode

sbrunner:


sbrunner:
Yes because `goog/asserts` is used in `olcs` ...

PR #3443, opened 2018-01-23
Examples gmf with webpack

ger-benjamin:
For: GSGMF-255


ger-benjamin:
(Post-merge comment: travis check example using webpack)

PR #3450, opened 2018-01-24
Handle 3D permalink

gberaudo:
Replaces #3448.


gberaudo:
@sbrunner, I added a commit about the extra require. Does it look good to you?

gberaudo:
Thanks for the review.

PR #3452, opened 2018-01-25
Add printNativeAngle parameter to the gmf-created print LayerGroup

llienher:
I set the parameter this way in a project controller via metadata 'printNativeAngle' in the admin interface applied to the layer group I want.

printNativeAngle value was not sent to the ngeo print component because gmf did not retrieve the original layerGroup but create a new one just before sending it to ngeo.

Desktop_alt updated to test this use case.

### DIFF
--- a/contribs/gmf/src/services/syncLayertreeMap.js
+++ b/contribs/gmf/src/services/syncLayertreeMap.js
@@ -163,6 +163,11 @@ gmf.SyncLayertreeMap.prototype.createGroup_ = function(treeCtrl, map,
   let layer = null;
   const isFirstLevelGroup = treeCtrl.parent.isRoot;
 
+  let printNativeAngle = true;
+  if (groupNode.metadata.printNativeAngle !== undefined) {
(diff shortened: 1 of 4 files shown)


contribs/gmf/src/services/syncLayertreeMap.js
```diff
@@ -213,16 +213,22 @@ gmf.SyncLayertreeMap.prototype.createLayerFromGroup_ = function(treeCtrl,
       ogcServer.credential ? 'use-credentials' : 'anonymous'
     );
     let hasActiveChildren = false;
+    let printNativeAngle = false;
     treeCtrl.traverseDepthFirst((ctrl) => {
+      console.log(ctrl);
```
sbrunner: To be removed

llienher:
Tested on my gmf internal instance, was working with the 'printNativeAngle' metadata added in the vars_project.

@sbrunner This is ready for review.

sbrunner:
Can you squash your commits...

llienher:
@sbrunner Done :-)

PR #3465, opened 2018-01-31
First step to fix the Webpack tests

sbrunner:
Related to GSGMF-195


test/spec/all.js
```diff
@@ -0,0 +1,28 @@
+// webpack: import 'jquery';
+// webpack: import 'angular';
+// webpack: import 'angular-mocks';
+import './beforeeach.js';
+import './search/createlocationsearchbloodhound.spec.js';
```
gberaudo: Can we find a mean to add all files automatically?

sbrunner:
+1 to do it later

PR #3469, opened 2018-01-31
Add require of goog.assert

ger-benjamin:
It will still fail because of:

Case 1: goog.exportSymbol
https://github.com/camptocamp/ngeo/blob/master/src/jstsexports.js#L14
Any idea ?

Case 2: goog.<type>
https://github.com/camptocamp/ngeo/blob/master/src/goog.asserts.js_#L139 (and brothers)
Can I use `typeof(value) === "number"` ?

What should I do ?

### DIFF
--- a/src/goog.asserts.js_
+++ b/src/goog.asserts.js_
@@ -136,9 +136,9 @@ exports.fail = function(opt_message, var_args) {
  * @throws {goog.asserts.AssertionError} When the value is not a number.
  */
 exports.assertNumber = function(value, opt_message, var_args) {
-  if (exports.ENABLE_ASSERTS && !goog.isNumber(value)) {
+  if (exports.ENABLE_ASSERTS && typeof value != 'number') {
     exports.doAssertFailure_(
-      'Expected number but got %s: %s.', [goog.typeOf(value), value],
+      'Expected number but got %s: %s.', [typeof value, value],
       opt_message, Array.prototype.slice.call(arguments, 2));
   }
   return /** @type {number} */ (value);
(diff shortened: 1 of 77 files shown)


sbrunner:
For me this file should be moved in option and will be used only by gcc or completely remove in the webpack migration ...
https://github.com/camptocamp/ngeo/blob/master/src/jstsexports.js#L14

sbrunner:
For the second case, yes, see:
https://github.com/camptocamp/saccas_suissealpine/blob/master/ui/suissealpine/static/js/goog/library/base.js#L1758-L1809

ger-benjamin:
For the last commit, I've taken what was done for saccas. But I've improvised methods for AssertFunction, nassertObject, assertArray and assertElement. Because the code was not existing.

src/goog.asserts.js_
```diff
@@ -191,9 +191,9 @@ exports.assertFunction = function(value, opt_message, var_args) {
  * @throws {goog.asserts.AssertionError} When the value is not an object.
  */
 exports.assertObject = function(value, opt_message, var_args) {
-  if (exports.ENABLE_ASSERTS && !goog.isObject(value)) {
+  if (exports.ENABLE_ASSERTS && (value === null || typeof value != 'object')) {
```
sbrunner: Array.isArray(value) && typeof value == 'object' && value != null || typeof value == 'function';

src/goog.asserts.js_
```diff
@@ -248,9 +248,9 @@ exports.assertBoolean = function(value, opt_message, var_args) {
  */
 exports.assertElement = function(value, opt_message, var_args) {
   if (exports.ENABLE_ASSERTS &&
-      (!goog.isObject(value) || value.nodeType != goog.dom.NodeType.ELEMENT)) {
+      (value === null || typeof value != 'object' || value.nodeType == undefined)) {
```
sbrunner: same

Issue #3278, opened 2017-12-15
Use of is_password_changed


svaabs:
In cgxp the purpose of the attribute _is_password_changed_ was to show a message and force the user to change his password if the value was set to false. How is this done in ngeo or is there no more use of this feature?

ger-benjamin:
It seems that this value exist in ngeo but is not used. So it looks like there is no implemented mechanism to force the user to change his password for now.

ybolognini:
To do

ger-benjamin:
I've added a `gmf-authentication-force-password-change` attribute. See #3346

svaabs:
Thanks for working on this. I had the chance to implement it on our gmf222-branch and unhappily it does not work as expected.
When I login with a user that has _is_password_changed_ set to false, I don't get any message (but I do expect one). As soon as the page is fully refreshed, the message appears.
Can you reproduce this on demo?

ger-benjamin:
:-/ Sadly I can reproduce the issue, I'll fix it soon

ger-benjamin:
Fixed with #3484

PR #3484, opened 2018-02-05
Debug login errors and force-password-change message

ger-benjamin:
First commit fix a $compile:multiling Linking Element Multiple Times angular error.

Second commit fix: https://github.com/camptocamp/ngeo/issues/3278
~~The problem is that the READY event was not emitted on login (but I need it, and it looks not problematic to emit it at the same time as the LOGIN event.) Then, the problem is that the "soft" reload of the page remove the notification, I've used a timeout to avoid this problem. If you know a better method...~~

### DIFF
--- a/contribs/gmf/examples/authentication.html
+++ b/contribs/gmf/examples/authentication.html
@@ -22,7 +22,7 @@
       This example shows how to use the <code>gmf-authentication</code>
       directive to insert an authentication panel in a GeoMapFish page.
     </p>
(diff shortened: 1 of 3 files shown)


ger-benjamin:
I thought that the reste was done in the abstract controller, and it this case, it should be hard to debug on your way @llienher . But in fact, the problem was just a reset function (in the auth directive) called before the event was launched. 

I've reworked the second commit. Now the solution is far more logical and suits me.

PR #3491, opened 2018-02-08
Fix gmf partials files

sbrunner:


gberaudo:
I was preparing the exact same PR!

PR #3455, opened 2018-01-25
[GSGMF-180] Remove module.requires

pfirpfel:

PR #3499, opened 2018-02-09
Activate and fix the tests that runs on webpack

sbrunner:
Based on #3455

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -835,6 +835,25 @@ $(EXTERNS_JQUERY): github_versions
 		--var "partials=ngeo:src gmf:contribs/gmf/src" \
 		--var "app=app" $< > $@
 
+.PRECIOUS: test/spec/templatecache.js
+test/spec/templatecache.js: buildtools/templatecache-webpack.mako.js \
+		.build/glob2.timestamp \
+		.build/python-venv/bin/mako-render \
+		$(NGEO_PARTIALS_FILES)
+	PYTHONIOENCODING=UTF-8 .build/python-venv/bin/mako-render \
+		--var "partials=ngeo:src" \
+		--var "app=./beforeeach.js" $< > $@
+
+.PRECIOUS: test/spec/gmftemplatecache.js
+test/spec/gmftemplatecache.js: buildtools/templatecache-webpack.mako.js \
+		.build/glob2.timestamp \
+		.build/python-venv/bin/mako-render \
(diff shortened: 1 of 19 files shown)


Makefile
```diff
@@ -835,6 +835,25 @@ $(EXTERNS_JQUERY): github_versions
...
+	PYTHONIOENCODING=UTF-8 .build/python-venv/bin/mako-render \
+		--var "partials=ngeo:src" \
+		--var "app=./beforeeach.js" $< > $@
+
+.PRECIOUS: test/spec/gmftemplatecache.js
+test/spec/gmftemplatecache.js: buildtools/templatecache-webpack.mako.js \
+		.build/glob2.timestamp \
+		.build/python-venv/bin/mako-render \
+		$(NGEO_PARTIALS_FILES) \
+		$(GMF_DIRECTIVES_PARTIALS_FILES)
```
gberaudo: This has been renamed.

test/spec/all.js
```diff
@@ -26,3 +28,4 @@ import './directives/filereader.spec.js';
 import './directives/scaleselector.spec.js';
 import './directives/layertree.spec.js';
 import './directives/grid.spec.js';
+import '../../contribs/gmf/test/spec/all.js';
```
gberaudo: Why do you add a dependency on gmf files?

test/spec/services/printutils.spec.js
```diff
@@ -2,11 +2,11 @@ goog.require('ngeo.print.Utils');
...
 describe('ngeo.print.Utils', () => {
 
-  let ngeoPrintUtils;
+  let ngeoPrintUtilsService;
 
   beforeEach(() => {
-    angular.mock.inject((_ngeoPrintUtils_) => {
-      ngeoPrintUtils = _ngeoPrintUtils_;
+    angular.mock.inject((ngeoPrintUtils) => {
+      ngeoPrintUtilsService = ngeoPrintUtils;
```
gberaudo: Is there a reason not to use anymore the underscore hack?

test/spec/all.js
sbrunner: Finally to don't have tow times Angular in the test scripts...

test/spec/services/printutils.spec.js
sbrunner: This is not for removing the _ but to not overriding the import in es6 mode...

PR #3500, opened 2018-02-09
Unify windows makefile

gberaudo:
With a tiny change of the main Makefile it looks possible to handle Windows build.
@kalbermattenm, would you be able to test this change before I merge it?


kalbermattenm:
I will test it this afternoon and give you a feedback.

Moreover, we discovered with @monodo that it was really easier to code for NGEO in a Linux environment, but I really appreciate being able to build it in Windows.

kalbermattenm:
@gberaudo, I tested this PR, and I had to apply following change for it to work:
image

`make` is not a Windows cmd, it is running inside a Cygwin environment and in Windows, the Python executable is `python.exe` whatever version you use. Thus if I leave the `-python python3`, then `make` tries to use the Python distribution which is installed with Cygwin and this one has not enough rights to write on the disk. So the only thing I did was to be sure to have Python3 in my path.

With these modifications, I am able to run `make dist` without any error.

But running `make serve` triggers an weird error and I do not know if that has something to do with Windows or if this is a general error:

image

gberaudo:
@kalbermattenm, I updated the PR, could you please test again?

kalbermattenm:
I'll test it again on Wednesday and give you a feedback.

kalbermattenm:
Weird, I get another error now when running `make dist` (after `make cleanall`)

image

gberaudo:
@kalbermattenm `make dist` is broken on all platforms. If it is the only issue, then I will merge.

gberaudo:
Travis passing, I am going to merge now. @kalbermattenm, let me know if there is something not working on your side and it will be fixed in a follow up PR.

PR #3506, opened 2018-02-12
Use istanbul coverage tool int the webpack coverage

sbrunner:


.travis.yml
```diff
@@ -99,6 +99,8 @@ jobs:
     - make lint
     - make check-examples-webpack
     - if [ ${IS_HOME} = TRUE ]; then JOB=publish-examples buildtools/deploy.sh; fi
+    after_success:
```
fredj: there's the same code at line 80, is is correct ?

.travis.yml
```diff
@@ -99,6 +99,8 @@ jobs:
...
+    - cat .build/coverage/lcov.info | node ./node_modules/coveralls/bin/coveralls.js
```
fredj: There's the same code line 80, is it correct?

sbrunner: Good question defectively, should I comment this line and we will uncomment it when we move on webpack?

fredj: let's keep it

PR #3508, opened 2018-02-12
Change type of binding to allow changes to mergetabs during runtime

pfirpfel:
see: GEO-977

Tested with desktop_alt. Changes to the bound object during runtime will be reflected in a new search.
This should not break existing code as long the feature is used like in the desktop_alt.

### DIFF
--- a/contribs/gmf/src/directives/displayquerygrid.js
+++ b/contribs/gmf/src/directives/displayquerygrid.js
@@ -90,7 +90,7 @@ gmf.displayquerygridComponent = {
     'removeEmptyColumnsFn': '&?gmfDisplayquerygridRemoveemptycolumns',
     'maxResultsFn': '&?gmfDisplayquerygridMaxresults',
     'maxRecenterZoomFn': '&?gmfDisplayquerygridMaxrecenterzoom',
-    'mergeTabsFn': '&?gmfDisplayquerygridMergetabs'
+    'mergeTabs': '=?gmfDisplayquerygridMergetabs'
   },
   templateUrl: gmfDisplayquerygridTemplateUrl
 };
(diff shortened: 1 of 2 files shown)


contribs/gmf/src/directives/displayquerygrid.js
```diff
@@ -90,7 +90,7 @@ gmf.displayquerygridComponent = {
     'removeEmptyColumnsFn': '&?gmfDisplayquerygridRemoveemptycolumns',
     'maxResultsFn': '&?gmfDisplayquerygridMaxresults',
     'maxRecenterZoomFn': '&?gmfDisplayquerygridMaxrecenterzoom',
-    'mergeTabsFn': '&?gmfDisplayquerygridMergetabs'
+    'mergeTabs': '=?gmfDisplayquerygridMergetabs'
```
ger-benjamin: `<` is better than `=` in this case. Because the change can only be from the main controller to this controller, never in the  other way. So that's useless to watch the change in this controller.

contribs/gmf/src/directives/displayquerygrid.js
```diff
@@ -274,7 +274,7 @@ gmf.DisplayquerygridController = function($injector, $scope, ngeoQueryResult, ng
 gmf.DisplayquerygridController.prototype.$onInit = function() {
   this.removeEmptyColumns_ = this['removeEmptyColumnsFn'] ? this['removeEmptyColumnsFn']() === true : false;
   this.maxRecenterZoom = this['maxRecenterZoomFn'] ? this['maxRecenterZoomFn']() : undefined;
-  this.mergeTabs_ = this['mergeTabsFn'] ? this['mergeTabsFn']() : {};
+  this.mergeTabs_ = this['mergeTabs'] ? this['mergeTabs'] : {};
```
ger-benjamin: You can remove this line and change the line 217 from:

````
   /**
    * @type {!gmfx.GridMergeTabs}
    * @private
    */
   this.mergeTabs_ = {};
````

to

````
   /**
    * @type {!gmfx.GridMergeTabs}
    */
   this.mergeTabs = {};
````
Then, adapt all `this.mergeTabs_` to `this.mergeTabs`;

File: contribs/gmf/src/directives/displayquerygrid.js [hunk 3]
```diff
@@ -274,7 +273,7 @@ gmf.DisplayquerygridController = function($injector, $scope, ngeoQueryResult, ng
 gmf.DisplayquerygridController.prototype.$onInit = function() {
   this.removeEmptyColumns_ = this['removeEmptyColumnsFn'] ? this['removeEmptyColumnsFn']() === true : false;
   this.maxRecenterZoom = this['maxRecenterZoomFn'] ? this['maxRecenterZoomFn']() : undefined;
-  this.mergeTabs_ = this['mergeTabs'] ? this['mergeTabs'] : {};
+  this.mergeTabs = this['mergeTabs'] ? this['mergeTabs'] : {};
```
ger-benjamin: Now you can completely remove this line. If nothing set the property, it will be the property declared in the constructor (so `{}`).

pfirpfel: true

PR #3510, opened 2018-02-12
Remove unused argument dimensions in gmf.Themes.prototype.getBgLayers

arnaud-morvan:


fredj:
note: the dimension argument is not used since 9cf2812b26ebb75089fb221297b76316758db146

PR #3512, opened 2018-02-13
 Deploy es6 version of ngeo 

sbrunner:


ger-benjamin:
6000 commits by the way :tada:

PR #3513, opened 2018-02-13
Fix the application module

sbrunner:


sbrunner:
Thanks for the review :-)

Issue #3524, opened 2018-02-14
Remove query service


ger-benjamin:
The query file is deprecated and unused since ngeo 2.2, right ? It is replaced by the querent and mapQuerent services ?

I want to remove it.

https://github.com/camptocamp/ngeo/blob/2.2/src/services/query.js

fredj:
fixed with #3529 and #3530

PR #3529, opened 2018-02-15
Remove useless ngeo Query Service

ger-benjamin:
/!\ For 2.2
Fix https://github.com/camptocamp/ngeo/issues/3524
Was replaced by ngeo Querent and MapQuerent

Query works as before on apps and examples.

The gmf QueryManager was not used anymore and was based on ngeoService (marked as deprecated and not more used)


ger-benjamin:
@rmichaelis @llienher It is still used by luxembourg, or by epfl ?
I don't think so, but to be sure that you are aware of this change...

llienher:
Looks like the EPFL doesn't use this file, so fine for me.

fredj:
> @fredj or by schweizmobil ?

not used by schweizmobil

fredj:
thanks @ger-benjamin

PR #3533, opened 2018-02-16
Update openlayers version used in travis

fredj:


.travis.yml
```diff
@@ -81,7 +81,7 @@ jobs:
     - JOB=webpack
     script:
     - npm install
-    - npm install github:openlayers/openlayers#2ee34b825c4b5d6edd3569e10b3f63510754e012
+    - npm install github:openlayers/openlayers#064c25f
```
sbrunner: Shouldn't we use master to directly fix the changes?

fredj: yes, good idea

fredj:
travis error is not related, merging

PR #3541, opened 2018-02-19
Up master with 2.2

ger-benjamin:
git merge origin/2.2


ger-benjamin:
I've copied manually the print and the authentication change because git can't follow the file over the changes (moved from src/directive to src/component/)

PR #3550, opened 2018-02-21
Remove unused vars 3

ger-benjamin:
For https://jira.camptocamp.com/browse/GSGMF-261

Still to fix (in a next PR):

90% -> Requires in the tests files,
10% -> Some require in apps (will be easy when apps will be transformed into modules)
~39 / 300+ warnings left

Then there is also some warnings (~20) in the "make lint" result without webpack.

### DIFF
--- a/contribs/gmf/apps/desktop/js/controller.js
+++ b/contribs/gmf/apps/desktop/js/controller.js
@@ -10,9 +10,6 @@ goog.require('app');
 goog.require('gmf.controllers.AbstractDesktopController');
 goog.require('ngeo.proj.EPSG2056');
 goog.require('ngeo.proj.EPSG21781');
-
-/** @suppress {extraRequire} */
-goog.require('ngeo.googlestreetview.component');
 goog.require('ol');
 
 app.desktop.module = angular.module('AppDesktop', [
--- a/contribs/gmf/apps/desktop_alt/js/controller.js
+++ b/contribs/gmf/apps/desktop_alt/js/controller.js
@@ -9,13 +9,16 @@ goog.provide('app.desktop_alt.Controller');
 goog.require('app');
 goog.require('gmf.controllers.AbstractDesktopController');
 /** @suppress {extraRequire} */
-goog.require('gmf.import.importdatasourceComponent');
+goog.require('gmf.import.module');
+/** @suppress {extraRequire} */
+goog.require('ngeo.googlestreetview.module');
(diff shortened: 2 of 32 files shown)


contribs/gmf/src/controllers/AbstractDesktopController.js
```diff
@@ -5,11 +5,13 @@ goog.require('gmf.controllers.AbstractAppController');
 goog.require('gmf.contextualdata.module');
 goog.require('gmf.drawing.module');
 goog.require('gmf.editing.module');
+goog.require('gmf.import.module'); // only for desktop_alt
```
sbrunner: Than is should be in desktop alt, no?

contribs/gmf/src/controllers/AbstractDesktopController.js
```diff
@@ -5,11 +5,13 @@ goog.require('gmf.controllers.AbstractAppController');
...
 goog.require('gmf.permalink.shareComponent');
 goog.require('gmf.print.component');
 goog.require('gmf.profile.module');
 goog.require('gmf.raster.component');
 goog.require('ngeo.draw.features');
+goog.require('ngeo.googlestreetview.module');
```
sbrunner: same?

ger-benjamin: Previously, both was importing googlestreetview but perhaps that was an error.

contribs/gmf/src/controllers/AbstractDesktopController.js
ger-benjamin: Yes but desktop_alt is not a module for now.
But you're right, I will remove it from here

contribs/gmf/src/controllers/AbstractDesktopController.js
sbrunner: As I remember streetview is only in the desktop alt, than he should be imported only in desktop alt :-)

contribs/gmf/src/controllers/AbstractDesktopController.js
sbrunner: ... :-/ effectively finally it should be, as the desktop one :-)

ger-benjamin:
That's nothing important.
~39 / 300+ = that was before the first "remove unused vars" PR, and that's the third PR. Compared to the master, that's more like ~39/~50 warn left

sbrunner:
OK, thanks :-)

PR #3553, opened 2018-02-21
Backgroundlayerselector opacityslider

llienher:
Add a way to get a static background layer linked to an opacity slider, via a parameter to include in the interface. I.E:
```
  /**
   * @type {string}
   * @export
   */
  this.bgOpacityOptions = 'Test aus Olten';
```
The string is the layer label (WMS name from admin interface).

The component is defined as follow in the html:
```
<gmf-backgroundlayerselector
    gmf-backgroundlayerselector-map="::mainCtrl.map"
    gmf-backgroundlayer-opacity-options="::mainCtrl.bgOpacityOptions"
    class="dropdown-menu">
</gmf-backgroundlayerselector> 
```
`gmf-backgroundlayer-opacity-options` is facultative and if ignored, the application will just set as it was without the feature.

First commit is remove correctly the dimension used in the past but not done completly.
The app now use a group containing the active layer and the one used as overlay for the opacity slider. This group is called "background" and is always index 0 with group "data" at index 1.
Tests are updated to this new structures.

### DIFF
--- a/contribs/gmf/src/directives/backgroundlayerselector.js
+++ b/contribs/gmf/src/directives/backgroundlayerselector.js
@@ -140,6 +140,12 @@ gmf.BackgroundlayerselectorController = function($scope, ngeoBackgroundLayerMgr,
    */
   this.backgroundLayerMgr_ = ngeoBackgroundLayerMgr;
 
+  /**
+   * @type {boolean}
+   * @private
+   */
+  this.isThemeLoadedOnce_ = false;
+
   this.listenerKeys_.push(ol.events.listen(
     this.backgroundLayerMgr_,
     ngeo.BackgroundEventType.CHANGE,
@@ -170,7 +176,11 @@ gmf.BackgroundlayerselectorController.prototype.handleThemesChange_ = function()
   this.gmfThemes_.getBgLayers().then((layers) => {
     this.bgLayers = layers;
 
-    if (this.opacityOptions !== undefined) {
+    if (this.opacityOptions !== undefined && !this.isThemeLoadedOnce_) {
+      // Angular initialize twice the component, this limit to set
+      // only once the background layer used by opacity slider
+      this.isThemeLoadedOnce_ = true;
+
       const opacityLayer = layers.find(layer => layer.get('label') === this.opacityOptions);
       if (opacityLayer !== undefined) {
         this.setOpacityBgLayer(opacityLayer);
@@ -190,7 +200,7 @@ gmf.BackgroundlayerselectorController.prototype.handleThemesChange_ = function()
  * @param {?number} val The opacity.
  * @returns {number} The background layer opacity.
  */
-gmf.BackgroundlayerselectorController.prototype.getSetOpacityBgLayer = function(val) {
+gmf.BackgroundlayerselectorController.prototype.getSetBgLayerOpacity = function(val) {
   if (val) {
     this.opacityLayer.setOpacity(val);
   }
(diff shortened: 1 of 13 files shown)


sbrunner:
What's this strange pull request for?

llienher:
@sbrunner EPFL and Schwytz want back the opacity slider for a background layer as it was available in CGXP. Ngeo implementation is extending the current background selector.

contribs/gmf/src/directives/backgroundlayerselector.js
```diff
@@ -166,9 +175,34 @@ gmf.BackgroundlayerselectorController.prototype.$onInit = function() {
 gmf.BackgroundlayerselectorController.prototype.handleThemesChange_ = function() {
   this.gmfThemes_.getBgLayers().then((layers) => {
     this.bgLayers = layers;
+
+    if (this.opacityOptions !== undefined && !this.isThemeLoadedOnce_) {
+      this.isThemeLoadedOnce_ = true;
```
ger-benjamin: Please add a comment to explain that

sbrunner:
We shouldn't do that in 2.2...

src/services/backgroundlayermgr.js
```diff
@@ -150,6 +162,36 @@ ngeo.BackgroundLayerMgr.prototype.set = function(map, layer) {
...
+/**
+ * Set a background layer used by the opacity slider.
+ * @param {ol.Map} map The map.
+ * @param {ol.layer.Base} layer The opacity background layer.
+ */
+ngeo.BackgroundLayerMgr.prototype.setOpacityBgLayer = function(map, layer) {
+  layer.setOpacity(0);
+  layer.setVisible(true);
+  const bgGroup = this.ngeoLayerHelper_.getGroupFromMap(map, gmf.BACKGROUNDLAYERGROUP_NAME);
+  bgGroup.getLayers().insertAt(1, layer);
```
gberaudo: You should remove all the extra layers (> 1) using a for (i=1; i<  bgGroup.getLayers().length) and bgGroup.getLayers().removeAt(i).

Then the method will be callable several times.

contribs/gmf/src/directives/backgroundlayerselector.js
```diff
@@ -166,9 +175,37 @@ gmf.BackgroundlayerselectorController.prototype.$onInit = function() {
 gmf.BackgroundlayerselectorController.prototype.handleThemesChange_ = function() {
   this.gmfThemes_.getBgLayers().then((layers) => {
     this.bgLayers = layers;
+
+    if (this.opacityOptions !== undefined && !this.isThemeLoadedOnce_) {
```
gberaudo: This method is called several times, it should be normal.
I would avoid messing with the ordering inside `getBgLayers` so I suggest to duplicate the array before changing it (use slice without arguments).
Then remove the `isThemeLoadedOnce` condition.

gberaudo:
@ybolognini, we are going to migrate EPFL to 2.3 starting this week.
If we merge in 2.2 it will need some work and delay to port it to 2.3.

Can we instead merge it in 2.3?

jwkaltz:
@gberaudo it is promised to Schwyz for an upcoming 2.2 version, too.

gberaudo:
@llienher, there is a test failing:
```
e 64.0.3282 (Linux 0.0.0) ngeo.BackgroundLayerMgr #set unsets the background layer FAILED
	TypeError: Cannot read property 'setZIndex' of null
	    at ngeo.BackgroundLayerMgr.set (src/services/backgroundlayermgr.js:9:2479)
	    at UserContext.it (test/spec/services/backgroundlayermgr.spec.js:61:30)
```

PR #3562, opened 2018-02-22
Make sure the lines from the contextual menu are never wrapped

### LINKED ISSUE #3496: Contextual data popup size
There is something strange with the popup size of the contextual data window.

The line containing the latitude / longitude data in DMS coordinates is always just a bit to small. We have the same behavior on the SITN project.

I did not see any parameter to set the width nor height, thus is it some kind of compiling bug ?

image

fredj:
fixes #3496

### DIFF
--- a/contribs/gmf/less/contextualdata.less
+++ b/contribs/gmf/less/contextualdata.less
@@ -4,5 +4,10 @@
 
   table {
     width: 100%;
+    tr {
+      td {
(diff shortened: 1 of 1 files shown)


fredj:
The width of the window is big enough but firefox still put the coordinates on a new line

fredj:
let's merge; we can still make the window bigger if needed

PR #3563, opened 2018-02-22
Stop using controllerAs in examples

fredj:
and better comments indentation


fredj:
> Do a last `git grep qrCtrl` before to merge

done: nothing left

PR #3566, opened 2018-02-23
Use the matrixSet value from the WMTS layer config

### LINKED ISSUE #3555: ngeo doesn't take into account "tilematrixset" of admin for WMTS layers
Version 2.2.2

When configuring a WMTS layer in admin interface and specifying the tilematrixset (for a layer with 2 tilematrixset available), this value is not taken into account in application, and the default tilematrixset is used. To reproduce : go to https://preprod.cartoriviera.ch/new/ -> the tiles of layer asitvd.fond_cadastral are loaded with tilematrixset 21781 :  

https://ows.asitvd.ch/wmts?layer=asitvd.fond_cadastral&style=default&tilematrixset=21781&Service=WMTS&Request=GetTile&Version=1.0.0&Format=image%2Fpng&TileMatrix=19&TileCol=24&TileRow=38&DIM1=default&ELEVATION=0

But the tiles should have 2056 instead of 21781 : 

https://ows.asitvd.ch/wmts?layer=asitvd.fond_cadastral&style=default&tilematrixset=2056&Service=WMTS&Request=GetTile&Version=1.0.0&Format=image%2Fpng&TileMatrix=19&TileCol=24&TileRow=38&DIM1=default&ELEVATION=0

So ngeo does a reprojection and the final quality is bad. The admin config is the followin : 
image

Here is the getcapabilities : https://ows.asitvd.ch/wmts?service=wmts&request=getcapabilities

fredj:
fixes #3555

### DIFF
--- a/src/services/layerHelper.js
+++ b/src/services/layerHelper.js
@@ -106,12 +106,13 @@ ngeo.LayerHelper.prototype.createBasicWMSLayer = function(sourceURL,
  * as key 'capabilitiesStyles' as param of the new layer.
  * @param {string} capabilitiesURL The getCapabilities url.
  * @param {string} layerName The name of the layer.
+ * @param {string=} opt_matrixSet Optional WMTS matrix set.
  * @param {Object.<string, string>=} opt_dimensions WMTS dimensions.
  * @return {angular.$q.Promise.<ol.layer.Tile>} A Promise with a layer (with source) on success,
  *     no layer else.
  * @export
  */
-ngeo.LayerHelper.prototype.createWMTSLayerFromCapabilitites = function(capabilitiesURL, layerName, opt_dimensions) {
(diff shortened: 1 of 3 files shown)


src/services/layerHelper.js
```diff
@@ -106,12 +106,13 @@ ngeo.LayerHelper.prototype.createBasicWMSLayer = function(sourceURL,
  * as key 'capabilitiesStyles' as param of the new layer.
  * @param {string} capabilitiesURL The getCapabilities url.
  * @param {string} layerName The name of the layer.
+ * @param {string|undefined} opt_matrixSet Optional WMTS matrix set.
```
sbrunner: I don't remember but the type shouldn't be `string=`?

fredj: yes, will change

PR #3583, opened 2018-02-27
Can set preload value of Tiles (default to 0)

ger-benjamin:
Improve performance.
(See: https://jira.camptocamp.com/projects/GEO/queues/custom/52/GEO-891)

### DIFF
--- a/src/services/layerHelper.js
+++ b/src/services/layerHelper.js
@@ -53,14 +53,6 @@ ngeo.LayerHelper.REFRESH_PARAM = 'random';
 
 
 /**
- * Load tiles up to preload levels. By default preload is 0, which means only load visible level.
- * See preload value in documentation for ol.Layer.Tile.
- * @const
- */
-ngeo.LayerHelper.TILE_PRELOAD = 0;
-
-
-/**
  * Create and return a basic WMS layer with only a source URL and a comma
  * separated layers names (see {@link ol.source.ImageWMS}).
  *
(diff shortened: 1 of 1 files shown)


fredj:
`preload` option should simply be removed

ger-benjamin:
(done with the Github editor ! )

sbrunner:
I think this this bag :-(, this is to have better performance when we zoom out ...

sbrunner:
In this ticker we don't see anything about preload?
Whant's the deference in therm of loading?

ger-benjamin:
Even on bad internet connection, the change is near invisible on the map, and the tiles around (90% of what you have under your eyes) must be loaded in every cases.

But on the initial load, that a lot of useless tile to load. In the cases of Basel, we load 2.3 Mo of tiled on a total of 4.5Mo (total size of initial load).

If you want, I can (re-)add a configuration value for this parameter. But with fredj  and gberaudo , we though that's useless.

sbrunner:
I just do a test with slow 3g,
An in booth case (tested on github.io 2.2/master) I get all the visible background in about 55s.
and when I zoom out on 2.2 it takes 23s. and on master 13s.
Then in therm of performance we should keep it, and if it's relay wanted we should add an option and the preload should be infinity by default ...

ger-benjamin:
>  I get all the visible background in about 55s.
More like in 55s for master and 51s in 2.2 for me. And in Basel, they load two background in the same time (one for basel, their tiles, and another for the whole switzerland (Swisstopo)). So, that make ~6-8 sec difference. And that's on the startup, so the slower part of our application. Each optimization on this part is interesting for our users.

But **ok**. I'll do a configuration value for that, defined to Infinity by default.

PR #3587, opened 2018-02-27
Don't load mainmodule in examples

fredj:
https://jira.camptocamp.com/browse/GSGMF-219


ger-benjamin:
(I've restarted the build, it was timed out)

sbrunner:
Why?

fredj:
Because we want the examples to load only what they need; not everything from gmf

sbrunner:
This is to have a big part in the commons part and It's also what we want ...

PR #3593, opened 2018-02-28
Add missing semicolon in css import

fredj:


fredj:
> Does it correct for you to have this imports in the examples files, shoudn't they be in the commons.js?

yes good idea, since they are imported in all examples, they can be moved into commons.js

sbrunner:
Thanks :-)

PR #3622, opened 2018-03-06
Use escaped label as id for querygrid, because id are not unique

ger-benjamin:
FIX: GEO-1022

The id in the querysources is not always unique (not in case of multiple server layers in one ol layer). And the querygrid already use label as ID to be able to merge the layers and because that make no sens to have multiple tabs with the same name, so to be based on id is a bad idea. Having a mix of id/label is also a bad idea.

Now all is based on labels. Note that we must escape the label to be able to use and select it as class or id (and avoid to fall again in this issue: https://github.com/camptocamp/ngeo/issues/3273)

### DIFF
--- a/contribs/gmf/src/directives/displayquerygrid.js
+++ b/contribs/gmf/src/directives/displayquerygrid.js
@@ -212,6 +212,7 @@ gmf.DisplayquerygridController = function($injector, $scope, ngeoQueryResult, ng
 
   /**
    * @type {!gmfx.GridMergeTabs}
+   * @export
    */
   this.mergeTabs = {};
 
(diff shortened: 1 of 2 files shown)


contribs/gmf/src/directives/displayquerygrid.js
```diff
@@ -212,6 +212,7 @@ gmf.DisplayquerygridController = function($injector, $scope, ngeoQueryResult, ng
 
   /**
    * @type {!gmfx.GridMergeTabs}
+   * @export
```
pfirpfel: Just a note: this fixes #3508

PR #3635, opened 2018-03-08
Fix ngeo examples

pfirpfel:
To be fixed later:

- asitvd.html: no layers for me (probably Olten-Problem)
- backgroundlayerdropdown.html: missing bootstrap
- colorpicker.html: (was broken before)
- datepicker.html: missing css import
- drawfeature.html: Draw Circle + DrawAzimut broken
- googlestreetview.html: google streetmap api is not loaded, need external loader like https://github.com/yuanqing/load-google-maps-api
- layerorder.html: sortable handles have a width of 0, no way to grab them. Current release broken as well.
- modifyrectangle.html: runtime error
- modifycircle.js: ol_geom_Polygon.fromCircle is not a function (despite ol/geom/Polygon.js being imported)
- simple3d.html: cesium import

### DIFF
--- a/buildtools/webpack.ngeoexamples.js
+++ b/buildtools/webpack.ngeoexamples.js
@@ -26,11 +26,18 @@ for (const filename of ls('examples/*.html')) {
 
 // move data folder
 plugins.push(new CopyWebpackPlugin(
-  [{
-    from: 'examples/data/*',
-    to: 'data',
-    flatten: true
-  }]
+  [
+    {
+      from: 'examples/data/*',
+      to: 'data',
+      flatten: true
+    },
+    {
+      from: 'examples/partials/*',
+      to: 'partials',
+      flatten: true
+    }
+  ]
 ));
 
 module.exports = {
--- a/buildtools/webpack.ngeoexamples.js
+++ b/buildtools/webpack.ngeoexamples.js
@@ -2,6 +2,7 @@ const path = require('path');
 const ls = require('ls');
 const webpack = require('webpack');
 const HtmlWebpackPlugin = require('html-webpack-plugin');
+const CopyWebpackPlugin = require('copy-webpack-plugin');
 
 const plugins = [];
 const entry = {};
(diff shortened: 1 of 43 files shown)


buildtools/webpack.ngeoexamples.js
```diff
@@ -2,23 +2,44 @@ const path = require('path');
...
   );
 }
 
+// move data folder
+plugins.push(new CopyWebpackPlugin(
+  [
+    {
+      from: 'examples/data/*',
+      to: 'data',
+      flatten: true
```
sbrunner: What is it for?

pfirpfel: Some examples use files from this folders. These must be copied to the compiled version for the examples to be usable.

sbrunner: Yes, but I just speak about the flatten...

pfirpfel: Without the flatten it preserves the source folder structure.

Instead of `data/file.ext` I get `data/examples/data/file.ext`

There could be another combination of parameters to get the same result, but this seems to work.

sbrunner: OK, thanks for your response :-)

buildtools/webpack.ngeoexamples.js
```diff
@@ -2,23 +2,44 @@ const path = require('path');
...
+    },
+    {
+      from: 'examples/partials/*',
```
sbrunner: And this one is relay needed?
The partials are loaded throw requires...
It's not working?

pfirpfel: I tested it again, and for the partials inside the examples, this is needed. Those HTML files were loaded during runtime and not included by webpack.

sbrunner: Looks strange, but I cans see it later...

PR #3639, opened 2018-03-09
Don't remove the spaces in template HTML

fredj:
Otherwise it breaks the fontawesome placement.

the doc for the `remove_all_empty_space` option says: This is almost guaranteed to break your HTML :-)

### DIFF
--- a/buildtools/templatecache.mako.js
+++ b/buildtools/templatecache.mako.js
@@ -23,8 +23,6 @@
               content = htmlmin.minify(
                   content,
                   remove_comments=True,
-                  remove_empty_space=True,
-                  remove_all_empty_space=True,
                   remove_optional_attribute_quotes=False,
               )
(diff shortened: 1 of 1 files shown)


buildtools/templatecache.mako.js
```diff
@@ -23,8 +23,6 @@
               content = htmlmin.minify(
                   content,
                   remove_comments=True,
-                  remove_empty_space=True,
-                  remove_all_empty_space=True,
```
sbrunner: Your sure that you should remove booth?

fredj: yes, I tried to remove only remove_all_empty_space but it didn't work

sbrunner: ???

sbrunner: OK, thanks :-)

Issue #3468, opened 2018-01-31
Filter: get data only works if layer is displayed in map


svades:
Project:
* GeoMapFish version: 2.2.0, can also be reproduced on 2.2.2
* URL: https://map.geo.test.bs.ch/ngeodesktop

Steps to reproduce the issue:
* open theme "Wald, Flora, Fauna", don't load/activate any layers
* open filter for "Bäume"
* choose some filter condition (e.g. spatial filter)
* choose get data
* same issue could also be reproduced on demo https://geomapfish-demo.camptocamp.net/2.2

Result: nothing happens!

Expected result: the filtered data should appear in the displayquerygrid

At least something should happen. Thinkable possibilitites:
- the filtered data appear in the displayquerygrid without the layer being activated/loaded in the map and is highlighted in the map (like cgxp)
- the layer is only chooseable in the filterpanel if the layer is activated/displayed in the map
- the layer gets activated/displayed by choosing "get data"

yjacolin:
on Demo, you can [test on the GMF 2.2 instance](https://geomapfish-demo.camptocamp.net/2.2/theme/Demo?lang=fr&baselayer_ref=OSM%20map&theme=Demo&tree_groups=OSM%20functions&map_x=632484&map_y=185457&map_zoom=3&tree_group_layers_Filters=osm_open&tree_group_layers_OSM%20functions=osm_open) (filter on "Ouverture OSM")

There is something similar with "apply filter" on background layer. User should know that the request has been sent and why nothing happen.

eleu:
reported to jira

ger-benjamin:
(reopen, wrong issue)

sbrunner:
Managed on JIRA

PR #3649, opened 2018-03-12
Print - Selected scale must not be auto. adapted

ger-benjamin:
Fix: GEO-1051

Summary: 
If the scale 1:1000 is never the "best calculated scale" for the print, the scale will be automatically adapted. Than means, we will never be able print in 1:1000. This PR correct that.

@jwkaltz can you test if that's the expected behaviors ? I've taken the same client-scale/print-scale than your, as explained in your comment, in the ticket.
@fredj can you valid the code ?

Demo here: https://gmf-test.sig.cloud.camptocamp.net/bge

### DIFF
--- a/contribs/gmf/src/directives/print.js
+++ b/contribs/gmf/src/directives/print.js
@@ -313,6 +313,12 @@ gmf.PrintController = function($rootScope, $scope, $timeout, $q, $injector,
   this.pointerDragListenerKey_;
 
   /**
+   * @type {ol.EventsKey}
+   * @private
+   */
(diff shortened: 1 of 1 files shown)


ger-benjamin:
@jwkaltz Tested by @eleu . I merge

jwkaltz:
@ger-benjamin I have updated https://gmf-test.sig.cloud.camptocamp.net/wkaltz22 and the test case I had works now (1:1000), thanks.

ger-benjamin:
Also close: #3468

PR #3663, opened 2018-03-15
Order print layers by z-index

llienher:
As the map is ordered by z-index with OL, the print need to replicate this otherwise layers are unsorted and given to mapfish in that state...

### DIFF
--- a/src/services/print.js
+++ b/src/services/print.js
@@ -7,6 +7,7 @@ goog.require('ngeo');
 goog.require('ngeo.LayerHelper');
 goog.require('ngeo.utils');
 goog.require('ol.color');
+goog.require('ol.array');
 goog.require('ol.format.GeoJSON');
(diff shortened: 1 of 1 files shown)


src/services/print.js
```diff
@@ -225,6 +225,17 @@ ngeo.Print.prototype.encodeMap_ = function(map, scale, object) {
   goog.asserts.assert(mapLayerGroup);
   this.printNativeAngle_ = !(mapLayerGroup.get('printNativeAngle') === false);
   let layers = this.ngeoLayerHelper_.getFlatLayers(mapLayerGroup);
+
+  // Sort the layer by ZIndex
+  layers.sort((layer_a, layer_b) => {
```
gberaudo: See https://github.com/openlayers/openlayers/blob/8f0ffe2043bc60e7caa86c00c28d636859c1fee7/src/ol/array.js#L200

PR #3682, opened 2018-03-20
Doc for gmf FTSearch's similarity

ger-benjamin:
Side PR of https://github.com/camptocamp/c2cgeoportal/pull/3565

For GEO-908

### DIFF
--- a/contribs/gmf/src/directives/search.js
+++ b/contribs/gmf/src/directives/search.js
@@ -65,7 +65,7 @@ gmf.module.value('gmfSearchTemplateUrl',
  *      <script>
  *        (function() {
  *          let module = angular.module('app');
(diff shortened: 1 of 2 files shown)


contribs/gmf/src/directives/search.js
```diff
@@ -95,6 +95,14 @@ gmf.module.value('gmfSearchTemplateUrl',
  *        })();
  *      </script>
  *
+ * The 'fulltextsearchUrl' value in the examples above set three "_query" parameters: "limit",
+ * "partitionlimit" and "ranksystem". For this last one "similarity" is the only effective value. It's used to
+ * order your search results with another rank system. Read the full-text search c2cgeoportal documentation
+ * to know more.
```
yjacolin: You should explain whait is the default value (if ranksystem is not set)

PR #3701, opened 2018-03-23
Add the import datasource module to desktop interface

llienher:
Fixes GSGMF-386

Was only present in desktop_alt, but it is a standard feature, so I added it to desktop.

### DIFF
--- a/contribs/gmf/apps/desktop/js/controller.js
+++ b/contribs/gmf/apps/desktop/js/controller.js
@@ -8,6 +8,7 @@ goog.provide('app.desktop.Controller');
 
 goog.require('app');
 goog.require('gmf.controllers.AbstractDesktopController');
+goog.require('gmf.import.module');
 goog.require('ngeo.proj.EPSG2056');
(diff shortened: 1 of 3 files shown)


contribs/gmf/apps/desktop/js/controller.js
```diff
@@ -8,15 +8,30 @@ goog.provide('app.desktop.Controller');
...
 goog.require('ngeo.proj.EPSG21781');
 goog.require('ol');
 
 app.desktop.module = angular.module('AppDesktop', [
   app.module.name,
   gmf.controllers.AbstractDesktopController.module.name,
+  gmf.import.module.name,
 ]);
 
+app.desktop.module.value('gmfExternalOGCServers', [{
```
sbrunner: What's those gmfExternalOGCServers ?

llienher: It is the prefilled value, I just took what was done in desktop_alt.

PR #3714, opened 2018-03-27
Load CSS through webpack

ger-benjamin:
For GSGMF-391

Should I commit somethings elsewhere (add dependencies ?, describe the "adapt 'less' things") ? See  https://jira.camptocamp.com/browse/GSEPF-47 to know what I've done directly in the final project.

### DIFF
--- a/docs/how_to_migrate_from_2.2_to_2.3.md
+++ b/docs/how_to_migrate_from_2.2_to_2.3.md
@@ -170,6 +170,7 @@ If you have no custom AngularJS element in your code, you may start directly wit
 
  1. [Transform your components to modules](#transform-your-components-to-modules)
  1. [Transform your applications to modules](#transform-your-applications-to-modules)
+ 1. [Adapt paths for 'less'](#Adapt-paths-for-less)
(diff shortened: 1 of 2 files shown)


ger-benjamin:
@gberaudo what about the less that I've adapted ? should it be documented ?

gberaudo:
Yes you should document that CSS must be handled by webpack.

docs/how_to_migrate_from_2.2_to_2.3.md
```diff
@@ -292,6 +293,16 @@ these values to the module of your controller like this:
 </script>
 ```
 
+### Adapt paths for 'less'
+
+We no more load all `.less` files in each application. You must import your `<interface.less>` in each
+application controller you have by somthing like: `import '../../less/<interface>.less'`.
```
sbrunner: `We no more load all `.less` files in each application. ` it was already the case...

fredj: `something` instead of `somthing`

PR #3719, opened 2018-03-28
Add routing service and directive

fredj:
work in progress, replaces #2669


src/misc/filters.js
```diff
@@ -321,3 +321,108 @@ ngeo.misc.filters.trustHtmlFilter = function($sce) {
...
+    HOURS: Symbol('hours'),
+    DAYS: Symbol('days')
+  });
+
+  /**
+   * @param {number} amount Amount of time.
+   * @param {symbol} unit Unit of time.
+   * @return {string} formatted and translated string
+   */
+  const pluralize = function(amount, unit) {
```
sbrunner: angular-gettext already have something to manage plurials, it's not working?

fredj:
everything works except the tests

src/misc/filters.js
fredj: yes, see bfc751a185488fed9362bf82b844e82b00cc575a

PR #3720, opened 2018-03-28
Fix closable attribute in modalcomponent

llienher:
Now the default value when nothing is provided will always allow to close a modal from the "close button" or clicking outside the modal.

### DIFF
--- a/src/message/modalcomponent.js
+++ b/src/message/modalcomponent.js
@@ -110,7 +110,7 @@ ngeo.message.modalComponent.Controller_ = class {
   }
 
   $onInit() {
-    this.closable = this.closable === true;
+    this.closable = this.closable !== false;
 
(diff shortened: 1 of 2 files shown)


src/message/modalcomponent.js
```diff
@@ -110,11 +110,11 @@ ngeo.message.modalComponent.Controller_ = class {
   }
 
   $onInit() {
-    this.closable = this.closable === true;
+    this.closable = this.closable === false;
```
sbrunner: `=== true` or `!== false` !

ger-benjamin: `!== false` would be better, effectively
(then keep the `!` in the if below)

llienher: @sbrunner @ger-benjamin Thanks, done.

PR #3733, opened 2018-04-03
Update FileSaver.js to version 1.3.8

llienher:
Fix GSGMF-399


sbrunner:
I hope that you back-port this change in c2cgeoportal?

PR #3780, opened 2018-04-18
Add missing webpack import for dropdown

llienher:
Not sure if it is added to the right file, but it is missing when building the app with webpack (consequence is EPFL theme selector is not visible when clicking the theme button).


fredj:
it should be in https://github.com/camptocamp/ngeo/blob/master/contribs/gmf/src/theme/selectorComponent.js

PR #3784, opened 2018-04-19
Fix tooltips in layertree and remove useless tooltips on buttons

llienher:


contribs/gmf/src/layertree/component.html
```diff
@@ -180,7 +183,6 @@
     <a
       class="fa fa-align-left gmf-layertree-legend-button"
       ng-if="::gmfLayertreeCtrl.getLegendsObject(layertreeCtrl) && layertreeCtrl.node.metadata.legend"
-      title="{{'Show/hide legend'|translate}}"
```
sbrunner: Where is this one?

PR #3785, opened 2018-04-19
Add example for ngeoDisplaywindow component

### LINKED ISSUE #3778: Add missing example for ngeo.message.displaywindowComponent
2.3 introduces ngeo.message.displaywindowComponent.
Unfortunately, there is no example for it. In comparison, the popup component (which is being replaced by the displaywindow component) has one.

ger-benjamin:
And Debug ngeo message displaywindowComponent (a lot of work when 'out of apps case')

FIX #3778 

Still any FontAwesome icon, but it's like in other examples

### DIFF
--- a/examples/displaywindow.css
+++ b/examples/displaywindow.css
@@ -1,0 +1,56 @@
+.windowcontainer {
+  background-color: #fbf5ef;
+  border: solid 1px black;
+  position: relative;
+  height: 100%;
+  width: 100%;
+  margin: 1rem;
+}
+
+.animation-container {
+  position: relative;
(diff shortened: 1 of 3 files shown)


ger-benjamin:
Ready to review, description updated

src/message/displaywindowComponent.js
```diff
@@ -209,7 +190,7 @@ ngeo.message.displaywindowComponent.Controller_ = class {
   get style() {
     return {
       'height': this.height || '240px',
-      'width': this.height || '240px'
```
fredj: nice catch !

PR #3797, opened 2018-04-23
[WIP] Transform code to es6

gberaudo:
This PR is about moving the code base to ES6 modules and switching to a webpack buildchain.
The Google Closure Compiler buildchain is abandonned (at least for now).
- [x] make Travis happy;
- [ ] rebase and reaply transform.


karma-conf.js
```diff
@@ -7,7 +7,6 @@ var isDebug = process.argv.some(function(argument) {
 });
 
 module.exports = function(config) {
-  var closureLibPath = 'node_modules/google-closure-library/';
```
sbrunner: This will be replace by the webpack version ...

ger-benjamin:
That's because of this kind of PR that Github stats are so useless :sweat_smile:

gberaudo:
+12,023 −22,632
I am a bit surprised there are 10k lines gone.

package.json
```diff
@@ -103,6 +103,7 @@
     "walk": "2.3.13",
     "webpack": "4.6.0",
     "webpack-cli": "2.0.15",
+    "webpack-dev-server": "^3.1.3",
```
sbrunner: Don't use `^`

PR #3799, opened 2018-04-24
Better 'no result' display in search

fredj:


contribs/gmf/less/search.less
```diff
@@ -108,6 +108,9 @@ gmf-search {
       }
     }
   }
+  .tt-menu.tt-open.gmf-search-no-results {
+    display: block !important;
```
sbrunner: We shouldn't use !important ...

fredj: I know but it's a hackish workaround for a typeahead limitation that will never be fixed upstream

PR #3800, opened 2018-04-24
Misc code cleanup

fredj:


src/search/createlocationsearchbloodhound.js
```diff
@@ -25,7 +25,7 @@ ngeo.search.createLocationSearchBloodhound = function(opt_options) {
    * @param {string} bbox Bbox string.
    * @return {?ol.Extent} Parsed extent.
    */
-  const parseBbox = function(bbox) {
+  const parseBbox = (bbox) => {
```
gberaudo: FWIW, there is no necessity to migrate function not using the `this` object to arrow functions.

sbrunner: For me it's better to use new syntax to avoid the `this` issues

PR #3801, opened 2018-04-24
 Add support for angularjs templates to ngeo-displaywindow

pfirpfel:
Allows angularjs templates inside the popups of ngeo-displaywindow.

Example in action: https://geomapfish-demo.camptocamp.com/mkuenzli/displaywindow.html

### DIFF
--- a/examples/displaywindow.html
+++ b/examples/displaywindow.html
@@ -21,6 +21,11 @@
             type="button"
             class="btn btn-default">Open / close window 3.</button>
 
+    <button ng-click="ctrl.window4IsOpen = !ctrl.window4IsOpen;"
+            title="Open window 4."
+            type="button"
+            class="btn btn-default">Open / close window 4.</button>
+
     <ngeo-displaywindow
       class="window1"
       url="::ctrl.window1Content"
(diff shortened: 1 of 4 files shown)


examples/displaywindow.html
```diff
@@ -36,5 +46,45 @@
       title="'Window 2 - A powerful window'">
     </ngeo-displaywindow>
 
+    <ngeo-displaywindow
+            class="window3"
+            clear-on-close="::false"
+            content-template="::ctrl.window3Template"
+            content-scope="::ctrl.windowScope"
+            draggable="::true"
+            resize="::true"
```
ger-benjamin: If draggable AND resize are `true`, that's the same as `desktop=true`.
And by default, desktop is `true`.
So you can remove both binding, that will be the same.
(read the doc :-p )

I think you can also remove height and width, that's not what we want to show (=noise).

examples/displaywindow.js
```diff
@@ -40,6 +41,56 @@ app.displaywindow.MainController = function() {
...
+   * @type {string}
+   * @export
+   */
+  this.window4Template = angular.element(document.getElementById('window4Template')).html();
+
+  /**
+   * @type {string}
+   * @export
+   */
+  this.window4TextBinding = 'This is an angular binding.';
```
ger-benjamin: AngularJS

src/message/displaywindowComponent.js
```diff
@@ -262,6 +299,8 @@ ngeo.message.displaywindowComponent.component('ngeoDisplaywindow', {
   bindings: {
     'clearOnClose': '<?',
     'content': '=?',
+    'contentTemplate': '<',
+    'contentScope': '<',
```
ger-benjamin: You must add `?` symbol to the bindings otherwise it's a mandatory bindings and you will get errors with example's window 1 and 2.

examples/displaywindow.html
pfirpfel: Window2 has both as well... But I agree on leaving out stuff, we don't want to show. I will keep the width/height on the last one though, otherwise you need to resize the window first to see what I want to show.

ger-benjamin: Yes, but in windows 2 I want to demonstrate it.
But ok, you can let it.

PR #3802, opened 2018-04-24
Use jQuery to synchronize rotation inputs

### LINKED ISSUE #3755: Print-Window : Negative values in Orientation-Field
In the Print-Window, we can configure the orientation of the map for printing. There is a slider and a number-field. But we cannot manually set negative numbers in the number-field, cause only numbers are allowed. So when we try to insert a value that begins with the char "-", the field will remain unchanged.
(Actually we can add the char "-" before the orientation value, but not replace the whole field with "-" alone)

This can be reproduced here : https://geomapfish-demo.camptocamp.net/2.2/

To correct this, I wanted to activate the allowInvalid option on the input.
But because of issues with HTML5 constraint validation, we cannot use input[type=number] with allowInvalid.  See: https://docs.angularjs.org/api/ng/input/input%5Bnumber%5D

I think a text-input should be used here, so that the character "-" will be allowed.

If further information is needed please see https://github.com/camptocamp/baselstadt_mapbs/commit/5d65f61fe3b6897ad6a3b78afc39710f72b87d6a

fredj:
fixes #3755

### DIFF
--- a/contribs/gmf/src/directives/print.js
+++ b/contribs/gmf/src/directives/print.js
@@ -149,6 +149,7 @@ gmf.module.directive('gmfPrint', gmf.printDirective);
 
 
 /**
+ * @param {angular.JQLite} $element Element.
  * @param {angular.Scope} $rootScope Angular root scope.
  * @param {angular.Scope} $scope Angular scope.
  * @param {angular.$timeout} $timeout Angular timeout service.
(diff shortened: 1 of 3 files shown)


contribs/gmf/src/directives/print.js
```diff
@@ -661,32 +675,22 @@ gmf.PrintController.prototype.isAttributeInCurrentLayout_ = function(name) {
...
-  if (opt_rotation !== undefined) {
-    let rotation = parseInt(opt_rotation, 10);
-    if (rotation > 180) {
-      rotation = -180;
-    } else if (rotation < -180) {
-      rotation = 180;
-    }
-    this.rotation = rotation;
+gmf.PrintController.prototype.setRotation = function(rotation) {
+  this.rotation = ol.math.clamp(rotation, -180, 180);
```
ger-benjamin: I don't know this function, are you sure it is supported by our browser we support ?

fredj: This function is from OpenLayers, see:
https://github.com/openlayers/openlayers/blob/v4.6.5/src/ol/math.js#L14

ger-benjamin: Agree, thanks

PR #3804, opened 2018-04-24
Restrict dragging to header handle of displaywindow component (+ imports)

pfirpfel:
- [minor change] Add draggable to webpack imports (like resizable)

- **Restrict dragging to header handle.**
If the whole window is the handle, the user cannot interact with certain elements of the window. For example, it is not possible to select text there. ~~A potential issue introduced with this change is, that there is no visual indicator, that the header is the dragging-handle. I'm open for suggestions.~~
Edit: The only visual indicator currently is the cursor, which changes if the window is draggable and the user is hovering over the header. This should be enough for the example, right?

### DIFF
--- a/src/message/displaywindowComponent.js
+++ b/src/message/displaywindowComponent.js
@@ -2,6 +2,7 @@ goog.provide('ngeo.message.displaywindowComponent');
 
 goog.require('ngeo'); // nowebpack
 // webpack: import 'jquery-ui/ui/widgets/resizable.js';
+// webpack: import 'jquery-ui/ui/widgets/draggable.js';
 // webpack: import 'angular-sanitize';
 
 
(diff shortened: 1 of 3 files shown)


src/message/displaywindowComponent.js
```diff
@@ -2,6 +2,7 @@ goog.provide('ngeo.message.displaywindowComponent');
 
 goog.require('ngeo'); // nowebpack
 // webpack: import 'jquery-ui/ui/widgets/resizable.js';
+// webpack: import 'jquery-ui/ui/widgets/draggable.js';
```
ger-benjamin: @gberaudo It is ok to import both lib here ? Because (for mobile) this component can disable drag and resize function. So in some case, we don't want to have it. So perhaps we should have these import here, perhaps it's better 
 to load them in the final app. What's your point of view ?

gberaudo: For the record, since the impact is small we decided to keep the dependency here for ease of use.

PR #3809, opened 2018-04-26
Some cleanup

gberaudo:


github_versions
```diff
@@ -1,2 +1 @@
 angular.js=704123a
```
sbrunner: Is it steel used?

gberaudo: If not I will remove it in a future PR. What is it for?

gberaudo@wrk29:~/dev/ngeo (cleanup *$)$ git grep github_versions 
Makefile:673:contribs/gmf/build/angular-locale_%.js: github_versions

PR #3811, opened 2018-04-26
More cleanup and port to OpenLayers master

gberaudo:


.eslintrc-es6.yaml
```diff
@@ -72,7 +72,7 @@ rules:
   # http://eslint.org/docs/rules/no-this-before-super
   no-this-before-super: error
 
-  no-unused-vars: warn
+  no-unused-vars: error
```
sbrunner: Sshouldn't we rename this file?

gberaudo: Not necessarily, since it contains essentially rules relevant to ES6 coding style.
Since we already have a .eslintrc.yaml it may be fine to keep it as-is.

PR #3814, opened 2018-04-27
Change debug port to 3000

fredj:
And remove extra `.html` in webpack server


buildtools/webpack.gmfapps.js
```diff
@@ -15,7 +15,7 @@ for (const filename of ls('contribs/gmf/apps/*/index.html')) {
     new HtmlWebpackPlugin({
       template: filename.full,
       chunksSortMode: 'manual',
-      filename: filenamePrefix + name + '.html',
+      filename: filenamePrefix + name,
```
sbrunner: Are you sure about that?

fredj: see #3818

PR #3827, opened 2018-05-01
Fix routing example + pass OLCS through babel

gberaudo:


examples/locationsearch.css
```diff
@@ -3,7 +3,7 @@
   width: 600px;
   height: 400px;
 }
-/* CSS stolen from https://github.com/bassjobsen/typeahead.js-bootstrap-css/ */
```
fredj: :-)

PR #3832, opened 2018-05-02
Move less files into modules

fredj:


contribs/gmf/apps/desktop_alt/less/main.less
```diff
@@ -1,4 +1,4 @@
-@import '../../../less/desktop.less';
+@import '../../../src/controllers/desktop.less';
```
sbrunner: @import 'gmf/controllers/desktop.less';

contribs/gmf/src/import/import.less
```diff
@@ -1,5 +1,6 @@
-@import "../../../node_modules/font-awesome/less/variables.less";
-@import "typeahead.less";
+@import "../../../../node_modules/font-awesome/less/variables.less";
```
sbrunner: ~font-awesome/less/variables.less

PR #3835, opened 2018-05-02
Fix gmf examples

llienher:
Fix several gmf examples:
- drawfeature
- profile
- mobilemeasure

Fix ngeo example:
- modifycircle

Also various fixes for gmf apps.

### DIFF
--- a/contribs/gmf/src/drawing/featureStyleComponent.js
+++ b/contribs/gmf/src/drawing/featureStyleComponent.js
@@ -3,7 +3,6 @@
  */
 import googAsserts from 'goog/asserts.js';
 import * as olEvents from 'ol/events.js';
-import olObject from 'ol/Object.js';
 import ngeoFormatFeatureProperties from 'ngeo/format/FeatureProperties.js';
 
 /** @suppress {extraRequire} */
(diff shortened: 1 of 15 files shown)


contribs/gmf/src/drawing/featureStyleComponent.js
```diff
@@ -162,7 +162,7 @@ exports.Controller_.prototype.handleFeatureSet_ = function(
       keys.push(
         olEvents.listen(
           newFeature,
-          olObject.getChangeEventType(propName),
+          getChangeEventType(propName),
```
gberaudo: We should create the string ourselves: `change-${propName}`.
See also the other files below where you should do for ex: 'change-dirty'.

src/map/FeatureOverlayMgr.js
```diff
@@ -7,7 +7,7 @@ import * as olBase from 'ol/index.js';
 import olLayerVector from 'ol/layer/Vector.js';
 import * as olObj from 'ol/obj.js';
 import olSourceVector from 'ol/source/Vector.js';
-import olStyleStyle, {toFunction as toStyleFunction} from 'ol/style/Style.js';
+import olStyleStyle, {toFunction as toStyleFunction, createDefaultStyle as defaultFunction} from 'ol/style/Style.js';
```
gberaudo: Better call it `olStyleDefaultFunction`.

PR #3837, opened 2018-05-02
Export IS_HOME for deploy

sbrunner:


buildtools/deploy.sh
```diff
@@ -8,28 +10,29 @@ fi
 
 # Cloning gh-pages into a local temporary directory
 TMP=".build/ngeo-${GITHUB_USERNAME}-gh-pages"
-git clone --single-branch --branch gh-pages $(GIT_REMOTE_URL) $TMP
+git clone --single-branch --branch gh-pages ${GIT_REMOTE_URL} ${TMP}
```
gberaudo: Looks like the GIT_REMOTE_URL is not exported.

gberaudo: I think we should directly put the value "git@github.com:camptocamp/ngeo.git" here.

sbrunner: Yes, I put is as befor by using GITHUB_USERNAME
=>
git@github.com:${GITHUB_USERNAME}/ngeo.git
:-)

PR #3842, opened 2018-05-03
Lidarprofile

kalbermattenm:
This PR superseeds #3464. It addresses the four remaining points which were not resolved until yet.

The LIDAR profile will be directly integrated into master (2.3) to avoid useless backporting.

@ger-benjamin, there is a demo here: https://sitn.ne.ch/dev_internet2/

We agreed with @ger-benjamin that he would take care of the 2.2 to 2.3 remaining migration steps.

PR #3464, opened 2018-01-30
Lidarprofile ngeo22

monodo:
TO BE DONE BEFORE MERING

- [x] Setup type for [d3.linearScale](https://github.com/monodo/ngeo/blob/lidarprofile_ngeo22/contribs/gmf/src/services/lidarprofileconfig.js#L33-L44)
- [ ] Fix altitude decoding issue [here](https://github.com/monodo/ngeo/blob/lidarprofile_ngeo22/contribs/gmf/src/d3-lidar-profile/lidarprofileLoader.js#L360)
- [x] Setup type for Pytree service response [here](https://github.com/monodo/ngeo/blob/lidarprofile_ngeo22/contribs/gmf/src/services/lidarprofileconfig.js#L86)
- [x] Move the [scaleX, scaleY and previousDomainX](https://github.com/monodo/ngeo/blob/lidarprofile_ngeo22/contribs/gmf/src/services/lidarprofileconfig.js#L33-L44) to a more pertinent place
- [x] Merge latest 2.2 branch into this one

Live demo [here](https://sitn.ne.ch/dev_internet2/desktop) 

@ger-benjamin please review 

---

Edited by ger-benjamin
New TODO:
 1. [ ] fix "FIXME missing description"
 1. [x] Create an example
 1. [x] Move the application example in the desktop_alt
 1. [x] Standardize name of pytree api (looks strange: `/profile_config_gmf2` looks strange no ? https://github.com/monodo/ngeo/blob/d2e8cb8a5496610eec93bbe5faef856e5ed9fd95/contribs/gmf/src/d3-lidar-profile/lidarprofileConfig.js#L65)
 1. [x] rework download functions in https://github.com/monodo/ngeo/blob/d2e8cb8a5496610eec93bbe5faef856e5ed9fd95/contribs/gmf/src/d3-lidar-profile/lidarprofileUtils.js
 1. [x] Add translation on all texts

Do all we can on available time to:
 - Improve general architecture
- Simplify functions when it's possible

---

Debug:

- [x] easy: Cleanall function (in lidarpanel controller) works not completely (line persist on the map)
- [x] easy: Fix some error that appears in the console
- [ ] not easy: Can't drag and drop profile to move along the line
- [ ] not easy: It's hard to do a measure
- [x] not easy: profile is compiled but doesn't work (first step: add missing require !)
[...]

PR #3855, opened 2018-05-07
WIP Lidarprofile 23

ger-benjamin:
This PR superseeds #3842, that superseeds original #3464. It convert lidarprofile to ngeo 2.3 (and is based with your last fixes @kalbermattenm , I've merged them into one commit)

WIP: The example works fine, but I can't test the app now, because gmf apps doens't work for now in ngeo.


ger-benjamin:
Works fine now on desktop apps except for two things:
- @sbrunner Where should I define this value in the new architecture ?: https://github.com/monodo/ngeo/blob/4a45c36dada9576733f4e37cd43abf3676dbd0e5/contribs/gmf/apps/desktop_alt/index.html#L443 
- @sbrunner, @kalbermattenm The lidarprofile wants coordinates in EPSG2056, our map in the demo app is in EPSG21781. Why should I do ?

sbrunner:
What about changing the projection in the desktop alt?

kalbermattenm:
We (SITN) cannot, for sure, provide data in EPSG:21781...

> @sbrunner Where should I define this value in the new architecture ?

What about creating a fake webservice, serving always the same points, which would be a bit odd, but which would removing a dependency from your webservice... (we could even add some doc about that in the example)

ger-benjamin:
> What about changing the projection in the desktop alt ?

As we have discussed, that's not so simple, because all widget that use the raster service (for elevation) will not work anymore, there is also also work with the background, a little bit for the search.
That's not impossible but that take a little bit time.

> Can't we re-projecte coordinates from the map to the lidar profile ?

That not so simple because we need to reproject the line AND the "mouseover" that display the information popup...

> What about creating a fake webservice ?

It's a possibility.

---

IMO, we should revert changes in the desktop_alt. So we will not have the Lidar Profile for now in our application desktop. We know that it works, you can see the revert commit to implement it into your application if you need it.
And if you want a demo, the example still works (contribs/gmf/example/lidarprofile).
And later, we re-add the Lidar profile in the desktop_alt.

ger-benjamin:
@sbrunner @kalbermattenm Is this idea ok for you ? (I hope merging this afternoon)

kalbermattenm:
+1, effectively, it looks like the easy way to resolve this problem.

ger-benjamin:
Done, I've made a specific commit that add, and another that revert what we need to add in the application files to have the Lidar profile (see the two lasts commits).

I'll merge :tada:

PR #3869, opened 2018-05-09
Override contextual data in gmf apps

gberaudo:
I could not test: there were 500 errors with dynamic.js.
Same as what we did in EPFL project.


gberaudo:
@fredj, I give you the PR since I will probably not have much time this afternoon.

PR #3884, opened 2018-05-16
Force GMF apps to be accessed through https

gberaudo:
See https://jira.camptocamp.com/browse/GSGMF-465.
Client applications can easily allow http access by removing the block from the html template.


gberaudo:
According to @fredj, it is possible to enforce this in the gh-pages config. So closing this PR.

PR #3899, opened 2018-05-22
Fix datepicker

gberaudo:
See https://jira.camptocamp.com/browse/GSGMF-462.

- [x] be able to display a range and select some dates;
- [x] use proper locale;
- [ ] do not restrict time range to the previous week (discussed with @ybolognini, see https://github.com/camptocamp/ngeo/blob/079a86b/src/filter/ruleComponent.js#L231-L232)

### DIFF
--- a/src/misc/datepickerComponent.js
+++ b/src/misc/datepickerComponent.js
@@ -7,6 +7,12 @@ import ngeoMiscTime from 'ngeo/misc/Time.js';
 import 'angular-ui-date';
 import 'jquery-ui/themes/base/all.css';
 
+// FIXME: import the locales in the applications
+import 'jquery-ui/ui/i18n/datepicker-fr.js';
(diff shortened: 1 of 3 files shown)


src/misc/datepickerComponent.js
```diff
@@ -7,6 +7,12 @@ import ngeoMiscTime from 'ngeo/misc/Time.js';
 import 'angular-ui-date';
 import 'jquery-ui/themes/base/all.css';
 
+// FIXME: import the locales in the applications
+import 'jquery-ui/ui/i18n/datepicker-fr.js';
+import 'jquery-ui/ui/i18n/datepicker-uk.js';
```
sbrunner: uk? => en_GB

gberaudo: You are right. Fixed.

PR #3901, opened 2018-05-23
[WIP] Follow up fix datepicker

gberaudo:


src/filter/ruleComponent.js
```diff
@@ -227,12 +227,12 @@ exports.RuleController_ = class {
      * @export
      */
     this.timeRangeMode = {
-      widget: /** @type {ngeox.TimePropertyWidgetEnum} */ ('datepicker'),
-      maxValue: this.createDate_(),
-      minValue: this.createWeekAgoDate_(),
```
sbrunner: For me it should be used as def values ...

src/filter/ruleComponent.js
```diff
@@ -424,24 +424,21 @@ exports.RuleController_ = class {
       this.unlisteners_.push(this.scope_.$watch(
         () => this.clone.getExpression(),
         (newVal) => {
-          const value = newVal === null ? this.createDate_() : newVal;
-          this.timeValueMode.minValue = value;
+          this.timeValueMode.minValue = newVal;
```
sbrunner: For that is this code for?

gberaudo: Not idea.
But since we now handle the null value correctly I suppose the ternary operation can be removed.

gberaudo:
@sbrunner I fixed the issue with Travis.

PR #3904, opened 2018-05-23
Add missing ng-if for share popup in desktop interface

llienher:
Was only added for desktop_alt... Fix the share popup for desktop interface.


sbrunner:
It should be done on branch 2.3 ...

llienher:
OK, I'm backporting it.

PR #3905, opened 2018-05-23
[2.3] Add missing ng-if to desktop interface

llienher:
Backport #3904 into 2.3 branch.


gberaudo:
There is also contribs/gmf/examples/share.html missing the ng-if.

llienher:
@gberaudo Good catch, I gonna fix the example in an other PR to not make this one (backport) diverge from the initial one and makes thing confusing.

PR #3909, opened 2018-05-23
Add NetworkStatus service for offline detection

fgravin:
Add a service that keeps network status up to date.

See comment:
```
   * This service watches the status of network connection.
   *
   * Currently it watches every $http and $.ajax requests errors, if an error
   * occurs we wait 2 sec then we make an http request on the checker file.
   * If the checker responds that means we are online, otherwise we make a
   * 2nd request 2 sec later, if the 2nd requests failed that means we
   * are offline.
   *
   * A timeout of 1 sec is set for the checker file, so if we have a bad
   * connection, we consider we are offline.
   *
   * During offline mode we test every 2 sec if we are back online.

```
I wonder why that have chosen to use ` $.ajax` instead of `$http`.
Some code is useless if we have no action while request are currently loading and not yet responded (ga use it for their `gaWaitCursor`).

Please have a first check @ger-benjamin or @gberaudo thanks

### DIFF
--- a/examples/offline.js
+++ b/examples/offline.js
@@ -25,13 +25,15 @@ app.offline.module = angular.module('app', [
   ngeo.offline.module.name
 ]);
 
+app.offline.module.value('ngeoOfflineTestUrl', '../../src/offline/component.html');
 
 /**
  * @param {ngeoFeatureOverlayMgr} ngeoFeatureOverlayMgr ngeo feature overlay manager service.
+ * @param {ngeoNetworkStatus} ngeoNetworkStatus ngeo network status service.
  * @constructor
  * @ngInject
  */
-app.offline.MainController = function(ngeoFeatureOverlayMgr) {
+exports.MainController = function(ngeoFeatureOverlayMgr, ngeoNetworkStatus) {
 
   /**
    * Save a square of 10 km sideways (Map's unit is the meter).
@@ -39,6 +41,12 @@ app.offline.MainController = function(ngeoFeatureOverlayMgr) {
    * @export
    */
   this.offlineExtentSize = 10000;
+
+  /**
+   * @type {ngeoNetworkStatus}
+   * @export
+   */
+  this.ngeoNetworkStatus = ngeoNetworkStatus;
 
   /**
    * @type {ol.Map}
--- a/src/offline/NetworkStatus.js
+++ b/src/offline/NetworkStatus.js
@@ -1,0 +1,209 @@
+goog.module('ngeo.offline.NetworkStatus');
+
+goog.require('ngeo.misc.debounce');
+
+
+class HttpInterceptor {
+
+  /**
+   * @ngInject
+   * @param {angular.$q} $q The Angular $q service.
+   * @param {ngeox.miscDebounce} ngeoDebounce ngeo debounce service.
+   * @param {ngeo.offline.NetworkStatus} ngeoNetworkStatus ngeo network status service.
+   */
+  constructor($q, ngeoDebounce, ngeoNetworkStatus) {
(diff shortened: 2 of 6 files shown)


src/offline/NetworkStatus.js
```diff
@@ -0,0 +1,200 @@
...
+   */
+  triggerChangeStatusEvent_(offline) {
+    this.offline = offline;
+    // this.$rootScope_.$broadcast('ngeoNetworkStatusChange', net.offline);
+    this.$rootScope_.$digest();
+  }
+
+};
+
+const name = 'ngeoNetworkStatus';
```
ger-benjamin: Adapt this whole end of file to non-webpack

src/offline/NetworkStatus.js
```diff
@@ -0,0 +1,200 @@
...
+     * @private
+     * @type {!angular.$q.Promise}
+     */
+    this.promise_;
+
+    this.initialize_();
+
+  }
+
+  initialize_() {
```
ger-benjamin: Add documentations line (`@private`)

src/offline/NetworkStatus.js
```diff
@@ -0,0 +1,200 @@
...
+        if (this.count_ > 2 && !this.offline) {
+          this.triggerChangeStatusEvent_(true);
+        }
+      }
+    });
+
+  }
+
+  /**
+   * @private
```
ger-benjamin: Add documentations line (`@param...`)

src/offline/NetworkStatus.js
```diff
@@ -0,0 +1,200 @@
...
+    this.$document_.ajaxError((evt, jqxhr, settings, thrownError) => {
+      // Filter out canceled requests
+      if (!/^(canceled|abort)$/.test(thrownError)) {
+        this.check(2000);
+      }
+    });
+
+  }
+
+  check(timeout) {
```
ger-benjamin: Add documentations line (quick description)
Add documentations line (`@param...`)

examples/offline.js
```diff
@@ -28,10 +28,11 @@ app.offline.module = angular.module('app', [
 
 /**
  * @param {ngeoFeatureOverlayMgr} ngeoFeatureOverlayMgr ngeo feature overlay manager service.
+ * @param {ngeoNetworkStatus} ngeoNetworkStatus ngeo network status service.
  * @constructor
  * @ngInject
  */
-app.offline.MainController = function(ngeoFeatureOverlayMgr) {
+exports.MainController = function(ngeoFeatureOverlayMgr, ngeoNetworkStatus) {
```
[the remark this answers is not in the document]
fgravin: Are you sure, my code is more ready for es6 module no ?
It comes from here at least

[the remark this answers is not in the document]
ger-benjamin: For the examples, that ~true. But I don't now what will happen with the transformation. I prefers to use the standard case, I think that will be easier to transform.

For the src code, we can't. Because the  app (and so luxembourg) still work without webpack.

fgravin:
@ger-benjamin this version of the PR works fine both in webpack example and in luxembourg project.
I didn't fix all your comments cause there are many ways to implement the module, i took example on olcsService and olcsModule, i think the code is more ready for future transformation.

ger-benjamin:
Before to merge, can you just do my first comment about "the style in the html -> move it to example.less" ?
I really don't want style in examples.html

fgravin:
> Before to merge, can you just do my first comment about "the style in the html -> move it to example.less" ?
> I really don't want style in examples.html

Yes sorry for this one, it's done !

PR #3926, opened 2018-05-31
Add an offline services manager to link ngeo and apps

ger-benjamin:
A basic concept to link ngeo-offline to apps-offline

See also https://github.com/Geoportail-Luxembourg/geoportailv3/pull/2042

### DIFF
--- a/src/offline/ServiceManager.js
+++ b/src/offline/ServiceManager.js
@@ -1,0 +1,94 @@
+goog.provide('ngeo.offline.ServiceManager');
+
+goog.require('ngeo');
+
+
+ngeo.offline.ServiceManager = class {
+
+  /**
+   * @param {angular.$injector} $injector Main injector.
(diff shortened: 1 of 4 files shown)


src/offline/ServiceManager.js
```diff
@@ -0,0 +1,80 @@
...
+    if (restoreServiceName && this.$injector_.has(restoreServiceName)) {
+      this.restoreService_ = this.$injector_.get(restoreServiceName);
+    } 
+  }
+
+  /**
+   * Ask the provided service to save the data to an offline purpose
+   */
+  save() {
+    if (this.saveService_) {
```
fgravin: Would be good for devs to have a warning if you call this method while not saveService is declared ?

ger-benjamin: Indeed

PR #3927, opened 2018-05-31
 Stop edit mode when leaving edit component

fredj:
To reproduce the issue:
 * go to edit component and start drawing a polygon
 * switch to the print component

To draw interaction is still active

### DIFF
--- a/contribs/gmf/src/editing/editFeatureSelectorComponent.js
+++ b/contribs/gmf/src/editing/editFeatureSelectorComponent.js
@@ -65,8 +65,7 @@ exports.component_ = function() {
 };
 
 
-exports.directive('gmfEditfeatureselector',
-  exports.component_);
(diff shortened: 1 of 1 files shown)


contribs/gmf/src/editing/editFeatureSelectorComponent.js
```diff
@@ -254,7 +252,7 @@ exports.Controller_.prototype.stopEditing = function() {
 exports.Controller_.prototype.handleActiveChange_ = function(active) {
   if (!active) {
     if (!this.dirty) {
-      this.selectedEditableNode = null;
```
fredj: `selectedEditableNode` variable is not defined (or used) !

sbrunner:
Shouldn't be on 2.3?

PR #3946, opened 2018-06-07
Offline goog modules

gberaudo:


gberaudo:
Hi @fredj and @sbrunner. The rest of the Luxembourg team is unavailable this end of week. Could you please look at this PR? It is simply:
- bug fixes;
- switch to goog.module;
- a basic debug flag (to ease offline work).

src/offline/NetworkStatus.js
```diff
@@ -132,7 +132,7 @@ const Service = class {
   /**
    * Check fir network status
    *
-   * @param {number|undefined} timeout Delay for timeout.
+   * @param {(number|undefined)=} timeout Delay for timeout.
```
gberaudo: Use simply number=

src/offline/NetworkStatus.js
```diff
@@ -1,50 +1,33 @@
...
-  }
-  response(response) {
-    return response;
-  }
-  responseError(rejection) {
-    this.ngeoDebounce_(() => this.ngeoNetworkStatus_.check(undefined), 2000, false);
-    return this.$q.reject(rejection);
-  }
-}
+const ngeoMiscDebounce = goog.require('ngeo.misc.debounce');
```
gberaudo: This does not look right. The debounce function is not a module.

gberaudo: Well, with the GCC we use it is actually working. Great!

PR #3964, opened 2018-06-13
[WIP] Offline localforage

gberaudo:


.eslintrc.yaml
```diff
@@ -7,8 +7,15 @@ env:
 rules:
     no-console: 0
     comma-dangle: 0
+    valid-jsdoc:
+      - error
+      -
+        requireReturn: false
+        requireParamDescription: false
+        requireReturnDescription: false
```
ger-benjamin: (We will pay for that later)

src/offline/DefaultConfiguration.js
```diff
@@ -2,17 +2,31 @@ goog.module('ngeo.offline.DefaultConfiguration');
...
-  constructor() {
+  /**
+   * @param {!angular.Scope} $rootScope The rootScope provider.
+   */
+  constructor($rootScope) {
     super();
+    localforage.config({
+      'name': 'ngeoOfflineStorage',
+      'version': 1.0,
+      'storeName': 'storage'
```
ger-benjamin: It is a problem if later, we have another storage in ngeo, and we change the name here ? (will the users loose their storage ?). If we have any doubt, we should call it "offlineStorage" or similar.

src/offline/DefaultConfiguration.js
```diff
@@ -21,6 +35,65 @@ exports = class extends ol.Observable {
         'progress': progress
       }));
     };
+
+    /**
+     * @private
+     */
+    this.rootScope_ = $rootScope;
```
ger-benjamin: missing type

PR #3965, opened 2018-06-13
Offline: add a mode for constant pixel size mask

fgravin:
See https://jira.camptocamp.com/browse/GSLUX-83

- fix mask size
If no `extentSize` is given to offline component, then the mask will not stick to an extent. 
The mask is then static, changing map resolution will change extent size, as on google map offline.
The size of the mask is a constant square, with a minimum margin defined by the user.

- zoom constraints
While the mask is open, the view zoom will be constraint with specified bounds.

### DIFF
--- a/src/offline/component.js
+++ b/src/offline/component.js
@@ -19,13 +19,14 @@ exports = angular.module('ngeoOffline', [
   ngeoMessageModalComponent.name
 ]);
 
+const MASK_MIN_MARGIN = 100;
 
 exports.value('ngeoOfflineTemplateUrl',
   /**
-     * @param {angular.JQLite} element Element.
-     * @param {angular.Attributes} attrs Attributes.
-     * @return {string} Template URL.
-     */
+   * @param {angular.JQLite} element Element.
+   * @param {angular.Attributes} attrs Attributes.
+   * @return {string} Template URL.
+   */
   (element, attrs) => {
     const templateUrl = attrs['ngeoOfflineTemplateurl'];
     return templateUrl !== undefined ? templateUrl :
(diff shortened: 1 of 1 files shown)


fgravin:
@gberaudo ready for final review.

fgravin:
Can anyone review it please ?
@gberaudo @ger-benjamin

src/offline/component.js
```diff
@@ -205,6 +211,41 @@ exports.Controller_ = class {
...
+     * @export
+     */
+    this.minZoom = this.minZoom || 10;
+
+    /**
+     * Maximum zoom where offline is enable.
+     * @type {number}
+     * @export
+     */
+    this.maxZoom = this.maxZoom || 15;
```
ger-benjamin: At this point, the default value will be always used. Because the values from the bindings are analyzed only from the "onInit" methode.
So perhaps you can write here `this.maxZoom = 15;` (and that's all.) And perhaps it's better to write `this.maxZoom || 15;` in the onInit.

ger-benjamin: Same for the other bindings

src/offline/component.js
```diff
@@ -362,6 +405,37 @@ exports.Controller_ = class {
...
+   * @private
+   */
+  addZoomConstraints_() {
+    const view = this.map.getView();
+    const zoom = view.getZoom();
+
+    this.originalMinZoom = view.getMinZoom();
+    this.originalMaxZoom = view.getMaxZoom();
+
+    if(zoom < this.minZoom) {
```
ger-benjamin: missing space : `if (...)`

src/offline/component.js
```diff
@@ -53,6 +52,10 @@ function ngeoOfflineTemplateUrl($element, $attrs, ngeoOfflineTemplateUrl) {
  *     <ngeo-offline
  *       ngeo-offline-map="ctrl.map"
  *       ngeo-offline-extentsize="ctrl.offlineExtentSize">
+ *       ngeo-offline-mask-margin="::100"
+ *       ngeo-offline-min_zoom="::11"
```
ger-benjamin: @fgravin 

I just see that here, it should be `ngeo-offline-min-zoom`, not `ngeo-offline-min_zoom`
Same for the max

PR #3975, opened 2018-06-20
Fix externs

gberaudo:


options/ngeox.js
```diff
@@ -2772,6 +2772,35 @@ ngeox.OfflineConfiguration.prototype.getCallbacks = function() {};
  */
 ngeox.OfflineConfiguration.prototype.createLayerMetadatas = function(map, userExtent) {};
 
+/**
+ * @param {ngeox.OfflineLayerMetadata} metadata
+ * @return {string}
+ */
+ngeox.OfflineConfiguration.prototype.getLayerKey = function(metadata) {};
```
ger-benjamin: I don't understand why you need that, why it doens't take the description from the class

gberaudo: Because we are working with interfaces.

PR #3987, opened 2018-06-26
Remove rotation and lang from the print spec

fredj:


contribs/gmf/src/print/component.js
```diff
@@ -830,13 +830,10 @@ exports.Controller_ = class {
...
     const viewResolution = this.map.getView().getResolution() || 0;
     const scale = this.layoutInfo.scale || this.getOptimalScale_(mapSize, viewResolution);
-    const rotation = this.rotateMask ? -this.rotation : this.rotation;
     const datasource = this.getDataSource_();
 
     const customAttributes = {
       'datasource': datasource,
-      'lang': this.gettextCatalog_.currentLanguage,
-      'rotation': rotation,
       'scale': scale
```
sbrunner: Shouldm't be also removed?

fredj: yes; it's already set in the map

PR #3991, opened 2018-06-27
Offline tile restoration

gberaudo:


src/offline/Restorer.js
```diff
@@ -20,15 +20,22 @@ class Restorer {
...
+    return this.configuration_.getItem('offline_content').then(offlineContent => this.doRestore_(map, offlineContent));
   }
 
   /**
+   * @param {ol.Map} map
    * @param {ngeox.OfflinePersistentContent} offlineContent
    * @return {ol.Extent}
    */
-  doRestore_(offlineContent) {
+  doRestore_(map, offlineContent) {
```
fgravin: this is were we could override what we do with the layers ...

gberaudo:
@fgravin, @ger-benjamin, I implemented storing/restoration of layer metadata and background layer.
It is working nicely in debug mode; not yet in compiled mode. I will see what is the reason for that.

src/offline/Restorer.js
ger-benjamin: add a `@private` or remove the final underscore please

gberaudo: OK, done. Anything else to do?

src/offline/Restorer.js
```diff
@@ -20,20 +30,33 @@ class Restorer {
    * @return {Promise<ol.Extent>}
    */
   restore(map) {
-    console.log('should restore something');
-    return this.configuration_.getItem('offline_content').then(offlineContent => this.doRestore_(offlineContent));
+    return this.configuration_.getItem('offline_content').then(offlineContent => this.doRestore(map, offlineContent));
   }
 
   /**
+   * @protected
```
ger-benjamin: equal to nothing I guess

fgravin:
looks good so far

gberaudo:
@ger-benjamin, thanks for the review. I am willing to merge as soon as it is green so that it can be tested this end of afternoon and delivered tomorrow.

PR #3999, opened 2018-07-02
Document $injector usage

### LINKED ISSUE #1865: Inconsistent use of $injector vs. injection as constructor argument
Currently we don't have clear rules when to use `$injector` and when to inject as constructor argument.

For example the [DatePickerController](https://github.com/camptocamp/ngeo/blob/b72175bad68ac4e25273235d4a97430c87a66fcb/src/directives/datepicker.js#L95) uses both techniques at the same time:

```
ngeo.DatePickerController = function($scope, $injector, ngeoTime) {
...
 this.ngeoTime_ = ngeoTime;
...
 this.gettextCatalog_ = $injector.get('gettextCatalog');
```

@sbrunner proposes to use $injector only for optional dependencies (like [here](this.gettextCatalog_ = $injector.get%28'gettextCatalog'%29;)).

In my opinion, it would also make sense to use `$injector` if you have many dependencies to avoid a long list of constructor arguments (but that is also usually a sign that your class is doing too much..).

sbrunner:
fix #1865

### DIFF
--- a/docs/guidelines.md
+++ b/docs/guidelines.md
@@ -13,7 +13,7 @@ or PR from persons outside of Camptocamp customers or developers.
 
 - [Main principle](#main-principle)
 - [Coding style guide](#coding-style-guide)
-- [Main changes in the code between ngeo 2.2 and 2.2](#main-changes-in-the-code-between-ngeo-22-and-23)
+- [Main changes in the code between ngeo 2.2 and 2.3](#main-changes-in-the-code-between-ngeo-22-and-23)
 - [AngularJS names ](#angularJS-names)
 - [Module management](#module-management)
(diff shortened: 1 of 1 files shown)


docs/guidelines.md
```diff
@@ -150,6 +150,11 @@ const myComponent = {
 };
 ```
 
+## `$injector` usage
+
+`$injector` should be use only for the optional requirements.
```
gberaudo: used

docs/guidelines.md
```diff
@@ -150,6 +150,11 @@ const myComponent = {
...
+The only exception is for the abstract controllers (in `gmf/controleurs`) for upgrade facilities.
```
gberaudo: to ease upgrade.

PR #4003, opened 2018-07-03
Add offline mode state

fgravin:
Angular service to dispatch offline mode state.
Offline mode is either web is disconnected or user has entered offline mode by restoring his data.

See https://jira.camptocamp.com/browse/GSLUX-95

- Change button label
- Add deactive offline mode button to reload the page
- Remove show extent if no data loaded


fgravin:
> Having both getOffline and isOffline methods is confusing.
> Actually you are not using the network status, what about removing it?

I'm using network status in `isOffline()`.
I can rename `getOffline` to `getOfflineMode`.

Yes it has been confusing since the beginning, offline means several things in lux context.....

gberaudo:
To make it less confusing I propose:
- we remove the networkstatus service from the offlinestate service and the current `isOffline` methode;
- we rename the current `getOffline` methode to  `isOffline`;
- application code use the appropriate service when they need;

fgravin:
I don't really like these changes.
The purpose of the state is to know what is the situation about all the offline things and to centralize the logic when code needs to know about the state.
The issue is just a naming issue.

Because disconnected means offline
offline mode means either we load offline data or we are disconnected
and we wan't to know if user has loaded offline data what we call offline mode ...

I would keep `isOffline` with either disconnect or offline mode because it's what we consider as offline.
I would change `getOffline` to `isOfflineDataLoaded`

fgravin:
@gberaudo i pushed changes following my previous comment.

gberaudo:
@fgravin, the concepts of disconnection and offline are disjoint. We can discuss this tomorrow.

PR #4013, opened 2018-07-06
Fix layer graphic with legend rule

sbrunner:
Partially fix https://jira.camptocamp.com/browse/GSGMF-488


src/map/LayerHelper.js
```diff
@@ -366,11 +366,13 @@ exports.prototype.getWMTSLegendURL = function(layer) {
  * @param {string} layerName The name of a wms layer.
  * @param {number=} opt_scale A scale.
  * @param {string=} opt_legendRule rule parameters to add to the returned URL.
+ * @param {number=} opt_legend_width the legend width.
```
gberaudo: Use camelcase

PR #4023, opened 2018-07-13
Fix local data import button

llienher:
Fix for GSGMF-559
image

There is somewhere a conflict with the bootstrap class "input-group-sm". My opinion is to remove the class, then the field looks normal again and there is actually no need to minifying the looks/height of it for a 3 elements panel ... The interface is not cluttered by a bigger button here from the initial idea.
image


kalbermattenm:
Thanks a lot @llienher for the quick reaction !

PR #4036, opened 2018-07-18
Rework the hasOfflineData mechanism

gberaudo:
The info is now stored in the configuration.
In addition, reload the page after a clear if we were in offline mode.

With this PR, the download performance is now great again.


gberaudo:
Thanks for the review.

PR #4037, opened 2018-07-18
Hide unusable options when disconnected

gberaudo:


gberaudo:
@ybolognini does it make sense to you?

PR #4038, opened 2018-07-18
Removes unwanted hardcoded mobile interface

sbrunner:
Fix https://jira.camptocamp.com/browse/GSGMF-564


sbrunner:
Effectively, yes :-)

PR #4041, opened 2018-07-18
Fix print layers when opacity background layer is a layer group

llienher:
Fixes an print issue with opacity background layer when this is a group of layer instead of a single layer element.

When we generate the print layers, we flatten the gmf layers to remote the group, but we did loose the opacity in the process when this value was only on the group. This PR fixes and apply back the opacity to the layer once flatten.

### DIFF
--- a/src/map/LayerHelper.js
+++ b/src/map/LayerHelper.js
@@ -4,7 +4,7 @@
 import googAsserts from 'goog/asserts.js';
 import * as olArray from 'ol/array.js';
 import olFormatWMTSCapabilities from 'ol/format/WMTSCapabilities.js';
-import olLayerGroup from 'ol/layer/Group.js';
+import OlLayerGroup from 'ol/layer/Group.js';
 import olLayerImage from 'ol/layer/Image.js';
 import olLayerTile from 'ol/layer/Tile.js';
 import * as olObj from 'ol/obj.js';
@@ -241,7 +241,7 @@ exports.prototype.createWMTSLayerFromCapabilititesObj = function(
  * @export
  */
 exports.prototype.createBasicGroup = function(opt_layers) {
-  const group = new olLayerGroup();
+  const group = new OlLayerGroup();
   if (opt_layers) {
     group.setLayers(opt_layers);
   }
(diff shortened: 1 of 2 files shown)


src/map/LayerHelper.js
```diff
@@ -296,17 +303,25 @@ exports.prototype.getFlatLayers = function(layer) {
  * of others groups.
  * @param {ol.layer.Base} layer The base layer, mostly a group of layers.
  * @param {Array.<ol.layer.Base>} array An array to add layers.
+ * @param {number|undefined} computedOpacity Opacity inherited from ancestor layer groups.
  * @return {Array.<ol.layer.Layer>} Layers.
  * @private
  */
-exports.prototype.getFlatLayers_ = function(layer, array) {
-  if (layer instanceof olLayerGroup) {
+exports.prototype.getFlatLayers_ = function(layer, array, computedOpacity) {
```
ger-benjamin: Can you use an optional arguments for "computedOpacity" instead of calling this function with "undefined" at line 297 ?

src/map/LayerHelper.js
```diff
@@ -296,17 +303,25 @@ exports.prototype.getFlatLayers = function(layer) {
...
+  }
+  if (layer instanceof OlLayerGroup) {
     const sublayers = layer.getLayers();
     sublayers.forEach((l) => {
-      this.getFlatLayers_(l, array);
+      this.getFlatLayers_(l, array, computedOpacity);
     });
   } else {
     if (array.indexOf(layer) < 0) {
+      layer.set('inheritedOpacity', computedOpacity, true);
```
ger-benjamin: Can you check in our code if there is a way to indicate (document) that this opacity and should be used ?

src/map/LayerHelper.js
llienher: After discussion with @gberaudo, we prefer that it is explicit that the value is called as 'undefined' instead of skipping the last argument.

src/print/Service.js
```diff
@@ -399,6 +399,21 @@ exports.prototype.getWmtsUrl_ = function(source) {
   return exports.getAbsoluteUrl_(urls[0]);
 };
 
+/**
+ * Return an opacity value for the specified layer.
+ * Usage: When we flatten a group (getFlatLayers() method), we get only the child layers.
+ * If opacity is defined on the group, this value is lost.
+ * Inherited opacity is a custom 'back-up' value that contains the parent element opacity.
```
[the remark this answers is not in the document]
gberaudo: This documentation should be attached to the `getFlatLayers_` method.

src/map/LayerHelper.js
ger-benjamin: Okay

PR #4045, opened 2018-07-20
Fix the offline extent computation fix

gberaudo:


gberaudo:
@sbrunner, you were right yesterday during the review: it was wrong to remove the division...

gberaudo:
Thanks

PR #4079, opened 2018-07-31
Workaround some wms edge effects

gberaudo:
See https://jira.camptocamp.com/browse/GSLUX-125.


gberaudo:
See https://github.com/openlayers/openlayers/pull/8477.

PR #4081, opened 2018-07-31
Debug IE

ger-benjamin:
Fix:
GSGMF-576
GSGMF-577
GSGMF-579
GSGMF-580

And GSGMF-578


gberaudo:
@ger-benjamin have you tried something like this (not tested):

```html
<!--[if IE ]>
    <script src="https://cdn.polyfill.io/v2/polyfill.min.js"></script>
<![endif]-->
```

gberaudo:
You might need to pass the `features=es6` parameter, see https://polyfill.io/v2/docs/examples.

ger-benjamin:
@gberaudo @sbrunner  I don't know, I think I've only 2 missing features

ger-benjamin:
@gberaudo I test that here: #4082

But I think that the [tags] won't work. we will see

ger-benjamin:
He want the rest params (...) arguments here: https://github.com/camptocamp/ngeo/pull/4081/files#diff-1c3729c9686c9bb355842fc03669676aR36

How to write that ? (that's so natural :-/)

ger-benjamin:
I close this PR, it continues here: #4082

PR #4082, opened 2018-07-31
Debug IE

ger-benjamin:
Supersede #4081 

Add polyfill for:
 - GSGMF-576
 - GSGMF-577
 - GSGMF-579
 - GSGMF-580

Some css for:
 - GSGMF-576
 - GSGMF-578

### DIFF
--- a/package.json
+++ b/package.json
@@ -23,6 +23,7 @@
     "compile-catalog": "buildtools/compile-catalog.js"
   },
   "devDependencies": {
+    "@babel/polyfill": "7.0.0-beta.55",
     "@camptocamp/babel-plugin-angularjs-annotate": "0.8.2",
     "@camptocamp/cesium": "1.45.0",
     "angular": "1.7.0",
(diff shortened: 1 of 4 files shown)


ger-benjamin:
I've choosed babel-polyfill after this advice: https://github.com/babel/babel/issues/2026#issuecomment-122330126

I've choosed "@babel-polyfill" over "babel-polyfill" even if it's tagged as beta because the second one is 1 year old now.

I still have to test, but with localhost, that looks good

package.json
```diff
@@ -108,5 +108,7 @@
     "webpack-dev-server": "3.1.4",
     "webpack-merge": "4.1.2"
   },
-  "dependencies": {}
+  "dependencies": {
+    "@babel/polyfill": "7.0.0-beta.55"
```
sbrunner: should be in dev dependecies ...

ger-benjamin: @sbrunner According to the babel-polyfill doc, no:
> Because this is a polyfill (which will run before your source code), we need it to be a dependency, not a devDependency
But honestly, I don't really know what that mean in our context and in the final code.

sbrunner: For us it mens nothing except adding a dependencies will break c2cgeoportal...

ger-benjamin:
Tested. It works well in IE11 with babel-polyfill

ger-benjamin:
Closed for #4086

PR #4098, opened 2018-08-06
Fix fontawsome 5

sbrunner:


src/filter/component.js
```diff
@@ -16,7 +16,7 @@ import ngeoRuleGeometry from 'ngeo/rule/Geometry.js';
 import ngeoMapFeatureOverlay from 'ngeo/map/FeatureOverlay.js';
 import * as olBase from 'ol/index.js';
 import * as olArray from 'ol/array.js';
-import '@fortawesome/fontawesome-free/scss/fontawesome.scss';
+import 'ngeo/ssss/font.scss';
```
ger-benjamin: `ssss` ? You means `scss` ?
looks also the other occurences

sbrunner: :-/

PR #4106, opened 2018-08-08
Simplify and pass everything through babel

sbrunner:


package.json
```diff
@@ -40,6 +40,8 @@
     "async": "2.6.1",
     "babel-core": "6.26.3",
     "babel-loader": "7.1.5",
+    "babel-plugin-transform-es2015-spread": "6.22.0",
+    "babel-plugin-transform-object-rest-spread": "6.26.0",
```
gberaudo: I think these plugins are not necessary. babel presets env should already handle this.

buildtools/webpack.commons.js
```diff
@@ -23,62 +23,33 @@ const babelPresets = [['env', {
...
+      plugins: ['@camptocamp/babel-plugin-angularjs-annotate'],
     }
   }
 };
 
-const olcsRule = {
-  test: /olcs\/.*\.js$/,
+
+const otherRule = {
+  test: /\/node_modules\/(?!ngeo\/|angular\/).*\.js$/,
```
gberaudo: Why do you pass node_modules in some cases an in other cases not?

PR #4107, opened 2018-08-08
Refresh snapping features after a feature has been modified, so the s…

ochriste:
…napping use the new feature geometry.


sbrunner:
Shouldn't be on 2.3?

ochriste:
I need to recreate the PR

PR #4111, opened 2018-08-09
Refresh snapping features after a feature has been modified, so the s…

ochriste:
…napping use the new feature geometry.


ochriste:
2.2 because I need it in epfl which is still using 2.2

contribs/gmf/src/editing/Snapping.js
```diff
@@ -449,6 +449,15 @@ exports.prototype.loadAllItems_ = function() {
 };
 
 
+/**
+ * Manually refresh all features
+ * @export
```
fredj: minor: `@export` not needed

PR #4119, opened 2018-08-10
Set the print legend DPI only on QGIS

sbrunner:


contribs/gmf/src/print/component.js
```diff
@@ -1130,11 +1128,12 @@ exports.Controller_ = class {
             // Don't add classes without legend url or from layers without any
             // active name.
             if (icons && name.length !== 0) {
-              classes.push({
+              classes.push(Object.assign({
                 'name': gettextCatalog.getString(name),
-                'dpi': dpi,
                 'icons': [icons]
-              });
+              }, layer.getSource().serverType_ == 'qgis' ? {
```
ger-benjamin: ===

ger-benjamin: And there is not a method, like "getServerType" or something like that ?

sbrunner: No :-/

ger-benjamin: :-/

PR #4122, opened 2018-08-13
some Bootstrap fix

sbrunner:


contribs/gmf/src/controllers/desktop.scss
```diff
@@ -69,6 +69,10 @@ main {
   overflow: hidden;
 }
 
+.btn {
+  padding: calc(5rem/16) calc(10rem/16)
```
ger-benjamin: 5rem/16 = 0.3125  = `$half-app-margin`
10rem/16 = 0.625 = `$app-margin`
So why not using these variables ?

sbrunner: Ignoring, thanks, fixed :-)

PR #4142, opened 2018-08-21
Fix Font Awesome icons

ger-benjamin:
GSGMF-418

Some icons needs to have specified `font-weight: 900` to be displayed

### DIFF
--- a/contribs/gmf/apps/mobile/index.html.ejs
+++ b/contribs/gmf/apps/mobile/index.html.ejs
@@ -60,7 +60,7 @@
       <button ngeo-mobile-geolocation=""
         ngeo-mobile-geolocation-map="::mainCtrl.map"
         ngeo-mobile-geolocation-options="::mainCtrl.mobileGeolocationOptions">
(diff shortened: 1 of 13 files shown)


contribs/gmf/apps/mobile/index.html.ejs
```diff
@@ -60,7 +60,7 @@
       <button ngeo-mobile-geolocation=""
         ngeo-mobile-geolocation-map="::mainCtrl.map"
         ngeo-mobile-geolocation-options="::mainCtrl.mobileGeolocationOptions">
-        <span class="fa fa-dot-circle-o"></span>
+        <span class="far fa-dot-circle"></span>
```
fredj: I'm not sure we should have a mix of `fa` and `far` classes

ger-benjamin: @fredj https://fontawesome.com/icons/dot-circle?style=regular The solid one don't match our needs. And we can't use always the "regular" police because the search is not free.

PR #4161, opened 2018-08-24
Fix the non WFS OGC server query, and the alias

sbrunner:


src/datasource/Helper.js
```diff
@@ -101,20 +101,21 @@ const exports = class {
...
-        googAsserts.assert(
-          featureType.complexType[0].name === element.type);
+        googAsserts.assertString(ogcLayerName, 'The data source should have only one ogcLayer.');
+        for (const element of featureType.element) {
+          if (element.name === ogcLayerName) {
+            googAsserts.assert(featureType.complexType[0].name === element.type);
 
-        const complexContent = featureType['complexType'][0]['complexContent'];
-        const attributes = new ngeoFormatWFSAttribute().read(complexContent);
+            const complexContent = featureType['complexType'][0]['complexContent'];
```
ger-benjamin: Can you use featureType.complexType instead of `featureType['complexType']` as two lines above ?

sbrunner: Thanks, fixed :-)

PR #4212, opened 2018-09-06
Fix dynamic.js endpoint for v2.4

loicgasser:
I just tested the apps and noticed an error was just introduced in https://github.com/camptocamp/ngeo/pull/4211


loicgasser:
Oh yes I didn't try those! :D

PR #4223, opened 2018-09-10
Add GmfLayer.dimensions_filters

arnaud-morvan:
Add dimensions_filters for handling dimensions through OGC filters with QGIS server.

ngeo part of https://github.com/camptocamp/aeroportdelyon_sig/issues/399
related with c2geoportal part : https://github.com/camptocamp/c2cgeoportal/pull/4125

### DIFF
--- a/contribs/gmf/externs/gmf-themes.js
+++ b/contribs/gmf/externs/gmf-themes.js
@@ -173,6 +173,14 @@ gmfThemes.GmfLayer.prototype.dimensions;
 
 
 /**
+ * The dimensions applied by filters on the layer configuration, if the value
+ * is null we will take the dimension from the application.
+ * @type {!ngeox.DimensionsFiltersConfig}
+ */
+gmfThemes.GmfLayer.prototype.dimensionsFilters;
+
+
+/**
  * @type {boolean|undefined}
  */
 gmfThemes.GmfLayer.prototype.editable;
--- a/contribs/gmf/src/datasource/Manager.js
+++ b/contribs/gmf/src/datasource/Manager.js
@@ -157,6 +157,14 @@ const exports = class {
     this.dimensions_;
 
     /**
+     * The function to call to unregister the `watch` event on the dimensions
+     * object properties.
+     * @type {?Function}
+     * @private
+     */
+    this.dimensionsWatcherUnregister = null;
+
+    /**
      * The cache of layertree leaf controller, i.e. those that are added to
      * the tree manager. When treeCtrl is added in this cache, it's given
      * a reference to its according data source.
@@ -199,7 +207,50 @@ const exports = class {
    *     object to keep a reference of in this service.
    */
   setDimensions(dimensions) {
+    if (this.dimensionsWatcherUnregister) {
+      this.dimensionsWatcherUnregister();
+    }
+
     this.dimensions_ = dimensions;
+
+    this.dimensionsWatcherUnregister = this.rootScope_.$watch(
+      () => this.dimensions_,
+      this.handleDimensionsChange_.bind(this),
+      true
+    );
+    this.handleDimensionsChange_();
+  }
+
+  /**
+   * Called when the dimensions change. Update all affected layer's filters.
+   * @private
+   */
+  handleDimensionsChange_() {
+
+    // Create a layer list to update each one only once
+    const layers = [];
+    const layerIds = [];
+
+    const dataSources = this.dataSources_.getArray();
+    for (const dataSource of dataSources) {
+      if (dataSource.dimensionsFiltersConfig) {
+        for (const key in dataSource.dimensionsFiltersConfig) {
+          if (dataSource.dimensionsFiltersConfig[key].value === null) {
+            const layer = this.getDataSourceLayer_(dataSource);
+            if (layer == undefined) {
+              return;
+            }
+            const id = olBase.getUid(layer);
+            if (layerIds.indexOf(id) == -1) {
+              layers.push(layer);
+              layerIds.push(id);
+            }
+          }
+        }
+      }
+    }
+
+    layers.forEach(this.updateLayerFilter_.bind(this));
   }
 
   /**
@@ -468,6 +519,7 @@ const exports = class {
     // (7) Dimensions
     const dimensions = this.dimensions_;
     const dimensionsConfig = node.dimensions || firstLevelGroup.dimensions;
+    const dimensionsFiltersConfig = node.dimensionsFilters;
 
     // (8) Time values (lower or lower/upper)
     let timeLowerValue;
@@ -494,6 +546,7 @@ const exports = class {
       copyable,
       dimensions,
       dimensionsConfig,
+      dimensionsFiltersConfig,
       gmfLayer,
       id,
       identifierAttribute,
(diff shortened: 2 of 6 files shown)


arnaud-morvan:
@ger-benjamin: Can you have a look at this PR ?

contribs/gmf/src/datasource/Manager.js
```diff
@@ -685,88 +724,119 @@ const exports = class {
...
+   */
+  updateLayerFilter_(layer) {
     googAsserts.assert(
       layer instanceof olLayerImage ||
       layer instanceof olLayerTile
     );
 
     const source = layer.getSource();
-    googAsserts.assert(
+    if (!(
```
ger-benjamin: no return to line here please

contribs/gmf/externs/gmf-themes.js
```diff
@@ -172,6 +172,14 @@ gmfThemes.GmfLayer = function() {};
 gmfThemes.GmfLayer.prototype.dimensions;
 
 
+/**
+ * The dimensions applied by filters on the layer configuration, if the value
+ * is null we will take the dimension from the application.
+ * @type {!ngeox.DimensionsFiltersConfig}
+ */
+gmfThemes.GmfLayer.prototype.dimensions_filters;
```
ger-benjamin: Please use camelCasing instead of `_`

contribs/gmf/src/datasource/Manager.js
```diff
@@ -685,88 +724,119 @@ const exports = class {
...
-        // parameter as well.
-        const filterParamValues = [];
-        for (let i = 0, ii = layersList.length; i < ii; i++) {
-          if (layersList[i] === filtrableLayerName) {
-            filterParamValues.push(`(${filterString})`);
-          } else {
-            filterParamValues.push('()');
+    const filterParamValues = [];
+    let hasFilter = false;
+    for (let i = 0, ii = layersList.length; i < ii; i++) {
```
ger-benjamin:
```
layersList.foreach((layer) {
...
})
```
?

contribs/gmf/src/datasource/Manager.js
```diff
@@ -685,88 +724,119 @@ const exports = class {
...
+    // the WMS ogcType, i.e. those that do not have an OpenLayers layer
+    // to update
+    if (dataSource.filtrable !== true ||
+        dataSource.ogcType !== ngeoDatasourceOGC.Type.WMS
+    ) {
+      return;
+    }
+
+    const layer = this.getDataSourceLayer_(dataSource);
+    if (layer == undefined) {
```
ger-benjamin: `===` ?

contribs/gmf/src/datasource/Manager.js
```diff
@@ -685,88 +724,119 @@ const exports = class {
...
+  /**
+   * Called when both the 'visible' and 'filterRules' properties of a data
+   * source change.
+   *
+   * If the data source is filtrable, then make sure that when it gets rules
+   * set to apply them as OGC filters to the OpenLayers layer, more precisely
+   * as a `FILTER` parameter in the layer's source parameters.
+   *
+   * @param {!gmf.datasource.OGC} dataSource Data source.
+   * @param {boolean} value Value.
```
ger-benjamin: Value ? What is it ? It is used ?

src/filter/RuleHelper.js
```diff
@@ -440,6 +440,23 @@ const exports = class {
       }
     }
 
+    if (options.incDimensions) {
+      const dimensionsFilter = this.createDimensionsFilterFromDataSource_(dataSource);
+      if (dimensionsFilter) {
+        if (mainFilter) {
+          mainFilter = olFormatFilter.and.apply(
+            null,
+            [
```
ger-benjamin: That's personal, but it it useful to have so many return to line ?
I find this harder to read

src/filter/RuleHelper.js
```diff
@@ -650,6 +667,38 @@ const exports = class {
     return filter;
   }
 
+  /**
+   * Create and return an OpenLayers filter object using the available
+   * dimensions filters configuration within the data source.
+   * @param {ngeo.DataSource} dataSource Data source from which to create the
+   *     filter.
+   * @return {?ol.format.filter.Filter} Filter
```
ger-benjamin: A filter or null.

contribs/gmf/src/datasource/Manager.js
```diff
@@ -199,7 +207,50 @@ const exports = class {
...
+    }
+
     this.dimensions_ = dimensions;
+
+    this.dimensionsWatcherUnregister = this.rootScope_.$watch(
+      () => this.dimensions_,
+      this.handleDimensionsChange_.bind(this),
+      true
+    );
+    this.handleDimensionsChange_();
```
ger-benjamin: handleDimensionsChange_ is only called here. So why do you need this watch, this register, etc... ?

ger-benjamin:
That a big change ! Please test well, perhaps also with EPFL that use a lot these filters

ger-benjamin:
Also, what is it this issue for ? (I means, is there a Jira ticket related ?)

contribs/gmf/src/datasource/Manager.js
arnaud-morvan: dimensions object is only set here, but we need to watch changes in dimensions objet properties ex:
{ "FLOOR": "+1" }

ger-benjamin: Oh, I undertsand. Thanks.
But then, are you sure it works with a simple `$watch` ? ? Don't you need a `$watchCollection`  Because I think that the reference will never change. Otherwise, it's more strange to have a private object that change like that.

contribs/gmf/src/datasource/Manager.js
arnaud-morvan: I would prefer `for (const layer of layersList) {`
Easier to step into with the debugger;

contribs/gmf/src/datasource/Manager.js
arnaud-morvan: You're right, value is not used, it was already the case before my commit. I will remove it.

src/filter/RuleHelper.js
arnaud-morvan: Yes, like createFilterFromRule_ and createTimeFilterFromDataSource_.
Is there something to change in annotations ?

contribs/gmf/src/datasource/Manager.js
arnaud-morvan: This works using the third parameter : true.
$watch(watchExpression, listener, [objectEquality]);
Property was already there, I just added a watcher to update layer's FILTER parameter.
this.dimensions_ is the same object as main controller dimensions property.

arnaud-morvan:
I added links in description for issue and server part.
@ger-benjamin: Are you telling me that I should run a local ngeo:qgis_dimensions-2.3 against EPFL server. But it seems that EPFL is still in v2.2

ger-benjamin:
> But it seems that EPFL is still in v2.2

You are right, sorry.
So if you're able to test it on aeroportdelyon, it must work well.

Thanks !

PR #4228, opened 2018-09-12
2.3.4 wms styles

svareg:
Hello,

In the Admin Tool of C2CGeoportal, we can set a "style" for a layer (i.e. the ClassGroup defined on MapServer). But in Ngeo, this value is never used. 

In this Pull Request, I've corrected this missing parameter evaluation in Ngeo, so that the "style" values that comes with the themes.json will be correctly passed to the layer configuration.

Is it possible for you to verify this code (I don't know ngeo really good, and I perhaps missed something) and merge it with the next 2.3.X Version ?

If you have any question, just ask!

Thanks,
Guillaume.


sbrunner:
The files 
/contribs/gmf/fonts/gmf-icons.eot
/contribs/gmf/fonts/gmf-icons.ttf
/contribs/gmf/fonts/gmf-icons.woff
Shouldn't be added in the repository...
the tag commit should be removed :-)

PR #4239, opened 2018-09-17
Gmf use default style when importing kml

julsbreakdown:
https://jira.camptocamp.com/browse/GSGMF-628

ExtractStyles options is true by default, if style is defined in the kml openlayers will use 
a white border and fill => https://openlayers.org/en/v4.6.5/apidoc/ol.format.KML.html

By setting ExtractStyles to false, the default ol.Style will be used  !


sbrunner:
Should be on branch 2.3

Issue #2601, opened 2017-05-09
First level panel title


rbovard:
When opening left and right panels, there is no title in the header

![image](https://cloud.githubusercontent.com/assets/4549666/25845465/4a1506b0-34ae-11e7-8952-44821490e482.png)

It appears only when a subpanel is open

![image](https://cloud.githubusercontent.com/assets/4549666/25845479/562426fc-34ae-11e7-91eb-10673a087c21.png)

My suggestion is to add default titles:
* Left panel: **Layers**
* Right panel: **Tools**

What do you think?

*BTW: a bold style for the title could be better*

ybolognini:
--> GMF 2.4

PR #4240, opened 2018-09-17
Gmf mobile keep text in header

julsbreakdown:
see https://github.com/camptocamp/ngeo/issues/2601
and https://jira.camptocamp.com/browse/GSGMF-605
capture du 2018-09-24 13-29-01


julsbreakdown:
Ok i will look into it thanks

julsbreakdown:
-_-'

PR #4241, opened 2018-09-17
Fix 652 - Bootstrap 4 dropdown-menu use 'show' instead of 'open'

adube:
Fixes the issue that occurs in the filter "rule component", i.e. when selecting an attribute in the filter tool, the dropdown menu was not opened.  Clicking on the drop-down list did show the menu, but in the background we had a broken state.

Cause: in Bootstrap 4, the class `show` has to be used on the element with the css class `dropdown-menu`, see (inspect the drop down menu when opened, you'll see the "show" css class name":

https://www.w3schools.com/bootstrap4/tryit.asp?filename=trybs_dropdown-menu&stacked=h

In Bootstrap 3, this was different.  The css class `open` had to be set on the element with the `dropdown` css class.

### DIFF
--- a/src/filter/rulecomponent.html
+++ b/src/filter/rulecomponent.html
@@ -1,6 +1,4 @@
-<div
-    class="dropdown"
-    ng-class="{open: $ctrl.rule.active}">
+<div class="dropdown">
   <button
       class="btn btn-default btn-sm btn-block dropdown-toggle"
       type="button"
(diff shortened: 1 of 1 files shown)


adube:
@arnaud-morvan Ready for review.

adube:
@fredj Would you please take care of the review?  This is a very small one.

adube:
Rebase onto master.

PR #4243, opened 2018-09-17
Fix 610 - Render drawn features from draw tool in mobile apps

adube:
Fixes GSGMF-610.

This fix makes it possible to draw features using the draw tool (a.k.a. red lining) from a desktop application, then copy the `rl_features=zzz` parameter from the url and then open a mobile application, paste the `rl_features` param in it and be able to see the features that were drawn, without the possibility to edit them.

It turned out to be a very simple fix.  Here's a complete explanation why.

## How it works

Ngeo provides a `ngeoFeatures` angular value, which is an `ol.Collection`.  That value is injected in the GMF draw feature and it is where features are pushed when drawn.

GMF has a permalink module, which is responsible of reading params from the url.  It also injects `ngeoFeatures`, and any features present in the url are automatically pushed in the collection.

Finally, the rendering of the features in a vector layer are not the responsability of the GMF draw tool.  The abstract desktop controller was responsible of injecting `ngeoFeatures` and create a "FeatureOverlay" out of it.  Moving this code to the abstract app controller was enough to make it work.  No changes required in the draw tool at all.

### DIFF
--- a/contribs/gmf/src/controllers/AbstractAppController.js
+++ b/contribs/gmf/src/controllers/AbstractAppController.js
@@ -64,6 +64,20 @@ const exports = function(config, $scope, $injector) {
   googAsserts.assertInstanceof(this.map, olMap);
 
   /**
+   * Collection of features for the draw interaction
+   * @type {ol.Collection.<ol.Feature>}
+   */
+  const ngeoFeatures = $injector.get('ngeoFeatures');
+
+  /**
+   * @type {ngeo.map.FeatureOverlay}
+   * @export
+   */
+  this.drawFeatureLayer = $injector.get('ngeoFeatureOverlayMgr')
+    .getFeatureOverlay();
+  this.drawFeatureLayer.setFeatures(ngeoFeatures);
+
+  /**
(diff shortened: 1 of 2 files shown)


sbrunner:
Should be on branch 2.3

adube:
> Should be on branch 2.3

Jira marks this as 2.4.  Also, this is a new feature.  Should it really be in 2.3?

adube:
For the record: this is really 2.4.  Merging...

sbrunner:
This
https://github.com/adube/ngeo/blob/2d95d6117aac6c300f52528c75b82474824c6e8e/contribs/gmf/src/controllers/AbstractDesktopController.js#L6
and this 
https://github.com/adube/ngeo/blob/2d95d6117aac6c300f52528c75b82474824c6e8e/contribs/gmf/src/controllers/AbstractDesktopController.js#L218
Should also be moved ...

adube:
@sbrunner I'll create a PR for this.

sbrunner:
Thanks :-)

PR #4246, opened 2018-09-18
Set nodata values to null and not 0

julsbreakdown:
camptocamp/c2cgeoportal#4121
https://jira.camptocamp.com/browse/GSGMF-629

### DIFF
--- a/src/profile/d3Elevation.js
+++ b/src/profile/d3Elevation.js
@@ -407,7 +407,9 @@ const exports = function(options) {
         // Configure the d3 line.
         line = d3.line()
           .x(d => x(distanceExtractor(d)))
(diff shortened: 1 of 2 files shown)


src/profile/d3Elevation.js
```diff
@@ -407,7 +407,9 @@ const exports = function(options) {
         // Configure the d3 line.
         line = d3.line()
           .x(d => x(distanceExtractor(d)))
-          .y(d => y(linesConfiguration[name].zExtractor(d)));
+          .y(d => y(linesConfiguration[name].zExtractor(d)))
+          .defined(d => d.value);
```
fredj: It would be clearer if the condition is changed to: `.defined(d => d.value !== null)`

PR #4244, opened 2018-09-17
Fix 645 - Draw feature, close menu

adube:
In the drawing tool, a.k.a. "redlining", when we right-click on a feature, make sure to close any existing menu.

PR #4251, opened 2018-09-18
Fix 613 - Delete feature vertex on right click with menu

adube:
~~Requires #4244 to be merged first.~~

This patch changes the way vertice can be removed from a feature.

## Originally

The original behaviour was: on left click, delete the vertex immediately.  This is the default behaviour in OpenLayers.

## From now on

This behaviour has been changed to two ways:

 1. on left click + ctrl (cmd on Mac), delete the vertex immediately.
 1. on right click, show a menu with the `Delete vertex` item that, when clicked, deletes the vertex

**Approval required** The first behaviour was not originally specifically asked.  It is a suggestion on my end for avanced users that are used to clicking on the vertice directly to remove them.

## Affected tools

 * GMF drawing tool (a.k.a. redlining)
 * GMF editing tool

## Technicallities

The way the detection of a vertex work is this: upon hitting a feature with the `contextual` menu event, we ask the ngeo featureHelper module to detect if the event occured on top of a vertex.  The style of the vertex is used alongside the current resolution of the map view to return a buffered extent around each vertex of the feature.

If one of these extents contains the location where the event occured, that's a vertex hit.  When that happens, we keep the indexes of the coordinate that hit to be able to delete it later.

The FeatureHelper has 2 new methods to help us in this way:

 * `getVertexInfoAtCoordinate`
 * `removeVertex`

### DIFF
--- a/src/misc/FeatureHelper.js
+++ b/src/misc/FeatureHelper.js
@@ -529,7 +529,6 @@ exports.prototype.createEditingStyles = function(feature) {
  * @param {!ol.Coordinate} coordinate Coordinate.
  * @param {number} resolution Current map view resolution.
  * @return {?Array.<number>} The indexes of the vertex (coordinate) that hits.
- * @export
  */
 exports.prototype.getVertexInfoAtCoordinate = function(
   feature, coordinate, resolution
@@ -720,7 +719,6 @@ exports.prototype.getVertexStyle = function(opt_incGeomFunc) {
  * @param {!ol.Feature} feature Feature.
  * @param {!Array.<number>} vertexInfo The indexes of the vertex
  *     (coordinate) to remove.
- * @export
  */
 exports.prototype.removeVertex = function(feature, vertexInfo) {
   let deleted = false;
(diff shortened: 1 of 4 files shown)


src/misc/FeatureHelper.js
```diff
@@ -508,6 +509,140 @@ exports.prototype.createEditingStyles = function(feature) {
...
+ * coordinates, so 2 indexes are required.
+ *
+ * If removing a vertex would make the geometry invalid, then the
+ * vertex info is not returned.
+ *
+ * @param {!ol.Feature} feature Feature.
+ * @param {!ol.Coordinate} coordinate Coordinate.
+ * @param {number} resolution Current map view resolution.
+ * @return {?Array.<number>} The indexes of the vertex (coordinate) that hits.
+ * @export
```
fredj: The `@export` is not needed (the function is not used in a template)

src/misc/FeatureHelper.js
```diff
@@ -579,6 +714,73 @@ exports.prototype.getVertexStyle = function(opt_incGeomFunc) {
 };
 
 
+/**
+ * Remove a vertex from a feature using the given information (indexes).
+ *
+ * @param {!ol.Feature} feature Feature.
+ * @param {!Array.<number>} vertexInfo The indexes of the vertex
+ *     (coordinate) to remove.
+ * @export
```
fredj: The `@export` is not needed (the function is not used in a template)

adube:
@fredj Fixes applied and rebased onto master.

adube:
Excellent, thank you very much for the review.

PR #4255, opened 2018-09-20
Remove GoogleStreetView from desktop app

arnaud-morvan:
https://jira.camptocamp.com/browse/GSGMF-641


arnaud-morvan:
+1, in my case, I've removed it from aeroportsdelyon_sig

arnaud-morvan:
In such case do we still need : https://github.com/camptocamp/ngeo/blob/453c721cb88b6cf234606f0f7c996f33f6b3217c/contribs/gmf/apps/desktop/index.html.ejs#L322

fredj:
+1 to have it only in desktop_alt

arnaud-morvan:
I've removed GoogleStreetView from desktop app. Ready for review now.

PR #4257, opened 2018-09-20
Fix 607 - Format http(s) links and emails as <a> in query results

adube:
This patch introduces a new filter in ngeo that allows to return a value as trusted HTML in addition to automatically converting http(s) or email strings into anchor html strings.

`StringToHtmlReplacements` is an angular constant that defines expressions that can be converted and a template to use when a string matches.

`ngeoTrustHtmlAuto` is the new filter that does like `ngeoTrustHtml`, but will replace strings with html if a match is found.

Note: Instead of the regex provided in 607 for the email, I used the one from: https://www.regular-expressions.info/email.html, which is lighter when used in a case-insensitive regex and more robust.

### DIFF
--- a/src/misc/filters.js
+++ b/src/misc/filters.js
@@ -339,7 +339,6 @@ exports.filter('ngeoTrustHtml', exports.trustHtmlFilter);
  * @param {angular.$sce} $sce Angular sce service.
  * @param {!Array.<!ngeox.StringToHtmlReplacement>}
  *     ngeoStringToHtmlReplacements List of replacements for string to html.
- *     html.
  * @ngname ngeoTrustHtmlAuto
  */
 exports.trustHtmlAutoFilter = function($sce, ngeoStringToHtmlReplacements) {
--- a/src/misc/filters.js
+++ b/src/misc/filters.js
@@ -324,6 +324,48 @@ exports.filter('ngeoTrustHtml', exports.trustHtmlFilter);
 
 
 /**
+ * A filter to mark a value as trusted HTML, with the addition of
+ * automatically converting any string that matches the
+ * StringToHtmlReplacements list to HTML.
+ *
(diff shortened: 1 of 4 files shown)


fredj:
We have a mix between `StringToHtmlReplacement` and `StringToHtmlReplacements`; can you change to have always `StringToHtmlReplacements`?

src/misc/filters.js
```diff
@@ -323,6 +323,48 @@ exports.trustHtmlFilter = function($sce) {
...
+ *    <p ng-bind-html="ctrl.someValue | ngeoTrustHtmlAuto"></p>
+ *
+ * If you use it, you don't require the "ngSanitize".
+ * @return {function(?):string} The filter function.
+ * @ngInject
+ * @ngdoc filter
+ * @param {angular.$sce} $sce Angular sce service.
+ * @param {!Array.<!ngeox.StringToHtmlReplacement>}
+ *     ngeoStringToHtmlReplacements List of replacements for string to html.
+ *     html.
```
fredj: minor: extra `html.` comment

fredj:
And is the addition in ngeox really needed? (cc @sbrunner )

adube:
Thanks for the reply.

> We have a mix between `StringToHtmlReplacement` and `StringToHtmlReplacements`; can you change to have always `StringToHtmlReplacements`?

StringToHtmlReplacements is the global value that is an array, i.e. `ngeoStringToHtmlReplacements`.

StringToHtmlReplacement is the ngeox definition of what a single entity of a "replacement" is, that is: the expression and the template when we have a match.  i.e.: `ngeox.StringToHtmlReplacement`.

It think this is okay as it is.  Please, let me know if you agree.

fredj:
@adube ok, understood. Please merge

PR #4263, opened 2018-09-24
Fix 612 - GMF Drawing, zoom out if feature doesn't fit map view extent

adube:
Fixes https://jira.camptocamp.com/browse/GSGMF-612

In the GMF Drawing tool, when selecting a feature from the list, if the feature would be too big for the current map view extent, then fit the view to the feature instead of just panning to it.


fredj:
@julsbreakdown please approve and merge if you're happy about this pull request !

PR #4267, opened 2018-09-25
Missing layer param in style function call

julsbreakdown:
https://jira.camptocamp.com/browse/GSGMF-658


fredj:
Thanks @julsbreakdown, I've push a commit to simple the code a bit: the `styleFunction.call? calls are now the same

julsbreakdown:
@fredj  Indeed, looks good =)

kalbermattenm:
Thanks a lot for solving that !

PR #4269, opened 2018-09-26
[WIP] Fix 608 - Support exclusiveGroup metadata for layer groups

adube:
Work in progress.

Fixes https://jira.camptocamp.com/browse/GSGMF-608

This patch introduces the support of the `exclusiveGroup` metadata that can be defined in a GMF tree node (theme), which tells its children that only one of them can be ON (and 'indeterminate') at a time.

The management occurs in the `SyncLayertreeMap`, upon `ngeo-layertree-state` event being fired.

## Todo

 * [ ] deal with loading / initializing multiple ON children


adube:
Ready for review.

contribs/gmf/src/layertree/SyncLayertreeMap.js
```diff
@@ -111,6 +124,100 @@ exports.prototype.sync_ = function(map, treeCtrl) {
...
+ * @param {ngeo.layertree.Controller} treeCtrl The ngeo layertree
+ *     controller that had its state changed.
+ * @private
+ */
+exports.prototype.syncExclusiveGroup_ = function(firstParent, treeCtrl) {
+
+  // No need to do anything if:
+  //  - we're not dealing with an exclusive group
+  //  - the treeCtrl that had its state changed was not turned ON
+  if (!firstParent.node.metadata.exclusiveGroup ||
```
arnaud-morvan: Use of firstParent here seems strange to me. I would expect use of treeCtrl.parent.
firstParent is not the direct parent of treeCtrl, it is the ancestor of treeCtrl directly under root.

arnaud-morvan: For example in case we have this : 
Root (Theme not shown in tree)
- G1 (firstParent <=> panel)
  - G11 (treeCtrl.parent)
    - L111 (treeCtrl)
Exclusive on G11 will be ignored
And exclusive on G1 will apply on G11 children.

adube: I understand.

Logically, this is correct.  But I wonder: how far are we supposed to allow such exclusivity?  It seems a bit confusing to have a group that's exclusive within an other group that isn't.  That's not much, but what happens if we raise this to let's say 6 parents, with a mix of exclusives and not exclusives?

My point it, shouldn't the "exclusive" property make only sense for first level parents?  That's what I had in mind.  If that's not the case, then it would be nice to have more complex examples on the demo server to be able to develop with such cases.

Also, it would be a bit more complex to support such behaviours.  When a node changes its state, the event is fired with a reference to the node itself and its first parent.  The node itself isn't aware it's in an exclusive group, nor what the origin of the group is.  I.E. if it's nested between lots more of other parents and one of them is excluve, we would have to go up in all parents every time a state changes, not only for one in a group, for all of them because we don't know if it's in an exclusive group.  And it would be required to always to up to the top.

I don't have a clear idea how this would be simple at the moment.  I'll have to think more about it...  Comments at this point would be welcome.  @arnaud-morvan Please, let me know what you think.

Oh, and it would also be nice to document how the "exclusiveGroup" property can be used.

@sbrunner ?

arnaud-morvan: It does not surprise me to have an exclusive group within an other group that isn't.
Suppose we have : 
- Group1:
  - Group11:
    - Layer111
    - Layer112
    - Layer113
  - Group12
    - Layer121
    ...

And it doesn't make sense to show multiple layers in Group11, but it make sense to show them in relation with Layer121. From my point of view, this is the most standard use case.

I agree that we need a more deep tree in demo for testing.

Concerning the propagation I pretty sure there not such big complexity to use. It may be handled directly by the events system.
=> When a node state change check the parent exclusive flag and set sibling node's state
If parent node state change, another event will be raised.

If a first level group is flagged exclusive, I expect it to apply only on direct children, not on leaves.

Also note that firstParent is wrongly named and may cause confusion as it is not the first parent, but the last ancestor, first level in the tree.

arnaud-morvan: After discussion with my neighbors, we may agree on "top level ancestor".

arnaud-morvan: @ybolognini : Can you read my last comment and give yours please.

adube:
@arnaud-morvan Ready for second review, with suggested fixes.  Upon merging, I'll squash the commits as one.

adube:
>Looks good to me, maybe @fredj or other ngeo dev could have a look before merge.

Sure.  I'll wait for a second "accept" to merge.

adube:
@fredj ping

adube:
Thanks for the review.

@sbrunner Can I merge this?  I heard discussions about an upcoming demo...

PR #4279, opened 2018-10-01
Add max-zoom setting for point geometries for oedit permalinks

pfirpfel:
see GEO-1465

Test link: http://localhost:3000/contribs/gmf/apps/oeedit/?&objectediting_geomtype=MultiPoint&objectediting_id=Le%20Mol%C3%A9son&objectediting_layer=113&objectediting_theme=ObjectEditing&objectediting_property=name

(You need to be logged into the demo 2.2 backend)

You can create new links here:
http://localhost:3000/contribs/gmf/examples/objecteditinghub.html

### DIFF
--- a/contribs/gmf/src/services/permalink.js
+++ b/contribs/gmf/src/services/permalink.js
@@ -705,9 +705,9 @@ gmf.Permalink.prototype.registerMap_ = function(map, oeFeature) {
   // (1) Initialize the map view with either:
   //     a) the given ObjectEditing feature
   //     b) the X, Y and Z available within the permalink service, if available
-  if (oeFeature && oeFeature.getGeometry()) {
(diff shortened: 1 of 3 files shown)


contribs/gmf/src/services/permalink.js
```diff
@@ -700,8 +707,16 @@ gmf.Permalink.prototype.registerMap_ = function(map, oeFeature) {
   //     b) the X, Y and Z available within the permalink service, if available
   if (oeFeature && oeFeature.getGeometry()) {
     const size = map.getSize();
+    const geom = oeFeature.getGeometry();
```
ger-benjamin: oeFeature.getGeometry() is already called two line upper. Can you call it once (use the variable) ?

PR #4280, opened 2018-10-01
Set inline svg width

fredj:
This is needed for IE11 support


fredj:
The idea is that if the height is 1 em, to keep the aspect ratio, the width should be: (width * 1 em) / height
Does it look correct to you?

ger-benjamin:
Yes, please merge

PR #4286, opened 2018-10-02
[WIP] Fix 609 - Make desktop apps left data panel resizable and collapsible

adube:
Work in progress

Fix 609 - https://jira.camptocamp.com/browse/GSGMF-609

This patch makes it possible to resize and collapse the left 'data-panel' that is in both GMF desktop apps.

We use the jQuery "resizable" widget to do so.

When the handle is clicked, the "collapse" takes effect, using the ngeo MapResize directive.

### Todo

 * [x] define a maximum resizable width
 * [x] adjust the CSS style of the resizable handle to make it visible
 * [x] add an arrow button for to have a UI for the collapsible component
 * ~change the style of the cursor while mouse hovering to have 2 arrows instead of one pointing right~ No need to do that.
 * [x] change the style of the cursor to "pointer" on the arrow button
 * [x] fix bug, missing icon

### DIFF
--- a/contribs/gmf/apps/desktop/index.html.ejs
+++ b/contribs/gmf/apps/desktop/index.html.ejs
@@ -17,7 +17,12 @@
       </div>
     </header>
     <main>
-      <div class="gmf-app-data-panel" ng-cloak>
+      <div class="gmf-app-data-panel"
+      <div class="gmf-app-data-panel"
+           ngeo-resizemap="mainCtrl.map"
+           ngeo-resizemap-state="mainCtrl.dataPanelActive"
+           ng-class="{'gmf-app-inactive': !mainCtrl.dataPanelActive}"
+           ng-cloak>
         <div class="gmf-app-header">
           <div class="dropdown">
             <a href class="btn btn-block prime dropdown-toggle"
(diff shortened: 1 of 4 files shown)


adube:
I use font-awesome to render "left" and "right" arrow icons in the collapse/expand "button".  I don't know why, but they don't render correctly, as if it was missing the font.  It CSS class names seem to be defined properly.

Anyone knows what I'm doing wrong?

adube:
Here's what it currently looks like:

screenshot_2018-10-03_11-30-23

fredj:
@adube it looks like that the icon must be inside the div. I've pushed a commit to fix that: 8fb011c

adube:
> @adube it looks like that the icon must be inside the div. I've pushed a commit to fix that: [8fb011c](https://github.com/camptocamp/ngeo/commit/8fb011c49237606c3d0f1639de7d125543ee180e)

Thank you very much.

adube:
Thanks.  I'm making some trivial enhancements to the interface for the button to be more centered, and I'll also use the same arrows as the south panel.  Changes should be coming soon.

PR #4290, opened 2018-10-05
API skeleton

fredj:


fredj:
Not complete yet but merging as step 1

api/Map.js
```diff
@@ -0,0 +1,209 @@
...
+import VectorLayer from 'ol/layer/Vector.js';
+import TileLayer from 'ol/layer/Tile.js';
+
+import {get as getProjection} from 'ol/proj.js';
+
+import * as constants from './constants.js';
+
+// FIXME: temporary
+import SwisstopoSource from '@geoblocks/sources/Swisstopo.js';
+import EPSG_21781 from '@geoblocks/sources/EPSG21781.js';
```
ger-benjamin: Can you use `EPSG21781` as in the rest of the project (to search and replace in the future for instance)

api/Map.js
```diff
@@ -0,0 +1,209 @@
...
+          src: options.icon
+        })
+      }));
+    }
+    this.vectorSource_.addFeature(marker);
+  }
+
+  /**
+   * @param {string} layer
+   * @param {Array.<string>} layer
```
ger-benjamin: ids ?

api/Map.js
```diff
@@ -0,0 +1,209 @@
...
+   */
+  constructor(options) {
+    /**
+     * @private
+     * @type {View}
+     */
+    this.view_ = new View({
+      projection: getProjection(constants.projection || 'EPSG:21781'),
+      resolutions: constants.resolutions,
+      zoom: options.zoom,
```
ger-benjamin: options.zoom = 10 ?

api/Map.js
```diff
@@ -0,0 +1,209 @@
...
+   * @param {string} type
+   * @param {string} name
+   * @param {string} url
+   * @param {Object} [options]
+   * @property {Array.<string>} [attr=['title', 'description']]
+   * @property {function()} [success]
+   * @property {function()} [error]
+   */
+  addCustomLayer(type, name, url, options = {}) {
+    if (type === 'text') {
```
ger-benjamin: no `else` ? So perhaps you can avoid this parameter ? (or add a FIXME if you plan to support more types)

api/Map.js
```diff
@@ -0,0 +1,209 @@
...
+      fetch(url)
+        .then((response) => response.text())
+        .then((text) => {
+          const attr = options.attr || ['title', 'description'];
+          const lines = text.split(/\r\n|\r|\n/);
+          const columns = lines.shift().split('\t');
+          for (const line of lines) {
+            if (line) {
+              const values = zip(columns, line.split('\t'));
```
ger-benjamin: (It is cross compatible ? It is always` \t` ?)

api/Map.js
fredj: done

api/Map.js
fredj: oups, thanks!

api/Map.js
fredj: done

api/Map.js
fredj: The old api supports text and gpx but the new version only supports text.
Because we need to be compatible with the old api, the type must stay; I've removed the if

api/Map.js
fredj: yes, it's always `\t`. It was like this in openlayers 2:
https://github.com/openlayers/ol2/blob/master/lib/OpenLayers/Format/Text.js#L91

ger-benjamin: Ok, I was just suspicious

ger-benjamin:
Please feel free to merge

PR #4297, opened 2018-10-08
Use prototype for array concat

julsbreakdown:


fredj:
we must have this one for the 2.3 branch

PR #4309, opened 2018-10-11
Add example usage of routing configuration [2.3]

pfirpfel:
I added the example configuration for the routing in the desktop_alt app, which is described here:
http://camptocamp.github.io/c2cgeoportal/2.3/integrator/routing.html#configuration


pfirpfel:
Looks like this would [be redundant](https://github.com/camptocamp/demo_geomapfish/blob/ac3f2698303b3da4265401423f4a503c748ab3e9/vars.yaml#L122-L136). Sorry.

PR #4310, opened 2018-10-11
Support columnsOrder in grid and window components

arnaud-morvan:
Add attribute columnsOrder to gmfDatasourceOGC base on columnsOrder metadata.
Add property ngeo_datasource_id to features in Querent.
Use columns order from datasource in grid and window components.

Client part of https://jira.camptocamp.com/browse/GSGMF-616

### DIFF
--- a/contribs/gmf/src/datasource/Manager.js
+++ b/contribs/gmf/src/datasource/Manager.js
@@ -488,11 +488,11 @@ const exports = class {
     const name = gmfLayer.name;
     const timeAttributeName = meta.timeAttribute;
     const visible = meta.isChecked === true;
-    const columnsOrder = meta.columnsOrder;
+    const attributesOrder = meta.attributesOrder;
 
     // Create the data source and add it to the cache
     this.dataSourcesCache_[id] = new gmfDatasourceOGC({
-      columnsOrder,
+      attributesOrder,
       copyable,
       dimensions,
       dimensionsConfig,
--- a/contribs/gmf/src/datasource/Manager.js
+++ b/contribs/gmf/src/datasource/Manager.js
@@ -488,9 +488,11 @@ const exports = class {
     const name = gmfLayer.name;
     const timeAttributeName = meta.timeAttribute;
     const visible = meta.isChecked === true;
+    const columnsOrder = meta.columnsOrder;
 
     // Create the data source and add it to the cache
     this.dataSourcesCache_[id] = new gmfDatasourceOGC({
+      columnsOrder,
       copyable,
       dimensions,
       dimensionsConfig,
--- a/contribs/gmf/src/datasource/OGC.js
+++ b/contribs/gmf/src/datasource/OGC.js
@@ -24,6 +24,11 @@ const exports = class extends ngeoDatasourceOGC {
      */
     this.gmfLayer_ = options.gmfLayer;
 
+    /**
+     * @type {?Array.<string>}
+     * @export
+     */
+    this.columnsOrder_ = options.columnsOrder;
   }
 
   // === Static property getters/setters ===
@@ -34,6 +39,14 @@ const exports = class extends ngeoDatasourceOGC {
    */
   get gmfLayer() {
     return this.gmfLayer_;
+  }
+
+  /**
+   * @return {?Array.<string>} Ordered list of column names
+   * @export
+   */
+  get columnsOrder() {
+    return this.columnsOrder_;
   }
 
 };
--- a/contribs/gmf/src/datasource/OGC.js
+++ b/contribs/gmf/src/datasource/OGC.js
@@ -25,10 +25,10 @@ const exports = class extends ngeoDatasourceOGC {
     this.gmfLayer_ = options.gmfLayer;
 
     /**
-     * @type {?Array.<string>}
+     * @type {Array.<string>|undefined}
      * @export
      */
-    this.columnsOrder_ = options.columnsOrder;
+    this.attributesOrder_ = options.attributesOrder;
   }
 
   // === Static property getters/setters ===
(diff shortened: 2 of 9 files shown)


adube:
@arnaud-morvan I should be able to review this tomorrow.  Meanwhile, would you please check the errors reported by travis and also include the github.io link that showcases this addition?  Thanks.

arnaud-morvan:
I fixed the CI, but for the moment, I did not manage to provide the required metadata on demo server.

contribs/gmf/src/datasource/Manager.js
```diff
@@ -488,9 +488,11 @@ const exports = class {
     const name = gmfLayer.name;
     const timeAttributeName = meta.timeAttribute;
     const visible = meta.isChecked === true;
+    const columnsOrder = meta.columnsOrder;
```
adube: We should add the `columnsOrder` property definition to the `gmfThemes.GmfMetaData` in `gmf-themes.js`

contribs/gmf/src/datasource/OGC.js
```diff
@@ -24,6 +24,11 @@ const exports = class extends ngeoDatasourceOGC {
      */
     this.gmfLayer_ = options.gmfLayer;
 
+    /**
+     * @type {?Array.<string>}
+     * @export
+     */
+    this.columnsOrder_ = options.columnsOrder;
```
adube: In a datasource, we the term "column" is not used.  We use "attribute".  I think it would make more sense to use `attributesOrder` here.

Also, the type shold be `Array.<string>|undefined`.

Also, we're missing the type definition of of the new option in `ngeox.datasource.DataSourceOptions`

contribs/gmf/src/datasource/OGC.js
```diff
@@ -36,6 +41,14 @@ const exports = class extends ngeoDatasourceOGC {
...
   }
 
+  /**
+   * @return {?Array.<string>} Ordered list of column names
+   * @export
+   */
+  get columnsOrder() {
+    return this.columnsOrder_;
+  }
+
```
adube: attributesOrder

undefined

contribs/gmf/src/query/gridComponent.js
```diff
@@ -675,13 +686,18 @@ exports.Controller_.prototype.makeGrid_ = function(data, source) {
...
   const clone = {};
   Object.assign(clone, data[0]);
   delete clone.ol_uid;
-  const columns = Object.keys(clone);
+
+  const dataSource = this.ngeoDataSources_.collection.getArray().find(
+    ds => ds.id == clone.ngeo_datasource_id);
+  delete clone.ngeo_datasource_id;
+
+  const columns = dataSource && dataSource.columnsOrder.filter(key => key in clone)
```
adube: if `dataSource.columnsOrder` is undefined (i.e. attributesColumn) then this would fail.  Let's add `&& dataSource.columnsOrder`.

Also, personnal opinion: I like one liner if/else when they are short. This one is pretty long, so I wouldn't mind a regular if/else statement instead to make the code easier to read.

contribs/gmf/src/query/windowComponent.js
```diff
@@ -210,6 +214,12 @@ exports.Controller_ = function($element, $scope, ngeoQueryResult, ngeoMapQuerent
    */
   this.ngeoFeatureOverlayMgr_ = ngeoFeatureOverlayMgr;
 
+  /**
+   * @type {ngeo.datasource.DataSources}
+   * @private
+   */
+  this.ngeoDataSources_ = ngeoDataSources;
+
```
adube: Okay, I understand why we use the ngeoDataSources to get the ordered attributes and bind the data source id in the feature later to find it amond this list.

Would it make more sense to bind the data source to the window component instead?

arnaud-morvan: Not sure to understand what you expect, note that there can be features coming from multiple datasources at the same time in the query result / window components features collection.

adube: Oh... I hadn't figured that.  I looked at the component more and it would indeed be too complex to try to implement something there...  We have the "sources", a.k.a. the query result sources, that could be a point of entry that would be "unique per data source", but let's not do that for now.  It would be too complex.

arnaud-morvan: Note that we could point the datasource directly instead of datasource_id in feature properties.
No need to search for datasource after that.

contribs/gmf/src/query/gridComponent.js
```diff
@@ -141,6 +141,7 @@ exports.component('gmfDisplayquerygrid', exports.component_);
  *     overlay manager service.
  * @param {!angular.$timeout} $timeout Angular timeout service.
  * @param {!ngeo.download.Csv} ngeoCsvDownload CSV download service.
+ *     data sources service.
```
adube: Remove this extra line

src/query/Querent.js
```diff
@@ -430,6 +430,7 @@ const exports = class {
       if (readFeatures.length > 0) {
         readFeatures.forEach((feature) => {
           feature.set('ngeo_feature_type_', type);
+          feature.set('ngeo_datasource_', dataSource);
```
adube: That makes sense.  Good catch.

adube:
@arnaud-morvan Did you successfuly manage to provide the required metadata on your demo server?  If so, please let me know.

arnaud-morvan:
@adube : this is waiting for specification now, as we might not use the same metadata for editing and querying as the field are not mandatory the sames.

adube:
> @adube : this is waiting for specification now, as we might not use the same metadata for editing and querying as the field are not mandatory the sames.

Thanks for letting me know.  If there is anything else, let me know.

arnaud-morvan:
This PR was opened to order fields in WFS query results in same order as editingColumnsOrder.
But we observed that editing is based on PostgreSQL table and query result are based on WFS service.
I was decided to reorder columns only for editing, see last comments on Jira ticket : https://jira.camptocamp.com/browse/GSGMF-616
Closing

PR #4339, opened 2018-10-26
Fix 617 - Iframe API new app

adube:
This patch introduces a new application called `iframe_api`, which is designed to be included within an iframe and doesn't have any tool in it.

This patch also introduces a new `AbstractAPIController`, which creates only the map, ensures that the bbox query tool can be implemented and defines some query options.  Desktop now inherits from it.

See also: https://jira.camptocamp.com/browse/GSGMF-617

### Attribution

~~The app includes a `olAttribution`, but it would seem that none of the layers created by GMF/NGEO are configured to work with it.  This is worth discussing...~~

We're using the disclaimer instead.

### LayerTree

While working on the disclaimer / attribution, I noticed that the disclaimer component was already there but nothing was shown.  I also noticed that there were not layers being created in the map.  Then, I remembered: the `<gmf-layertree>` is mandatory if we want our layers to be created in GMF.

So, as part of this app, it is created, but hidden using CSS.  As soon as I did that, I saw the disclaimers appear and the features on the map as well from the layers loaded in the default theme.


contribs/gmf/apps/iframe_api/Controller.js
```diff
@@ -0,0 +1,73 @@
...
+/**
+ * @param {angular.Scope} $scope Scope.
+ * @param {angular.$injector} $injector Main injector.
+ * @constructor
+ * @extends {gmf.controllers.AbstractDesktopController}
+ * @ngInject
+ * @export
+ */
+const exports = function($scope, $injector) {
+  gmfControllersAbstractDesktopController.call(this, {
```
sbrunner: I it a good idea to use the AbstractDesktopController?
For me he contains many things that not useed on API interface ...

adube: That's a good point.  Indeed, AbstractDesktopController contains more things that are not required by IframeAPI those required:

 * map creation
 * feature helper / vector features (if we want to load vector features in the map)
 * isDesktop true
 * query options

If you think it is best to have those copied in the controller of the app and extend AbstractAppController, please let me know.

sbrunner: Isn't the best to create an AbstractAPIController and AbstractDesktopController will extend it?

adube: Sure, I'll do that.

sbrunner:
Thanks :-)

adube:
@sbrunner Thanks.

(Before merging): Should we address the attribution issue in an other task, then?  (see description above)

sbrunner:
@adube Effectively it seem that's not used, can you remove it on all the applications?
Thanks :-)

sbrunner:
And merge it :-)

PR #4356, opened 2018-11-06
Update fontawesome classes to work with FT5

julsbreakdown:
https://jira.camptocamp.com/browse/GSGMF-698


contribs/gmf/src/drawing/drawFeatureComponent.js
```diff
@@ -674,11 +674,11 @@ exports.Controller_.prototype.handleMapContextMenu_ = function(evt) {
           type == ngeoGeometryType.POLYGON ||
           type == ngeoGeometryType.RECTANGLE) {
         actions = actions.concat([{
-          cls: 'fa fa-arrows',
```
fredj: Note: `fa-arrows` exists but is in the pro "pack"

PR #4370, opened 2018-11-08
Fix 693 - Add recenterOnObjects in JS API

adube:
Support the recenterOnObjects method in the JS API.

### DIFF
--- a/api/src/Map.js
+++ b/api/src/Map.js
@@ -15,11 +15,17 @@ import {createStringXY} from 'ol/coordinate.js';
 import ScaleLine from 'ol/control/ScaleLine.js';
 import OverviewMap from 'ol/control/OverviewMap.js';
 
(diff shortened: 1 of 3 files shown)


api/src/Map.js
```diff
@@ -15,11 +15,17 @@ import {createStringXY} from 'ol/coordinate.js';
 import ScaleLine from 'ol/control/ScaleLine.js';
 import OverviewMap from 'ol/control/OverviewMap.js';
 
-import {getCenter} from 'ol/extent.js';
+import {
```
fredj: personal taste but what about changing it to:
```
import * as extent from "ol/extent.js";
```
and then use `extent.getCenter()` ?

adube: I haven't actually checked if it affects the size of the build in the end if you use * instead of importing each method individually, but that's the thought behind this: don't import things you don't need.

PR #4376, opened 2018-11-13
Fix error message modal dialog in editFeatureComponent

arnaud-morvan:
I don't know why `$parent` was there, it probably works some time.
But it does not works anymore and `efCtrl` is used directly everywhere in this file since 2 years.


arnaud-morvan:
@julsbreakdown: Can you have a look here, I need an approve to merge.

PR #4378, opened 2018-11-13
Be able to run the Webpack dev server in https

sbrunner:


gberaudo:
@sbrunner, what domain your certificates use?
I would be reluctant to whitelist a localhost certificate, but a dedicated domain (ngeo, to wire in /etc/hosts would be fine).

sbrunner:
localhost ...

sbrunner:
For me a requirement to change /etc/hosts is not fine...

gberaudo:
I am approving this change: while it is not a good habit to whitelist certificates, one should not input sensitive data to https://localhost...

PR #4389, opened 2018-11-16
Use between as temporal operator with QGIS

arnaud-morvan:
QGIS does not support "During" operator.

### DIFF
--- a/src/filter/RuleHelper.js
+++ b/src/filter/RuleHelper.js
@@ -1,6 +1,7 @@
 import angular from 'angular';
 import ngeoFilterCondition from 'ngeo/filter/Condition.js';
 import ngeoFormatAttributeType from 'ngeo/format/AttributeType.js';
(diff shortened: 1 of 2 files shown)


arnaud-morvan:
@sbrunner : it was much simpler than I thought

sbrunner:
:-)

arnaud-morvan:
@sbrunner : I need it in 2.2 for ADL and there is big problem with typing (between take two numbers).
As we might restore type checking, here is a new version with new filter operator that accepts strings.

sbrunner:
I don't think that there is a big problem, but it can be good to have tow different pull requests :-)

arnaud-morvan:
@sbrunner : I've rebase this old PR, still ok for merge ?

fredj:
looks code but `src/ol-ext/format/filter.js` needs to be converted to es6 modules / classes

arnaud-morvan:
Thanks, I forgot to remove this file during rebase, done now.

PR #4396, opened 2018-11-20
668 - Mobile measure area

adube:
Fixes https://jira.camptocamp.com/browse/GSGMF-668

Introduce the mobile "area" measure tool.

Since it was very similar to the "length" measurement, its controller was broke down into a base class for lenght and area to inherit from it.  They also share the same template.

 * [x] fix the GMF mobile measure example tooltip (was not showing)
 * [x] implement a parent class for length and area measure components to extend from
 * [x] implement an area component
 * [x] support polygons in mobile draw interaction (ngeo)
 * [x] update the GMF mobile measure example to include the area
 * [x] update the GMF mobile_alt app to include the area


contribs/gmf/apps/mobile_alt/index.html.ejs
```diff
@@ -31,6 +31,14 @@
         gmf-displayquerywindow-featuresstyle="::mainCtrl.queryFeatureStyle"
         gmf-displayquerywindow-defaultcollapsed="false">
       </gmf-displayquerywindow>
+      <div
+        class="gmf-mobile-measure"
+        gmf-mobile-measurearea
+        gmf-mobile-measurearea-active="mainCtrl.measureAreaActive"
+        gmf-mobile-measurearea-precision="2"
```
sbrunner: In 2.5 it's plane to remove all those kink of configuration to replace the by config object injection ...

adube: Okay, but what about now?  Should I change the config of this component and make it not the same as the others?  I think this should all be done in 2.5.

Thoughts?

sbrunner: No problem for me :-)

contribs/gmf/src/mobile/measure/areaComponent.js
```diff
@@ -1,6 +1,7 @@
 /**
  * @module gmf.mobile.measure.areaComponent
  */
+import * as angular frmo 'angular';
```
adube: from

fredj: oups, I pushed by mistake, sorry

PR #4399, opened 2018-11-21
IE11 distored layer images - backport #4230

kalbermattenm:
#4230 was merged into master, but we need that in 2.3.

I tested that locally since I am now able to use a local NGEO distribution out of Docker (and this is a little revolution, thanks @sbrunner !!!)

See comments in https://jira.camptocamp.com/browse/GSGMF-625#add-comment for more info !

Please review


sbrunner:
Thanks @kalbermattenm

PR #4406, opened 2018-11-23
Remove jsdoc, add tsdoc config file

fredj:


.gitignore
```diff
@@ -1,5 +1,6 @@
 *.pyc
 .*.swp
+apidoc/
```
sbrunner:
```suggestion
/apidoc/
```

tsconfig.json
```diff
@@ -0,0 +1,18 @@
+{
+    "compilerOptions": {
+      /* Basic Options */
+      "target": "ES6",
+      "module": "system",
+      "lib": ["es2017", "dom"],
+      "allowJs": true,
+      "checkJs": false,
```
sbrunner: Shouldn't be true?

tsconfig.json
```diff
@@ -0,0 +1,18 @@
...
+      "noEmit": true
+    },
+    "include": [
+      "src/**/*.js",
+      "contribs/gmf/src/**/*.js",
+      "node_modules/@types/openlayers/*",
+      "types/**/*"
```
sbrunner: Those tow lines are relay needed, wanted?

tsconfig.json
fredj: the doc generation doesn't work if this option sis set to true (too many errors)

PR #4427, opened 2018-11-29
Fix filter on layer with different value in name and layers

sbrunner:


contribs/gmf/src/datasource/Manager.js
```diff
@@ -782,7 +782,8 @@ const exports = class {
           continue;
         }
         if (olBase.getUid(dsLayer) == olBase.getUid(layer) &&
-            dataSourceName === dataSource.name)  {
+            layer.get('querySourceIds').indexOf(dataSource.id) >= 0 &&
+            dataSource.gmfLayer_.layers.split(',').indexOf(wmsLayerName) >= 0)  {
```
fredj: `gmfLayer` instead of `gmfLayer_`

PR #4434, opened 2018-12-03
[wip] More code cleanup

fredj:
Don't export private classes, directive and components.


contribs/gmf/src/authentication/module.js
```diff
@@ -1,7 +1,7 @@
 /**
  * @module gmf.authentication.module
  */
-import * as angular from 'angular';
+import angular from 'angular';
```
sbrunner: Your sure that's correct?

PR #4471, opened 2018-12-18
Improve display window template management

kalbermattenm:
Actually @ger-benjamin did that... I leave the complete explanation up to him...

FIX: GEO-1113


ger-benjamin:
The aim is to update the content of the windowComponent at any time, and not only at the onInit of the windowComponent. With this change we can display the content of a component even if it's initialized after the windowComponent.
(The watch is a little bit heavy, but that's the easiest solution, and it's not SO heavy )

kalbermattenm:
Thanks a lot for having merged that ! I am really looking forward to see this PR in the next coming GMF version !

PR #4483, opened 2018-12-21
Fix #4471 - Call the function in watcher

kalbermattenm:
#4471 is not working, because, IHMO, we need to call the function and not just define it in the watcher.

I tested that in my project and it seems to work.

Could somebody please review that and eventually merge it, because I really need it inside 2.3...

Thanks a lot.


ger-benjamin:
Oups, thanks, please merge !

PR #4507, opened 2019-01-08
Fix Babel and Uglify config for the API build

fredj:


package.json
```diff
@@ -6,7 +6,7 @@
     "build-ngeo-examples": "TARGET=ngeo-examples webpack --mode development --progress --debug",
     "build-gmf-examples": "TARGET=gmf-examples webpack --mode development --progress --debug",
     "build-gmf-apps": "TARGET=gmf-apps webpack --mode development --progress --debug",
-    "build-api": "webpack --config buildtools/webpack.api.js --mode development --library demo",
+    "build-api": "webpack --config buildtools/webpack.api.js --mode production",
```
sbrunner: Any reason that all build- do a development build and this one a production one?

fredj: yes, we can have a development build here and a production build in c2cgeoportal

fredj: fixed

PR #4522, opened 2019-01-16
Use the OnlineResource from the WMS if the url is not provided

fredj:


contribs/gmf/src/datasource/ExternalDataSourcesManager.js
```diff
@@ -322,6 +322,8 @@ const exports = class {
     const id = exports.getId(layer);
     const service = capabilities['Service'];
 
+    url = url !== undefined ? url : service['OnlineResource'];
```
sbrunner: Why not?
```suggestion
    url = service['OnlineResource'] || url;
```

PR #4533, opened 2019-01-18
More jsdoc type fixes

fredj:


sbrunner:
For me the commit `Remove extends jsdoc tags  is wrong` ...

PR #4538, opened 2019-01-21
Fix the type defs

sbrunner:
And in WfsPermalink


contribs/gmf/src/objectediting/toolsComponent.js
```diff
@@ -26,7 +26,7 @@ import {getUid as olUtilGetUid} from 'ol/util.js';
 /**
  * Additional configuration options for the object editing tools directive.
  * @typedef {Object} ObjectEditingToolsOptions
- * @property regularPolygonRadius {number|undefined} The radius of the shapes created by the regular polygon
+ * @property {number} [regularPolygonRadius] The radius of the shapes created by the regular polygon
```
fredj:
```suggestion
 * @property {number} [regularPolygonRadius=100] The radius of the shapes created by the regular polygon
```
To have the default value in the API doc

PR #4553, opened 2019-01-25
Fix types in message and misk

sbrunner:


package.json
```diff
@@ -100,6 +103,7 @@
     "moment": "2.22.2",
     "node-sass": "4.11.0",
     "node-sass-importer": "1.0.0",
+    "npm": "^6.7.0",
```
fredj: is it correct to have npm here ?

package.json
```diff
@@ -79,6 +81,7 @@
     "fs-extra": "7.0.1",
     "html-webpack-include-assets-plugin": "1.0.6",
     "html-webpack-plugin": "3.2.0",
+    "install": "^0.12.2",
```
fredj: ?

package.json
sbrunner: ?

package.json
sbrunner: ???

PR #4570, opened 2019-01-31
Add a option to send bbox as GET param of getfeature requests

ger-benjamin:
To be able to access the bbox of the queried zone in a MapFile.
GEO-1746

### DIFF
--- a/src/services/mapquerent.js
+++ b/src/services/mapquerent.js
@@ -92,6 +92,12 @@ ngeo.MapQuerent = class {
       options.tolerance : 3;
 
     /**
+     * @type {boolean}
+     * @private
+     */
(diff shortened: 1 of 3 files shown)


src/services/mapquerent.js
```diff
@@ -91,6 +91,12 @@ ngeo.MapQuerent = class {
     this.tolerancePx_ = options.tolerance !== undefined ?
       options.tolerance : 3;
 
+    /**
+     * @type {boolean}
+     * @private
+     */
+    this.bboxAsGETParam_ = options.bboxAsGETParam ? || false;
```
fredj:
```suggestion
    this.bboxAsGETParam_ = options.bboxAsGETParam !== undefined ? options.bboxAsGETParam : false;
```

PR #4573, opened 2019-01-31
Fix typecheck (mostly for gmf lidar profile)

adube:
Typecheck fixes, mostly for lidarprofile.  Not done yet, but if everything here is okay then we could merge as-is.


fredj:
merging to include this in the next release

PR #4588, opened 2019-02-04
Even more typecheck fixes for gmf

adube:
As title says.

I noticed that the getter/setters `filterCondition` for ngeo/datasource/OGC have been removed.  Why?

### DIFF
--- a/contribs/gmf/src/datasource/Helper.js
+++ b/contribs/gmf/src/datasource/Helper.js
@@ -3,7 +3,7 @@ import gmfEditingEnumerateAttribute from 'gmf/editing/EnumerateAttribute.js';
 import ngeoDatasourceHelper from 'ngeo/datasource/Helper.js';
 import ngeoFormatAttributeType from 'ngeo/format/AttributeType.js';
 
-class Helper {
+export class Helper {
 
   /**
    * A service that provides utility methods to manipulate or get GMF data
--- a/contribs/gmf/src/datasource/Helper.js
+++ b/contribs/gmf/src/datasource/Helper.js
@@ -12,7 +12,7 @@ export class Helper {
    * @param {angular.IQService} $q The Angular $q service.
    * @param {import("gmf/editing/EnumerateAttribute.js").EditingEnumerateAttributeService} gmfEnumerateAttribute The Gmf enumerate
    *     attribute service.
-   * @param {import("ngeo/datasource/Helper.js").Helper} ngeoDataSourcesHelper Ngeo data
+   * @param {import("ngeo/datasource/Helper.js").DatasourceHelper} ngeoDataSourcesHelper Ngeo data
    *     source helper service.
    * @ngdoc service
    * @ngname gmfDataSourcesHelper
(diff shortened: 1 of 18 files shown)


contribs/gmf/src/datasource/OGC.js
```diff
@@ -43,3 +43,5 @@ export default class extends ngeoDatasourceOGC {
   }
 
 }
+
+export default gmfDatasourceOGC;
```
sbrunner: Why this change?

contribs/gmf/src/layertree/component.js
```diff
@@ -173,8 +175,9 @@ module.component('gmfLayertree', component);
  * @ngdoc controller
  * @ngname gmfLayertreeController
  */
-function Controller($element, $scope, ngeoLayerHelper, gmfDataSourceBeingFiltered,
-  gmfExternalDataSourcesManager, gmfPermalink, gmfTreeManager, gmfSyncLayertreeMap, ngeoWMSTime, gmfThemes) {
+function Controller($element, $scope, ngeoLayerHelper,
+  gmfDataSourceBeingFiltered, gmfExternalDataSourcesManager, gmfPermalink,
+  gmfTreeManager, gmfSyncLayertreeMap, ngeoWMSTime, gmfThemes) {
```
sbrunner: Why this change We can have line of 110 char length...

src/datasource/OGC.js
```diff
@@ -476,6 +476,22 @@ class OGC extends ngeoDatasourceDataSource {
     return this.dimensions_;
   }
 
+  /**
+   * @return {string} Filter condition
+   * @export
+   */
+  get filterCondition() {
```
fredj: why do we need to have getter and setter for this variable ?

adube: 1) This variable needs to be changed.  In the filter component, one can set OR instead of AND, which affect this variable, thus it needs a setter
2) It needs a getter because we bind UI components to it
3) If you look around the other properties of the data source classes, you'll notice that they have getters/setters as well and those we're removed.  If this one was removed, then why not the others?
4) Back then, when this was first implemented, it was my first time writing ES6 classes, and the use of getters/setters seemed quite nice and it was working with Angular well.  This was a good choice.
5) The typecheck complains that there is not a filterCondition property set.  Is this because inner properties can't be accessed?

Please, comment.  Thanks.

fredj: ok, so it was changed ("simple" public variable to getter / setter) for the typecheck ?

contribs/gmf/src/datasource/Helper.js
```diff
@@ -3,7 +3,7 @@ import gmfEditingEnumerateAttribute from 'gmf/editing/EnumerateAttribute.js';
 import ngeoDatasourceHelper from 'ngeo/datasource/Helper.js';
 import ngeoFormatAttributeType from 'ngeo/format/AttributeType.js';
 
-class Helper {
+export class Helper {
```
sbrunner: Please rename it to `DatasourceHelper` to don't have to rename ii if we import it :-)

src/datasource/OGC.js
adube: I don't know why it was changed.  It was using a getter/setter before and it was working fine.

adube:
Please, let me know if I can merge this.  Thanks.

PR #4593, opened 2019-02-05
Fix types

sbrunner:


contribs/gmf/src/filters/filterselectorComponent.js
```diff
@@ -248,7 +248,7 @@ class Controller {
     /**
      * The data source ready to be filtered, after it has been selected and
      * prepared.
-     * @type {?import("gmf/datasource/OGC.js").default}
+     * @type {?import("gmf/datasource/OGC.js").OGCOptions}
```
adube: This is invalid.  The type of this property is not a hash of option. It's an instance of a class.

contribs/gmf/src/filters/filterselectorComponent.js
```diff
@@ -541,8 +541,7 @@ class Controller {
   }
 
   /**
-   * @param {?import("gmf/datasource/OGC.js").default} dataSource Newly selected data source
-   *     object.
+   * @param {?import("gmf/datasource/OGC.js").OGCOptions} dataSource Newly selected data source object.
```
adube: idem

contribs/gmf/src/filters/filterselectorComponent.js
```diff
@@ -563,9 +562,7 @@ class Controller {
       this.active = true;
     }
 
-    this.gmfDataSourcesHelper_.prepareFiltrableDataSource(
-      dataSource
-    ).then((dataSource) => {
+    this.gmfDataSourcesHelper_.prepareFiltrableDataSource(dataSource).then((dataSource) => {
```
adube: Why was this changed?  It was easier to read before.

contribs/gmf/src/filters/filterselectorComponent.js
sbrunner: Right, fixed be removing one @extent :-), Thanks.

contribs/gmf/src/datasource/OGC.js
```diff
@@ -4,7 +4,6 @@ import ngeoDatasourceOGC from 'ngeo/datasource/OGC.js';
 /**
  * The options required to create a `gmf.datasource.OGC`.
  * @typedef {Object} OGCOptions
- * @extends import('ngeo/datasource/OGC.js').OGCOptions
```
adube: Hum...  I don't understand how this is supposed to affect the type checking yet, but I won't prevent you from merge just because of this.  I'll figure it out later.

adube:
Thanks for the fix.  Check what's wrong with Travis then merge.

fredj:
Build restarted

contribs/gmf/src/datasource/OGC.js
sbrunner: One good example of using extends is the following:
```
/**
 * @extends A<number>
 */
class B extends A {...}
```
To the value of the template.
In this case it's incoherent, then I don't know what it can mean...

PR #4606, opened 2019-02-06
Fix remaining typechecks in gmf

adube:
This patch fixes the remaining type issues in GMF.

Following @fredj recommendation, I copied the properties from ("ngeo/datasource/OGC.js").OGCOptions to ("gmf/datasource/OGC.js").OGCOptions, because typescript doesn't currently support inheritence of objects for typing purpose.

There are also 2 remaining errors that I don't know what to do about.  They are about the use of DocumentTouch property, which seems to be deprecated.  I suspect this is something that needs to be addressed (and not just ignored using ts-ignore). Here they are:

```
contribs/gmf/src/controllers/bootstrap.js:24:65 - error TS2339: Property 'DocumentTouch' does not exist on type 'never'.

24       if (small_screen && (('ontouchstart' in window) || window.DocumentTouch && document instanceof DocumentTouch)) {
                                                                   ~~~~~~~~~~~~~

contribs/gmf/src/controllers/bootstrap.js:24:102 - error TS2304: Cannot find name 'DocumentTouch'.

24       if (small_screen && (('ontouchstart' in window) || window.DocumentTouch && document instanceof DocumentTouch)) {
         
```

### DIFF
--- a/contribs/gmf/src/controllers/AbstractAppController.js
+++ b/contribs/gmf/src/controllers/AbstractAppController.js
@@ -561,7 +561,7 @@ export function AbstractAppController(config, map, $scope, $injector) {
   /**
    * @export
    */
-  // @ts-ignore: todo
+  // @ts-ignore: We do want to define a new property on `window`.
   window.gmfx = gmfx;
 
   /**
@@ -614,7 +614,7 @@ export function AbstractAppController(config, map, $scope, $injector) {
   /**
    * @export
    */
-  // @ts-ignore: todo
+  // @ts-ignore: We do want to define a new property on `window`.
   window.cgxp = cgxp;
   /**
    * @export
--- a/contribs/gmf/src/controllers/AbstractAppController.js
+++ b/contribs/gmf/src/controllers/AbstractAppController.js
@@ -40,7 +40,7 @@ import {findThemeByName} from 'gmf/theme/Themes.js';
  * @property {import("ol/style/Style.js").default} [positionFeatureStyle]
  * @property {import("ol/style/Style.js").default} [accuracyFeatureStyle]
  * @property {number} [geolocationZoom]
- * @property {boolean|undefined} [autorotate]
+ * @property {boolean} [autorotate]
  * @property {olx.ViewOptions} [mapViewConfig]
  * @property {import("ol/Collection.js").default.<import('ol/control/Control.js').default>|Array.<import('ol/control/Control.js').default>} [mapControls]
  * @property {import("ol/Collection.js").default.<import('"ol/interaction/Interaction.js').default>|Array.<import('ol/interaction/Interaction.js').default>} [mapInteractions]
--- a/contribs/gmf/src/controllers/AbstractAppController.js
+++ b/contribs/gmf/src/controllers/AbstractAppController.js
@@ -30,7 +30,7 @@ import olStyleFill from 'ol/style/Fill.js';
 import olStyleStroke from 'ol/style/Stroke.js';
 import olStyleStyle from 'ol/style/Style.js';
 import {ThemeEventType} from 'gmf/theme/Manager.js';
-import gmfThemeThemes from 'gmf/theme/Themes.js';
+import {findThemeByName} from 'gmf/theme/Themes.js';
 
 
 /**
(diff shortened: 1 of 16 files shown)


contribs/gmf/src/controllers/AbstractAppController.js
```diff
@@ -40,7 +40,7 @@ import gmfThemeThemes from 'gmf/theme/Themes.js';
  * @property {import("ol/style/Style.js").default} [positionFeatureStyle]
  * @property {import("ol/style/Style.js").default} [accuracyFeatureStyle]
  * @property {number} [geolocationZoom]
- * @property {string} [autorotate]
+ * @property {boolean|undefined} [autorotate]
```
fredj: you normally don't need to add `undefined`: if the property name is optional (marked by using `[]`) typescript should accept `undefined`

contribs/gmf/src/controllers/bootstrap.js
```diff
@@ -1,12 +1,13 @@
...
 /* global DocumentTouch */
 
 import $ from 'jquery';
 import angular from 'angular';
 
-
 function bootstrap(module) {
-  const interface_ = $('meta[name=interface]')[0].attributes.content.value;
-  const dynamicUrl_ = $('meta[name=dynamicUrl]')[0].attributes.content.value;
+  const interface_ = $('meta[name=interface]')[0].getAttribute('content');
```
fredj: Not sure it works: if I run: `$('meta[name=interface]')[0].getAttribute('content')` on https://camptocamp.github.io/ngeo/master/examples/contribs/gmf/apps/desktop.html/

I get the following error:
```
Uncaught TypeError: Cannot read property 'getAttribute' of undefined
    at <anonymous>:1:29
```

fredj: Should be:
```suggestion
  const interface_ = $('meta[name=interface]').getAttribute('content');
```

fredj:
See #4609 for the DocumentTouch warning

contribs/gmf/src/controllers/bootstrap.js
adube: I noticed that too.  From what I understand, if you execute the code in a console then it doesn't work.  Try putting a breakpoint after and it works.

I suspect that the scope of `$` is not the same while we execute, because the code does with `[0]` and it was like this before and it worked.  I'm not sure if we should remove it (I'll try).

contribs/gmf/src/controllers/AbstractAppController.js
adube: Oh, I didn't know that.  OK, I'll fix this.

contribs/gmf/src/controllers/bootstrap.js
adube: As expected, if I remove the `[0]`I get a thrown JavaScript error.  I won't touch this, because it works already.

adube:
All done.  Ready merge?

contribs/gmf/src/controllers/AbstractAppController.js
```diff
@@ -591,6 +561,7 @@ export function AbstractAppController(config, map, $scope, $injector) {
   /**
    * @export
    */
+  // @ts-ignore: todo
```
sbrunner: todo?

contribs/gmf/src/controllers/AbstractAppController.js
```diff
@@ -643,6 +614,7 @@ export function AbstractAppController(config, map, $scope, $injector) {
...
```
sbrunner: same

contribs/gmf/src/controllers/AbstractAppController.js
adube: Yes, todo.  I suspect we don't want to leave this like this forever.  If you have a better idea for a comment, please let met know what to write.

contribs/gmf/src/controllers/AbstractAppController.js
adube: same

contribs/gmf/src/controllers/AbstractAppController.js
sbrunner:
```suggestion
  // @ts-ignore: wanted to define new property on window
```
:-)

adube: Thanks for the suggestion.  I'll apply them, hold on...

PR #4623, opened 2019-02-11
Don't use angular.element for notify target element

fredj:
`target` can be a string


fredj:
examples only, merging

PR #4626, opened 2019-02-11
Add new ngeo-draggable-handle option to modalComponent

fredj:
ping @gberaudo @llienher

### DIFF
--- a/src/message/modalComponent.js
+++ b/src/message/modalComponent.js
@@ -36,6 +36,8 @@ const module = angular.module('ngeoModal', []);
  *
  * See our live example: [../examples/modal.html](../examples/modal.html)
  *
+ * @htmlAttribute {string} ngeo-draggable-handle The jquery selector that define the element
+ *     that can starts the dragging sequence. Defaults to `.modal-header`.
  * @htmlAttribute {boolean} ngeo-modal-resizable Whether the modal can be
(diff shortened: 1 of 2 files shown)


gberaudo:
@sbrunner, @ybolognini, we found out the issue with have with the Luxembourg app is due to a bug in ngeo. The chrome browser in Android does not support our clicks inside draggable boxes.

Currently modals are draggable: the user can drag any part of the modal and move it around.

A reasonable workaround would be to:
- restrict drag to the modal headers (the text part only).

Bonus questions:
- does it make sense for modals to be draggable?
- are there other parts of ngeo/gmf that are impacted by this issue?
- can this be fixed for 2.4?

src/message/modalComponent.js
```diff
@@ -123,7 +131,9 @@ class Controller {
     this.resizable = !!this.resizable;
 
     const dialog = this.modal_.find('.modal-dialog');
-    dialog.draggable();
+    dialog.draggable({
+      handle: this.draggableHandle || '.modal-header'
```
gberaudo: To allow clicking on the close buttons, shouldn't the default be `.modal-title`?

gberaudo: The `.modal-title` is not always present, so what you implemented is better.

PR #4632, opened 2019-02-12
 Only activate swipe up/down on the collapse button

fredj:
demo: https://camptocamp.github.io/ngeo/GSGMF-758/examples/contribs/gmf/apps/mobile.html


fredj:
validated by @kalbermattenm, merging

PR #4633, opened 2019-02-12
GMF Print - support sending printed file to email

adube:
This patch introduces sending a printed file using the GMF Print tool to an email.  The specs are described in this task: https://jira.camptocamp.com/browse/GSGMF-738

It is not yet completed, as there are things remaining to do:

 * [x] show a message about the file that is going to be sent by email
 * [ ] the request, when using the email, currently doesn't work. More details to come in a comment.

### DIFF
--- a/contribs/gmf/src/authentication/Service.js
+++ b/contribs/gmf/src/authentication/Service.js
@@ -16,6 +16,7 @@ import olEventsEventTarget from 'ol/events/Target.js';
 
 /**
  * @typedef {Object} User
+ * @property {string|null} email User's email address
  * @property {AuthenticationFunctionalities|null} functionalities Configured functionalities of the user
  * @property {boolean|null} is_password_changed True if the password of the user has been changed. False otherwise.
  * @property {number|null} role_id the role id of the user.
@@ -207,6 +208,13 @@ export class AuthenticationService extends olEventsEventTarget {
     return this.$http_.post(url, $.param({'login': login}), {
       headers: {'Content-Type': 'application/x-www-form-urlencoded'}
     }).then(successFn);
+  }
+
+  /**
+   * @return {string|null} User's email
+   */
+  getEmail() {
+    return this.user_.email || null;
   }
 
   /**
--- a/contribs/gmf/src/controllers/AbstractDesktopController.js
+++ b/contribs/gmf/src/controllers/AbstractDesktopController.js
@@ -3,7 +3,7 @@ import gmfControllersAbstractAPIController, {AbstractAPIController} from 'gmf/co
 import gmfContextualdataModule from 'gmf/contextualdata/module.js';
 import gmfEditingModule from 'gmf/editing/module.js';
 import gmfPermalinkShareComponent from 'gmf/permalink/shareComponent.js';
-import gmfPrintComponent from 'gmf/print/component.js';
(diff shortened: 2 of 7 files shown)


adube:
@sbrunner When trying to print using the parameters, it didn't work.  I got this error response from the server.

```
javax.net.ssl.SSLException: Unrecognized SSL message, plaintext connection?
```

Would you please try the patch and see if you get the same error?

contribs/gmf/src/authentication/Service.js
```diff
@@ -16,6 +16,7 @@ import olEventsEventTarget from 'ol/events/Target.js';
 
 /**
  * @typedef {Object} User
+ * @property {string|null} email User's email address
```
sbrunner:
```suggestion
 * @property {string} [email] User's email address
```
same for other properties

sbrunner:
```suggestion
 * @property {string} [email] User's email address
```
same for other properties...
Not completely sure but looks better :-)

adube: That's not entirely true.  If you look at the code, you'll notice that it gets set to `null`.  `[email]` means `string|undefined`, not `string|null`.  It has not effect if one use a thruthy comparision, like `if (user.email)`, but would not work if `if (user.email === null)`.

For the sake of not breaking anything I'd leave the others as-is and this one like this too.

adube: Same comment.

sbrunner: OK, I will have a look :-)

sbrunner:
I fix the error on sending email, but the client shouldn't pool the print to wait on the job...

adube:
> I fix the error on sending email, but the client shouldn't pool the print to wait on the job...

How do you suggest we do that?

adube:
I just tried it and it didn't work.

I got an error:

`console.js:35 org.locationtech.jts.util.AssertionFailedException: Expected 200 OK but encountered 400 BAD_REQUEST: Http status code for https://geomapfish-demo-dc.camptocamp.com/2.4/mapserv_proxy?MAP_RESOLUTION=…34%2C179503.875%2C637358.7916666667%2C191410.125&VERSION=1.1.1&SERVICE=WMS was not OK.  It was: 400 BAD_REQUEST. The response message was: 'Bad Request'`

I did not get an e-mail.﻿
￼
@sbrunner Any idea?

sbrunner:
When we click on `image` or `PDF` and the `send email` is checker we just send the request on `report.pdf`, and display the message :-)

adube:
> When we click on `image` or `PDF` and the `send email` is checker we just send the request on `report.pdf`, and display the message :-)

OK, I think I understand what you mean.  However, I'm not that familiar with the code so I don't know what that implies yet.  I also suspect that regardless if I make those changes or not, the request asking for a file being sent should have worked and it didn't.

I suggest we make sure that the current version works with the file, then we make your suggested changes.

sbrunner:
OK, looks good to me :-)

adube:
> OK, looks good to me :-)

Ok thanks.  Do you want to take a look at the error before we merge this?  I would like to see this feature working before merging it.

sbrunner:
Now the error, is as I see, completely fixed :-)

PR #4649, opened 2019-02-18
Don't send datasource when it's not available

sbrunner:
Fix: https://jira.camptocamp.com/browse/GSGMF-778


rbovard:
Thanks @sbrunner, it's working.

PR #4685, opened 2019-02-26
[2.2] Update current theme after loadThemes call is done

pfirpfel:
Currently, the updateCurrentTheme_ call happens before the  loadThemes call. This is problematic if you log in as an user who has a default theme that is non-public. At the time updateCurrentTheme_ gets called, this non-public theme is not available. It will be, after the loadThemes call has finished. That's why updateCurrentTheme_ should happen after loadThemes.

### DIFF
--- a/contribs/gmf/src/controllers/abstract.js
+++ b/contribs/gmf/src/controllers/abstract.js
@@ -152,15 +152,10 @@ gmf.AbstractController = function(config, $scope, $injector) {
     this.gmfThemes_.loadThemes(roleId);
 
     if (evt.type !== gmf.AuthenticationEventType.READY) {
-      let listenerKey;
-      const themesLoaded = () => {
-        ol.events.unlistenByKey(listenerKey);
+      this.gmfThemes_.getThemesObject().then(() => {
         this.updateCurrentTheme_(previousThemeName);
-      };
-      listenerKey = ol.events.listen(this.gmfThemes_,
-        gmf.ThemesEventType.CHANGE, themesLoaded, this);
+      });
     }
-
     this.setDefaultBackground_(null);
     this.updateHasEditableLayers_();
   };
--- a/contribs/gmf/src/controllers/abstract.js
+++ b/contribs/gmf/src/controllers/abstract.js
@@ -152,9 +152,7 @@ gmf.AbstractController = function(config, $scope, $injector) {
     this.gmfThemes_.loadThemes(roleId);
 
     if (evt.type !== gmf.AuthenticationEventType.READY) {
-      this.gmfThemes_.getThemesObject().then(() => {
-        this.updateCurrentTheme_(previousThemeName);
-      });
+      this.updateCurrentTheme_(previousThemeName);
     }
     this.setDefaultBackground_(null);
     this.updateHasEditableLayers_();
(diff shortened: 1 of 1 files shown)


contribs/gmf/src/controllers/abstract.js
```diff
@@ -147,11 +147,20 @@ gmf.AbstractController = function(config, $scope, $injector) {
...
     // Reload themes and background layer when login status changes.
     this.gmfThemes_.loadThemes(roleId);
+
+    if (evt.type !== gmf.AuthenticationEventType.READY) {
+      let listenerKey;
+      const themesLoaded = () => {
+        ol.events.unlistenByKey(listenerKey);
+        this.updateCurrentTheme_(previousThemeName);
+      };
+      listenerKey = ol.events.listen(this.gmfThemes_,
```
fredj: use `listenOnce` instead ?

contribs/gmf/src/controllers/abstract.js
```diff
@@ -147,11 +147,20 @@ gmf.AbstractController = function(config, $scope, $injector) {
...
+        gmf.ThemesEventType.CHANGE, themesLoaded, this);
+    }
+
```
sbrunner: Shouldn't we use:
```suggestion
if (evt.type !== gmf.AuthenticationEventType.READY) {
  this.gmfThemes_.getThemesObject().then(() => {
    this.updateCurrentTheme_(previousThemeName);
  })
}
```

contribs/gmf/src/controllers/abstract.js
```diff
@@ -147,11 +147,15 @@ gmf.AbstractController = function(config, $scope, $injector) {
...
-    }
+
     // Reload themes and background layer when login status changes.
     this.gmfThemes_.loadThemes(roleId);
+
+    if (evt.type !== gmf.AuthenticationEventType.READY) {
+      this.gmfThemes_.getThemesObject().then(() => {
+        this.updateCurrentTheme_(previousThemeName);
+      });
+    }
```
sbrunner: After a secound alalysis in can just be:
```suggestion
    if (evt.type !== gmf.AuthenticationEventType.READY) {
      this.updateCurrentTheme_(previousThemeName);
    }
```
Just change the order from the initial version ...

sbrunner:
@fredj Do you know why we have this 
```if (evt.type !== gmf.AuthenticationEventType.READY) {```

File: contribs/gmf/src/controllers/abstract.js (hunk 3 above)
Comment (pfirpfel) [reply contribs/gmf/src/controllers/abstract.js]: According to my debugging this won't work. updateCurrentTheme_ happens still before the daa gets updated in the loadThemes promise.

Comment (sbrunner) [reply contribs/gmf/src/controllers/abstract.js]: `updateCurrentTheme_` happens before the data gets of the theme, but in `updateCurrentTheme_` we have:
```js
this.gmfThemes_.getThemesObject().then((themes) => {
   // this should happens after we get the theme
```
And if this not working we should know why...

pfirpfel: I set my stops at a wrong point then. I deployed your simplified version on my problematic client's dev server and it works.

pfirpfel:
> @fredj Do you know why we have this
> `if (evt.type !== gmf.AuthenticationEventType.READY) {`

One reason I see, is that every login and logout produces a second call with gmf.AuthenticationEventType.READY.

May be we can remove this if clause and change the  listener to listenOnce below:

`ol.events.listenOnce(gmfAuthentication, gmf.AuthenticationEventType.READY, userChange);`

contribs/gmf/src/controllers/abstract.js
pfirpfel: Found a simpler solution

PR #4690, opened 2019-02-26
Force theme reload

adube:
This hack forces theme to be obtained upon reloading after a login/logout.

The solution was taken from Martin's answer in this question on Stack Overflow: https://stackoverflow.com/questions/16098430/angular-ie-caching-issue-for-http

I tried the first solution and it didn't fix the issue. Trying the second one did.

I don't understand what suddenly made the previous code stop working...


fredj:
we found the cause of this issue with @sbrunner : the problem is that `role_id` is never defined because of the multi-role system. New PR to come.

sbrunner:
Thanks @adube for the initial investigations :-)

PR #4698, opened 2019-02-27
793 - Fix GMF edit feature stop editing while dirty

adube:
In the GMF feature editing tool, while trying to stop editing when there a feature is selected and dirty, a modal opens asking for confirmation.  In the process of applying the cancel, the modal gets hidden with an animation for a short period of time.  If we cancel right away, the modal gets destroyed while animating and the result is we have the modal backdrop still there.

This patch lets the modal do its animation then apply the cancel.


adube:
@sbrunner or @fredj - ready for review.

PR #4704, opened 2019-03-01
Fix font name used in OpenLayers text style

fredj:
Use `Font Awesome 5 Free` instead of `FontAwesome` and weight `900` instead of `normal` (alias to `400`)


arnaud-morvan:
@fredj: For curiosity, it works with '900 30px "Font Awesome 5 Free"' but not with 'normal 30px "Font Awesome 5 Free"' ?

fredj:
@arnaud-morvan yes, because `normal` s equal to `400`and we do not have this weight

PR #4705, opened 2019-03-01
Workaround for error in tooltip config type checking with IE11 

arnaud-morvan:


arnaud-morvan:
@fredj, @sbrunner : I did not manage to please typecheck (npm run typecheck).

arnaud-morvan:
Seems I got it with @ts-ignore.
Note that typecheck compare bs.tooltip config to jqueryui.TooltipOption, which is completely different object.

fredj:
try to force the type with:
```js
 body.tooltip(/** @type {Bootstrap.TooltipOption} */ ({
      container: 'body',
      trigger: 'hover',
      selector: '[data-toggle="tooltip"]',

      // Avoid error in config type checking with IE11
      sanitizeFn: $injector.get('$sanitize')
    }));
```

arnaud-morvan:
@fredj: perfect, thanks.

PR #4764, opened 2019-03-21
Minor code cleanup

fredj:


sbrunner:
This shouldn't be on 2.3?

fredj:
> This shouldn't be on 2.3?

not needed because `contribs/gmf/apps/iframe_api/index.html.ejs` in not in 2.3

PR #4792, opened 2019-04-03
Code cleanup

fredj:


gberaudo:
Looks like we have 123 import * in the code.
Some day these should be cleaned up to be made explicit and improve treeshaking:

```
gberaudo@wrk29:~/dev/ngeo (master $%)$ git grep 'import \* as' | wc -l
123
```

sbrunner:
Thanks :-)

PR #4817, opened 2019-04-11
In flush mode load current theme private groups on login

arnaud-morvan:
To keep things simple, I simply reload current theme when in flush mode and no default_theme for the user.
This have no impact on "not flush" mode.


sbrunner:
It doesn't work in this case:
- Load theme Mixt
- Search for filtres
- Add data filtres

=> Groups: Filtres, Function OSM

- Login with demo

=> Groups: Private, Function OSM
Expected groups (Order?): Filtres, Function OSM, Private

arnaud-morvan:
Note that this also occurs when there is a default_theme configured for the user.

sbrunner:
Yes but it's wanted in the case of default_theme ...

arnaud-morvan:
Should be better now, only add private first level group at end of layer tree. Finally not so complex.

sbrunner:
For me it's OK, I let @fredj  to have a third look :-)

PR #4852, opened 2019-04-26
Waiting cursor for map query [2.3]

pfirpfel:
I wanted to show a different cursor if there is a pending map query. But query/windowComponent (formerly displayquerywindow) keeps clearing ngeoQueryResult right after MapQuerent.issue() gets called. This clearing is not distinguishable from a unsuccessful query.

For my feature, I need the first commit in ngeo (the one that removes the clear). GMF seems to work fine without it.

The second commit is the cursor feature, which does not need to be part of gmf.

### DIFF
--- a/contribs/gmf/apps/desktop/Controller.js
+++ b/contribs/gmf/apps/desktop/Controller.js
@@ -27,12 +27,13 @@ if (!window.requestAnimationFrame) {
 /**
  * @param {angular.Scope} $scope Scope.
  * @param {angular.$injector} $injector Main injector.
+ * @param {!ngeox.QueryResult} ngeoQueryResult ngeo query result.
  * @constructor
  * @extends {gmf.controllers.AbstractDesktopController}
  * @ngInject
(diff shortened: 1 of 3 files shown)


pfirpfel:
@fredj
Since you did not question the feature itself, I have to add more thoughts. I'm not sure, if the code is at the right place in the controller of the demo desktop app.
I tried putting it in the abstract desktop controller (makes no sense on mobile?), but that abstract controller was loaded before ngeoQueryResult was initialized. That's why it is there right now in the demo app...

Would it be possible to do this in the abstract desktop controller?

Moving it there would make it a default feature, is this even a desired outcome?

Alternatively, I could move id to the desktop_alt demo app. This is more the home of those weird non-default features. Having it in the regular desktop app could create wrong expectations from clients, since it would not available by default.

pfirpfel:
Moved the bugfix to new PR #4887 

Will implement the waiting cursor as a custom feature.

PR #4868, opened 2019-05-02
 Fix the JSTS OpenLayers link

sbrunner:


sbrunner:
@eleu How to use the oeview, do you have a valid URL?

eleu:
oeview is not used at all AFAIK

sbrunner:
OK :-/

sbrunner:
Should we remove the example application?

PR #4888, opened 2019-05-10
Have less version upgrade

sbrunner:


gberaudo:
contribs/gmf/examples/search.js:120:13 - error TS7014: Function type, which lacks return-type annotation, implicitly has an 'any' return type.

PR #4904, opened 2019-05-22
[WIP] merge 2.3 into 2.4

llienher:


.eslintrc.yaml
```diff
@@ -18,3 +18,10 @@ rules:
   max-len:
     - error
     - code: 110
+globals:
+    angular: false
+    Cesium: false
+    google: false
+    Bloodhound: false
+    DateFormatter: false
+    localforage: false
```
sbrunner: Shouldn't be removed?

src/message/modalComponent.js
```diff
@@ -1,6 +1,9 @@
-import angular from 'angular';
+/**
+ * @module ngeo.message.modalComponent
+ */
```
sbrunner: It should be removed

gberaudo:
@sbrunner, see #4913.

PR #4907, opened 2019-05-23
Add IE11 workaround for click on the theme selector scrollbar

fredj:


fredj:
doesn't work, closing

PR #4954, opened 2019-06-17
Add back backgroundLayers options in the API

fredj:


api/src/Map.js
```diff
@@ -116,8 +120,10 @@ class Map {
 
     // Get background layer first...
     themes.getBackgroundLayers().then((layers) => {
+      // The options is an array for backward compatibility reason.
+      const backgroundLayer = options.backgroundLayers || [constants.backgroundLayer];
```
arnaud-morvan: backgroundLayer => backgroundLayer**s**

fredj: The option should be without the s but we need to keep it for  backward compatibility.
The map can have only one background layer

PR #4970, opened 2019-06-25
 Add ESLint Plugin TypeScript 

sbrunner:


contribs/gmf/src/raster/component.js
```diff
@@ -109,6 +109,7 @@ function rasterComponent() {
       'layer': '<gmfElevationLayer',
       'map': '=gmfElevationMap'
     },
+    // @ts-ignore
```
gberaudo: Why these ts-ignore?

PR #4981, opened 2019-06-28
Some fix for the GMF applications

sbrunner:


src/geolocation/mobile.js
```diff
@@ -335,10 +335,9 @@ Controller.prototype.autorotateListener = function() {
       if (!(event instanceof DeviceOrientationEvent)) {
         throw new Error('Wrong event type');
       }
-      if (!event.alpha) {
-        throw new Error('Missing event.alpha');
+      if (event.alpha) {
```
fredj: cmp with undefined

PR #5021, opened 2019-07-16
Simplify file input in import datasource component

fredj:
By using the custom-file-input from Bootstrap:
https://getbootstrap.com/docs/4.3/components/input-group/#custom-file-input


fredj:
I've checked and the title is automatically displayed

PR #5059, opened 2019-08-13
1029(2) Map swipe example done

RBcote:


examples/mapswipe.js
```diff
@@ -0,0 +1,60 @@
...
+  ngeoMapswipeModule.name
+]);
+
+/**
+ * @constructor
+ * @ngInject
+ */
+function MainController() {
+
+  const openStreatMap = new olLayerTile({
```
adube: openStreetMapLayer

examples/mapswipe.js
```diff
@@ -0,0 +1,60 @@
...
+    source: new olSourceOSM()
+  });
+
+  /**
+   * @type {import('ol/layer/Tile.js').default}
+   */
+  this.openSeaMap = new olLayerTile({
```
adube: openSeaMapLayer

examples/mapswipe.html
```diff
@@ -0,0 +1,16 @@
...
+  <head>
+    <title>Map swipe example</title>
+    <meta charset="utf-8">
+    <meta name="viewport"
+          content="initial-scale=1.0, user-scalable=no, width=device-width">
+    <meta name="mobile-web-app-capable" content="yes">
+  </head>
+  <body ng-controller="MainController as ctrl">
+   <div id="map" ngeo-map="::ctrl.map"></div>
+   <ngeo-mapswipe map="::ctrl.map" layer="::ctrl.openSeaMap"></ngeo-mapswipe>
```
adube: openSeaMapLayer

src/map/swipe.html
```diff
@@ -0,0 +1,4 @@
+<div>
+    <label>Swipe</label>
+  <input id="swipe" type="range" class="swipe-input" style="width: 100%;">
```
adube: style should not be added in the element directly, but in the CSS of the example instead

src/map/swipe.js
```diff
@@ -0,0 +1,147 @@
...
+    /**
+     * @type {JQuery}
+     */
+    this.swipeInput_ = $element.find('.swipe-input');
+
+    /**
+     * @type {?import("ol/events.js").EventsKey}
+     * @private
+     */
+    this.layerKey_ = null;
```
adube: This needs to be an array of listener keys

src/map/swipe.js
```diff
@@ -0,0 +1,147 @@
...
+  }
+
+  /**
+   * Init the controller
+   */
+  $onInit() {
+    this.layerKey_ = this.layer.on('prerender', this.handleLayerPrerender_.bind(this));
```
adube: push the result of this in the listener key array

instead of .bind, add `this` as 3rd argument

src/map/swipe.js
```diff
@@ -0,0 +1,147 @@
...
+    this.swipeInput_.on('input change', event => {
+      this.swipeValue = Number($(event.target).val());
+      this.map.render();
+    });
+    this.layer.on('postrender', this.handleLayerPostrender_.bind(this));
```
adube: push the result in the listener key

remove bind

add `this` as 3rd argument

src/map/swipe.js
```diff
@@ -0,0 +1,147 @@
...
+     */
+    this.layerKey_ = null;
+  }
+
+  /**
+   * Init the controller
+   */
+  $onInit() {
+    this.layerKey_ = this.layer.on('prerender', this.handleLayerPrerender_.bind(this));
+    this.swipeInput_.on('input change', event => {
```
adube: this event is not unlistened

src/map/swipe.js
```diff
@@ -0,0 +1,147 @@
...
+   * @private
+   */
+  handleLayerPostrender_(evt) {
+    const ctx = evt.context;
+    ctx.restore();
+  }
+
+  $onDestroy() {
+    if (this.layerKey_ !== null) {
+       olEvents.unlistenByKey(this.layerKey_);
```
adube: unlistenByKey should also work when given an array

after that, you should empty the array

no need for the `if` there

finally, you need to unlisten the jquery event on the input as well using `off` https://api.jquery.com/off/ . Doing that will remove all events, which is okay in our case.

src/map/swipe.js
```diff
@@ -0,0 +1,145 @@
...
+     */
+    this.scope_ = $scope;
+
+    /**
+     * @type {number}
+     */
+    this.swipeValue = 50;
+
+    /**
+     * @type {JQuery}
```
adube: Add `@private`

src/map/swipe.js
```diff
@@ -0,0 +1,145 @@
...
+     */
+    this.swipeInput_ = $element.find('.swipe-input');
+
+    /** @type {(olEvents.EventsKey | olEvents.EventsKey[])[] | null} */
```
adube: Add `@private`

The correct type should be:

```
olEvents.EventsKey[]
```

src/map/swipe.js
```diff
@@ -0,0 +1,145 @@
...
+   */
+  $onInit() {
+    this.layerKeys_.push(olEvents.listen(this.layer, 'prerender', this.handleLayerPrerender_, this));
+
+    this.swipeInput_.on('input change', event => {
+      this.swipeValue = Number($(event.target).val());
+      this.map.render();
+    });
+
+    this.layerKeys_.push(olEvents.listen(this.layer, 'postrender', this.handleLayerPostrender_, this));
```
adube: move under the other one that looks like it

src/map/swipe.js
```diff
@@ -0,0 +1,145 @@
...
+   * @private
+   */
+  handleLayerPostrender_(evt) {
+    const ctx = evt.context;
+    ctx.restore();
+  }
+
+  $onDestroy() {
+    olEvents.unlistenByKey(this.layerKeys_);
+    this.swipeInput_.off();
```
adube: `off` requires an argument, the event (which is `''input change''`)

src/map/swipe.js
```diff
@@ -0,0 +1,149 @@
+import angular from 'angular';
+import * as olEvents from 'ol/events.js';
+import '../../examples/mapswipe.css';
```
adube: I hadn't noticed that one.

Remove this.  This should be imported in the example itself, not in the component.

adube:
@fredj Would you please take a look at this patch and see if we can merge it?

src/map/swipe.js
```diff
@@ -0,0 +1,156 @@
...
+    if (evt instanceof RenderEvent) {
+      const ctx = evt.context;
+    if (!ctx) {
+      return;
+    }
+    const width = ctx.canvas.width * (this.swipeValue / 100);
+    ctx.save();
+    ctx.beginPath();
+    ctx.rect(0, 0, width, ctx.canvas.height);
+    ctx.clip();
```
sbrunner: Wrong indentation

examples/mapswipe.css
```diff
@@ -0,0 +1,8 @@
+#map {
+    width: 100%;
+    height: 400px;
+  }
+
+.swipe-input {
+  width: 100%;
+}
```
sbrunner: Missing endline :-)

adube:
@fredj Build failed, but I'm not sure it is relevant to the code here...  Would you please check?

src/map/swipe.js
```diff
@@ -0,0 +1,157 @@
...
+  /**
+   * @param {?Event|import("ol/events/Event.js").default} evt OpenLayers object event.
+   * @private
+   */
+  handleLayerPrerender_(evt) {
+    if (!(evt instanceof RenderEvent)) {
+      return;
+    }
+    const ctx = evt.context;
+      if (!ctx) {
```
sbrunner: Wrong indentation here ...

PR #5067, opened 2019-08-16
2FA auth support

fredj:


sbrunner:
- link lost password not visible anymore ...
- after the first password change the theme should be reloaded
- after a reload we are not logged anymore (maybe a server issue)
- after a login succeed the page is reloaded ...

PR #5069, opened 2019-08-19
WIP Introduce swipe component CSS to style slider in example

RBcote:
This patch introduce CSS to the swipe component, this one look like a vertical slider.


adube:
This is the UI it should look like:

Map_swiper

This is how it looks in the ngeo example as part of this patch:

Screenshot_2019-08-19_13-38-14

 1. the "zone" it hides does not match the slider (it's too much on the right), and you can't go all the way to the left
 2. the bar doesn't look like the one above
 3. we're missing a section there
 4. we're missing the middle section (I suspect this is going to be hard to do, don't put too much efforts on this yet)

Also, I think the positionning should be made to the ngeo-mapswipe component, not on the input itself.  See if you can fix this.

examples/mapswipe.css
```diff
@@ -3,6 +3,30 @@
...
 
-.swipe-input {
+ngeo-mapswipe .swipe-input {
+  background: none;
+  height: 0;
+  opacity: 0.7;
+  outline: none;
+  pointer-events: none;
+  position: absolute;
+  top: 20%;
```
adube: Let's not use positions with percentage.

examples/mapswipe.css
```diff
@@ -3,6 +3,30 @@
...
+
+ngeo-mapswipe .swipe-input:hover {
+  opacity: 1;
+}
+
+ngeo-mapswipe .swipe-input::-webkit-slider-thumb {
+  appearance: none;
+  background: #206020;
+  cursor: pointer;
+  height: 415px;
```
adube: height should be 100% (not tested, but normally it should follow the container).

PR #5070, opened 2019-08-19
Set higher z-index for map messages

rbovard:
In some case, the close button for disclaimer is hidden behind button.

Before
image

After
image


sbrunner:
What's the status of this pull request?

rbovard:
See new PR #5914

PR #5071, opened 2019-08-20
GSGMF-1085: api query fix

llienher:
Fix https://jira.camptocamp.com/browse/GSGMF-1085

### DIFF
--- a/api/src/Map.js
+++ b/api/src/Map.js
@@ -40,7 +40,7 @@ import * as themes from './Themes.js';
 /**
  * @type {Array<string>}
  */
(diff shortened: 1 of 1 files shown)


api/src/Map.js
```diff
@@ -40,7 +40,7 @@ import * as themes from './Themes.js';
 /**
  * @type {Array<string>}
  */
-const EXCLUDE_PROPERTIES = ['geom', 'geometry', 'boundedBy'];
+const EXCLUDE_PROPERTIES = ['geom', 'geometry', 'boundedBy', 'the_geom'];
```
sbrunner: We should ignore geometry object not all possible geometry names ...

PR #5085, opened 2019-08-23
Increase query tolerance on 'touch' devices

fredj:


contribs/gmf/src/controllers/AbstractDesktopController.js
```diff
@@ -279,5 +280,8 @@ module.controller('AbstractDesktopController', AbstractDesktopController);
 module.value('ngeoMeasurePrecision', 3);
 module.value('ngeoMeasureDecimals', 0);
 
+module.value('ngeoQueryOptions', {
+  'tolerance': hasCoarsePointingDevice() ? 10 : 3
```
sbrunner: To be configurable id should be:
```suggestion
  'tolerance': {'click': 3, 'touch': 10}
```
and the hasCoarsePointingDevice function should be used in the component :-)

PR #5094, opened 2019-08-27
Use a custom layer for the print mask

fredj:


contribs/gmf/src/print/component.js
```diff
@@ -628,8 +628,9 @@ export class PrintController {
       getRotationFn = () => this.rotation;
     }
 
-    this.postcomposeListener_ = this.ngeoPrintUtils_.createPrintMaskPostcompose(
-      getSizeFn, this.getScaleFn.bind(this), getRotationFn);
```
sbrunner: Shouldn't we remove createPrintMaskPostcompose?

fredj: it's still used in one example, I will change it

fredj: done

sbrunner: ok, thanks :-)

PR #5098, opened 2019-08-28
Fix exlusion condition in api query

llienher:
Fix: GSGMF-1085


llienher:
@sbrunner I did the change but now the CI don't pass... I don't know why (it is not errors related to my work).

sbrunner:
Just restarted ...

PR #5100, opened 2019-08-29
Add data-sentry string to the breadcrumb message

fredj:


contribs/gmf/src/controllers/AbstractAppController.js
```diff
@@ -820,6 +823,23 @@ export function getLocationIcon() {
...
+ * @param {Sentry.BreadcrumbHint} hint
+ */
+function augmentBreadcrumb(breadcrumb, hint) {
+  if (breadcrumb.category === 'ui.click') {
+    const target = hint.event.target;
+    if (target) {
+      const message = target.dataset.sentry;
+      if (message) {
+        breadcrumb.message = `${breadcrumb.message} | ${message}`;
+      }
```
sbrunner: Shouldn't we have more information with something like:?
```js
const tags = []

function browse(element) {
    if (element) {
        const message = target.dataset.sentry;
        if (message) {
            tags.append(`${element.tagName}.${message}`);
        } else {
            tags.append(element.tagName);
        }
        browse(element.parentElement);
    }
}

browse(hint.event.target);
if (tags) {
  breadcrumb.message = `${breadcrumb.message} | ${tags.join('-')}`;
}

```

PR #5104, opened 2019-09-04
Translate layers names in API layer switcher

arnaud-morvan:
GSGMF-1103

Server part: https://github.com/camptocamp/c2cgeoportal/pull/5262


arnaud-morvan:
@sbrunner : Would you like to have a look on this PR

PR #5108, opened 2019-09-06
Fix filter attributes

llienher:
Fix https://jira.camptocamp.com/browse/GSGMF-1106


sbrunner:
@fredj do you agree with the second commit that disable strictNullChecks?

sbrunner:
For CI see:
https://github.com/openlayers/openlayers/pull/9888
https://github.com/openlayers/openlayers/pull/9897
...

PR #5109, opened 2019-09-10
Fix overlay position in API

fredj:


api/src/Map.js
```diff
@@ -341,14 +341,15 @@ class Map {
...
    */
-  selectObject(id, table = false) {
+  selectObject(id, position = null, table = false) {
     const feature = this.vectorSource_.getFeatureById(id);
     if (feature) {
-      const coordinates = /** @type {import('ol/geom/Point.js').default} */(
-        feature.getGeometry()
-      ).getCoordinates();
+      if (!position) {
+        position = feature.getGeometry().getCoordinates();
```
sbrunner: `feature.getGeometry()` should be cast (in SimpleGeometry?)...

PR #5111, opened 2019-09-10
Be able to change the logo in the simple project

sbrunner:


sbrunner:
I will try :-)

Mobile has no header bar, I will also migrate desktop_alt and oe_edit :-)

sbrunner:
Simplify, also test other desktop interfaces :-)

PR #5114, opened 2019-09-12
Introduce a new query component and example

adube:
## GSGMF 1047 - phase 1 

This patch is the first step towards implementing a new way to query the map by drawing polygons.

This patch introduces:

* a new `QueryModeSelector` service that:
   * will replace the `QueryKeyboard` service
   * stores the **mode** of the queries, i.e. "click", "draw_box", "draw_polygon"
   * stores the **action** of the queries, i.e. "add", "remove", "replace"
   * listens to the keyboard key pressed to temporarily change the mode or action
* a new `<ngeo-query>` component that:
  * will replaces the `<ngeo-map-query>` and `<ngeo-bbox-query>` directives
  * uses the new query mode selector
  * uses a Draw interaction to draw the boxes (instead of DragBox in the original mapquery directive)
* a new ngeo "query" example that features the component, in which you can see it in action, with all the modes enabled

### Limit option

There is a `limit` option that can be set in the `ngeoQueryOptions` angular value.

There is also a `ngeo-bbox-query-limit` attribute that can be set in the `ngeo-bbox-query.

As part of this refactoring, I suggest that only the `limit` option in `ngeoQueryOptions` should be used.

### autoclear option

There is a `ngeo-map-query-autoclear` attribute that can be set in the `ngeo-map-query` directive.  It was added as a possible "binding attribute" in the new component.

I'm mentionning it, because we do have `ngeoQueryOptions` and it could go there instead.  If you think the option should go there instead, please let me know.

**Update**: it was decided that the autoclear option should go in the `ngeoQueryOptions` instead.  This will come in an other PR.

### Next step

The next step will be to remove the 2 old directives, examples and replace the query tool in all examples (ngeo, gmf) and applications (gmf).

### DIFF
--- a/examples/query.css
+++ b/examples/query.css
@@ -1,0 +1,5 @@
+
+#map {
+  width: 600px;
+  height: 400px;
+}
--- a/examples/query.html
+++ b/examples/query.html
@@ -32,9 +32,13 @@
           href="../apidoc/ngeo.misc.ToolActivateMgr.html"
           title="Read our documentation">ngeo.misc.ToolActivate</a>
       </code>
-      can be used to make the directive inactive when an other tool
+      can be used to make the component inactive when an other tool
       becomes active.  The "Dummy" button here does nothing, but when
-      toggled the query directive becomes inactive.
+      toggled the query component becomes inactive. Click on the map to
+      issue a query. Hold down the "ctrl" key (meta on MAC) then click twice
+      in the map to draw a box instead. Hold the A key to add the results to
+      existing ones. Hold the X key to remove the results from the existing
+      ones.
     </p>
 
     <button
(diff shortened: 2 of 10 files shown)


fredj:
@adube I've pushed a commit that should fix the typing checks

adube:
@fredj or @sbrunner any idea why Travis is complaining?

Thanks Fred for the fixes btw

fredj:
the travis error is not related to your changes. I've restarted the build

sbrunner:
@adube The CI pass :-)

adube:
Alright!  This is ready for review, then.

adube:
@fredj I'll take care of moving the "autoclear" option of the component to the `ngeo.QueryOptions` angular value in an other PR.

This is ready for review.

PR #5081, opened 2019-08-22
Introduces swipe component to desktop alt

RBcote:
Introduces layer being swipe into desktop alt.

PR #5121, opened 2019-09-17
Introduces swipe component to desktop alt

fredj:
#5081 was not correctly commited to the master branch


contribs/gmf/apps/desktop_alt/sass/desktop_alt.scss
```diff
@@ -14,3 +13,53 @@ div .ngeo-displaywindow {
...
+    opacity: 1;
+  }
+}
+
+ngeo-mapswipe div {
+  vertical-align: middle;
+  width: 100%;
+  position: absolute;
+  top: 0;
+}
```
sbrunner: This styles should be in a separate file... (like this it's impossible to include it in the desktop)

sbrunner:
The map swiper should be visible in the standard desktop (for me it should be in booth)...

fredj:
> The map swiper should be visible in the standard desktop (for me it should be in booth)...

agree, I'll change that

PR #5125, opened 2019-09-18
Introduce polygon query

adube:
## GSGMF 1047 - phase 4

This patch is the last step towards implementing a new way to query the map by drawing polygons, as it is included in the patch.

The query component now includes a drawfeature interaction that uses polygon type. A button is also added in the query panel tool.

In the Querent service, the polygon is given as an option (like a coordinate or box could be given) and transformed into a Spatial Filter that is added to the WFS request sent.

In the filter/RuleHelper, a method is introduced to "join filters", i.e. given two filters (both of which can be null), wrap them in an AND filter, if necessary.

### DIFF
--- a/src/datasource/OGC.js
+++ b/src/datasource/OGC.js
@@ -595,20 +595,27 @@ class OGC extends ngeoDatasourceDataSource {
   }
 
   /**
-   * @param {?string} layer The layer name.
+   * @param {string=} layer The layer name.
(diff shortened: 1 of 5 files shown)


adube:
@fredj Ready for review.

adube:
@fredj Travis is happy!

fredj:
> @fredj Travis is happy!

so do I ! please merge

PR #5126, opened 2019-09-18
Fix query example projection - 2056

adube:
This patch fixes the ngeo query example to use projection 2056.


adube:
Travis is complaining for something that's not related to my changes.  Can I merge this even if Travis is complaining?

sbrunner:
@adube Travis pass :-)

PR #5131, opened 2019-09-19
Some little fix

sbrunner:


contribs/gmf/examples/backgroundlayerselector.js
```diff
@@ -7,6 +7,7 @@ import gmfMapComponent from 'gmf/map/component.js';
 
 import gmfThemeThemes from 'gmf/theme/Themes.js';
 import EPSG21781 from '@geoblocks/proj/src/EPSG_21781.js';
```
fredj: `EPSG21781` is not used anymore: this line can be removed

PR #5173, opened 2019-09-26
Add search in WMS browser

arnaud-morvan:


arnaud-morvan:
@sbrunner : I do not understand why typecheck complains about:
```
const index = layer.Title.toLowerCase().indexOf(this.searchText.toLowerCase());
      if (index >= 0) {
        layer._searchPrefix = layer.Title.substring(0, index);
        layer._searchMatch = layer.Title.substring(index, index + this.searchText.length);
        layer._searchSuffix = layer.Title.substring(index + this.searchText.length);
        layer._visible = true;
      }
```
```
/home/amorvan/dev/ngeo/contribs/gmf/src/import/importdatasourceComponent.js
  444:59  error  Operands of '+' operation must either be both strings or both numbers  @typescript-eslint/restrict-plus-operands
  445:53  error  Operands of '+' operation must either be both strings or both numbers  @typescript-eslint/restrict-plus-operands
``` 
String.indexOf and String.length always return integers.

PR #5188, opened 2019-10-02
Add notifications

sbrunner:
Do we want notification on slack?


fredj:
> Do we want notification on slack?

maybe in a specific room

PR #5205, opened 2019-10-08
Support intranet users

sbrunner:
Requires: https://github.com/camptocamp/c2cgeoportal/pull/5356

### DIFF
--- a/contribs/gmf/src/authentication/component.html
+++ b/contribs/gmf/src/authentication/component.html
@@ -1,3 +1,8 @@
+<div ng-if="$ctrl.gmfUser.is_intranet">
+  <div class="form-group">
+    <span>{{'You are recognise as an intranet user.' | translate}}</span>
+  </div>
+</div>
 <div ng-if="$ctrl.gmfUser.username">
   <div class="form-group">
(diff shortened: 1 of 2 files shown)


sbrunner:
@fredj this is ready to be reviewed :-)

contribs/gmf/src/authentication/component.html
```diff
@@ -1,3 +1,8 @@
+<div ng-if="$ctrl.gmfUser.is_intranet">
+  <div class="form-group">
+    <span>{{'You are recognise as an intranet user.' | translate}}</span>
```
jwkaltz: You are recognized as an Intranet user.

sbrunner: Intranet don't take a capital letter (in opposition of Internet), see e.-g.:
https://en.wikipedia.org/wiki/Intranet

jwkaltz: OK... but you still need a "d" at the end of "recognize".

sbrunner: Already done ...

PR #5210, opened 2019-10-09
Merge 2.4 into master

fredj:


fredj:
@sbrunner  is the change in `package.json` correct ?

sbrunner:
No it's wrong, on master we use the ~versions :-)

fredj:
I've reverted the changes in `package.json`

PR #5215, opened 2019-10-10
Update Cesium

gberaudo:
image


sbrunner:
The example is not working, we get a 404 on Cesium.js ...

gberaudo:
@sbrunner, looks broken for a long time already!
I changed to use the CDN instead.

PR #5216, opened 2019-10-10
1121- Map swiper improvements

RBcote:
- Vertical bar is narrower, so its easier to compares two maps.
- The arrow symbol is the same as the one used in the layer tree.
- Vertical bar displayed below query results window and below search field. Still above zoom buttons....


fredj:
Thanks @RBcote !

PR #5221, opened 2019-10-11
1121-2- Remove spacing around arrows (WIP)

RBcote:
This patch removing extra padding around the 'arrows' symbol in the layer tree.


fredj:
the spacing was fixed in an other pull request, closing

PR #5233, opened 2019-10-17
While filter is active, change toolbar icon

adube:
In the "desktop" and "desktop_alt" applications, when there is an actual filter being set onto a layer in the filter tool, then change the icon of the filter in the toolbar.

This replaces the behaviour of having the background being changed instead.

Also, the icon is only changed while the filter is in action, where the background color was always there even if the filter was not active.

##Technical details

In order to do so, we look for the "gmfDataSourceBeingFiltered.dataSource" value, which is the data source being filtered and if it has filters, then its `filterRules` won't be null.


sbrunner:
Looks good to me, @fredj do you want to have a look?

PR #5249, opened 2019-10-24
Use same swipe symbol everywhere, fix swipe div color

fredj:


contribs/gmf/src/sass/swipe.scss
```diff
@@ -14,22 +14,30 @@ ngeo-mapswipe {
...
-      padding: 5px 4px;
-      background-color: #f00;
+      top: calc(50% - 1.25rem);
+      background-color: $brand-primary;
       color: #fff;
+      border: $border;
+      svg {
+        height: 100%;
+        display: block;
+        margin: auto;
```
sbrunner: For me if we add `font-size: 1rem;` the result is better (tested on Firefox)

PR #5250, opened 2019-10-24
Support layer group filtering

adube:
The main goal of this patch is to introduce the support of filtering a WMS layer that uses a group as layer name.

To accomplish this, the following change was made.

The creation of attributes within an OGC data source object is made by first checking if there are attributes set for the layers used by the data source in the inner `ogcLayers_` property.  Then, they are combined together, i.e. only the attributes that are commonly shared in all layers are kept.

From there, the rest works as usual.

## Known issue

~~There is an issue with this approach.  To resume: it is currently not supported by MapServer to give the name of the group in the LAYERS parameter while giving only one filter in the FILTER property.  Therefore, this patch breaks the MapServer WMS layers in those cases.~~

~~More about this issue in GSGMF-1133.~~

The above issue has been fixed in MapServer: https://github.com/mapserver/mapserver/commit/ba27152e140f881e11fc1a97b60086ec4d7d1ee3

### DIFF
--- a/contribs/gmf/src/themes.js
+++ b/contribs/gmf/src/themes.js
@@ -150,6 +150,9 @@
 
 /**
  * @typedef {Object} GmfOgcServerAttribute
+ * @property {string} [alias] The attribute alias.
+ * @property {string} [minOccurs] If not '0', then the attribute is
+ *     considered required.
  * @property {string} namespace The attribute namespace.
  * @property {string} type The attribute type (in namspace).
  */
--- a/src/datasource/Helper.js
+++ b/src/datasource/Helper.js
@@ -91,37 +91,52 @@ export class DatasourceHelper {
    */
   getDataSourceAttributes(dataSource) {
 
-    const wfsDescribeFeatureTypeDefer = this.q_.defer();
+    const getDataSourceAttributesDefer = this.q_.defer();
 
     if (dataSource.attributes) {
-      wfsDescribeFeatureTypeDefer.resolve(dataSource.attributes);
+      getDataSourceAttributesDefer.resolve(dataSource.attributes);
     } else {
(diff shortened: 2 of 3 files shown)


adube:
The patch on MapServer has been merge.  Please, write a comment here once the GMF demo server has its MapServer version up-to-date.

adube:
@sbrunner or @fredj : Ready for review.

sbrunner:
Any reason why we don't have the geom filter?

sbrunner:
When we click on get data we should only have the features from the layer Hotel (trigged with the GetData metadata...)

adube:
> When we click on get data we should only have the features from the layer Hotel (trigged with the GetData metadata...)

Here's a snippet from the JSON returned by the themes service:

```
{
  "maxResolutionHint": 999999999,
  "minResolutionHint": 0,
  "name": "hotel_label",
  "queryable": true
},
```

To my understanding, the label layer returns features because it is queryable.

If that's an issue, then it should be logged as a separate task in JIRA, as it is not related to this PR.

adube:
> Any reason why we don't have the geom filter?

That's an issue intruduced in this PR.  I'll look into it.

adube:
> > Any reason why we don't have the geom filter?
> 
> That's an issue intruduced in this PR. I'll look into it.

The issue has been fixed.  Ready for review.

sbrunner:
> If that's an issue, then it should be logged as a separate task in JIRA, as it is not related to this PR.

=> https://jira.camptocamp.com/browse/GSGMF-1171

sbrunner:
> The issue has been fixed. Ready for review.

Confirmed, thanks :-)

adube:
@sbrunner or @fredj Just a reminder: this PR is ready for review.

PR #5275, opened 2019-11-04
Introduces close button to the swiper

RBcote:
This patch introduce the close button to the map swiper. By clicking on it, the swiper closes

### DIFF
--- a/src/map/swipe.js
+++ b/src/map/swipe.js
@@ -140,6 +140,7 @@ class SwipeController {
    */
   deactivate() {
     this.layer = null;
+    this.map.render();
   }
 
   /**
--- a/src/map/swipe.js
+++ b/src/map/swipe.js
@@ -136,6 +136,13 @@ class SwipeController {
   }
 
   /**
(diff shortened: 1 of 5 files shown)


src/map/swipe.js
```diff
@@ -135,6 +135,10 @@ class SwipeController {
     this.resizeObserver_.observe(this.draggableElement_.parent().get(0));
   }
 
+  deactivate() {
```
adube: Missing comment block

src/map/swipe.js
```diff
@@ -135,6 +135,13 @@ class SwipeController {
     this.resizeObserver_.observe(this.draggableElement_.parent().get(0));
   }
 
+  /**
+   * Allows you to deactivate the swiper on click of the close button.
+   */
+  deactivate() {
+    this.layer = null;
```
fredj:
```suggestion
    this.layer = null;
    this.map.render();
```
Otherwise the layer is now fully displayed after the deactivation.

PR #5283, opened 2019-11-06
Fix duplicate WMS query results

adube:
~~This tiny patch fixes the duplicates that we had in the query results due to the fact that the WFS layers were not being pushed in the proper array.~~

~~So, if the wms and wfs layers were the same, they were duplicated.  In some cases, wms use a layer "group" as name, but wfs uses layer names, so this may explain why there were not duplicates sometimes.~~

**Update**: This patch now properly fixes the listing of wms layer names to be used for query result.

### DIFF
--- a/src/array.js
+++ b/src/array.js
@@ -1,0 +1,17 @@
+import {includes as olArrayIncludes} from 'ol/array.js';
+
+
+/**
+ * Push an object in an array, unless already there.
+ * @param {Array.<*>} arr The array to push the element.
+ * @param {*} obj The object for which to test.
+ * @return {boolean} The object has been pushed in the array.
+ */
+export function pushUnlessIncluded(arr, obj) {
+  let ret = false;
+  if (!olArrayIncludes(arr, obj)) {
+    arr.push(obj);
+    ret = true;
+  }
+  return ret;
+}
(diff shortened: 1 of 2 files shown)


sbrunner:
@adube I rebase your branch to make the CI pass...

sbrunner:
* On ESRI it's working correctly :-)
* On 'OSM scale' and 'OSM open' in 'Function OSM mixed' I get duplicated results...
* On 'OSM scale' and 'OSM open' in 'Function OSM' I don't get any results...

adube:
> * On ESRI it's working correctly :-)
> * On 'OSM scale' and 'OSM open' in 'Function OSM mixed' I get duplicated results...
> * On 'OSM scale' and 'OSM open' in 'Function OSM' I don't get any results...

Yeah, I was thinking about this fix yesterday evening, wondering if this was a typo or intentional.

Turns out it seems to be intentional: with WMS, when using "group" names, the query asks for the group name but the result contains the layer name.  Therefore, I think this is why the wfs layer names were added to the wms ones, but caused issues when the WMS layer doesn't use groups.

I'll come with an other fix and make more tests.  I'll also leave a comment there, to let devs know that this is not a typo, it is intentionnal.

adube:
Thanks. Can I merge this, knowing that the demo is tomorrow?

sbrunner:
+1 :-)

PR #5295, opened 2019-11-13
Change query panel icon

llienher:


sbrunner:
@ybolognini does it look good to you?

ybolognini:
> @ybolognini does it look good to you?

Yes! Much better like this, thanks

PR #5334, opened 2019-11-25
GSGMF-1141: Fix print with filters

llienher:
https://jira.camptocamp.com/browse/GSGMF-1141


src/print/VectorEncoder.js
```diff
@@ -358,8 +358,8 @@ VectorEncoder.prototype.encodeVectorStyleStroke = function(symbolizer, strokeSty
       throw new Error('Wrong strokeColor type');
     }
     const strokeColorRgba = asColorArray(strokeColor);
-    if (!(Array.isArray(strokeColor))) {
-      throw new Error('Wrong strokeColor type');
+    if (!(Array.isArray(strokeColorRgba))) {
+      throw new Error('StrokeColor is not an array');
```
gberaudo:
```suggestion
      throw new Error('parsed StrokeColor is not an array');
```

PR #5335, opened 2019-11-26
GSGMF-1151: Fix error on arcgis layers when printing

llienher:
Printed with this fix:
[GeoMapFish_2019-11-26_10-08-35.pdf](https://github.com/camptocamp/ngeo/files/3891416/GeoMapFish_2019-11-26_10-08-35.pdf)

### DIFF
--- a/src/print/Service.js
+++ b/src/print/Service.js
@@ -308,13 +308,18 @@ PrintService.prototype.encodeWmsLayer_ = function(arr, layer, url, params) {
   delete customParams.SERVERTYPE;
   delete customParams.VERSION;
 
+  let serverType = undefined;
+  if (params.SERVERTYPE !== 'arcgis') {
+    serverType = params.SERVERTYPE;
+  }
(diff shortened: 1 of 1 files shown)


src/print/Service.js
```diff
@@ -320,6 +320,12 @@ PrintService.prototype.encodeWmsLayer_ = function(arr, layer, url, params) {
     version: params.VERSION,
     useNativeAngle: this.printNativeAngle_,
   };
+
+  // Quickfix to GSGMF-1151 for arcgis layers
+  if (object.serverType === 'arcgis') {
+    delete object.serverType;
+  }
```
sbrunner: I prefer to remove the line 317 and do:
```
if (params.SERVERTYPE !== 'arcgis') {
    object.serverType = params.SERVERTYPE;
}

llienher: You're right, it's better :+1

PR #5345, opened 2019-12-04
Support defining size of features in GMF draw tool

adube:
This patch introduces a new component in GMF to specify sizes (length, width, height, radius) while drawing features in the draw tool.

## gmf-drawfeatureoptions

A new component is born: gmf-drawfeatureoptions

The parameters it must have are:

 * the OpenLayers map
 * a single interaction responsible to draw the feature (can be a measure interaction from ngeo, or an OpenLayers Draw interaction directly)

The behaves according to the type of interaction it was given.  The possible ones are:

 * the ngeo length measure interaction
 * the ngeo area measure interaction
 * the ngeo azimut measure interaction
 * a draw interaction (used to draw rectangles)

## Use of the Snap interaction

The Snap interaction of OpenLayers is used as main way to force the drawing tool to "snap" to an invisible grid while drawing and while sizes are set.

A single feature is created and its geometry is set in a way that the mouse always snaps to it while drawing.  This is done using an enormous pixel tolerance of 10,000 pixels.

## Hack to allow draw end on double-click

Double-clicking ends the drawing of a LineString or Polygon.  While snapping with a 10,000 pixel tolerance, the event is eaten up by the Snap interaction, and messes with the Draw interaction.

To fix this, we manually end the drawing on double clicking, when necessary.

## All Draw interactions are required

In order for the feature used by the Snap interaction to have its geometry created, we need to have access to the feature being drawn.  A feature being drawn is only accessible from its Draw interaction.

There was currently no way to have access to those, which are created and nested in the NGEO drawfeature directive.

To obtain them, they were given unique ids, and then obtained through the map itself, in the GMF drawfeature directive, by using their given unique ids as a way to find them.

### DIFF
--- a/contribs/gmf/src/drawing/drawFeatureOptionsComponent.js
+++ b/contribs/gmf/src/drawing/drawFeatureOptionsComponent.js
@@ -1,0 +1,517 @@
+import angular from 'angular';
+
+import {listen as olEventsListen} from 'ol/events.js';
+import OLFeature from 'ol/Feature.js';
+import {
+  Circle as OLGeomCircle,
+  GeometryCollection as OLGeomGeometryCollection,
+  LineString as OLGeomLineString,
+  Polygon as OLGeomPolygon
+} from 'ol/geom.js';
+import {fromExtent as olGeomPolygonFromExtent} from 'ol/geom/Polygon.js';
+import OLInteractionDraw from 'ol/interaction/Draw.js';
+import OLInteractionPointer from 'ol/interaction/Pointer.js';
+import OLInteractionSnap from 'ol/interaction/Snap.js';
+import OLMapBrowserEvent from 'ol/MapBrowserEvent.js';
+import OLSourceVector from 'ol/source/Vector.js';
+
+import {unlistenByKeys as ngeoEventsUnlistenByKeys} from 'ngeo/events.js';
+
+/**
+ * @type {angular.IModule}
+ * @hidden
+ */
+const module = angular.module('GmfDrawFeatureOptionsComponent', [
+]);
+
+
+module.run(
+  /**
+   * @ngInject
+   * @param {angular.ITemplateCacheService} $templateCache
+   */
+  ($templateCache) => {
+    $templateCache.put(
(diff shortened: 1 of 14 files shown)


contribs/gmf/src/drawing/drawFeatureOptionsComponent.js
```diff
@@ -0,0 +1,517 @@
...
+    this.snapSource_ = new OLSourceVector({
+      features: [this.snapFeature_]
+    });
+
+    /**
+     * @type {!OLInteractionSnap}
+     * @private
+     */
+    this.snapInteraction_ = new OLInteractionSnap({
+      // @ts-ignore: webpack
```
adube: I had to add this line to ignore an error reported by the typecheck.  I couldn't find why or what it was complaining about.  Any hint would be welcome.

fredj: the `handleEvent` option is missing in OpenLayers, that's fine to have a `ts-ignore`

adube:
Thanks.  Was there anything more to review?

sbrunner:
Just tow remarks on testing the example:
- I enforce the line length to 1000m and I get (in the tooltip) 1km vertically and 997m horizontally. It's probably a projection - geodetic issue, but we should do the same on booth cases...
- I don't know what's specified but when we draw a rectangle we can't enforce only the width or only the height...

adube:
>I enforce the line length to 1000m and I get (in the tooltip) 1km vertically and 997m horizontally. It's probably a projection - geodetic issue, but we should do the same on booth cases...

I highly suspect that's an OpenLayers issue.  I don't think this should be addressed as part of this PR.

>I don't know what's specified but when we draw a rectangle we can't enforce only the width or only the height...

It was not specified.  I'll ask the question on JIRA.

adube:
>I don't know what's specified but when we draw a rectangle we can't enforce only the width or only the height...

Enforcing only one side at a time is not required, but we need to let the user know about it.  I'll add a message to the panel.

adube:
@sbrunner or @fredj Ready for review.

PR #5355, opened 2019-12-09
Additional search actions

svaabs:
As discussed with @sbrunner we from Basel-Stadt are creating a pull request to get this additional gmf functionality within our project.

Our aim is to include tools to the search so a user can search for a tool and with a click on the search result the tool will be opened. The implementation in our project can be found here: https://github.com/camptocamp/baselstadt_mapbs/commit/a2d95c78976a8fa30d4ecb47c71699429ac3e18e

How do we have to proceed to get this to GMF 2.5 as base functionality for futher versions?

### DIFF
--- a/contribs/gmf/src/search/component.js
+++ b/contribs/gmf/src/search/component.js
@@ -346,12 +346,6 @@ class SearchController {
     this.searchActionCallback;
 
     /**
-     * @type {function(): void}
-     * @private
-     */
-    this.executeSearchAction_;
-
-    /**
      * Whether or not to show a button to clear the search text.
      * Default to true.
      * @type {boolean}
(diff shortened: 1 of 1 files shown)


contribs/gmf/src/search/component.js
```diff
@@ -338,6 +340,17 @@ class SearchController {
      */
     this.onInitCallback;
 
+    /**
+     * @type {function(): void}
```
sbrunner:
```suggestion
     * @type {function(any): void}
```

contribs/gmf/src/search/component.js
```diff
@@ -993,6 +1009,10 @@ class SearchController {
             const silent = !!featureGeometry;
             this.gmfTreeManager_.addGroupByLayerName(actionData, true, silent);
           }
+        } else {
+          if (this.executeSearchAction_) {
+            this.executeSearchAction_(action);
+          }
```
sbrunner: With es6 arrow function we should be able to call ``searchActionCallback`` (the ``executeSearchAction_`` seem not be necessary) directly, and here:
https://github.com/camptocamp/baselstadt_mapbs/blob/ff51369ec2ab256738e2d51aeb83580614708d3f/mapbs/static-ngeo/js/ngeodesktop.js#L543 the intermediate function can be removed.

svaabs:
Thanks for the review. I fixed your suggestions.

PR #5358, opened 2019-12-13
Close associated popup submenu at layertree removal

llienher:
https://jira.camptocamp.com/browse/GSGMF-1161

This fix allows a node to be removed from layertree as well as its associated popup menu if existing.

I had to add a new ID to HTML element to easily find the attributes I needed to extract. Popup identifier is created randomly at each new opening and is not consistent when closing/opening the menu.

I added this popup identifer to a optional custom attribute in the GMF group and retrieve it when I want to target the existing popup.

I use UID over ID, as these last are duplicated when a single layer is in several groups loaded in the layertree. If I don't I'm not able to select the correct html element.

### DIFF
--- a/contribs/gmf/src/layertree/component.js
+++ b/contribs/gmf/src/layertree/component.js
@@ -25,6 +25,7 @@ import olSourceImageWMS from 'ol/source/ImageWMS.js';
 import olSourceTileWMS from 'ol/source/TileWMS.js';
 import olSourceWMTS from 'ol/source/WMTS.js';
 import LayerBase from 'ol/layer/Base.js';
+import {getUid} from 'ol/util.js';
 
 import 'bootstrap/js/src/collapse.js';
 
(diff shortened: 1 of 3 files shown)


contribs/gmf/src/layertree/component.js
```diff
@@ -759,21 +760,69 @@ Controller.prototype.afterReorder = function() {
...
+};
+
+
+/**
+ * @param {import('gmf/themes.js').GmfGroup} node Layer tree node to remove.
+ * @private
+ */
+Controller.prototype.removePopup_ = function(node) {
+  const popupId = node.popupId;
+  $(`div#${popupId}`).remove();
```
fredj:
```suggestion
  $(`#${popupId}`).remove();
```

contribs/gmf/src/layertree/component.js
```diff
@@ -759,21 +760,69 @@ Controller.prototype.afterReorder = function() {
...
 
 
+/**
+ * @param {import('gmf/themes.js').GmfGroup} node Layer tree node to tag popup identifier.
+ */
+Controller.prototype.tagPopup = function(node) {
+  const uid = getUid(node);
+
+  // Find the random id associated with the new popup being opened.
+  const popupId = $(`span.extra-actions#popup-id-${uid}`).attr('aria-describedby');
```
fredj:
```suggestion
  const popupId = $(`#popup-id-${uid}`).attr('aria-describedby');
```

contribs/gmf/src/layertree/component.js
```diff
@@ -759,21 +760,69 @@ Controller.prototype.afterReorder = function() {
...
 };
 
 
-/**
- */
 Controller.prototype.removeAllNodes = function() {
+  const nodes = this.root.children;
+  nodes.forEach(node => {
+    this.parseTreeNodes(node);
+  });
```
gberaudo: You can do `this.parseTreeNodes(this.root)`, right?

llienher: Right, forgot to change that after doing my recursive function. Now fixed.

PR #5403, opened 2020-01-06
Add spinner for WFS query window

llienher:
Implements the WFS query loading spinner.
https://jira.camptocamp.com/browse/GSGMF-1168

### DIFF
--- a/contribs/gmf/src/query/windowComponent.html
+++ b/contribs/gmf/src/query/windowComponent.html
@@ -1,3 +1,9 @@
+<div class="spinner-window" ng-if="ctrl.isLoading">
+  <i class="fa fa-spin">
+    <%=require('gmf/icons/spinner.svg?viewbox&height=3rem')%>
+  </i>
+</div>
+
 <div
   class="gmf-displayquerywindow"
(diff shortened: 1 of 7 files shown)


contribs/gmf/src/query/windowComponent.html
```diff
@@ -1,3 +1,9 @@
+<div id="spinner-window" ng-if="ctrl.isLoading">
```
fredj: don't use an id but a CSS class (eg. `gmf-displayquerywindow-spinner`)

contribs/gmf/src/query/windowComponent.js
```diff
@@ -267,9 +267,20 @@ export function QueryWindowController(
...
 
+  /**
+   * @type {boolean}
+   */
+  this.isLoading = false;
+
   $scope.$watchCollection(
     () => ngeoQueryResult,
     (newQueryResult, oldQueryResult) => {
+      if (newQueryResult.pending) {
```
fredj: can be simplified: `this.isLoading = newQueryResult.pending`

PR #5404, opened 2020-01-06
Do not copy 'visible' property for WMTS layers

adube:
The WMTS layers are created in a way that their properties get asynchronously set, i.e. after WMTS capabilities are obtained.

The setting of the layer's initial visibility can be done before that, which can result in an issue of overriding the visibility after the capabilities are read, i.e. the created layer while reading the capabilities has its visible property set to true, but the original one may have its visible property to false.

To fix this issue, an utility method was added in the LayerHelper of ngeo to be able to exclude the keys we know we don't want to copy.

### DIFF
--- a/contribs/gmf/src/permalink/Permalink.js
+++ b/contribs/gmf/src/permalink/Permalink.js
@@ -1342,16 +1342,6 @@ PermalinkService.prototype.initLayers_ = function() {
             }
             this.gmfLayerBeingSwipe_.layer = treeCtrl.layer;
           }
-
-          if (
-            treeCtrl.layer.getLayers &&
-            treeCtrl.layer.getLayers().item(0) &&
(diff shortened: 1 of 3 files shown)


contribs/gmf/src/permalink/Permalink.js
```diff
@@ -1342,6 +1342,16 @@ PermalinkService.prototype.initLayers_ = function() {
             }
             this.gmfLayerBeingSwipe_.layer = treeCtrl.layer;
           }
+
+          if (
+            treeCtrl.layer.getLayers &&
+            treeCtrl.layer.getLayers().item(0) &&
+            treeCtrl.layer.getLayers().item(0).get('layerNodeName') === 'ch.are.alpenkonvention'
```
sbrunner: Debug that should be removed?

adube: Oops, indeed.  I'll do that.

PR #5408, opened 2020-01-08
Read queryLayers metadata for WMS data source

adube:
## Issue

With the introduction of #5250 (Support layer group filtering), a WMS data source that use a "group" as WMS layer can have multiple inner WFS layers, each of which get filtered properly.

However, an issue occurs when we try to query such a data source: we get duplicate records.  More precisely, we have 1 set or records per WFS layer.  For example, in a group "hotel_filter", which groups 2 layers: "hotel" and "hotel_layer", we have 2 sets of results, one for "hotel" and one for "hotel_layer".

We should have only one.

## Assumption

~~It is assumed that a WMS data source that uses a group layer should only have ONE of its inner WFS layer to return data when queries are issued.~~ It's okay to have multiple inner layers within a group that can return data.

It is assumed that all WFS layer within a group can be filtered through the WMS layer in the map.

## Fix

This patch implements a way to only have 1 set of records returned for a WMS data source that use a group: by using the `queryLayers` metadata.

This metadata was already introduced for WMTS, and used to be only used by WMTS.  Now, if defined in a WMS data source, it means that only the layers defined in there should be queried.

That metadata is read in the GMF data source manager, and is translated as a new property of the `GmfLayerChildLayer` object: `getData`.  That new property is not an official property supported by the Themes service, but is rather a calculated value to have a nice way to know at the "child layer level" if it is queryable and can have data fetched (i.e. if it can be used to issue requests).  That property is assumed to be `true`, if not defined.

### DIFF
--- a/contribs/gmf/src/datasource/Manager.js
+++ b/contribs/gmf/src/datasource/Manager.js
@@ -524,19 +524,40 @@ export class DatasourceManager {
           break;
         }
       }
+
+      // Read 'meta.queryLayers' for WMS data source to set 'getData'
+      //
+      // If a WMS data source has the `queryLayers` property set in
+      // its metadata, then it lists the layers that should be used
+      // when queries are issued with that data source, i.e. when a
+      // WFS GetFeature request is issued. The 'queryable' property
+      // still needs to be true for the purpose of the filter tool to
+      // work properly, i.e. a data source must be queryable to be
+      // filtrable.
+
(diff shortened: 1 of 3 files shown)


adube:
@fredj or @sbrunner Ready for review.

adube:
Fix applied.  Ready for code review: @fredj

PR #5445, opened 2020-01-21
Vertically align 'Result x/y' at bottom of display window result mobile

adube:
As title says.


adube:
@sbrunner Ready for review.

adube:
@fredj Ready for review.

PR #5447, opened 2020-01-21
Revert regression introduced in ngeo displayWindow component

adube:
This patch fixes an other regression introduced by the automatic changes in: 6d38eed1c249a8d3624eeb8ab2a3248c5b508d16

This time it's the ngeo `displayWindowComponent`.

## Explanation

An angular component gets its its binding properties (i.e. those defined in the html) being set their values in the $onInit method.  In the constructor, their values are not yet set but we do need to type them:

```
   /**
     * @type {boolean}
     */
    this.resizable;
```
The issue introduced by the auto-patch is that they got values assigned according to their type:

```
    /**
     * @type {boolean}
     */
    this.resizable = false;
```

This breaks the setting of the value by Angular in $onInit.  Therefore, we should not do that.  It's okay to leave the type as `boolean`, because the property itself in `this` will be set to a boolean in $onInit.  The angular option (through the biding) is optional but that doesn't matter.

There may be more regressions like this.

### DIFF
--- a/src/message/displaywindowComponent.js
+++ b/src/message/displaywindowComponent.js
@@ -79,7 +79,7 @@ class Controller {
     /**
      * @type {?string}
      */
-    this.content ;
+    this.content;
 
     /**
      * @type {?string}
--- a/src/message/displaywindowComponent.js
+++ b/src/message/displaywindowComponent.js
@@ -124,7 +124,7 @@ class Controller {
     /**
      * @type {?string}
      */
-    this.title ;
+    this.title;
 
     /**
      * @type {?string}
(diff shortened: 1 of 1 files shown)


src/message/displaywindowComponent.js
```diff
@@ -74,67 +74,67 @@ class Controller {
...
      * @type {boolean}
      */
-    this.resizable = false;
+    this.resizable;
 
     /**
      * @type {?string}
      */
-    this.title = null;
+    this.title ;
```
sbrunner: Extra space :-)

PR #5450, opened 2020-01-23
Add search input to the api

fredj:


package.json
```diff
@@ -39,6 +39,7 @@
     "@geoblocks/proj": "0.4.2",
     "@sentry/browser": "5.10.2",
     "@sentry/types": "5.10.0",
+    "@trevoreyre/autocomplete-js": "^2.1.0",
```
sbrunner: extra ^

api/dist/apihelp/apihelp.html
```diff
@@ -207,6 +207,22 @@ <h2>Load data from a text file</h2>
...
+                    <h2>Search input</h2>
+                    <div id='map10search' style="width: 400px;"></div>
+                    <div id='map10' class="map"></div>
+                    <pre data-language="javascript">
+var map = new demo.Map({
+    div: 'map10',
+    zoom: 3,
+    center: [2544500, 1210100],
+    addSearch: true,
+    searchElementId: 'map10search'
```
sbrunner: Do we relay needs the `addSearch`?
To be coherent with `div` shouldn't it be `searchDiv`?

sbrunner:
Thanks :-)

PR #5470, opened 2020-01-28
Introduce pending in query ModeSelector

adube:
This patch introduces a `pending` property in the ngeo query ModeSelector to manage the changing of query mode while a request is already pending.

The idea is to keep the mode "active" while the request is pending to prevent the issue that was occuring while pressing the Ctrl key to draw a box, make the query tool issue a request and while it was still pending releasing the Ctrl key immediately changed the mode.


sbrunner:
Looks good :-)

PR #5472, opened 2020-01-29
Test and add copyright notice in all files

sbrunner:


buildtools/copyright/package-lock.json
```diff
@@ -0,0 +1,4 @@
+{
```
fredj: we should not commit this file

PR #5491, opened 2020-02-05
Fix mobile and mobile_alt spinner notification on query

llienher:
The spinner was out of the interface, but working. I fixed the css.

https://jira.camptocamp.com/browse/GSGMF-1193


sbrunner:
Looks good to me :-)

PR #5517, opened 2020-02-12
Fix legend show/collapse, and legend for QGIS

adube:
Fixes the regression introduced in dd38023b8ad9a45370babc90e18a215c9f161d57 that caused the show/hide to no longer works for any legend in the layer tree. (@llienher Please, check what I restored and see if/why the change was made).

Then, this patch also fixes the issue with QGIS legends.  It was simply a matter of nothing being handled for them at all.

### DIFF
--- a/contribs/gmf/src/layertree/component.js
+++ b/contribs/gmf/src/layertree/component.js
@@ -643,16 +643,19 @@ Controller.prototype.getLegendsObject = function(treeCtrl) {
     const scale = this.getScale_();
     // QGIS can handle multiple layers natively. Use Multiple URLs for other map
     // servers
-    if (gmfOgcServer.type !== ServerType.QGISSERVER) {
-      const layerNamesList = layersNames.split(',');
-      layerNamesList.forEach((layerName) => {
-        const wmtsLegendURL = this.layerHelper_.getWMSLegendURL(gmfOgcServer.url, layerName, scale);
-        if (!wmtsLegendURL) {
-          throw new Error('Missing wmtsLegendURL');
-        }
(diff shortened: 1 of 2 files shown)


contribs/gmf/src/layertree/component.js
```diff
@@ -933,7 +936,7 @@ Controller.prototype.toggleSwipeLayer = function(treeCtrl) {
 Controller.prototype.toggleNodeLegend = function(legendNodeId) {
   const div = document.querySelector(legendNodeId);
   if (div) {
-    div.classList.toggle('d-none');
+    div.classList.toggle('show');
```
llienher: @adube Do you need to make change ? I tested and mobile is actually not working at all to open the legend div if using "show", that why I made this change in the first place... So I don't know where the issue is with that interface.

Is it a class that come from jquery-ui initially ?

adube: I'm not sure where it comes from.

That line was recently changed from 'show' to 'd-none', which causes the desktop version to no longer work.

I'll check the mobile version too and see if both are required for both versions to work.

adube: Ok, to my understanding the small "cog" icon tries to do 2 different things on mobile and desktop and the way it is currently implemented only one of them can work.

As a potential fix, I'll try to see if I can separate the container as two different elements instead, one for mobile and one for desktop, to make them both work individually.

adube: I've made the fix.  Now both apps work properly.

llienher: Great, looks good to me!

PR #5553, opened 2020-02-21
[api] Fix getFeaturesFromIds when GMF and WMS layers do not have the …

arnaud-morvan:
…same name

Need to be ported in master

### DIFF
--- a/api/src/Querent.js
+++ b/api/src/Querent.js
@@ -54,14 +54,20 @@ export function getFeaturesFromIds(layer, ids) {
         return;
       }
 
-      const featureIds = ids.map(id => `${layer}.${id}`);
+      const gmfLayer = /** @type import('gmf/themes.js').GmfLayerWMS */ (overlayDef.layer);
+      const childLayerNames = [];
+      let featureIds = [];
+      for (const childLayer of gmfLayer.childLayers) {
+        childLayerNames.push(childLayer.name);
+        featureIds = featureIds.concat(ids.map(id => `${childLayer.name}.${id}`));
+      }
 
       const params = {
         'FEATUREID': featureIds.join(','),
-        'MAXFEATURES': ids.length,
+        'MAXFEATURES': featureIds.length,
         'REQUEST': 'GetFeature',
         'SERVICE': 'WFS',
-        'TYPENAME': layer,
+        'TYPENAME': childLayerNames.join(','),
         'VERSION': '1.0.0'
       };
       const url = olUriAppendParams(overlayDef.ogcServer.urlWfs, params);


api/src/Querent.js
```diff
@@ -54,14 +54,17 @@ export function getFeaturesFromIds(layer, ids) {
...
+      }
 
       const params = {
         'FEATUREID': featureIds.join(','),
-        'MAXFEATURES': ids.length,
+        'MAXFEATURES': featureIds.length,
         'REQUEST': 'GetFeature',
         'SERVICE': 'WFS',
-        'TYPENAME': layer,
+        'TYPENAME': overlayDef.layer.layers,
```
arnaud-morvan, line 67: Use childLayers names instead

api/index.js
```diff
@@ -2,7 +2,7 @@ import config from './src/constants.js';
 import Map from './src/Map.js';
 
 config.themesUrl = 'https://geomapfish-demo-2-4.camptocamp.com/themes?' +
-  'version=2&background=background&interface=api';
+  'version=2&background=background&interface=desktop';
```
sbrunner, line 5: This should not be committed...

api/src/Querent.js
```diff
@@ -54,14 +54,19 @@ export function getFeaturesFromIds(layer, ids) {
         return;
       }
 
-      const featureIds = ids.map(id => `${layer}.${id}`);
+      const childLayerNames = [];
+      let featureIds = [];
+      for (const childLayer of overlayDef.layer.childLayers) {
```
sbrunner, line 59:
```suggestion
      for (const childLayer of overlayDef.childLayers) {
```

sbrunner: wrong - removing ...

sbrunner, line 59: ``overlayDef.layer`` should be casted into ``GmfLayerWMS``...

api/index.js
arnaud-morvan: Commited by mistake, but note that interface API does not exists on demo 2.4.

api/src/Querent.js
```diff
@@ -42,7 +42,7 @@ export function getFeaturesFromIds(layer, ids) {
 
       /** Array<import('ol/Feature.js').default> */
       let features = [];
-      const overlayDef = overlayDefs.get(layer);
+      const overlayDef = /** @type import('gmf/themes.js').GmfLayerWMS */ overlayDefs.get(layer);
```
sbrunner, line 45:
```suggestion
      const overlayDef = overlayDefs.get(layer);
      const gmfLayer = /** @type import('gmf/themes.js').GmfLayerWMS */ (overlayDef.layer);
```

api/index.js
sbrunner: Just created :-)

sbrunner:
Nice :-)

arnaud-morvan:
Thanks @sbrunner

sbrunner:
@arnaud-morvan do you see that the backports failed?

arnaud-morvan:
Yes, already opened backport PRs. Master one is failing, but is it master branch which is broken.

arnaud-morvan:
Sorry, master one is green now.

PR #5567, opened 2020-02-26
Initialize map with opacity 0 for background layer slider

llienher:


contribs/gmf/src/backgroundlayerselector/component.js
```diff
@@ -265,7 +265,10 @@ Controller.prototype.setLayer = function(layer, opt_silent) {
  * @param {import("ol/layer/Base.js").default} layer The opacity background layer.
  */
 Controller.prototype.setOpacityBgLayer = function(layer) {
-  const opacity = this.opacityLayer ? this.opacityLayer.getOpacity() : layer.getOpacity();
+  let opacity = 0;
+  if (this.opacityLayer) {
+    opacity = this.opacityLayer.getOpacity();
+  }
```
sbrunner:
```suggestion
  const opacity = this.opacityLayer ? this.opacityLayer.getOpacity() : 0;
```
let should be avoided in this kind of cases.

sbrunner: In which case the `layer.getOpacity()` is not `0`?

llienher: At page loading, the value was set to 1. For example if you load the page and the baselayer_opacity parameter was missing from the permalink, the value was always set to 100 (1 in layer).

PR #5593, opened 2020-03-05
Show loading screen after login while themes are loading

adube:
This patch introduces the following behaviour:  In the "desktop" applications, after the user successfully logs in, the "login" panel will stay open until:

 * the login is successful (as before)
 * and after the new themes have been loaded (this is new)

Also, after loging in, while the new themes are being loaded, a message is shown in the login panel along with a spinner.


adube:
Ready for review.

PR #5620, opened 2020-03-17
Set default layout for print component

svaabs:
The aim of this change is to allow the configuration of the default print layout in the print component.


sbrunner:
@svaabs Can't it be in the 'gmfPrintOptions'?

svaabs:
> @svaabs Can't it be in the 'gmfPrintOptions'?

@sbrunner Sure, if that is the prefered way, I will change the code.

sbrunner:
@svaabs your tow pull request will be in conflict...

PR #5621, opened 2020-03-17
Allow to set goodness of fit parameter for labels

svaabs:
With this pull request we would like to add the configuration of the MapFishPrint goodnessOfFit parameter to ngeo. The issue was already discussed here:
- https://github.com/mapfish/mapfish-print/issues/969
- https://github.com/mapfish/mapfish-print/issues/981

PR #5661, opened 2020-03-27
Port default print layout

llienher:
Port 2.4 features in master branch

### DIFF
--- a/src/print/Service.js
+++ b/src/print/Service.js
@@ -153,17 +153,20 @@ PrintService.prototype.cancel = function(ref, opt_httpConfig) {
  * @param {string} format Formats.
  * @param {Object<string, *>} customAttributes Custom attributes.
  * @param {string=} email Email to send the file to.
+ * @param {number=} goodnessOfFit Goodness of fit.
  * @return {import('ngeo/print/mapfish-print-v3.js').MapFishPrintSpec} The print spec.
  */
 PrintService.prototype.createSpec = function(
-  map, scale, dpi, layout, format, customAttributes, email) {
+  map, scale, dpi, layout, format, customAttributes, email, goodnessOfFit) {
 
   const specMap = /** @type {import('ngeo/print/mapfish-print-v3.js').MapFishPrintMap} */ ({
     dpi: dpi,
     rotation: customAttributes.rotation
   });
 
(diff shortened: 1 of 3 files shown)


llienher:
@sbrunner I have an issue in the console, testing this port:

```
vendor.js?dev:2 org.mapfish.print.ExtraPropertyException: Extra properties found in the request attributes
Extra Properties: 

	* goodnessOfFit

All allowed properties are: 

	* northArrow
	* comments
	* debug
	* requestHeaders
	* legend
	* datasource
	* title
	* map
	* scalebar
```

Do you have an idea what is missing ? I'm pretty sure we already had this problem in the past but cannot figure out what we did to fix it...

sbrunner:
@llienher Where do you found that?

llienher:
@sbrunner It's in my console, error when trying to print

sbrunner:
It's not already present on master?

llienher:
Yes, it is actually the previous merged PR that broke that, not related to this one. So visible on master.

sbrunner:
C'est aussi le cas sur 2.4 :-(

sbrunner:
This line looks wrong: https://github.com/camptocamp/ngeo/pull/5621/files#diff-253425453dea4e54ec1f5fde0c68b1d5R1000
Shouldn't it be removed?

llienher:
@sbrunner Nice catch, seems to fix it if we remove it.

sbrunner:
But now this line will not work...
https://github.com/camptocamp/ngeo/pull/5621/files#diff-8ba1e081de38ec354b188acb2fea6454R144

src/print/Service.js
```diff
@@ -163,7 +163,8 @@ PrintService.prototype.createSpec = function(
     rotation: customAttributes.rotation
   });
 
-  this.goodnessOfFit_ = customAttributes['goodnessOfFit'];
+  this.goodnessOfFit_ = customAttributes.goodnessOfFit;
+  delete customAttributes.goodnessOfFit;
```
sbrunner: This is a big hack, it shouldn't be passed throw the customAttributes, but throw a new optional? attribute.

llienher:
@sbrunner I did pass the parameter to the service from the component via an optional parameter in the createSpec method. That's cleaner. :)

PR #5693, opened 2020-04-03
Remove typecheck to show wmts layer name

svareg:
Hello,

When using the "import external WMS" functionality, the layername won't be displayed correctly if a WMTS layer is chosen :

image

This small change should allow the WMTS layername to be displayed instead of just "string".

Thanks,

### DIFF
--- a/contribs/gmf/src/datasource/ExternalDataSourcesManager.js
+++ b/contribs/gmf/src/datasource/ExternalDataSourcesManager.js
@@ -412,8 +412,8 @@ export class ExternalDatSourcesManager {
       dataSource = this.extDataSources_[id];
     } else {
 
-      const name = typeof layer['Title'];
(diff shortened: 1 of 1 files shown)


contribs/gmf/src/datasource/ExternalDataSourcesManager.js
```diff
@@ -412,8 +412,8 @@ export class ExternalDatSourcesManager {
       dataSource = this.extDataSources_[id];
     } else {
 
-      const name = typeof layer['Title'];
-      const wmtsLayer = typeof layer['Identifier'];
+      const name = layer['Title'];
+      const wmtsLayer = layer['Identifier'];
```
sbrunner, line 416: Shouldn't we use the 
`layer.Identifier`
syntax?

llienher:
We did the merge in our own PR. I close this one.

PR #5718, opened 2020-04-14
Fix adding external layer

sbrunner:
Fix: https://jira.camptocamp.com/browse/GSGMF-1259


sbrunner:
I don't understand how the URL can come from, @fredj can you verify that there is not some other thing something broken?

sbrunner:
Merging to be able to do the final tests

PR #5812, opened 2020-05-12
Fix for WMTS layer name

llienher:
Fix: GEO-3141
Port of https://github.com/camptocamp/ngeo/pull/5693


sbrunner:
Shouldn't it be back ported on 2.5 and master?
If yes add the corresponding labels...

sbrunner:
The issue number looks to be wrong...

sbrunner:
=> Updated :-)

PR #5867, opened 2020-05-28
Don't include dragPan interaction twice

fredj:


sbrunner:
And it's working on all tinterfaces ?
api, iframe_api, mobile, desktop?

fredj:
I've just checked: working with all interfaces

sbrunner:
:-)

PR #5868, opened 2020-05-28
Keep style of feature when selected

RenataMuellerC2C:
Fix GSGMF-1278


sbrunner:
@fredj does it look good?

PR #5912, opened 2020-06-09
Fix API overlays with not existing WMTS layers

arnaud-morvan:


sbrunner:
@fredj does it looks good to you? for me yes :-)

PR #5914, opened 2020-06-10
Set higher z-index for map messages

rbovard:
Port of #5070 in 2.4 branch


sbrunner:
@fredj can you review it, thanks :-)

PR #5922, opened 2020-06-11
Data sources without dimension filters are considered equal

adube:
When 2 OGC data sources don't have dimension filters set, they should be considered as having the same filters.

This patch makes it so.

### DIFF
--- a/src/datasource/OGC.js
+++ b/src/datasource/OGC.js
@@ -1166,11 +1166,8 @@ class OGC extends ngeoDatasourceDataSource {
    * @override
    */
   haveTheSameActiveDimensionsFilters(dataSource) {
-    const myConfig = this.dimensionsFiltersConfig;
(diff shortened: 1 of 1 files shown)


adube:
@arnaud-morvan Do you think this fix should go in 2.5 as well?

sbrunner:
@adube If you need that now we can add a 'backport <branch>' label to do that automatically, in any case a pull request is open, and in case of error some instructions will be added in the pull request :-)

arnaud-morvan:
Yes, this can be considered a bug fix so I think it's a good idea to backport this on 2.5.

adube:
> @adube If you need that now we can add a 'backport ' label to do that automatically, in any case a pull request is open, and in case of error some instructions will be added in the pull request :-)

Ohh, that's a nice "automated" feature!  Didn't know about that one.

PR #5934, opened 2020-06-15
Update API's css to match OpenLayers version

fredj:


sbrunner:
Should be rebased ...

fredj:
> Should be rebased ...

done

PR #5992, opened 2020-07-03
Support snapping on features from Draw and Import tools

adube:
This patch introduces the support of snapping to features from the Draw/Measure tool and from the imported vector files (Import Layer / Local).

## What we currently have

Before detailing the changes of this patch, I would like to first describe what we had that handled the Snapping.

In GMF, we have a service called `Snapping` (in contrib/gmf/src/editing).  It hooks with the layer tree root node to get all the tree controlers that have node (i.e. layer node from theme) with the "snappingConfig" metadata.  The nodes that have it are WMS, and support WFS.  When snapping to those, WFS is used to fetch the features.

### Support features from the draw tool

The features from the gmf Draw tool are added in the ngeoFeatures collection, which is also an "angular value" that can be injected from anywhere.  That part was the easy one, as we only needed to add a Snap interaction bound to this features collection and that was it.

### Support features from "Import Layer / Local"

That as the tricky part.  There was nothing that could be hooked onto to get those features.

Introducing: `gmfDatasourceFileGroup`, as an "angular value" that can be injected from anywhere to get the FileGroup data source in which File data sources are added when adding vector files (kml, etc.) using the "Import Layer / Local" tool.

The FileGroup data source is still initialized where it was created, and it also sets it in the `gmfDatasourceFileGroup` "angular value".

Then, in Snapping, we hook on the FileGroup and listen to when File data sources are added/removed in order to bind a Snap interaction for each one of them.  That's pretty much the idea behind this.  Like their WMS counterparts, File data source are registered a watcher in Snapping onto their "visible" property, to make sure that the snapping only occurs on the features of an imported data source only if the data source (and OpenLayers layer) is visible.

## Snaping while editing

In the GSGMF-1292 task, this was asked:

>Not available in the editing

As it turns out, the Snapping service does not currently differenciate when we draw/edit using the Draw or Edit tool.  In other words, even prior this patch, it was possible to "snap" to features on the map while editing or drawing (using both tools).

So, this patch introduces the same logic, i.e. it is possible to "snap" on features that were drawn using the Draw tool or on features that were imported from a file, regardless of the tool we're using (both Draw and Edit).

This behaviour is the opposite of what was asked for the Edit tool, i.e. "Not available in the editing".

**Question**: It would require more efforts to prevent snapping for the edit tool.  Do we need to put efforts into this?

## Todo

 * [x] implement gmfSnappingConfig (only used by features from the Draw tool and from the "Import Layer / Local")
 * [x] fix all issues reported by Travis
 * [x] reply to the **Question** in the "Snapping while editing" section (see above)

### DIFF
--- a/contribs/gmf/src/datasource/fileGroup.js
+++ b/contribs/gmf/src/datasource/fileGroup.js
@@ -1,0 +1,18 @@
+import angular from 'angular';
+
+/**
+ * @type {angular.IModule}
+ * @hidden
+ */
+const module = angular.module('gmfDatasourceFileGroup', []);
+
+/**
+ * @typedef {Object} DatasourceFileGroup
+ * @property {import("ngeo/datasource/FileGroup.js").default|null} fileGroup
+ */
+
+module.value('gmfDatasourceFileGroup', {
+  fileGroup: null,
+});
+
+export default module;
--- a/contribs/gmf/src/datasource/fileGroup.js
+++ b/contribs/gmf/src/datasource/fileGroup.js
@@ -1,3 +1,24 @@
+// The MIT License (MIT)
+//
+// Copyright (c) 2017-2020 Camptocamp SA
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy of
+// this software and associated documentation files (the "Software"), to deal in
+// the Software without restriction, including without limitation the rights to
+// use, copy, modify, merge, publish, distribute, sublicense, and/or sell copies of
+// the Software, and to permit persons to whom the Software is furnished to do so,
+// subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in all
+// copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY, FITNESS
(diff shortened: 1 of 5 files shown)


adube:
@fredj @sbrunner Ready for review.  Please, read the **Question** in the description too.

adube:
@fredj @sbrunner I'm back from my vacations.  This is ready for review.

Please, read the **Question** in the description too.

adube:
Hello @ger-benjamin .  Maybe you'd like to do this review?

contribs/gmf/src/datasource/fileGroup.js
```diff
@@ -0,0 +1,39 @@
...
+ * @hidden
+ */
+const module = angular.module('gmfDatasourceFileGroup', []);
+
+/**
+ * @typedef {Object} DatasourceFileGroup
+ * @property {import("ngeo/datasource/FileGroup.js").default|null} fileGroup
+ */
+
+module.value('gmfDatasourceFileGroup', {
```
ger-benjamin, line 35: Also, can you describe in 1-2 lines what this value is for ? (I know, generally we (sadly) don't describe values but as this file have nothing else than this value, a short description could be helpful)
(And, or add a description in the DatasourceFileGroup typedef)

adube: That's an excellent idea.  I'll do this right away.

adube:
I'll wait for a second "approval" and a second opinion about the question before merging.

ger-benjamin:
@adube Nobody else ? Bah, please merge :-)

adube:
>But please write in the jira ticket that it would take more effort to prevent snapping for edit tools

Will do.

PR #6004, opened 2020-07-08
Better regularShape handling in print

fredj:


fredj:
These changes improves the 'cross' Regular Shape but the result is not very satisfying.
An alternative could be to send the image of the symbol to the print (using `imageStyle.getImage().toDataURL()`)

PR #6009, opened 2020-07-10
Gui for swipe on layertree groups gsgmf 1336

ger-benjamin:
FIX GSGMF-1336

It's not easy to open a popup on the parent node with a link. I give up to achieve that for now.

I add the message `More options at the group level` for layers with parents having a "filter" possibility or a "set opacity" possibility.
Is the message understandable ?

I add this message as first option item in the mobile because the legend must be last.

UPDATE
Tests are failing, but It looks like it's not this PR fault. I've done a fix here: https://github.com/camptocamp/ngeo/pull/6013
And that could be good to have too: https://github.com/camptocamp/ngeo/pull/6002

### DIFF
--- a/contribs/gmf/src/layertree/common.scss
+++ b/contribs/gmf/src/layertree/common.scss
@@ -4,6 +4,11 @@
   padding: $micro-app-margin;
 }
 
+.more-options-info {
+  font-style: italic;
+  color: $color;
+}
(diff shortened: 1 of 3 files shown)


sbrunner:
I think that if you rebase that will fix the CI ...

PR #6011, opened 2020-07-11
Remove snapping icon for mobile

RenataMuellerC2C:


ger-benjamin:
@sbrunner I don't know why it still fails.
Locally, all test fails because of some different pixels on texts probably, so I can't reproduce this precis problem :-/

PR #6013, opened 2020-07-13
Update images with new disclaimer

ger-benjamin:
Tests are failing because of a new disclaimer on the apps.
I've updated the images. it is the good thing to do ?


sbrunner:
Normally I remove it ..., restarted :-)

PR #6026, opened 2020-07-14
Correct scaleselector with constrainResolution to false

ger-benjamin:
For GSGMF-1315

Get zoom give a value that can be a scales index or a value between scales indexes.
With the `map.view.constrainResolution` to true (default), a zoom value thta is not an index is possible, but the map will automatically adapt the zoom the have a defined scale.

With this PR, the app supports having `map.view.constrainResolution` to false. The scale will be calculated from the zoom value.


src/map/scaleselector.js
```diff
@@ -252,6 +252,28 @@ class ScaleselectorController {
...
+    if (scale !== undefined) {
+      return scale;
+    }
+
+    // If zoom is not an exact index of scales, try to determine the current scales from the zoom value.
+    const flooredZoom = Math.floor(zoom);
+    const lowerScale = this.scales[flooredZoom];
+    const upperScale = this.scales[flooredZoom + 1];
+    scale = lowerScale - (lowerScale - upperScale) * (zoom - flooredZoom);
+    return isNaN(scale) ? undefined : Math.round(scale);
```
sbrunner, line 274: Why not
```suggestion
    const inchesPerMeter = 39.37;
    const dpi = 25.4 / 0.28;
    return olView.getResolution() * inchesPerMeter * dpi;
```
Then it will also work with an empty scales :-)
See also: https://github.com/openlayers/openlayers/blob/2dd9342a982e14c068cdd8c6492d2c6f79e8f81f/src/ol/control/ScaleLine.js#L463-L473

ger-benjamin: I prefer my solution, I don't depend on hard-coded value (like dpi - even if these values should not changes) and it's easy to test.
But I can try to adapt the code. Do you want that I change ? I can't do it immediate (I've no time to do it right now).

ger-benjamin: \+ it looks not working as expected, the scale value obtained is more than the double of the real scale. But I've not tried to know why.

sbrunner: I just verify and It's the current values that are completely wrong... => new proposal

src/map/scaleselector.js
```diff
@@ -232,7 +232,7 @@ class ScaleselectorController {
    */
   handleResolutionChange_(e) {
     const view = this.map_.getView();
-    const currentScale = this.scales[/** @type {number} */ (view.getZoom())];
+    const currentScale = this.getCalculateScale(view.getZoom());
```
sbrunner, line 235: Shouldn't we reuse the getScale function?

src/map/scaleselector.js
sbrunner, line 274:
```suggestion
    const inchesPerMeter = 39.37;
    // See: 7.2.4.6.9 Scale denominators in WMS standard: http://portal.opengeospatial.org/files/?artifact_id=14416
    const standardPixelSize = 0.28;
    const dpi = 25.4 / standardPixelSize;
    const calculatedScale = olView.getResolution() * inchesPerMeter * dpi;
    
    // If zoom is an index of scales, return the matching scale.
    const scale = this.scales[zoom];
    if (scale !== undefined) {
      if (Math.exp(Math.abs(Math.log(calculatedScale/scale))) > 1.5)  {
        console.warn(`Big deferance between configured scale '${scale}' and real calculated scale '${calculatedScale}'.`);
      }
      return scale;
    } else {
      return calculatedScale;
    }
```

PR #6065, opened 2020-07-22
Fix travis 2 4

ger-benjamin:
Try to reduce length of logs because travis fail with "The job exceeded the maximum log length, and has been terminated."


ger-benjamin:
I've to merge to test. So, let's try

ger-benjamin:
It doesn't work, I've missed the right command, I've added a commit (directly in 2.4, not using a PR): https://github.com/camptocamp/ngeo/commit/ea2a20e2341a094d8488b4975923fb3a8cfd1db2

PR #6082, opened 2020-07-28
Corrections after release v2.4.2 geo 3403

ger-benjamin:
After: GEO-3403

# commit 1
In upgrade of a GeoMapFish project, the line:
```
module.value('gmfLayertreeOptions', {'legendIcon': {'width': '30', 'height': '30'}});
```
becomes (with two slashes at the end):
```
module.value('gmfLayertreeOptions', {'legendIcon': {'width': '30', 'height': '30'\}\});
```
I try to correct that, but I think the real problem is not in ngeo.

# commit 2
Add an example in desktop_alt in https://github.com/camptocamp/ngeo/pull/5971

### DIFF
--- a/contribs/gmf/apps/desktop_alt/index.html.ejs
+++ b/contribs/gmf/apps/desktop_alt/index.html.ejs
@@ -46,7 +46,8 @@
         <div class="gmf-app-content">
           <gmf-layertree
             gmf-layertree-dimensions="mainCtrl.dimensions"
-            gmf-layertree-map="::mainCtrl.map">
+            gmf-layertree-map="::mainCtrl.map"
+            gmf-layertree-isexpanded="::true">
           </gmf-layertree>
         </div>
       </div>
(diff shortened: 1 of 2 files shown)


contribs/gmf/apps/desktop_alt/index.html.ejs
```diff
@@ -46,7 +46,8 @@
         <div class="gmf-app-content">
           <gmf-layertree
             gmf-layertree-dimensions="mainCtrl.dimensions"
-            gmf-layertree-map="::mainCtrl.map">
+            gmf-layertree-map="::mainCtrl.map"
+            gmf-layertree-isexpanded="::true">
```
sbrunner, line 50: This looks to not working well...

ger-benjamin: I don't see any errors on. What do you means ?

sbrunner: => all the group are open but the icon is on close position
open => close and open effect...

ger-benjamin: Ah, now it's clear.
But what should we do ? The feature is not developed by Camptocamp and it was approved here: https://github.com/camptocamp/ngeo/pull/5971
Who have to debug that ?

sbrunner: Why not keep it disabled?

PR #6083, opened 2020-07-28
Support external WMS in query tool

adube:
For GSGMF-1290.

This patch allows connected (imported) WMS layers to be queried.  It used to work before, but an extra validation about "layers being in range", which requires WFS to work, was preventing the regular WMS layers to work.

Also, this patch includes a note in the ngeo-query-panel that is now shown for each possible action: click, drawbox, drawpolygon.  This informs the user about limitations when using a box or polygon (i.e. only queryable layers that support WFS will return results).

## How to test

 * run a desktop app (I like the desktop_alt one)
 * remove all layers from the layer tree and switch base map to "blank"
 * use the "Import Layer" tool to the left:
  * choose "Swiss Topo WMS" server
  * search for "dégâts"
  * add the layer "Carte des dégâts aux forêts" to the map
  * click on a cluster of points on the map
  * --> results should be shown at the bottom


adube:
>Which layer do you use to test ?

I've updated the description with explanations how to test this feature.

>Also do you have done something for (comment in the jira ticket):

Not as part of this patch, no.  At least not yet.

How about we add informative text here:

Screenshot_2020-07-29_08-03-48

So, while the "query with click" is enabled, the text would explain that all queryable layers will be queried, with both WMS and WFS.  While the "query by box" or "by polygon" is enabled, then only the layers that support WFS will be queried.

adube:
@ger-benjamin I updated the patch to include info text for each action.

adube:
I don't understand why there a Continuous integration test that fails.  Any hint would be appreciated.

ger-benjamin:
The `make check-examples` target take a screenshot of every current apps and examples. Then compare them to references images.
Perhaps the reference image must be updated with your new changes.

... But I can't run `make check-examples` on my side. I always get errors because the font are slightly different... So I can't help you.
If you can run `make check-examples` and it only fail at this PR's related examples. Please update the references images of related examples. Then it should passes

adube:
>If you can run make check-examples and it only fail at this PR's related examples. Please update the references images of related examples. Then it should passes

Running `make check-examples` immediately throws an error on my side, so I guess that I can't do anything about it either.

You've accepted the PR, but I suppose I can't merge yet, right?

Thanks.

ger-benjamin:
Yes you have to wait on @sbrunner .
Or maybe@fredj or @arnaud-morvan know something ?

arnaud-morvan:
Hello @adube.

I also get an error with .build/examples-hosted/contribs/gmf/apps/mobile.html.png so I didn't reach locally the test that fails on CI.

But note that you can download produced images on CI in GitHub Action: https://github.com/camptocamp/ngeo/pull/6083/checks?check_run_id=924766307
Click on "Artifact" button, next to "Re-run jobs" on the right above log panel.

Then you can get, in this archive, the image produce by CI for the considered test: `query.html.png`
Note that the diff image is also available (differences are reprensented with blue color), please take a look before replacing the expected image.

Then you could replace the expected image in your branch: `examples/query-ref.png`

And hope this will fix the CI for that test.

adube:
@arnaud-morvan I worked!  Thanks for the explanation.  I wasn't aware of this new type of CI using images. That's pretty cool.

Please, let me know if I can merge this now.

PR #6104, opened 2020-08-02
Client-side validation

RenataMuellerC2C:


RenataMuellerC2C:
This branch contains:
- disabled save button if there are invalid fields
- red border on invalid fields but only if they are touched (Don't have all fields in red when we create a new field)
- only a work-around for allow only numbers on number fields in other browsers than Chrome
Open:
- Does the behaviour correspond to the wanted one?
- What is the right place for the css, which is now in editing.css?
- Better solution for number input or test for Safari/Edge
- missing features?
- ci-errors

PR #6141, opened 2020-08-18
Don't use the pixelRatio to create the mask

fredj:


fredj:
It works with the Chrome emulation but needs to be tested on a real high dpi device

PR #6155, opened 2020-08-21
Add the ngeo options in the documentation 

sbrunner:

PR #6158, opened 2020-08-21
Add the gmf options in the documentation 

sbrunner:
Same as  #6155 but for GMF


ger-benjamin (approved):
As far as I know, it looks good

PR #6173, opened 2020-08-25
Fix controls3d 2.5

gberaudo:


sbrunner:
Looks like that the layer order ref image should be regenerated ...

llienher (approved):

PR #6184, opened 2020-08-26
Use constant gmfDisplayQueryGridOptions

sbrunner:


contribs/gmf/src/options.js
```diff
@@ -157,4 +162,97 @@
...
+ * @property {Stroke} [stroke] The stoke config.
+ * @property {Circle} [circle] The circle config.
+ * @property {number} [zIndex] The z index.
+ */
+
+/**
+ * @param {olStyle|Style} styleDescriptor The description of the style
+ * @returns {olStyle}
+ */
+export function buildStyle(styleDescriptor) {
```
ger-benjamin, line 214: @fredj Is there nothing to do that in ol ? Do you know ?

ger-benjamin (approved):
Looks good

contribs/gmf/src/options.js
fredj: non, nothing like that in OpenLayers

ger-benjamin: Thanks (could be an useful tool)

PR #6192, opened 2020-08-27
Use option gmfLayerTreeOptions

sbrunner:


contribs/gmf/src/options.js
```diff
@@ -376,4 +376,11 @@ export function buildStyle(styleDescriptor) {
  * @property {Style} sketchStyle A style for the measure point.
  */
 
+/**
+ * @typedef {Object} gmfLayerTreeOptions
+ * @property {boolean} openlinksinnewwindow if true, open
+ *     metadataURLs in a new window. Otherwise open them in a popup.
+ * @property {boolean} isExpanded Is expanded.
```
ger-benjamin, line 383: Can you remove the comment at: https://github.com/camptocamp/ngeo/pull/6192/files#diff-c27c1737f28b221201d1fd6a0c4b8061R184

And move it here ? That's currently not correct. Thanks

ger-benjamin (approved):

contribs/gmf/src/options.js
sbrunner: Thanks for the precision, fixed :-)

PR #6194, opened 2020-08-27
Use new style descrition in the permalink

sbrunner:


contribs/gmf/src/options.js
```diff
@@ -48,28 +48,10 @@ import {createDefaultStyle} from 'ol/style/Style.js';
...
- * @property {number} [rotation=0] Rotation in degree (positive rotation clockwise).
- * @property {boolean} [rotateWithView=false] Whether to rotate the shape with the view.
- */
-
 /**
  * Configuration options for the permalink service.
- * @typedef {Object} gmfPermalinkOptions
- * @property {RegularShapeOptions[]} [crosshairStyle] An alternate
+ * @typedef {object} gmfPermalinkOptions
+ * @property {StyleLike} [crosshairStyle] An alternate
```
ger-benjamin, line 54: I think you can write this comment on a single row

ger-benjamin (approved):
Looks good but not tested (I'm not sure how to test that
)

contribs/gmf/src/options.js
sbrunner: No, it just misses a little...

PR #6216, opened 2020-09-01
Again more delay to wait that the effect ends

sbrunner:


ger-benjamin (approved):
If there is a 3th PR about this, I will suggest you using another strategy... :-)
LGTM

PR #6223, opened 2020-09-02
Fix duplicated STYLES parameter in print requests

asaunier:
As for now STYLES param is passed in the print requests in the ``customParams`` object. This leads to having a duplicated STYLES parameter in the GetMap requests sent by the print tool.
It seems no problem with Mapserver but causes a "Can't parse XML request." error with ESRI.


sbrunner:
Looks good but the tests are failing ...

asaunier:
@sbrunner Fixed by a414d6c
I will squash the fixup commit soon.

asaunier:
@sbrunner Commits squashed and tests passed :)

sbrunner (approved):
Thanks :-)

PR #6226, opened 2020-09-02
Clean routing gsgmf 1361

ger-benjamin:
Some cleaning in the routing component I found during adding per error arrows to this component (see #6225 ).
I keep the cleaning with this PR.


sbrunner (approved):
Looks good :-)

PR #6234, opened 2020-09-04
Add arrow possibilities to line drawing

ger-benjamin:
GSGMF-1361

Add arrow lines of drawing tools (only).

Still TODO: be able to print the arrows
--> Printing routing is also not working

### DIFF
--- a/contribs/gmf/src/drawing/drawFeatureOptionsComponent.js
+++ b/contribs/gmf/src/drawing/drawFeatureOptionsComponent.js
@@ -120,7 +120,7 @@ class DrawFeatureOptionsController {
     this.length = NaN;
 
     /**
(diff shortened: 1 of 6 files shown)


sbrunner:
It should be aligned with « Stroke », except that it looks good :-)

ger-benjamin:
I've created a ticket for the alignement (draw, edit...): GSGMF-1382
And one for the print: GSGMF-1381

Review is ready, except for these two points

sbrunner (approved):

ger-benjamin:
I've seen right now that if you export the line to kml or gpx, and then import it, you will not have the arrrow on the layer. That's the same as for texts on line. They are exported but not read.

PR #6237, opened 2020-09-04
Fix the query window (nothing was displayed)

sbrunner:


src/statemanager/WfsPermalink.js
```diff
@@ -41,6 +41,7 @@ import olFormatWFS from 'ol/format/WFS.js';
  *    configured limit.
  * @property {number} [totalFeatureCount] If `tooManyResults` is `true`, this contains the total number of
  *    features.
+ * @property {string} [identifierAttributeField] The fild to use to identify the feature.
```
ger-benjamin, line 44:
```suggestion
 * @property {string} [identifierAttributeField] The field to use to identify the feature.
```

contribs/gmf/src/query/window.scss
```diff
@@ -11,8 +11,8 @@ div.gmf-displayquerywindow {
   button.collapse-button {
     pointer-events: all;
   }
-  .details {
-    display: none;
+  div.animation-container.next {
```
ger-benjamin, line 14: `div.animation-container` or `div.animation-container {&.ext, &.previous {...}}`
Because if you click on "previous" results disappears

ger-benjamin (changes requested):

src/statemanager/WfsPermalink.js
sbrunner: :-)

contribs/gmf/src/query/window.scss
sbrunner: OK, thaks :-)

ger-benjamin (approved):
Should be good (demo not updated - not tested)

PR #6241, opened 2020-09-08
Fix error TS2341: Property 'getArrowLineStyles_' is private and only …

sbrunner:
…accessible within class 'FeatureHelper'.


ger-benjamin (approved):
Sorry, thanks

sbrunner:
No problème :-)

PR #6257, opened 2020-09-14
Print style at coordinate

ger-benjamin:
GSGMF-1381

 - First commit create a new feature for each style with a geometry (otherwise the style is always at the center of the whole feature, resulting as an issue on, for instance, line with an arrow at start or end)
 - Second commit replace FontAwesome (FA) icons to html entities to be able to print (MapFishPrint sadly didn't load FA)
 - Third commit change the encoding to keep geometry on sfeatures sharing the same style (we are now able to priint things like the routing line)
 - Fourth add tests (it doesn't cover 100%, but that already something)

For the second point it's not perfect. We need an offset to align the real point with the edge of the symbol. Sadly, it looks than the print set the offset before the rotation, and OL set the rotation before the offset. That make a small visual issue.
That was less the case with FA because their icons have a better "centroid"


ger-benjamin:
Edit: Already re-viewable. But I'll continue to ad test for the EncodeVector

sbrunner:
Shouldn't this pull request be rebased?

ger-benjamin:
After more than three weeks to ask for review, yes obviously

sbrunner (approved):

PR #6260, opened 2020-09-16
Add option to trigger geolocation at loading

asaunier:
May be tested in https://localhost:3000/contribs/gmf/apps/mobile_alt.html

When the desktop_alt app has loaded, the user is directly prompted to allow (or not) to be geolocated. If they accept, the screen looks like:
Screenshot from 2020-10-02 15-29-45

Tested with Firefox and Chromium

Related to https://jira.camptocamp.com/browse/GEO-3577


examples/mobilegeolocation.html
```diff
@@ -12,6 +12,7 @@
       <div id="map" ngeo-map="::ctrl.map"></div>
       <button ngeo-geolocation=""
         ngeo-geolocation-map="::ctrl.map"
+        ngeo-geolocation-mainctrl="::ctrl"
```
sbrunner, line 15: For me, you don't need the full main controller =>
```suggestion
        ngeo-geolocation-loading="::ctrl.loading"
```

asaunier: I agree, it would make more sense.
But I tried to do so (in GMF 2.4 or maybe 2.5) and the ``loading`` property value was passed once for all and never updated.
Could it be different in master? Or perhaps the attribute syntax is not correct? I had tried removing the one-time binding ``::`` but with no more success.

By the way would you have a suggestion where this functionality should be demonstrated? The current ``mobilegeolocation`` has a very simple main controller with no ``loading`` attribute...

sbrunner: I just remove the one-time binding from the suggestion :-)
@fredj @ger-benjamin any input?

src/options.js
```diff
@@ -339,6 +339,7 @@ export function buildStyle(styleDescriptor) {
  * @property {number} [zoom] If set, in addition to recentering the map view at the location, determines
  * the zoom level to set when obtaining a new position
  * @property {boolean} [autorotate] Autorotate.
+ * @property {boolean} [atloadingtime] Toggle geolocating at loading time if true.
```
sbrunner, line 342:
```suggestion
 * @property {boolean} [atloadingtime=false] Toggle geolocating at loading time if true.
```

examples/mobilegeolocation.html
fredj: it will not work as it because the scope is not bound to the controller.
- add `bindToController: true` in `geolocationComponent`
- add `$onInit` function to the controller and move the `getMapFn` function call
- handle `loading` value change in `$onInit`

sbrunner: Then @fredj what do you preconize to do?
Keep like it or bindToController, ...?

asaunier:
@fredj Thanks for your tips!
I have updated the PR in order to use the ``bindToController`` approach you suggested.

As a result, the ``map`` property passed to the directive is also impacted and I had to move most of the map related code from the constructor to the ``$onInit`` method. I have only kept the injected objects and their local representations in the constructor.
Looks good to you?

Thanks!

asaunier:
There is a CI error:
Screenshot from 2020-10-02 15-46-45

Is it really related to the suggested changes?

asaunier:
I see also the following error. Is there something to update in the tests?

Screenshot from 2020-10-02 15-53-19

src/geolocation/component.js
```diff
@@ -162,20 +182,20 @@ function Controller(
         let msg;
         switch (error.code) {
           case 1:
-            msg = gettextCatalog.getString('User denied the request for Geolocation.');
+            msg = this.gettextCatalog_.getString('User denied the request for Geolocation.');
```
sbrunner, line 185: This will not work for the string collection...

src/geolocation/component.js
```diff
@@ -110,37 +112,55 @@ function Controller(
...
+  /**
+   * @type {angular.gettext.gettextCatalog}
+   */
+  this.gettextCatalog_ = gettextCatalog;
+
   /**
    * @type {import("ol/Map.js").default}
    */
-  this.map_ = map;
+  this.map;
```
sbrunner, line 138: Why removing the _?

src/geolocation/component.js
```diff
@@ -230,10 +250,21 @@ function Controller(
     }
   );
 
-  const view = map.getView();
+  const view = this.map.getView();
 
   listen(view, 'change:center', this.handleViewChange_, this);
   listen(view, 'change:resolution', this.handleViewChange_, this);
+
+  if (this.options.atloadingtime && typeof this.loading != 'undefined') {
```
sbrunner, line 258: `'undefined'` shouldn't be `undefined` (without quote)?

sbrunner:
Don't worry if an upload-artifacts filed after a failing task :-)

sbrunner:
You can download the "Examples hosted" artifact to see what's wrong with the examples checks.

src/geolocation/component.js
asaunier: Because with ``bindToController: true``, the ``map``` attribute of the directive defined at https://github.com/camptocamp/ngeo/blob/bf6fdd7fdab4e56f8adcf73099ffc9481fbaceea/src/geolocation/component.js#L77 is automatically set to ``this.map``. Not ``this.map_``.

src/geolocation/component.js
asaunier: The string collection only works with ``gettextCatalog``?
This variable is not available outside of the constructor... Unless if we create an alias  this way:
```
const gettextCatalog = this.gettextCatalog_;
```
Do I really have to do so?

src/geolocation/component.js
asaunier: Do you mean
```
if (this.options.atloadingtime && this.loading !== undefined) {
```
?

asaunier:
@sbrunner Thanks for your review and tips!

contribs/gmf/apps/mobile_alt/Controller.js
```diff
@@ -58,4 +58,21 @@ const module = angular.module('Appmobile_alt', [appBase.name, gmfControllersAbst
...
+    },
+  },
+  accuracyFeatureStyle: {
+    fill: {color: 'rgba(100, 100, 230, 0.3)'},
+    stroke: {color: 'rgba(40, 40, 230, 1)', width: 2},
+  },
+  zoom: 8,
+  autorotate: true,
+  atloadingtime: true,
+});
```
sbrunner, line 76: This should be defined here: https://github.com/camptocamp/demo_geomapfish/commit/8459afe6c805f4ba272ce02a12e7618a0c553a1e

src/geolocation/component.js
sbrunner: +1 Yes :-)

contribs/gmf/apps/mobile_alt/Controller.js
asaunier: @sbrunner There is no ``vars.yaml`` file in the ngeo repo: how should I proceed then?

sbrunner: No, I do it on the demo, then it's done :-)

src/options.js
```diff
@@ -340,6 +340,7 @@ export function buildStyle(styleDescriptor) {
  * @property {number} [zoom] If set, in addition to recentering the map view at the location, determines
  * the zoom level to set when obtaining a new position
  * @property {boolean} [autorotate] Autorotate.
+ * @property {boolean} [atloadingtime=false] Toggle geolocating at loading time if true.
```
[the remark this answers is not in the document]
asaunier: Yes. I will change this but you'll have to update it in https://github.com/camptocamp/demo_geomapfish/commit/8459afe6c805f4ba272ce02a12e7618a0c553a1e as well

asaunier:
> Don't worry if an upload-artifacts filed after a failing task :-)
> You can download the "Examples hosted" artifact to see what's wrong with the examples checks.

@sbrunner I have done so but have no clue what to look for. Are artifacts/hosted examples documented anywhere?

> Screenshot saved at: .build/examples-hosted/contribs/gmf/apps/mobile_alt.html.png
[ "$(gm compare -metric RMSE -highlight-style xor .build/examples-hosted/contribs/gmf/apps/mobile_alt.html.png contribs/gmf/apps/mobile_alt-ref.png -file .build/examples-hosted/contribs/gmf/apps/mobile_alt-diff.png 2>&1 | tail --lines=1 | sed 's/.* \([0-9\.]\+\) .*/\1/g')" \< 0.005 ]
Makefile:346: recipe for target '.build/contribs/gmf/apps/mobile_alt.check.timestamp' failed
make: *** [.build/contribs/gmf/apps/mobile_alt.check.timestamp] Error 1

Does this mean a screenshot of the mobile_alt UI is done at some point but no longer match with the expectations as there is now a geolocation prompt window showing up?

sbrunner:
> @sbrunner I have done so but have no clue what to look for. Are artifacts/hosted examples documented anywhere?

Yes: https://docs.github.com/en/free-pro-team@latest/actions/guides/storing-workflow-data-as-artifacts

> Does this mean a screenshot of the mobile_alt UI is done at some point but no longer match with the expectations as there is now a geolocation prompt window showing up?

By seeing the generated image and the diff image in the artifact, yes :-)

asaunier:
> By seeing the generated image and the diff image in the artifact, yes :-)

OK Got it! Thanks!

The currently expected image:
mobile_alt-ref

The one generated with the PR changes:
mobile_alt html

Commit e54a9a2 push the second instead of the first.

contribs/gmf/apps/mobile_alt/Controller.js
```diff
@@ -58,4 +58,21 @@ const module = angular.module('Appmobile_alt', [appBase.name, gmfControllersAbst
...
+    },
+  },
+  accuracyFeatureStyle: {
+    fill: {color: 'rgba(100, 100, 230, 0.3)'},
+    stroke: {color: 'rgba(40, 40, 230, 1)', width: 2},
+  },
+  zoom: 8,
+  autorotate: true,
+  atLoadingTime: true,
+});
```
sbrunner, line 76: Steel should be removed...

src/geolocation/component.js
sbrunner: > The string collection only works with gettextCatalog?

Yes

> Do I really have to do so?

Yes

sbrunner (approved):
Just some little comments :-)

contribs/gmf/apps/mobile_alt/Controller.js
asaunier: I don't understand where ``atLoadingTime: true`` then?

sbrunner: Here: https://github.com/camptocamp/demo_geomapfish/blob/prod-2-6/geoportal/vars.yaml#L392

asaunier:
@sbrunner I have removed the config as you suggested it, since it's done in demo_geomapfish.
By the way I notice that some parameters are somehow duplicated https://github.com/camptocamp/demo_geomapfish/blob/prod-2-6/geoportal/vars.yaml#L389-L392 => is it on purpose to support casing that have changed in the meantime (eg. "atloadingtime" => "atLoadingTime")?

I have also squashed the fixups.

@fredj OK for you?

asaunier:
@sbrunner A question/remark though: with the current settings in the demo_geomapfish, when the geolocation is done, the circle diameter is too big and is not fully viewable with the zoom value used in combination, I think?!

sbrunner:
@asaunier The circle diameter is related to the localization precision, then the thing we can do is to zoom out on low precision, but it's another point...

asaunier:
@sbrunner OK, thanks for explaining!

PR #6265, opened 2020-09-17
[Fix master] Replace ls package with an internal function

sbrunner:


buildtools/ls.js
```diff
@@ -0,0 +1,62 @@
+// The MIT License (MIT)
+//
+// Copyright (c) 2020 Camptocamp SA
```
fredj, line 3: if the code is from the `ls` package, the copyright and license must be from the project

sbrunner: But the project is in copy left: https://github.com/awnist/ls/blob/master/UNLICENSE
And our CI requires to have a copyright header...

arnaud-morvan (approved):
Thanks @sbrunner, please merge ASAP.
It might be possible to create a dedicated folder for third party libraries excluded from copyright check, but it seems not stricly necessary in this case.

PR #6270, opened 2020-09-18
Add missing dependency required by ls

sbrunner:


arnaud-morvan (approved):
Thanks @sbrunner 
I do not really understand why gmf app bundle did not fail in ngeo CI before this.

sbrunner:
Good question effectively ...

PR #6284, opened 2020-09-22
NgeoTilesPreloadingLimit, remove default value, fix null value

sbrunner:


src/map/LayerHelper.js
```diff
@@ -62,7 +62,7 @@ export function LayerHelper($q, $http, ngeoTilesPreloadingLimit) {
    * The Tiles Preloading Limit value
    * @type {number}
    */
-  this.tilesPreloadingLimit_ = ngeoTilesPreloadingLimit ? ngeoTilesPreloadingLimit : Infinity;
+  this.tilesPreloadingLimit_ = ngeoTilesPreloadingLimit === null ? ngeoTilesPreloadingLimit : Infinity;
```
ger-benjamin, line 65: Mhhh, so if the value is null, we use null, otherwise Infinity ? It is not better with:
```suggestion
  this.tilesPreloadingLimit_ = ngeoTilesPreloadingLimit === null ? Infinity : ngeoTilesPreloadingLimit;
```

ger-benjamin (commented):
Is there any description for theses constants ? (`null` means... `#` means...)

src/map/LayerHelper.js
sbrunner: +1 Thanks :-)

ger-benjamin (approved):

PR #6290, opened 2020-09-22
List of WMS/WMTS services when setting focus on text box

RenataMuellerC2C:
Fix GSGMF-1372


sbrunner:
This should be on branch 2.5 :-)

sbrunner (approved):
@ybolognini do you agree with that?

PR #6316, opened 2020-10-05
Be less tolerant with the possibility to have a mask

sbrunner:


buildtools/check-example
```diff
@@ -0,0 +1,49 @@
+#!/usr/bin/env python3
+
+import os
+import subprocess
+import sys
+import tempfile
+
+
+def main():
```
arnaud-morvan, line 9: Would be helpful to have a short description of what does this script do and an example usage.
A idea about how to create a mask, some example using CLI ?

buildtools/check-example
```diff
@@ -0,0 +1,49 @@
...
+    diff = current.replace(".html.png", "-diff.png")
+
+    if os.path.exists(mask):
+        print("Apply mask")
+        old_current = current
+        old_ref = ref
+        current = tempfile.NamedTemporaryFile(suffix="-current.png").name
+        ref = tempfile.NamedTemporaryFile(suffix="-ref.png").name
+        subprocess.check_call(
+            ["gm", "composite", "-compose", "Multiply", old_current, main, current]
```
arnaud-morvan, line 22: Shouldn't it be "mask" instead of "main" here ?
Same on next line.

arnaud-morvan (approved):
Seems a good move.

buildtools/check-example
sbrunner: +1 Thanks :-)

buildtools/check-example
sbrunner: Just done :-)

buildtools/check-example
```diff
@@ -0,0 +1,62 @@
...
+import sys
+import tempfile
+
+
+def main():
+    """
+    Compare similarity of tow images, we can use an optional mash file.
+    On error a diff image is generated.
+
+    Usage: check-examle <generated image> <referanc image>
```
arnaud-morvan, line 14:
```suggestion
    Usage: check-example <generated image> <reference image>
```

sbrunner: Thanks :-)

PR #6356, opened 2020-10-13
Server logout message

RenataMuellerC2C:


RenataMuellerC2C:
Are the condition for disconnection right?
Should the message be different if the request returns an error?
Should the previous username be set (like it is done now)?

sbrunner:
For me, the condition should be:
```
this.user_.username && this.user_.username !== resp.data.username 
```
> Should the message be different if the request returns an error?

For me, we should just log an error message...

> Should the previous username be set (like it is done now)?

I don't understand this question...

contribs/gmf/src/authentication/Service.js
```diff
@@ -168,6 +176,44 @@ export class AuthenticationService extends olEventsEventTarget {
...
+   * Check whether the user is connected or not like on load.
+   * @private
+   */
+  checkConnection_() {
+    if (!this.connectionLost_) {
+      const url = `${this.baseUrl_}/${RouteSuffix.IS_LOGGED_IN}`;
+      this.$http_
+        .get(url, {withCredentials: true})
+        .then((resp) => {
+          if (this.user_.username && this.user_.username !== resp.data.username) {
```
sbrunner, line 195: checkConnection_ soulf be arround the $http.get

sbrunner: on place of connectionLost_ that can be removed :-)

sbrunner (approved):

PR #6368, opened 2020-10-19
Metadata popup too big on mobile

RenataMuellerC2C:


ger-benjamin (approved):
Thanks.
It looks good for the popup from osm_open, but not so good for Theme "edition"-> layer "point" (click on the `i` of the layer in the layertree to open the popup). This popup is too small.
Can you take a look ? but if that's looks too complicated to change, perhaps you have to speak with the team to know if it's worth it to fix it.

contribs/gmf/src/controllers/AbstractMobileController.js
```diff
@@ -79,6 +79,16 @@ export class AbstractMobileController extends AbstractAppController {
...
 
+    /**
+     * @type {?string}
+     */
+    this.displaywindowHeight = '100vh';
+
+    /**
+     * @type {?string}
+     */
+    this.displaywindowWidth = '100vh';
```
ger-benjamin, line 90:
```suggestion
    this.displaywindowWidth = '100vw';
```
is better no ?

ger-benjamin (approved):
One comment, otherwise perfect it's now

contribs/gmf/src/controllers/AbstractMobileController.js
sbrunner: +1

PR #6369, opened 2020-10-19
Be able to customise the colors with css only

sbrunner:
Fix: GSGMF-1385
Desktop alt work with: https://github.com/camptocamp/demo_geomapfish/blob/prod-2-6/geoportal/vars.yaml#L169-L173


contribs/gmf/src/sass/vars.scss
```diff
@@ -3,3 +3,37 @@
 @import "~bootstrap/scss/_functions.scss";
 @import "~bootstrap/scss/_variables.scss";
 @import "~@fortawesome/fontawesome-free/scss/_variables.scss";
+
+
+$table-border-color: $border-color !default;
+
+:root {
+  --brand-primary: #{$brand-primary};
```
ger-benjamin, line 11: missing at least:
`--danger`
`--back-color`

(also in vars_only file)

ger-benjamin (approved):
Was it not possible to simply set variables from a provided (or customised) scss file ?

contribs/gmf/src/sass/vars.scss
sbrunner: Not in vars only because they are defined by bootstrap, thanks :-)

PR #6397, opened 2020-10-26
Fix Firefox ctrl key

ger-benjamin:
Fix point 3 of GSGMF-1405

I'll also provide a fix (at least for discussions) in ol. => https://github.com/openlayers/openlayers/pull/11693

`evt.ctrlKey` sometimes don't work on firefox (You can test on: https://codepen.io/ger-benjamin/pen/ZEOeyRb). I switch to use the W3C Modifier Keys Attributes value to detect ctrl key


sbrunner (approved):
Relay nice thanks :-)

PR #6401, opened 2020-10-27
Fix selectObject in API

ger-benjamin:
Fix GSGMF-1404

Debug 'selectObject' example in api (success callback was applied on the wrong map)


sbrunner (approved):
Thanks :-)

PR #6432, opened 2020-11-09
Use Font Awesome for routing and drawing

ger-benjamin:
For GSGMF-1416

Can't currently be be printed. Wait on GSGMF-1391


ger-benjamin:
@sbrunner It still didn't work :-/
I've updated the PR, I hope the CI passes now.

ger-benjamin:
@sbrunner Works fine now. In the vectorEncoder file, I remove the quotes on the font-family before to send it to the print. (we need the quotes in the font object of OpenLayers)

sbrunner (approved):
Nice thanks :-)

PR #6438, opened 2020-11-10
Fix theme selectior css

arnaud-morvan:
May be a good practice to always put the sass variable as default value in var() instead of duplicting the rule.
Seems that var parameter declaration-value is there exactly for that.

### DIFF
--- a/contribs/gmf/src/controllers/desktop.scss
+++ b/contribs/gmf/src/controllers/desktop.scss
@@ -300,7 +300,7 @@ gmf-search {
 
 gmf-themeselector {
   width: $half-app-margin * 2 + $theme-selector-columns * $theme-selector-column-width;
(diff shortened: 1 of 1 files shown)


ger-benjamin (approved):
+1, there is a lot occurrence (probably harmless) of these "double-definition" that we should replace with this use of the var.

sbrunner:
The first line is for IE then it shouldn't be removed...

sbrunner (changes requested):
There is an issue that --theme-selector-columns is not always defined?

arnaud-morvan:
@sbrunner:
Yes theme selector is broken on GMF 2.6 desktop app.
Seems that calc, with calls to `var()`, with not existing properties, does not return an empty value but a wrong value.
By adding `declaration-value` (default) parameterto var calls, it return the good value.

You're right, this PR breaks theme selector layout on IE.
I've restored the lines without call to var for IE.

PR #6445, opened 2020-11-13
Click tolerence per layer

ger-benjamin:
For GSGMF-1294

### Variation from the specs
I've named the `queryTouchMin` option to `toleranceTouch` to be coherant with the `tolerance` for desktop and `tolerance`for the edition (which is use basically for the the same behavior).
I've not seen any `queryToleranceTouch` to remove.

### OpenLayers:
Done, PR here : https://github.com/openlayers/openlayers/pull/11740
**We have to wait on a new ol realease**

### C2cgeoportal:
It looks already done: https://github.com/camptocamp/c2cgeoportal/pull/6512/files
There is no `tolerance` or `toleranceTouch` in ngeoQueryOptions from constants of CONST_vars.yaml . And none is needed. A default value exists in ngeo (the same default value as before).

### Ngeo:
I've take care of: https://github.com/camptocamp/ngeo/pull/4570 (still and always send the simple click with tolerance bbox as GET parameter. A custom bbox can be done at the mapserver side, and I won't sent one bbox per layer).

### DIFF
--- a/src/query/Querent.js
+++ b/src/query/Querent.js
@@ -85,9 +85,8 @@ import olSourceImageWMS from 'ol/source/ImageWMS.js';
  * @property {QueryableDataSources} [queryableDataSources] A hash of queryable data sources, which must meet
  *    all requirements. The querent service requires either the `dataSources` or `queryableDataSources`
  *    property to be set.
- *
- * @property {number} [tolerancePx] A tolerance value in pixels used to create an extent from a coordinate
- *    to issue WFS requests.
+ * @property {number} [tolerance] A minimal buffer value in pixels to ensure a minimal bbox around a
+ *    coordinate to issue WFS requests.
  * @property {boolean} [wfsCount] When set, before making WFS GetFeature requests to fetch features,
  *    WFS GetFeature requests with `resultType = 'hits'` are made first. If
  *    the number of records for the request would exceed the limit, then
@@ -565,10 +564,9 @@ export class Querent {
     let bbox;
     const coordinate = options.coordinate;
     if (coordinate) {
-      const tolerancePx = options.tolerancePx;
-      console.assert(tolerancePx);
-      const tolerance = tolerancePx * resolution;
-      bbox = olExtent.buffer(olExtent.createOrUpdateFromCoordinate(coordinate), tolerance);
+      const tolerance = options.tolerance;
+      console.assert(tolerance);
+      bbox = olExtent.buffer(olExtent.createOrUpdateFromCoordinate(coordinate), tolerance * resolution);
     } else {
       bbox = options.extent;
     }
@@ -576,7 +574,7 @@ export class Querent {
     // (2) Launch one request per combinaison of data sources
     const wfsFormat = new olFormatWFS();
     const xmlSerializer = new XMLSerializer();
-    let hasOneQueryIconPosition = false;
+    let hasAtLeastOneQueryIconPosition = false;
     for (const dataSources of combinedDataSources) {
       /** @type {?import('ol/format/WFS.js').WriteGetFeatureOptions} */
       let getFeatureCommonOptions = null;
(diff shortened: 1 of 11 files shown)


sbrunner:
Looks good but the CI should pass to make it possible to test it on pages :-)

ger-benjamin:
So, we have to wait on an OpenLayers release.
If I target a commit in the `package.json`, I get some errors with the dist.

src/query/Querent.js
```diff
@@ -771,6 +798,38 @@ export class Querent {
...
+   * @return {!import("ol/extent.js").Extent} The extended bbox or null if the queryIconPosition param
+   * is not valid.
+   * @private
+   */
+  extendBboxWithQueryIconPosition_(queryIconPosition, resolution, bbox) {
+    const buffers = queryIconPosition.map((value) => value * resolution);
+    const length = buffers.length;
+    if (!length || length > 4) {
+      // Bad format.
+      return null;
```
arnaud-morvan, line 815: Might log something here.
What if lengh == 3 ? We will get an bad index error some lines below

src/query/Querent.js
```diff
@@ -771,6 +798,38 @@ export class Querent {
...
+  /**
+   * Extend the given bbox with the queryIconPosition values.
+   * @param {!number[]} queryIconPosition The values in px to buffer the bbox.
+   * @param {number} resolution The map view resolution to define the px size correctly.
+   * @param {!import("ol/extent.js").Extent} bbox The bbox to extend.
+   * @return {!import("ol/extent.js").Extent} The extended bbox or null if the queryIconPosition param
+   * is not valid.
+   * @private
+   */
+  extendBboxWithQueryIconPosition_(queryIconPosition, resolution, bbox) {
```
arnaud-morvan, line 810: Here tolerance is not a min bbox size, but a buffer around queryIconPosition.
The ticket says : A new config will be added queryTouchMin for the min box size for touch devices.
This might create a big bbox with a queryIconPosition + 10px on each side.

Did you reversed the QueryIconPosition somewhere ?
The ticket says : The queryIconPosition specify the positions of the icon => inverse to get the query position.

We do not want the IconPosition bbox, but the bbox to query the real geometry as we now that the use click somewhere in the icon.

=> If the center of the icon center is above the geometry, we need to query below the click. Same for left and right.

src/query/Querent.js
ger-benjamin: I've added more explicit comments (case length === 3 is case length !==2 and length !==4).
Please take a look at my unit test if you doubt.
I would avoid the log here but I do an assert out of this function to check if the result is null.

src/query/Querent.js
ger-benjamin: I've renamed this function to bufferBboxWithQueryIconPosition_ which is more correct

ger-benjamin:
What's left is the ol update and this "invert position" requirement.
But I still don't understand this requirement.

First, what do you means by `invert` ? In the context, I understand `flip vertically`. it is correct ?
And by, `The queryIconPosition specify the positions of the icon` you means that the center of the `icon` is the center (centroid) of the queryIconPosition ? Right ?

Then, okay, the user knows the size of its icon. And he wants to apply the queryIconPosition from the center of this icon. So in this case, I've to `shift` (and not `flip vertically`) the queryIconPosition from the point (queryable point). And the amount of pixel to shift would be the computed bottom value of the queryIconPosition ?
It is correct ?

IMO That's complicated for the user. It's easier and more flexible to build the bounding box from the geometry point. But the specs are done, let me know what I've to do. 

I don't see a scenario where to `flip` is the solution. (even if  by `invert` you means flip horizontally, or reverse the bbox).

arnaud-morvan:
Invert queryIconPosition : flip vertically and horizontally (invert bottom and top and invert left and right).
That is, the queryIconPosition should represent the box relative to the geometry (geometry => icon).
To query the geometry after click in the icon we need to do the inverse (icon => geometry).

ger-benjamin:
Thanks for the precision. With this comment and a draw I've understand.
I'll invert the queryIconPosition values logic ( to `[bottom, left, top, right]` for 4, or,`[bottom, left-right, top]` for 3, or `[bottom-top, left-right]` for 2, or `[all]` for one value)

arnaud-morvan:
Note that I'm not sure it is a good idea to mix bbox calculation from css definition, inversion and buffer application.
For comprehension, it might be better to do thing one by one.

ger-benjamin:
@arnaud-morvan I've splitted the calculation but in the same function.
I hope it suits to you.

arnaud-morvan (approved):
Yes it suit to me, IMHO it is easier to understand now, thanks.

PR #6456, opened 2020-11-17
Auto fill timezone attribute in print

arnaud-morvan:
Here is a proposition to fill timezone in print.
This solution only fill timezone if attribute is present in layout.

I firstly think to add it directly in print service in ngeo part (easier), but this may raise errors in case attribute is not present in print config but throwErrorOnExtraParameters is activated.

May be discussed before merge.

### DIFF
--- a/contribs/gmf/src/print/component.js
+++ b/contribs/gmf/src/print/component.js
@@ -829,8 +829,23 @@ export class PrintController {
     // The attributes without 'clientParams' are the custom layout information (defined by end user).
     this.layout_.attributes.forEach((attribute) => {
       this.layoutInfo.attributes.push(attribute.name);
+
       if (!attribute.clientParams) {
         const name = `${attribute.name}`;
+
+        // Special case for timezone, hide it and fill it with browser timezone.
+        if (name == 'timezone') {
+          if (this.options.hiddenAttributes.includes('timezone')) {
+            this.options.hiddenAttributes.push('timezone');
+          }
+          this.layoutInfo.simpleAttributes.push({
+            name: 'timezone',
+            type: 'text',
+            value: Intl.DateTimeFormat().resolvedOptions().timeZone,
+          });
+          return;
+        }
+
         const defaultValue = attribute.default;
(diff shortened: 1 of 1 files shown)


jwkaltz:
I'm a bit confused by this, because from my experience the timezone is automatically filled, when using this processor is used in the print: http://mapfish.github.io/mapfish-print-doc/processors.html#!dateFormat
In fact, I had the opposite need, in my project the customer wanted to suppress the display of the timezone (which appeared automagically). For the record, this can be done with  gmf-print-hiddenattributes="::['timezone']">

sbrunner:
@jwkaltz you speak about hard-coding the timezone on the server, here we set the timezone with a value from the client.

arnaud-morvan:
@jwkaltz 
Object of this PR is to fill the timezone with the user browser timezone.

I agree that actually, there will be really not much users outside CET, and it might be possible to set a different timezone on a project / layout basis. And this might be enough.

If you want to hide the timezone, you can already, just add it in vars.yaml / interfaces_config / constants / gmfPrintOptions / hiddenAttributes.
I agree that we could put this by default in CONST_vars.yaml.
I suppose this is what you expect.

jwkaltz:
Thanks for the explanations @arnaud-morvan @sbrunner , makes sense to me.

arnaud-morvan:
Note we may only set the value of timezone on client side if it is present but not set on server side (as a default).
With this it would be possible to force a timezone on server side.
Even if the name on server side is default and it force the value in facts.

arnaud-morvan:
@sbrunner : need an approuve here, can you also have a look at my last comment ^

sbrunner (approved):
As you do the client-side value will win, the server-side value is just a default...
Did I respond to your question?

arnaud-morvan:
Yes, this is what I've done, but the server side is not really a default is this case, it does nothing.
But we may change the client side (make it a default) to give priority to the server side default, because for now the server default has no effect at all.
=> It would give the possibility to force timezone on server side (using print config default).

PR #6467, opened 2020-11-20
Use actions group

sbrunner:
See: https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-commands-for-github-actions#grouping-log-lines


sbrunner:
Nor working as expected :-(

PR #6476, opened 2020-11-23
Update reference base images

arnaud-morvan:


ger-benjamin (approved):
Thanks please merge.

ger-benjamin (commented):
It is related to you polygon changes ?
Should we update the 2.5 too ?

sbrunner (approved):

arnaud-morvan:
@ger-benjamin :  Pretty sure geodata are not shared between GMF 2.5 / 2.6 demos.

ger-benjamin:
Yes, very probably not shared, but I was wondering if you have done some modification on the 2.5 too.
I guess no

arnaud-morvan:
My modifications on polygons was to solve this issue, but seems there was other differences.
We might discuss at the daily how to avoid such issues, for example by using dedicated layers for the example checks, here maybe on 2.6 and 2.5.

PR #6481, opened 2020-11-24
Fix examples

ger-benjamin:
CI 2.5 fail on new PR because the layerstree image "expect" a translation for ch.astra.hauptstrassennetz. But there is none and the po files was not updated since month... I don't get why we have this error only now


arnaud-morvan (approved):
If CI pass it looks good, thanks

PR #6486, opened 2020-11-25
WMTS opacity on background layers

RenataMuellerC2C:
Fix GSGMF-1392


ger-benjamin (approved):
Looks good. not tested

PR #6531, opened 2020-12-08
Retablish query without queryIconPosition

ger-benjamin:
Fix GSGMF-1435 and https://github.com/camptocamp/ngeo/pull/6445

Oups... : If there is no queryIconPosition, use a simple array of names and a common geom.
We can query every queriable layer again now.


sbrunner (approved):
Looks good :-)

PR #6624, opened 2020-12-24
Use getDataSource to check if layer is filtrable

vuilleumierc:
Hi, it seems that since PR #6009, the filter button which appears next to the layer (when clicking the gear icon) does not work for some layer, although they still appear in the drop down menu in the right panel. See for example in the demo the layer "OSM map" in the group "Filters mixed". Thanks to @RenataMuellerC2C we found out that the button re-appears by reverting one change of the commit b37759d78d048d173138a8fa6a294c49817741ae

### DIFF
--- a/contribs/gmf/src/layertree/component.html
+++ b/contribs/gmf/src/layertree/component.html
@@ -192,7 +192,7 @@
               {{'Swipe with this layer'| translate}}
             </a>
           </li>
-          <li ng-if="::gmfLayertreeCtrl.isFiltrable(layertreeCtrl)">
+          <li ng-if="gmfLayertreeCtrl.isFiltrable(layertreeCtrl)">
             <i class="fa fa-filter fa-fw"></i>
             <a
               ng-click="gmfLayertreeCtrl.toggleFiltrableDataSource(layertreeCtrl.getDataSource())"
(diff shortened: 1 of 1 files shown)


ger-benjamin (commented):
The previous code, on call, do exactly the same thing. Look here: https://github.com/camptocamp/ngeo/blob/addFilterButton/contribs/gmf/src/layertree/component.js#L982

The difference is the `::` (`one-time binding`, an angularjs things). This means that the value or function is evaluated only once, at the creation. this symbol is useful for performance; to avoid to evaluate too much time a value that never change. But I think that the `filterable` parameter on the datasource is not set at the creation of the layertree but later.  And with these `::` the icon never appear. 

Can you please revert your change and try without the `one-time binding` ?

ger-benjamin (approved):

sbrunner:
Shouldn't be backported to 2.5?

sbrunner:
3 commits just to remove 2 chars?

ger-benjamin:
It's in 2.5, I've added the `backport master` label. (It is too late ?)

> 3 commits just to remove 2 chars?

Next time, a merge would be better yes.

sbrunner:
No, you just do `git push --force`...

vuilleumierc:
It was not a problem to push, I said that my local branch was broken. The merge couldn't complete nor be cancelled, maybe because of the reverted commit? I forgot the message. If this is really a problem to you let me know how to start again

sbrunner:
To clean the commits of a branch you shouldn't use `git merge` it will just add a new commit, you can do an interactive rebase e.-g.: `git rebase -i origin master` :-)

PR #6649, opened 2021-01-05
Export controllers

ger-benjamin:
Add some export, mainly for `controllers` classes.
It doesn't cost anything (or perhaps only a little bit in code comprehension) and it make easier the customization of the application.
For instance, with an exported class, we can import it in a customer application and modify slightly one method.
To do the same without the export, we have to redefine the whole component AND the html partials where the component is called. 

I've also suppressed some `@private` tag on exported class.
They looks not useful (and 1/3 are already missing...)

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,6 +1,6 @@
 The MIT License (MIT)
 
-Copyright (c) 2013-2020 Camptocamp SA
+Copyright (c) 2013-2021 Camptocamp SA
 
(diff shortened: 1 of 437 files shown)


sbrunner (approved):

ger-benjamin:
@sbrunner Finally, I've updated the copyright everywhere (that took the same time with `sed`)

PR #6652, opened 2021-01-05
Debug print timezone

ger-benjamin:
Fix https://github.com/camptocamp/ngeo/pull/6456

I remove two line concerning the timezone. It makes no sens like that and throw an error on the desktop interface (includes on a not existing options.hiddenAttributes element => error and unable to print anymore).

I think you can still hide the timezone input by adding this `timezone` in the vars of <interface>->print->hiddenattributes.

### DIFF
--- a/contribs/gmf/src/print/component.js
+++ b/contribs/gmf/src/print/component.js
@@ -833,11 +833,8 @@ export class PrintController {
       if (!attribute.clientParams) {
         const name = `${attribute.name}`;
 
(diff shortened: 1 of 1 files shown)


sbrunner (approved):
Looks good to me if you agree with https://github.com/camptocamp/c2cgeoportal/pull/7570 :-)

ger-benjamin:
Deal :-)

PR #6660, opened 2021-01-07
Be able to switch between query grid and query window with config

sbrunner:


contribs/gmf/src/options.js
```diff
@@ -319,6 +319,11 @@
  * @typedef {Object<string, string[]>} GridMergeTabs
  */
 
+/**
+ * Use the query grit to display the query result.
```
ger-benjamin, line 323:
```suggestion
 * Use the query grid instead the query window to display the query result.
```

ger-benjamin (approved):
Looks good

arnaud-morvan (commented):
LGTM

PR #6679, opened 2021-01-14
Set boolean attributes before adding the propertychange listener

RenataMuellerC2C:
GSNAT-21 (editing)
Setting default value of checkbox was not working for more than one checkbox because of the error.


ger-benjamin (approved):
Looks good.
That's a little bit dangerous because we do things before to listen changes. But in this case we do the same in the `handleFeaturePropertyChange_`. So it works but can broke if we extend the code without being aware of this change.
Perhaps you can add a little comment at the (new) line 179 to say `Set properties before to listen changes to avoid multi AngularJS apply error`.

PR #6680, opened 2021-01-14
Fix query grid

sbrunner:
I have a start of pull request But I don't understand why this pull request brake the query:
https://github.com/camptocamp/ngeo/pull/6660/files

It looks like that everything work fine except the class `gmf-query-grid-active` is not set here https://github.com/camptocamp/ngeo/blob/master/contribs/gmf/apps/desktop_alt/index.html.ejs#L14
If I set it manually it's working ...

### DIFF
--- a/contribs/gmf/src/controllers/AbstractDesktopController.js
+++ b/contribs/gmf/src/controllers/AbstractDesktopController.js
@@ -70,6 +70,12 @@ export class AbstractDesktopController extends AbstractAPIController {
     this.queryGrid = false;
 
     /**
+     * To toggle the activation query grid in the HTML, shouldn't be touch (it's managed by angular).
+     * @type {boolean}
+     */
+    this.queryGridActive = false;
+
+    /**
      * @type {boolean}
      */
     this.dataPanelActive = true;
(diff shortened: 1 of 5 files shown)


contribs/gmf/src/controllers/AbstractDesktopController.js
```diff
@@ -69,6 +69,12 @@ export class AbstractDesktopController extends AbstractAPIController {
      */
     this.queryGrid = false;
 
+    /**
+     * To togle the query grinf in the HTML, shouldn't be touch
```
ger-benjamin, line 73:
```suggestion
     * To toggle the activation query grid in the HTML, shouldn't be touch (it's managed by angular).
```

ger-benjamin (approved):

ger-benjamin:
Note about this dev:
We have to add the `queryGridActivate` in the maincontroller because otherwise, with a ng-if, the `gmf-displayquerygrid-active` "html variable" will be deconnected from the context and the grid will be not displayed (the class at the top will be never added as the `queryGridActivate` will be always false)

arnaud-morvan (approved):
LGTM

PR #6701, opened 2021-01-19
Avoid dropdown menus expanding out of viewport

arnaud-morvan:
Fix GEO-3437

### DIFF
--- a/contribs/gmf/src/sass/base.scss
+++ b/contribs/gmf/src/sass/base.scss
@@ -171,3 +171,9 @@ i, cite, em, var, address, dfn {
   font-size: 2rem;
   margin: 20px 45%;
 }
+
+// Avoid dropdown menus expanding out of viewport
+.dropdown-menu {
+  max-height: 80vh;
+  overflow-y: auto;
+}


sbrunner (approved):

ger-benjamin (commented):
Not tested. It's a little bit dangerous to apply that on every dropdown (that have no more precise rule). But it's okay to me if you have tested carefully

arnaud-morvan:
@ger-benjamin : In demo, I do not see any dropdown menu that take more than 80% of viewport, so it is hard to test.

But I think this is a good default (adaptation for desktop and mobile app which limit their height to the viewport size).

Pretty sure this cannot override another max-height or overflow rule on dropdown-menu class as there is no such rule yet.

If someone has or will place another similar rule for a specific dropdown-menu, it will be more precise and will take predecence.

If some project header take more that 20% of the viewport, they should override the default rule.

The question I see is : is it ideally placed in base.scss or should I move it in desktop.scss and mobile.scss.

arnaud-morvan:
Note that this is a fix for a precise dropdown (attribute list in filter panel) on a precise project (agrola).
But I think it is good to apply this as a default for all dropdowns.

PR #6709, opened 2021-01-20
Fix blank line on grid (do resize map)

ger-benjamin:
Same fix as for https://jira.camptocamp.com/browse/GEO-3866

Under the grid, a blank line can appear as the map-resize is disconnected to the queryGridActivate value.
Appear almost always with the filter tool (and interfaces with the grid). You can test the bug on the GMF demo 2-6
This fix correct that.


sbrunner (approved):
Thanks :-)

sbrunner:
This should be probably back ported?

ger-benjamin:
No, we don't have the `ng-if` and in 2.5 And we have attached the queryGridActivate variable to the mainCtrl for this ng-if

sbrunner:
OK, Thanks :-)

PR #6711, opened 2021-01-20
Make displayquerygrid height overridable

asaunier:
(tested in GMF 2.5 but seems the same with GMF master)

When trying to customize GMF's gmf-displayquerygrid height, overridding ``$grid-height`` in the project's ``vars_desktop.scss`` as for the colors (eg. ``$brand-primary``) has no effect.

It seems because ``$grid-height`` is directly set in https://github.com/camptocamp/ngeo/blob/master/contribs/gmf/src/query/grid.scss#L26 (right before using it in the grid styles) instead of being added to https://github.com/camptocamp/ngeo/blob/master/contribs/gmf/src/sass/vars_only.scss 

This PR moves the var in the ``vars_only.scss`` and rename it.

Tested successfully in a GMF 2.5 project.

### DIFF
--- a/contribs/gmf/src/sass/vars_only.scss
+++ b/contribs/gmf/src/sass/vars_only.scss
@@ -61,6 +61,7 @@ $hover-background-color: lighten($brand-primary, $standard-variation) !default;
 $brand-secondary-dark: darken($brand-secondary, $standard-variation) !default;
 
 $displayquerywindow-desktop-height: 15rem !default;
+$displayquerygrid-height: 15.62rem !default;
 
 // Z-indexes
 $below-content-index: 1;
--- a/contribs/gmf/src/query/grid.scss
+++ b/contribs/gmf/src/query/grid.scss
@@ -23,14 +23,13 @@
 
 @import '~gmf/sass/vars.scss';
 
-$grid-height: 15.62rem;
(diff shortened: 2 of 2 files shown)


contribs/gmf/src/sass/vars_only.scss
```diff
@@ -38,6 +38,7 @@ $hover-background-color: lighten($brand-primary, $standard-variation) !default;
 $brand-secondary-dark: darken($brand-secondary, $standard-variation) !default;
 
 $displayquerywindow-desktop-height: 15rem !default;
+$displayquerygrid-height: 15.62rem !default;
```
asaunier, line 41: I have not used the "desktop" keyword here because I don't know if it is really a desktop-only style. Should I?

sbrunner: no :-)

sbrunner (approved):
Looks good :-)

asaunier:
@sbrunner Thanks for reviewing!

I have just force-pushed the branch because I have rebased it against master (my local master was not up-to-date).

asaunier:
@sbrunner @ger-benjamin Merged. Could this fix be backported to GMF 2.5 please?

PR #6712, opened 2021-01-20
Handle QGIS nil value

sbrunner:
With QGIS value can be
```
{
    "_content_": undefined
    "xsi:nil": "true"
}
```

Fix: GSGMF-1454


ger-benjamin (approved):
Looks good, not tested

PR #6713, opened 2021-01-20
Fix offline mask

llienher:
When disabling the offline mask, the view was "zoom" locked.

### DIFF
--- a/src/offline/component.js
+++ b/src/offline/component.js
@@ -260,28 +260,28 @@ export const Controller = class {
      * @type {number}
      * @export
      */
-    this.minZoom = 0;
+    this.minZoom;
 
     /**
      * Maximum zoom where offline is enable.
      * @type {number}
      * @export
      */
-    this.maxZoom = 0;
+    this.maxZoom;
 
     /**
      * Map view max zoom constraint.
      * @type {number}
      * @export
      */
-    this.originalMinZoom = 0;
+    this.originalMinZoom;
 
     /**
      * Map view min zoom constraint.
      * @type {number}
      * @export
      */
-    this.originalMaxZoom = 0;
+    this.originalMaxZoom;
 
     /**
      * @type {number}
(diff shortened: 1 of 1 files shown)


src/offline/component.js
```diff
@@ -260,28 +260,28 @@ export const Controller = class {
      * @type {number}
      * @export
      */
-    this.minZoom = 0;
+    this.minZoom = undefined;
```
ger-benjamin, line 263:
```suggestion
    this.minZoom;
```
Should be the same. (I don't know if the typing accept that).

ger-benjamin: By the way that's exported but looks not used in the html

src/offline/component.js
```diff
@@ -260,28 +260,28 @@ export const Controller = class {
...
-    this.maxZoom = 0;
+    this.maxZoom = undefined;
 
     /**
      * Map view max zoom constraint.
      * @type {number}
      * @export
      */
-    this.originalMinZoom = 0;
+    this.originalMinZoom = undefined;
```
ger-benjamin, line 277: Same comments as for the min/maxZoom

src/offline/component.js
```diff
@@ -508,8 +508,10 @@ export const Controller = class {
    */
   removeZoomConstraints_() {
     const view = this.map.getView();
-    view.setMaxZoom(this.originalMaxZoom);
-    view.setMinZoom(this.originalMinZoom);
+    if (this.originalMaxZoom) {
```
ger-benjamin, line 511:
```suggestion
    if (this.originalMaxZoom !== undefined) {
```
Otherwise, 0 would be true.
Also, it is possible to have a minZoom but not a maxZoom ? In that's the  case, the test is bad.

llienher: I don't think it is possible, but I modified the condition to make sure both are undefined.

ger-benjamin (approved):
Thanks !

PR #6741, opened 2021-01-27
Set ghost style

RenataMuellerC2C:
As noticed by @ger-benjamin for Mapilary, the ghost for left and right panel do not look the same. The reason is that the resizing of the right panel is not done on the panel itself, because it contains the app bar which should  not be stretched, but on its content. This PR proposes a solution to make them look more similar.


ger-benjamin (approved):
thanks

PR #6742, opened 2021-01-27
Fix panel close

RenataMuellerC2C:
Fix the close of the left panel when using Desktop-version on Android.


sbrunner:
@ger-benjamin What do you think, for me, it looks good :-)

ger-benjamin (approved):
Hard to say, I can't reproduce the issue on our 2.5 demo. Nor with my android device, nor on browserstack...

@RenataMuellerC2C can you reproduce this issue on your device ? In all cases, your change doesn't break the normal behavior.

But this would fix only one issue of the 3 listed in https://github.com/camptocamp/GeoMapFish/issues/42

RenataMuellerC2C:
@ger-benjamin Yes, I can reproduce it on the demo 2.5. And yes it is only for the first point!

PR #6745, opened 2021-01-28
Min and max time can be null

sbrunner:
Fix GEO-4084

### DIFF
--- a/src/misc/Time.js
+++ b/src/misc/Time.js
@@ -74,12 +74,6 @@ Time.prototype.getOptions = function (time) {
 
   const minTime = this.getTime(minDate);
   const maxTime = this.getTime(maxDate);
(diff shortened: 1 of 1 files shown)


asaunier:
@sbrunner What's the connection between a date selector in italian and null min/max times?

sbrunner:
null min/max times throws an error that skip the language switch ...

asaunier:
Thanks! Good to know for reviewers :)

ger-benjamin (approved):
That's what we should avoid throwing error, at least in AngularJS.
But should we not do a `console.error('Missing minTime')` ? Otherwise, we don't know there is an error in our configuration

sbrunner:
But it's not an error, the field can be initialized with no value...

PR #6746, opened 2021-01-28
Change attribute types to ngeoFormatAttributeTypes

RenataMuellerC2C:
For filtering only the default rules were available for numbers, because the attribute types were not changed to ngeoFormatAttributeTypes (double -> number).


sbrunner:
What do you think @ger-benjamin, for me, it looks correct

ger-benjamin (approved):

sbrunner (approved):

arnaud-morvan (approved):

PR #6770, opened 2021-02-03
Ignore permalink on manual background selection

ger-benjamin:
Fix GSGMF-1249

We can't determine when app is fully loaded. So I've applied the invert logic: Specify when a user touch the background selection And from then, start to ignore the permalink.

I'll do more cleaning on 2.6


sbrunner:
> We can't determine when app is fully loaded. 

Can't we catch the same "event" then we used to hide the loading panel?

ger-benjamin:
No, the event you think only means that the theme is loaded.
But a lot of promises are connected to this event and there is no order of execution.
Plus, other events could append after that (like the login, or even the same event but loaded by a component not loaded before).

sbrunner (approved):
Thanks :-)

PR #6780, opened 2021-02-05
Fix permalink: avoid using ng-include

ger-benjamin:
Fix GSGMF-1469

### Why permalink was broken and fix

The commit that break the permalink is surprisingly this innocent one: https://github.com/camptocamp/ngeo/pull/6375

It looks than using `ng-include` create problem with the `$location` service. Maybe it's related to this warning: https://github.com/camptocamp/ngeo/blob/779251e31115d77ffd2451deac5670336e9fa97f/src/statemanager/Location.js#L29-L33
I've not investigate more in depth. ng-include breaks our permalink.

To avoid using `ng-include`, I've create a directive for the header. It's fare more line of code (even without every comments), but it works and it's still possible to override the partial.

Another possibility would be to investigate into the permalink and the `$location` service to know why we have this issue but it could cost us a lot of time...

### Others

 1. I've added the custom header in the oeedit view.
 1. In some components we have one other way to override the html (via attributes). I think that's an old way to do things, useless and I'm not even sure it works. I don't have implemented the others way here. (For instance see: https://github.com/camptocamp/ngeo/blob/779251e31115d77ffd2451deac5670336e9fa97f/contribs/gmf/src/backgroundlayerselector/component.js#L36-L47)

### TODO

I'll update the c2cgeoportal changelog

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -269,6 +269,8 @@ contribs/dist: .build/build-dll.timestamp
 	cp ./contribs/gmf/apps/desktop/header.html .build/examples-hosted/contribs/gmf/apps/desktop
 	mkdir -p .build/examples-hosted/contribs/gmf/apps/desktop_alt
 	cp ./contribs/gmf/apps/desktop_alt/header.html .build/examples-hosted/contribs/gmf/apps/desktop_alt
+	mkdir -p .build/examples-hosted/contribs/gmf/apps/oeedit
+	cp ./contribs/gmf/apps/oeedit/header.html .build/examples-hosted/contribs/gmf/apps/oeedit
 	touch $@
 
 .build/examples-hosted/index.html: \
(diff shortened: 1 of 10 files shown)


sbrunner:
Then how do we update the header in simple application mode?

ger-benjamin:
In the desktop controller, to use myCustomHeader.html:

```
import gmfHeaderComponent from 'gmf/header/component.js';

/**
 * Override gmf/header/component partial with a custom one.
 */
gmfHeaderComponent.run(/* @ngInject */ ($templateCache) => {
  // @ts-ignore: webpack
  $templateCache.put('gmf/header/component', require('./myCustomHeader.html'));
});
```

sbrunner:
In simple application mode we can't write JavaScript, we can't change the HTML...

ger-benjamin:
ah...

Well, I could add this code in our desktops interfaces to read a `header.html` at the same place you had it. But... that's ugly.
I no time right that, I would like to see how it's done for the contextual data partial, perhaps I can find some hints. I'll see that Monday.

sbrunner:
It's not a solution, can we call about it on monday?

ger-benjamin:
Finally, as discussed, I use the attributes "template-url" (see "Other", point 2 in this PR's description) to give the path to the template. It looks to working well.

sbrunner (approved):
Thanks :-)

PR #6782, opened 2021-02-05
Fix versions in CI, add 2.6

ger-benjamin:
It is the correct way to correct check version on 2.6 ?

And it is useful to have the 2.6 in the audit ?

### DIFF
--- a/ci/config.yaml
+++ b/ci/config.yaml
@@ -2,8 +2,7 @@ checks:
   required_workflows:
     main.yaml: False
     rebuild.yaml: False
-  versions:
-    rebuild: False
+  versions: False
   codespell:
     ignore_re:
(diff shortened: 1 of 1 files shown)


sbrunner:
The audit in 2.6 is never run (only the cron in master will be taken in account), and only the SECURITY.md of the master is maintained...

ger-benjamin:
@sbrunner  But the addition of 2.6 in the SECURITY fix the CI in the 2.6

sbrunner:
Effectively, for that, we should simply replace those lines
https://github.com/camptocamp/ngeo/blob/2.6/ci/config.yaml#L5-L6
by: `versions: False` thanks :-)

ger-benjamin:
Ok, thanks. Let's see

sbrunner (approved):

PR #6783, opened 2021-02-05
Simplify background selection

ger-benjamin:
After GSGMF-1249

Cleaning on 2.6 only as it's a sensible point.
This method do the same (except thta it's missing to search the background layer in the theme functionalities) as the method `setDefaultBackground_` under:  https://github.com/camptocamp/ngeo/blob/26c7180f9079d78fdeb7faefac744c0e253471ec/contribs/gmf/src/controllers/AbstractAppController.js#L704

I don't call it here because it's already called multiple times during the process. Notably by the initial unavoidable "login".


sbrunner (approved):
I'm confident to you

PR #6795, opened 2021-02-08
Add crossorigin=anonymous on ressources that don't required an autent…

sbrunner:
…ication
Fix: https://jira.camptocamp.com/browse/GSGMF-1473


ger-benjamin (approved):

arnaud-morvan (approved):
LGTM

PR #6804, opened 2021-02-10
Use alert instead of modal to show changed password (simple message)

ger-benjamin:
Fix GSGMF-1478

Authentication modal do not work on mobile, as the right menu use "position: fixed" (as the modal black screen), the childen elements can't appears above...

Solution: the modal is not really useful. I'll display the message in a simple alert, in the component.

From 2.6, I'll have another modal to move to a simple message (lost connection message).

### DIFF
--- a/contribs/gmf/src/authentication/component.html
+++ b/contribs/gmf/src/authentication/component.html
@@ -79,8 +79,6 @@
           ng-click="$ctrl.logout()" />
     </div>
   </form>
-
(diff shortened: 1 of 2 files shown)


sbrunner:
The build should be fixed by: #6805

ger-benjamin:
Thanks @sbrunner , yes the CI passes now

sbrunner (approved):

PR #6810, opened 2021-02-10
Rename Rule.expression to Rule.literal

arnaud-morvan:
And change type from {number|string} to {number|string|string[]}.
To avoid using comma as separator in "select" rules.

### DIFF
--- a/src/filter/RuleHelper.js
+++ b/src/filter/RuleHelper.js
@@ -244,6 +244,10 @@ export class RuleHelper {
    * @return {import("ngeo/rule/Rule.js").default} Rule.
    */
   createRule(options) {
+    // Compatibility with filters saved before version 2.6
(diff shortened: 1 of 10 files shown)


src/filter/RuleHelper.js
```diff
@@ -254,6 +258,11 @@ export class RuleHelper {
         rule = new ngeoRuleGeometry(options);
         break;
       case ngeoFormatAttributeType.SELECT:
+        // Compatibility with filters saved before version 2.6
+        if (options.expression) {
+          const expression = /** @type {string} */ options.expression;
```
sbrunner, line 263:
```suggestion
          const expression = /** @type {string} */ (options.expression);
```

sbrunner (approved):
Looks good to me :-) Thanks :-)

PR #6811, opened 2021-02-10
Rename 'Apply' button in rule creating form

asaunier:
Because the button only adds a rule and does not apply it yet.

I have noticed several users confused because filters were not actually applied to the map when pressing the "Apply" button available in the form to add filters. Instead, one has to eventually press the global "Apply filter" button to have the set of filters applied.

Screenshot from 2021-02-10 18-32-33

I suggest to remove this confusion by simpling renaming the button to "Add". Eventually the users will press the "Apply filter" button when they are done setting rules.

### DIFF
--- a/src/filter/rulecomponent.html
+++ b/src/filter/rulecomponent.html
@@ -207,7 +207,7 @@
         <button
             class="btn btn-sm btn-default"
             ng-click="$ctrl.apply()"
-            type="button">{{'Add' | translate}}</button>
+            type="button">{{'OK' | translate}}</button>
         <button
             class="btn btn-sm btn-default"
             ng-click="$ctrl.cancel()"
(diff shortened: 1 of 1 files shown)


sbrunner:
Add is also wrong, it can be Set, or just OK...

asaunier:
@sbrunner Updated!

ger-benjamin (approved):

ger-benjamin:
@asaunier the CI can't currently passe, we have to wait on a fix on our QGIS config

ger-benjamin:
It should pass now, I've restarted the build

sbrunner (approved):

asaunier:
@sbrunner @ger-benjamin Thanks! Merging!

PR #6840, opened 2021-02-17
Fix print text on feature

ger-benjamin:
Fix GSGMF-1479

If a feature have a style with a geometry, and this geometry has a style AND a text, we now push both symbolizers.


sbrunner (approved):

jwkaltz:
thanks @ger-benjamin: can this easily be backported to 2.5? (we do need a backport)

ger-benjamin:
I'll try to backport this fix in 2.5. But if that takes time, I'll not put too much effort in it.

PR #6905, opened 2021-03-04
Fix point highlight and console error

RenataMuellerC2C:
Fix: GSGMF-1499

### DIFF
--- a/contribs/gmf/src/editing/editFeatureComponent.js
+++ b/contribs/gmf/src/editing/editFeatureComponent.js
@@ -1307,7 +1307,16 @@ Controller.prototype.openFeatureMenu_ = function (coordinate, features) {
   );
 
   this.map.addOverlay(this.menuMultiple_);
-  this.menuMultiple_.open(coordinate);
+  if (features[0].getGeometry().getType() === 'MultiPoint') {
+    const resolution = this.map.getView().getResolution();
+    if (resolution === undefined) {
+      throw new Error('Missing resolution');
+    }
+    const offset = resolution * 10;
+    this.menuMultiple_.open([coordinate[0] + offset, coordinate[1] + offset]);
+  } else {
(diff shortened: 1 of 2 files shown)


contribs/gmf/src/editing/editFeatureComponent.js
```diff
@@ -563,6 +563,12 @@ Controller.prototype.$onInit = function () {
     throw new Error('Missing map');
   }
 
+  this.options_.highlightStyle.circle = {
+    radius: 7,
+    fill: this.options_.highlightStyle.fill,
+    stroke: this.options_.highlightStyle.stroke,
+  };
```
sbrunner, line 570: Why this?
Looks like that it should be moved here: https://github.com/camptocamp/c2cgeoportal/blob/2.6/geoportal/c2cgeoportal_geoportal/scaffolds/update/geoportal/CONST_vars.yaml_tmpl#L319

sbrunner (commented):
Looks good :-)

contribs/gmf/src/options.js
```diff
@@ -465,7 +465,7 @@
 
 /**
  * @typedef {Object} gmfEditFeatureOptions
- * @property {import('ngeo/options.js').StyleLike} highlightStyle A style object for highlighting of feature.
+ * @property {import('ngeo/options.js').Style} highlightStyle A style object for highlighting of feature.
```
llienher, line 468: What is this change for ? OL use StyleLike (https://openlayers.org/en/latest/apidoc/module-ol_style_Style.html#~StyleLike) and there is 9 others occurance of it in this same file.

sbrunner: +1 it should be `StyleLike`

llienher (approved):
Looks good, thanks !

PR #6916, opened 2021-03-05
Fix the event name

sbrunner:
According https://github.com/camptocamp/c2cgeoportal/blob/master/.github/workflows/ngeo-26.yaml#L7


arnaud-morvan (approved):
LGTM

PR #6924, opened 2021-03-08
Too long drop down gsgmf 1506

ger-benjamin:
Fix GSGMF-1506

If the dropdown is in the bottom half of the app, it will be a drop-up. With 50vh, the dropdown should be always shorter than the page.


llienher (approved):
Looks good to me

PR #6907, opened 2021-03-04
Add LIDAR profile to desktop alt

sbrunner:

PR #6934, opened 2021-03-08
Fix LIDAR

ger-benjamin:
To try on desktop alt, over canton Neuchâtel.

From ticket  GSGMF-1486 :

- Fix refresh button
- Fix Lidar export-png size (was shrunk because of a smaller size of canvas than in the original)

From #6907 :

- Fix css (don't re-use drawProfile class)
- Use more standard open/close system panel (fix always closing panel)

Other:

- Set the title of the tool's panel to say that it works only on Neuchâtel.

### DIFF
--- a/contribs/gmf/apps/desktop_alt/index.html.ejs
+++ b/contribs/gmf/apps/desktop_alt/index.html.ejs
@@ -226,10 +226,10 @@
               </div>
             </div>
           </div>
-          <div ng-show="mainCtrl.drawLidarprofilePanelActive" class="row profile-panel">
+          <div ng-show="mainCtrl.drawLidarprofilePanelActive" class="row lidar-profile-panel">
             <div class="col-sm-12">
               <div class="gmf-app-tools-content-heading">
-                {{'Lidar profile'|translate}}
+                {{'Lidar profile (only on the canton of Neuchâtel)'|translate}}
                 <a class="btn close" ng-click="mainCtrl.drawLidarprofilePanelActive = false">&times;</a>
               </div>
               <gmf-lidarprofile-panel
--- a/contribs/gmf/src/controllers/AbstractDesktopController.js
+++ b/contribs/gmf/src/controllers/AbstractDesktopController.js
@@ -242,7 +242,7 @@ export class AbstractDesktopController extends AbstractAPIController {
      * The ngeo ToolActivate manager service.
      * @type {import("ngeo/misc/ToolActivateMgr.js").ToolActivateMgr}
      */
-    const ngeoToolActivateMgr = $injector.get('ngeoToolActivateMgr');
+    this.ngeoToolActivateMgr = $injector.get('ngeoToolActivateMgr');
 
     /**
      * @type {import('gmf/datasource/DataSourceBeingFiltered.js').DataSourceBeingFiltered}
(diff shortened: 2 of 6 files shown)


contribs/gmf/apps/desktop_alt/index.html.ejs
```diff
@@ -226,10 +226,10 @@
               </div>
             </div>
           </div>
-          <div ng-show="mainCtrl.drawLidarprofilePanelActive" class="row profile-panel">
+          <div ng-show="mainCtrl.drawLidarprofilePanelActive" class="row lidar-profile-panel">
             <div class="col-sm-12">
               <div class="gmf-app-tools-content-heading">
-                {{'Lidar profile'|translate}}
+                {{'Lidar profile (only on the canton of Neuchâtel)'|translate}}
```
asaunier, line 232: Why adding this mention? The tool works outside of Neuchatel, doesn't it? Or is it because the "desktop_alt" app only has lidar data in canton NE?

contribs/gmf/src/lidarprofile/Utils.js
```diff
@@ -260,13 +260,7 @@ export default class {
...
     const profileCanvas = d3select('#gmf-lidarprofile-container .lidar-canvas');
     const profileCanvasEl = /** @type {HTMLCanvasElement} */ (profileCanvas.node());
-    ctx.drawImage(
-      profileCanvasEl,
-      margin.left,
-      margin.top,
-      w - (margin.left + margin.right),
-      h - (margin.top + margin.bottom)
-    );
+    ctx.drawImage(profileCanvasEl, margin.left, margin.top);
```
asaunier, line 263: Is it the change that actually fixes the margin issue?

contribs/gmf/apps/desktop_alt/index.html.ejs
ger-benjamin: It's only for our desktop_alt demo. It's not common to add info related to the config here. But desktop_alt changes is not exported in customer projects and we have only data on NE.

contribs/gmf/src/lidarprofile/Utils.js
ger-benjamin: Yes

contribs/gmf/src/controllers/AbstractDesktopController.js
```diff
@@ -242,7 +242,7 @@ export class AbstractDesktopController extends AbstractAPIController {
      * The ngeo ToolActivate manager service.
      * @type {import("ngeo/misc/ToolActivateMgr.js").ToolActivateMgr}
      */
-    const ngeoToolActivateMgr = $injector.get('ngeoToolActivateMgr');
+    this.ngeoToolActivateMgr = $injector.get('ngeoToolActivateMgr');
```
sbrunner, line 245: Why this change?

ger-benjamin: Otherwise, I've to redo `$injector.get('ngeoToolActivateMgr');` in the `desktop_alt/Controller.js` Nothing big, but I think that not cast anything to reference it as variable of `AbstractDesktopController.js` (to be able to do `this.ngeoToolActivateMgr....` in the `desktop_alt/Controller`)

sbrunner (approved):
LGTM

PR #6952, opened 2021-03-09
Better texts for Raster component

ger-benjamin:
For GSGMF-1502

That would be strange to have only the loading spinner in the button. I've put `Raster` instead. That's the same behaviors as the next button `Coordinate`. If no data is displayed (like out of CH), it displays `Coordinate`.

### DIFF
--- a/contribs/gmf/src/raster/widgetComponent.html
+++ b/contribs/gmf/src/raster/widgetComponent.html
@@ -16,12 +16,12 @@
       <span ng-show="ctrl.elevationLoading" class="fa custom-spinner-generic">
           <%=require('gmf/icons/spinner.svg?viewbox&height=1rem')%>
       </span>
-      <span ng-show="!ctrl.elevationValue && !ctrl.elevationLoading" translate>Elevation…</span>
+      <span ng-show="!ctrl.elevationValue && !ctrl.elevationLoading" translate>Raster</span>
     </span><span class="caret" ng-if="::ctrl.options.layers.length > 1"></span>
   </a>
   <ul class="dropdown-menu dropdown-menu-right" role="menu"
       ng-if="::ctrl.options.layers.length > 1">
-    <li class="dropdown-header" translate>Elevation data sources</li>
+    <li class="dropdown-header" translate>Data source</li>
     <li ng-repeat="elevationItem in ::ctrl.options.layers">
       <a href ng-click="ctrl.selectedElevationLayer = elevationItem">
         <span class="fa fa-fw" ng-class="{'fa-check': ctrl.selectedElevationLayer === elevationItem}"></span>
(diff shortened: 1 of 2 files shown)


contribs/gmf/src/raster/widgetComponent.html
```diff
@@ -16,12 +16,12 @@
       <span ng-show="ctrl.elevationLoading" class="fa custom-spinner-generic">
           <%=require('gmf/icons/spinner.svg?viewbox&height=1rem')%>
       </span>
-      <span ng-show="!ctrl.elevationValue && !ctrl.elevationLoading" translate>Elevation…</span>
+      <span ng-show="!ctrl.elevationValue && !ctrl.elevationLoading" translate>Raster</span>
```
sbrunner, line 19: => ?
```
      <span ng-show="ctrl.elevationLoading || !ctrl.elevationValue" class="fa custom-spinner-generic">
          <%=require('gmf/icons/spinner.svg?viewbox&height=1rem')%>
      </span>
```

ger-benjamin: I'll need words...

sbrunner: ?

ger-benjamin: I don't understand what you means with `=> ?`

ger-benjamin: If you mean "I've said that it's better to always have a loading spinner", I've replied in the first comment of the PR

sbrunner: OK I understand better =>
+ For Raster when the mouse is out of the map
But actually, it's strange to have when we move the mouse on the maps
'Elevation...' then loading icon then the value
For me, we should just have the loading icon then the value

ger-benjamin: Yes but it's strange to have the loading symbol forever when you are on tools panel or on the layertree... It looks like a bug.
I don't know... we need a third opinion @ybolognini @arnaud-morvan @RenataMuellerC2C ?

sbrunner: > Yes but it's strange to have the loading symbol forever when you are on tools panel or on the layertree... It looks like a bug.

On my last comment I don't say that... (+1 For Raster when the mouse is out of the map)

ger-benjamin: Ok, I'll display the 'loading spinner' always on the map when it has no value. And `Raster` when the mouse pointer is not on the map.

ger-benjamin:
Done (+ some cleaning => move part of a big function into sub functions for clarity)

sbrunner (approved):
Excellent, thanks :-)

PR #6958, opened 2021-03-11
Fix first segment lenght on mobile

RenataMuellerC2C:
Fix GSGMF-1497


sbrunner (approved):
I didn't see a regression on Firefox :-)

PR #6969, opened 2021-03-15
Trigger CI and debug it after new i18n

ger-benjamin:
After GSGMF-1508 https://github.com/camptocamp/demo_geomapfish/pull/219

Ci should fail, I'll debug it here.


ger-benjamin:
Ah, the OSM time gorup is closed, so, the CI passes. sorry....

PR #6979, opened 2021-03-17
Fix API Help instructions

jwkaltz:


ger-benjamin (approved):
Thanks, (this change is already in 2.5, it take us time to find out where's the problem)

PR #6987, opened 2021-03-18
Adapt WFS-version for reading of GML2 results

RenataMuellerC2C:
GSGMF-1505


llienher (approved):
Looks good to me

sbrunner (changes requested):
For what reason do you do this change?

RenataMuellerC2C:
@sbrunner The reading of the features fails for polygon layers with output format GML2 like landuse_alias, if the version of olFormatWFS is set to the default 1.1.0. Or should we change that on the qgis side?

sbrunner (approved):
No, that looks good :-)

PR #7002, opened 2021-03-23
Fix Sentry

sbrunner:
GSGMF-1455


ger-benjamin (approved):
Looks technically good as far as I know... But I don't know a lot about that...

PR #7050, opened 2021-04-07
Move apihelp marker in dist

ger-benjamin:
Fix GSGMF-1558

I think I didn't need to backport that to 2.5 or less.


sbrunner:
Using relative URL will be hard to make it working everywhere, then shouldn't we just add the images on the demo and use the with absolute URL?

ger-benjamin:
Okay, the code is updated and should works with: https://github.com/camptocamp/demo_geomapfish/pull/229

It will work as long as we have a 2-6 demo server.

sbrunner (approved):

PR #7051, opened 2021-04-07
Better printed legend for qgis

ger-benjamin:
For GSGMF-1560

Somehow better now, but with scale bigger than 25'000, I still get wrong results. (until 25'000, it looks working).

Like this request: https://geomapfish-demo-2-6.camptocamp.com/mapserv_proxy?ogcserver=QGIS+server&cache_version=d8714cbf6e4c4b33841550cc2c25f2a5&FORMAT=image%2Fpng&TRANSPARENT=true&SERVICE=WMS&VERSION=1.1.1&REQUEST=GetLegendGraphic&LAYER=osm&DPI=254&BBOX=2530379.3079208867%2C1156552.5068461096%2C2540168.9108334263%2C1168458.7806586574&SRS=EPSG%3A2056&SRCWIDTH=1958&SRCHEIGHT=2381&ITEMFONTFAMILY=DejaVu%20Sans&ITEMFONTSIZE=8&LAYERFONTFAMILY=DejaVu%20Sans&LAYERFONTSIZE=10

Is at the nord of lausanne. There are not "pompier" feature in the area, and even not near. But it still appears in the legend...

### DIFF
--- a/src/map/LayerHelper.js
+++ b/src/map/LayerHelper.js
@@ -541,8 +541,9 @@ LayerHelper.prototype.getWMSLegendURL = function (
     ) {
       queryString.BBOX = opt_bbox.join(',');
       queryString.SRS = opt_srs;
-      queryString.WIDTH = Math.round(((opt_bbox[2] - opt_bbox[0]) / opt_scale) * 39.37 * opt_dpi);
-      queryString.HEIGHT = Math.round(((opt_bbox[3] - opt_bbox[1]) / opt_scale) * 39.37 * opt_dpi);
+      queryString.SRCWIDTH = Math.round(((opt_bbox[2] - opt_bbox[0]) / opt_scale) * 39.37 * opt_dpi);
+      queryString.SRCHEIGHT = Math.round(((opt_bbox[3] - opt_bbox[1]) / opt_scale) * 39.37 * opt_dpi);
+      delete queryString.SCALE; // QGIS calculate it from the BBOX the SRCWIDTH and the SRCHEIGHT.
     }
   }
   if (opt_additionalQueryString) {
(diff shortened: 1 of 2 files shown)


src/map/LayerHelper.js
```diff
@@ -541,8 +541,9 @@ LayerHelper.prototype.getWMSLegendURL = function (
     ) {
       queryString.BBOX = opt_bbox.join(',');
       queryString.SRS = opt_srs;
-      queryString.WIDTH = Math.round(((opt_bbox[2] - opt_bbox[0]) / opt_scale) * 39.37 * opt_dpi);
-      queryString.HEIGHT = Math.round(((opt_bbox[3] - opt_bbox[1]) / opt_scale) * 39.37 * opt_dpi);
+      queryString.SRCWIDTH = Math.round(((opt_bbox[2] - opt_bbox[0]) / opt_scale) * 39.37 * opt_dpi);
+      queryString.SRCHEIGHT = Math.round(((opt_bbox[3] - opt_bbox[1]) / opt_scale) * 39.37 * opt_dpi);
+      delete queryString.SCALE; // QGIS calculate it from the BBOX the SRCWIDTH and the SRCHEIGHT.
```
sbrunner, line 546: Looks like that the lines 538 and 539 should also be removed

sbrunner:
Looks like to be a QGIS issue then...

src/map/LayerHelper.js
ger-benjamin: Why? They are still used.

sbrunner: Oups, you're right :-)

sbrunner (approved):

ger-benjamin:
>  Looks like to be a QGIS issue then...

Should we open an issue somewhere?
Or document this problem ? I don't really know where. Maybe directly in the code ?

PR #7055, opened 2021-04-08
Do not clear search field

RenataMuellerC2C:
Fix GSGMF-1561

### DIFF
--- a/contribs/gmf/src/search/component.js
+++ b/contribs/gmf/src/search/component.js
@@ -938,7 +938,7 @@ export class SearchController {
     // based on https://github.com/twitter/typeahead.js/issues/780#issuecomment-251554452
     // FIXME: remove this workaround when https://github.com/corejavascript/typeahead.js/issues/60 is fixed
 
-    //if there is no clear button clean the map and colorpicker
+    // If there is no clear button clean the map and colorpicker.
     if (!this.clearButton) {
       this.featureOverlay_.clear();
       this.displayColorPicker = false;
--- a/contribs/gmf/src/search/component.js
+++ b/contribs/gmf/src/search/component.js
@@ -345,7 +345,7 @@ export class SearchController {
      */
     this.clearButton = this.options.clearButton !== false;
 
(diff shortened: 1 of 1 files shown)


sbrunner:
The fts on mobile applications didn't work any more...

contribs/gmf/src/search/component.js
```diff
@@ -345,13 +345,6 @@ export class SearchController {
      */
     this.clearButton = this.options.clearButton !== false;
 
-    if (this.clearButton) {
```
ger-benjamin, line 348: This must be `!this.clearButton`. It's what has changed, by my fault, since the 2.5. And it's why the search works bad on both desktop and desktop_alt interfaces.

The aim of the block in the if is to clear the input on focus and on blur (blur means "leave focus", you could add a comment about that here).
We don't want to change this behavior. I would keep this block.

contribs/gmf/src/search/component.js
```diff
@@ -911,9 +904,6 @@ export class SearchController {
    * @private
    */
   leaveSearch_() {
-    if (!this.clearButton) {
```
ger-benjamin, line 914: This has the same purpose as in my previous comment

contribs/gmf/src/search/component.js
```diff
@@ -938,6 +928,12 @@ export class SearchController {
     // based on https://github.com/twitter/typeahead.js/issues/780#issuecomment-251554452
     // FIXME: remove this workaround when https://github.com/corejavascript/typeahead.js/issues/60 is fixed
 
+    //if there is no clear button clean the map and colorpicker
```
ger-benjamin, line 931:
```suggestion
    // If there is no clear button clean the map and colorpicker.
```
I think that's a good improvement, but it was not asked by any customer. If you merge that, we should present that carefully during the demo.

ger-benjamin (changes requested):

ger-benjamin (approved):
Looks good, thanks !

PR #7105, opened 2021-04-27
Be able to inject custom Javascript or Stylesheet for multi-organisations

sbrunner:


ger-benjamin (approved):
Looks good.
Eventually, `gmfCustomJavascriptUrl` would be a better name. Because in a constant, I expected a value or a dict but not an url.

PR #7111, opened 2021-04-30
Search result highlight

llienher:
Not sure what is the best way to set the value correctly.
`this.option.highlight` is the value that the component will use in the end.
The initial value was set to `this.typeaheadOptions` object but not used so I re-added the Object.assign used in the 2.5 code. We could as well just set directly the `this.option` object with the correct highlight value.


sbrunner (approved):
Thanks :-)

PR #7115, opened 2021-05-07
Set styles related to the number of layers to print

llienher:


sbrunner (approved):
It's not the best but for me, we can merge...

PR #7164, opened 2021-05-21
External WMS query

sbrunner:
Fix: GSGMF-1520


llienher (approved):
I tested and it works for me.

PR #7197, opened 2021-06-03
Be able to open more than 5 Dependabot pull requests

sbrunner:


ger-benjamin (approved):
Some thought:
If you set the limit to 20, with npm, you'll have a lot of PR (near to 20 probably :stuck_out_tongue_closed_eyes: ) per day to handle.
That means X times the CI, and probably more once you start to merge.

In some (smaller) projects, I've gave up to manage these PR and limit PR request limit to 1. If I have one PR open, I manually check the updates. (as the check is monthly, I've always a PR to do, but that's a good remember to me)

It's probably not suitable to do that on a daily bases. But instead, couldn't we install and run [npm-check](https://www.npmjs.com/package/npm-check) to make one PR per day with all npm dependencies to update ?

sbrunner:
> If you set the limit to 20, with npm, you'll have a lot of PR (near to 20 probably stuck_out_tongue_closed_eyes ) per day to handle.
> That means X times the CI, and probably more once you start to merge.

In the major cases we don't have to do anything to do if the CI pass it will be merged automatically...

>In some (smaller) projects, I've gave up to manage these PR and limit PR request limit to 1. If I have one PR open, I manually check the updates. (as the check is monthly, I've always a PR to do, but that's a good remember to me)

And if you have 1 pending pull request it block all the other updates...
Actually, we have 7 pending pull request on ngeo, including the one done by the old version...
Or did I miss something?

> It's probably not suitable to do that on a daily bases. But instead, couldn't we install and run npm-check to make one PR per day with all npm dependencies to update ?

With more than 10 dependencies that we can't immediately update, this is relay time-consuming...

ger-benjamin:
> And if you have 1 pending pull request it block all the other updates...
Actually, we have 7 pending pull request on ngeo, including the one done by the old version...
Or did I miss something?

In this case, you can exclude some library for the daily updates.

> With more than 10 dependencies that we can't immediately update, this is relay time-consuming...

Not sure to understand what you mean here, but I understand that's not easy to have a good system in the case of GMF, with so much dependencies. My reflexion aim was just to limit a little bit the usage of the CI.

Maybe there is no good solution...
Anyway, you can merge

PR #7198, opened 2021-06-03
If zoom is undefined or null condition

llienher:
Fix GSGMF-1610

When the zoom param is not set, null value should go in the "else" instead of "if".

### DIFF
--- a/src/geolocation/component.js
+++ b/src/geolocation/component.js
@@ -329,7 +329,7 @@ Controller.prototype.setPosition_ = function () {
 
   if (this.follow_) {
     this.viewChangedByMe_ = true;
-    if (this.options.zoom !== undefined) {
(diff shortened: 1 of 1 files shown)


src/geolocation/component.js
```diff
@@ -329,7 +330,7 @@ Controller.prototype.setPosition_ = function () {
 
   if (this.follow_) {
     this.viewChangedByMe_ = true;
-    if (this.options.zoom !== undefined) {
+    if (this.options.zoom) {
```
sbrunner, line 333:
```suggestion
    if (this.options.zoom or this.options.zoom === 0) {
```

llienher: "or" is not a valid operator in javascript, it's not python. Plus zoom = 0 is not equal to what I want check, if the value is null.

sbrunner (approved):

PR #7236, opened 2021-06-16
Allow to draw elements without the 1 meter limit

llienher:


contribs/gmf/src/drawing/drawFeatureOptionsComponent.html
```diff
@@ -3,7 +3,7 @@
   <div class="input-group">
     <input
         class="form-control"
-        min="1"
+        min="0.1"
```
sbrunner, line 6: Any reason why we need a minimum?

llienher: Nothing from a pure technical point of view, I don't know if there were something in the spec.
Maybe we can just remove entirely the attribute.

sbrunner: Probably we shouldn't remove the attribute but set it to 0 to avoid having negate values

sbrunner (approved):

PR #7274, opened 2021-06-28
Fix static legend on printing

sbrunner:
Get the right node when the layer is in a subgroup

Fix: GSGMF-1638


ger-benjamin (approved):
quick review, looks good

PR #7275, opened 2021-06-28
Use the demo 2.7

sbrunner:


buildtools/check-example.js
```diff
@@ -227,6 +227,7 @@ function loaded(page, browser) {
     const url = request.url();
     if (
       !url.startsWith('https://www.camptocamp.com/') &&
+      !url.startsWith('https://o330647.ingest.sentry.io/') &&
```
ger-benjamin, line 230: o330647 ... it is a fix address ? it is safe ?

ger-benjamin (approved):

buildtools/check-example.js
sbrunner: It's the id of the demo project on Sentry, for me, it's safe enough because it's testing code, and if it changes the tests failed ...

PR #7394, opened 2021-08-24
Add a store based on rxjs

ger-benjamin:
GSGMF-1669
(see first commit to know how to implement it)


src/store/user.ts
```diff
@@ -0,0 +1,119 @@
...
+
+export class UserModel {
+  constructor() {
+    this.config_ = new BehaviorSubject<User>(null);
+  }
+
+  // Notation one. Shorter, less clear, hard to follow (grep...)
+  // get config () {
+  //  return this.config_;
+  // }
```
sbrunner, line 80: Is anyone against this chose, if not, we can remove those comments

src/store/user.ts
```diff
@@ -0,0 +1,119 @@
...
+//   * Will immediately log the current value (null)
+//   */
+// user.getConfig().subscribe({
+//  next: (config) => console.log(config)
+// });
+
+//  /**
+//   * Set the config, will be logged by the subscribe above.
+//   */
+// user.setConfig({username: 'toto'});
```
sbrunner, line 119: Shouldn't be in the pull request component?

sbrunner (approved):

PR #7396, opened 2021-08-25
Webpack config for typescript + lit-element installation

llienher:
Add webpack config and dependencies to support typescript (and lit-element).

Uses ts-loader 8 in order to work with webpack 4.
The tsconfig.json makes the transition to Typescript checking instead of Javascript (as discussed with Stéphane).


package.json
```diff
@@ -79,6 +79,7 @@
     "angular-ui-slider": "0.4.0",
     "babel-loader": "8.2.2",
     "babel-plugin-angularjs-annotate": "0.10.0",
+    "babel-plugin-transform-runtime": "^6.23.0",
```
sbrunner, line 82:
```suggestion
    "babel-plugin-transform-runtime": "6.23.0",
```

tsconfig.json
```diff
@@ -6,14 +6,13 @@
     "lib": ["es2020", "dom"],
     "esModuleInterop": true,
     "allowJs": true,
-    "checkJs": true,
-    "noEmit": true,
-    "noImplicitAny": true,
```
sbrunner, line 11: Can you try to re add this rule?

sbrunner (commented):
Don't you want to also add your example component?

sbrunner:
Can you add `--noEmit` here https://github.com/camptocamp/ngeo/blob/poc_2.7/package.json#L17 ?

tsconfig.json
llienher: Done (for noImplicitAny).

llienher:
I've added the basic implementation I did for test, that way we can see if the CI works with real typsecript to compile.

src/auth/auth.scss
```diff
@@ -0,0 +1,3 @@
+.btn-lit-element {
+    background: red !important;
+}
```
sbrunner, line 3: Should be a CSS, we plan to remove all SCSS...

sbrunner (commented):
@ger-benjamin do you know how to remove the shadow DOM?

PR #7404, opened 2021-08-30
Authentication web component

llienher:
panel.ts -> dispatch an event to angular html/controller with ng-on (could be generic at some point)
form.ts -> re-implement the gmf/authentication/component.js

Todo and known issues:
- Password change form and handling
- Fetch the logged status at page loading
- First error message is not displayed after page loading
- Logout seems to reload the page

Services used:
gmf/authentication/Service.js
ngeo/message/Message.js
ngeo/message/Notification.js


contribs/gmf/apps/service.js
```diff
@@ -0,0 +1,41 @@
...
+
+    /** @type {import("gmf/authentication/Service.js").AuthenticationService} */
+    auth: null,
+
+    /** @type {import('gmf/authentication/Service.js').User} */
+    user: null,
+
+    /** @type {import("ngeo/message/Notification.js").MessageNotification} */
+    notification: null,
+};
```
sbrunner, line 41: For me, this file should be in the folder src...

contribs/gmf/apps/appmodule.js
```diff
@@ -26,8 +26,9 @@
  */
 
 import angular from 'angular';
+import 'regenerator-runtime/runtime';
```
sbrunner, line 29: Should be in `AbstractAppController.js`

src/auth/panel.ts
```diff
@@ -0,0 +1,45 @@
...
+  }
+  // Disable shadow DOM
+  protected createRenderRoot() {
+    return this;
+  }
+
+  closePanel() {
+    this.dispatchEvent(new CustomEvent('close-panel', {detail: false}));
+  }
+}
```
sbrunner, line 45: `component.ts` isn't a better name?

sbrunner (commented):
For me, that looks good, I just purpose some renaming, reorganization, @ger-benjamin @llienher  do you agree with that?

contribs/gmf/apps/service.js
ger-benjamin: For me too, also because we have to build the gmf-apps (as it's in apps) to run the (ongoing) storybook example

src/auth/panel.ts
ger-benjamin: He's next to the ngeo-auth-form. Which one should be named component ? For me, it's more the "form" than the "panel". And we could have a `panel.ts` for each component in a panel in the apps.

llienher: Yes it's my thought, I will rename form.ts and keep the panel.ts filename.

ger-benjamin (commented):
Some comments, but looks good.

contribs/gmf/apps/appmodule.js
llienher: Done.

contribs/gmf/apps/desktop_alt/index.html.ejs
```diff
@@ -137,6 +141,11 @@
               </div>
             </div>
           </div>
+          <ngeo-auth-panel ng-show="mainCtrl.ngeoAuthActive"
+          .loginInfoMessage="mainCtrl.loginInfoMessage"
```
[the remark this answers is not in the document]
llienher: It's property binded to an expression (value from the Angular controller here).
https://lit.dev/docs/templates/expressions/#property-expressions

contribs/gmf/apps/service.js
llienher: As explained in the meeting, I will use a more usefull name but not put it in auth folder as it will be the name of the current authentication/service.js that is in angular. I'm going to use root of src/ and import it as 'AngularServices' in the files.

sbrunner (approved):

PR #7417, opened 2021-09-01
RxJS User implementation

ger-benjamin:
A complete integration of the new User based on the rxjs store.
We can discuss on that Friday. There is other implementation possibility, I think this one is good.

Some interesting points:

- It was already working on the (small) first commit. But the original implementation is strange (because `gmfUser` was a constante).
- Simplify the code at some points: no more mix of OpenLayer and ngeo events, no injection... 
- My last commit set back the userState (logged in, logged out, ready...) to a simple string.  Reason in the commit message:
```
With two observable, if you subscribe to one, you have
no guarantee that the other is already up to date. Even
if they are updated at the same time !
```
- I've seen that the (old/gmf) auth component is initialized three times by AngularJs, but only destroyed one time ! There is only one component, I don't get why we have two time this component initialized at the same time. It's risky...

### DIFF
--- a/contribs/gmf/apps/appmodule.js
+++ b/contribs/gmf/apps/appmodule.js
@@ -54,7 +54,6 @@ appModule.config([
  */
 appModule.run(($injector) => {
   AngularServices.auth = $injector.get('gmfAuthenticationService');
-  AngularServices.user = $injector.get('gmfUser');
   AngularServices.notification = $injector.get('ngeoNotification');
 });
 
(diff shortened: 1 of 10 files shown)


fredj (approved):
Looks very good !

sbrunner (approved):

ger-benjamin:
Finally I've decided to fix the test. That was good idea. I've found errors (in my code, but also in the old test, and bad code in the service)

PR #7422, opened 2021-09-03
Auth web-component

llienher:
Implement the missing bits of the first implementation.

Uses the RxJS implementation to set the gmfUser object and get the updates and state.
Fixes few issues with the stories and add missing property to user store.

Untested: change password form.


vuilleumierc (commented):
Tested the login and it works here :+1: (though the logout does not seem to work)

llienher:
@vuilleumierc @sbrunner The logout is now fixed (finally).

ger-benjamin (approved):
Looks good :-)

sbrunner (approved):

sbrunner (approved):

PR #7423, opened 2021-09-03
Convert js code with jsdoc as ts

sbrunner:
To use it, use e.g.:
`npm run tots -- --dry --print api/src/Map.js` # to test
`npm run tots -- api/src/Map.js`

Actually, the variable, cast are not completely converted....

Not working:
- convert more than one file in one shot
- all the type cast are not working
- `@type {{ ... }}`
- `function(Type1|Type2): ResultType`


sbrunner:
Should we remove all the .js in the imports?
I think that we can do that with another transformer to facilitate the script renaming...

sbrunner:
I just notes that `@param {function(angular.IAttributes): string}` is not working...

sbrunner:
Should we do something with the enum?

gberaudo:
> Should we do something with the enum?

Let's try without them and see.

ger-benjamin:
Nice work !

> Should we remove all the .js in the imports?
I think that we can do that with another transformer to facilitate the script renaming...

Yes let do that in another script if that's easier. Or here after the addImport/removeImport functions. As you want. 
We just need enough documentation to know that we have to run two command. (or create one command that run both scripts). That will be useful for projects with custom dev based on ngeo.

> Should we do something with the enum?

It would be great if we can transforms them to typescript enums: https://www.typescriptlang.org/docs/handbook/enums.html

sbrunner:
Ready for review :-)

buildtools/codeshift-js2ts.js
```diff
@@ -0,0 +1,991 @@
...
+      .forEach((path) => {
+        visit('', path.value, (indent, node) => {
+          // if (node.comments) {
+          //   console.log(node.comments[0].value);
+          // }
+          // if (node.innerComments) {
+          //   console.log('innerComments');
+          //   console.log(node.innerComments[0].value);
+          // }
+          // console.log(indent + indent + node.type);
```
llienher, line 587: Is it usefull ?

docs/codeshift.md
```diff
@@ -0,0 +1,30 @@
+# JSCodeShift scripts
+
+We provide tow jscodeshift scripts not migrate the code from JavaScript with types in JSDoc to TypeScript.
```
llienher, line 3:
```suggestion
We provide two jscodeshift scripts to migrate the code from JavaScript (with types in JSDoc) to TypeScript.
```

docs/codeshift.md
```diff
@@ -0,0 +1,30 @@
...
+
+The first one removes the '.js' in the import, then when we convert a script from `.js` to `.ts` we didn't
+need to change all the relative import.
+Then we should run it first on all the repository with: e.-g.
```
llienher, line 7:
```suggestion
The first script removes the '.js' in the import. That way, when we convert a script from `.js` to `.ts` we don't
need to change all the relative imports.
We should run it first on all the repository with: e.-g.
```

docs/codeshift.md
```diff
@@ -0,0 +1,30 @@
...
+git mv <file_path>.js <file_path>.ts
+npm run cs-tots -- <file_path>.ts
+```
+
+Conversion limitations:
+
+- I use global variable then we shouldn't convert more than one file in one shot.
+- Not all the type cast are working.
+- `@type {{ ... }}` are not working, it should be converted into an `@typedef` first.
+- `function(Type1|Type2): ResultType` is not working the parsing with `doctrine` return an incorrect result.
```
llienher, line 30:
```suggestion
- I use global variable so we shouldn't convert more than one file in at a time.
- Not all the type casts are working.
- `@type {{ ... }}` is not working, it should be converted into an `@typedef` first.
- `function(Type1|Type2): ResultType` is not working, the parsing with `doctrine` return an incorrect result.
```

buildtools/codeshift-js2ts.js
```diff
@@ -0,0 +1,991 @@
...
+        quoteProps: 'preserve',
+        parser: 'babel',
+      };
+      return prettier.format(result, config);
+    } catch (error) {
+      console.log('Fail to format the result');
+      return result;
+    }
+    //  } catch (error) {
+    //    return result;
```
llienher, line 985: Is it normal to catch error in the finally{} ?

llienher (changes requested):
I focused my review mainly on the doc and how to use the scripts. The doc need a few changes to be easily readable. :-)

buildtools/codeshift-js2ts.js
sbrunner: To investigate yes :-)

buildtools/codeshift-js2ts.js
sbrunner: It's not in the finally, but I remove it...

ger-benjamin (approved):
Quick review, looks good
Thanks a lot

llienher (approved):

PR #7432, opened 2021-09-07
Better check that user exists

ger-benjamin:
Fix for the new auth component (GSGMF-1651)

In some cases, the server return a user with only some properties (like no username). I've added some checks to be sure to set the model with a complete User object.

One observation:
As RXJS make a good works, I've tried to add back the Observable on the user state (instead of a string) (in my first analyze, I was thinking that observing the state and getting the user's properties can give you only an half object of the user. It's wrong, that was the server's fault). But RXJS works procedural. In the model I set the state before the properties. If I observer the state, and try to get the properties, I get the old properties not the new. So finally I keep the state as a single string.

### DIFF
--- a/contribs/gmf/src/authentication/Service.js
+++ b/contribs/gmf/src/authentication/Service.js
@@ -55,15 +55,6 @@ import user, {UserState} from 'ngeo/store/user.ts'
  */
 
 /**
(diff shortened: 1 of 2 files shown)


llienher (approved):
Looks good !

PR #7440, opened 2021-09-10
Consider *.ts files as js in the Makefile

sbrunner:


src/store/user.ts
```diff
@@ -100,31 +122,33 @@ export class UserModel {
...
 
   /**
    * Set the current User's properties and state.
+   * @param properties
+   * @param state
    */
   setUser(properties: User, state: UserState) {
     this.checkUserProperties_(properties);
-    console.assert(state == null);
+    console.assert(state === null);
```
ger-benjamin, line 145: Well, `==` was the aim. But I was surprised to no have an error.
Anyway. Merge it like that. I'll set that to `state !== undefined && state !== undefined` in my PR about the tests

ger-benjamin (approved):

src/auth/component.stories.ts
```diff
@@ -1,4 +1,24 @@
-import {html} from 'lit-html';
```
llienher, line 1: My guess nothing should work if you remove this line ... :-)

llienher: Oups, no it's the stories not the web-component ... So it's fine.

PR #7443, opened 2021-09-13
Fix properties

sbrunner:


ger-benjamin (approved):
Looks good to me.
One remark (but more for @llienher ), we have lost these attributes ?
https://github.com/camptocamp/ngeo/blob/master/contribs/gmf/src/authentication/component.js#L121-L122

llienher:
> One remark (but more for @llienher ), we have lost these attributes ?
> https://github.com/camptocamp/ngeo/blob/master/contribs/gmf/src/authentication/component.js#L121-L122

@ger-benjamin passwordValidator is in the file, but the other not. It seems I did forget to do the binding in the current implementation to the web-component...

llienher (approved):

ger-benjamin:
@sbrunner the example will "fail" we only have the property "loginInfoMessage" now:
You probaly should remove https://github.com/camptocamp/ngeo/blob/poc_2.7/src/auth/component.stories.ts#L32-L35

PR #7452, opened 2021-09-14
No more use OpenLayers eslint config

sbrunner:
- Remove dependency on `eslint-config-openlayers`
- Test all the js files
- Set some warning on js files as error on ts files (is the IDE you will see them as warning)
- Apply all the fix


ger-benjamin (commented):
Can you add the transformation to `returns` instead of `return` in the `buildtools/codeshift-js2ts.js` ?

sbrunner:
> Can you add the transformation to `returns` instead of `return` in the `buildtools/codeshift-js2ts.js` ?

It's already the case :-)

PR #7458, opened 2021-09-15
Bindings for passwordValidator

llienher:
PasswordValidator addition was straight-forward.

Final decision to drop onSuccessfulLogin, we can override it from the angular controller if needed.

Previously:
I have to assign the value of onSuccessfulLogin when the Angular controller is done initializing. Otherwise trying to do it outside the store getProperties().subscribe() get the value undefined and will not change later, as the Angular controller is not ready before the DOM of the web-component and there is no compatible watcher to lit-element.

On the matter of not using updated (https://lit.dev/docs/components/lifecycle/#reactive-update-cycle-completing):
The updated(changedProperties) works as watcher on attributes/properties in general, but not on @property({attribute: false}) as we need to bind the variable to a function and Lit only support string, boolean, number, array or object types.

### DIFF
--- a/src/auth/component.ts
+++ b/src/auth/component.ts
@@ -26,7 +26,6 @@ import AngularServices from 'ngeo/services';
 import {MessageType} from 'ngeo/message/Message.js';
 import {unsafeSVG} from 'lit/directives/unsafe-svg.js';
 import loadingSvg from 'gmf/icons/spinner.svg';
-import {AuthenticationLoginResponsePromise} from 'gmf/authentication/Service';
 import {gmfBackgroundlayerStatus} from 'gmf/backgroundlayerselector/status.js';
 import {Subscription} from 'rxjs';
 import user, {User, UserState} from 'ngeo/store/user';
@@ -44,9 +43,6 @@ type PasswordValidator = {
 export default class ngeoAuthComponent extends LitElementI18n {
   @property({type: String}) loginInfoMessage = '';
   @property({type: Object}) private passwordValidator: PasswordValidator = null;
-  @property({attribute: false}) onSuccessfulLogin: (
-    resp: AuthenticationLoginResponsePromise
-  ) => AuthenticationLoginResponsePromise = undefined;
   @state() private isLoading = false;
   @state() private disconnectedShown = false;
   @state() private resetPasswordShown = false;
@@ -70,7 +66,6 @@ export default class ngeoAuthComponent extends LitElementI18n {
           this.setOtpImage_();
           this.checkUserMustChangeItsPassword_();
           this.onUserStateUpdate_(user.getState());
-          this.onSucessfullLoginCallback_();
         },
       })
     );
@@ -309,15 +304,6 @@ export default class ngeoAuthComponent extends LitElementI18n {
     this.userMustChangeItsPassword = true;
   }
 
-  /**
-   * @private
-   */
-  onSucessfullLoginCallback_() {
-    if (this.onSuccessfulLogin) {
-      AngularServices.auth.onSuccessfulLogin = this.onSuccessfulLogin;
-    }
-  }
-
   // METHODS THAT CALL THE AUTHENTICATION SERVICE METHODS
 
   /**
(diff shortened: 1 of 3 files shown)


ger-benjamin (commented):
I'm not sure to understand this `onSuccessfulLogin`. Even on the ngeo code, it looks badly implemented.

As far as I understand, we can give a `onSuccessfulLogin` function to the component to override the one in the service (that's already pretty strange).

To do the same, we should call `onSucessfullLoginCallback_` in the `connectedCallback` of the component, but **not inside** de `user.getProperties().subscribe` function.
No ?

llienher:
Yes, it overrides the one in the service. I don't know who need such an implementation, but yes it is done the same in ngeo...

And as much I would do as you said, this is not really working. The DOM and the web-component are loaded before Angular controller. When the controller is called, the function is available to the $scope and passed trough the DOM. But the web-component already did the `connectedCallback` and will not do it again, so the value is not available and stay to undefined.

Usually if the property is not set to `attribute: false`, the `updated` or `firstUpdated` could be used to detect the change on the watched properties. But as I use a function I had to disable that, because `function` type is not supported. The life-cycle methods that are available are not working out of the box for this use-case. As I have no way to makes it work, I used the side-effect that `user.getProperties().subscribe` is runned after the controller initialization to retrieve the value ... But of course I'm open to proposal to get a better implementation.

`loginInfoMessage` has the same issues and limitations, but it is not visible because if the value changes, Lit watch on it and trigger an update of the UI to render the change. `loginInfoMessage` is `attribute: true`, the default value and type `string`, which is supported

ger-benjamin:
We will have this problem all along the project. We must find a standard and clean way to solve this.

Delay the loading of web-component is probably not possible.

Maybe we could launch an event when AngularJS is ready. Then we could import and observe this event once in the component to do the stuff we have to do once AngularJS is ready.

ger-benjamin:
So something like that maybe:

```
// File AngularJsReady.ts
import { Subject } from 'rxjs';

const ready = new Subject();
export default ready;

```
```
// In a main controller
import ready from 'AngularJsReady';

...
export function AbstractAppController($scope, $injector, mobile) {
...
ready.complete();
...
}
....

```
```
// In a component
import ready from 'AngularJsReady';
...
@customElement('ngeo-auth-component')
export default class ngeoAuthComponent extends LitElementI18n {
  ...
  ready.subscribe({complete: ()=> this.onAngularJSReady_()});
  ...
}
```

llienher:
Ok, I will look if this is possible tomorrow, thanks of the idea.

ger-benjamin:
Well, if the next step is to remove all services, we could also remove leave it like this...

llienher:
We will not pass onSuccessfulLogin from the controller to the web-component, as it is not UI-related.
We can still assign it from the controller until we remove Angular completely.

I will just use this PR for the binding of passwordValidator.

src/auth/component.ts
```diff
@@ -102,14 +97,18 @@ export default class ngeoAuthComponent extends LitElementI18n {
...
+                          value=${i18next.t('Logout') as string}
+                        />
                       </div>
                       <div class="form-group">
                         <input
                           ?hidden="${!this.allowPasswordChange}"
                           type="button"
                           class="form-control btn btn-default"
-                          value=${i18next.t('Change password')}
+                          value=${i18next.t('Change password') as string}
```
llienher, line 111: @ger-benjamin @sbrunner It removes errors in my IDE as the i18next object is not typed as a string as needed by the HTML spec, so casting it like that remove the error. It initially doesn't break the build or running the app, just getting dozens of errors that aren't relevant in my IDE makes it hard to spot real errors while working.

It is okay for you to do that for the translated string when needed ?

llienher: Actually the CI does not like it ...

llienher: It is an initial error from the lit-plugin, my fix made an typescript error...

sbrunner (approved):

PR #7483, opened 2021-09-22
Fix the clean of gh-pages

sbrunner:


ger-benjamin (approved):
What's the issue ? Can't we clean the branch ?
(Well, then, amending the last commit each time is not a problem for this branch)

sbrunner:
> What's the issue ? 

It seems what when we push the content to the gh pages branch we reset to the first commit of the branch then we lost all the delete...

sbrunner:
> Can't we clean the branch ?

I don't understand this question ...

ger-benjamin:
> Can't we clean the branch ?
> I don't understand this question ...

Forget that, it doesn't make sense now that you explained me the issue

PR #7500, opened 2021-10-01
Be able to configure the api name in simple application mode

sbrunner:


llienher (commented):
Correct me if I'm wrong but it seems to me that the constructor is executed before the timeout in the constant is done getting the constants, looks like the constants.package is always empty in Map.js ? Maybe I did interpret it wrongly as it should work ?

PR #7511, opened 2021-10-06
Include required polyfills with Babel

sbrunner:


ger-benjamin (commented):
Why do you remove the 3d examples ?

sbrunner:
Good question, It's the only example that have an error with regenerator-runtime event if I add it directly to the example, and this is not used by the UserGroup...

ger-benjamin:
Maybe it would be better to comment them and create a ticket to restore it. That's annoying to trash them... (as long as we don't have a storybook example instead)

ger-benjamin (approved):
As discussed, let's remove these examples. They are not used anymore and not official supported.
LGTM

ger-benjamin (approved):

llienher (approved):

PR #7549, opened 2021-10-13
Reset time from timeslider like datepicker on component reset

ger-benjamin:
Fix GSGMF-1687

Bug: deactivate and reactive a layer with a timeslider makes the slider reset, but not the state of the map.

Now: Reset the layer time state accordingly with the map/permalink.

Edit will be backported for GEO-4876


ger-benjamin:
Mhh it doesn't works in some cases. If you set the timeslider, and you refresh the page, then deactivate a reactivate the layer, the slider will be desynchronized. I'll work further on this one.

PR #7560, opened 2021-10-19
Add vector tiles

ger-benjamin:
For  GSGMF-1684

Works, but print is currently not set because printing in general doesn't work


contribs/gmf/apps/build/locale/webcomponent/en/app.json
```diff
@@ -0,0 +1,38 @@
...
+  "Incorrect old password.": "Incorrect old password.",
+  "The username is required.": "The username is required.",
+  "The password is required.": "The password is required.",
+  "Incorrect credentials or disabled account.": "Incorrect credentials or disabled account.",
+  "Could not log out.": "Could not log out.",
+  "Please, input a login...": "Please, input a login...",
+  "An error occurred while resetting the password.": "An error occurred while resetting the password.",
+  "Loading themes, please wait...": "Loading themes, please wait...",
+  "Login": "Login"
+}
```
sbrunner, line 38: This should not be committed...

contribs/gmf/src/datasource/Manager.js
```diff
@@ -611,7 +617,7 @@ export class DatasourceManager {
       }
     }
 
-    // (5) ogcServer
+    // (6) ogcServer
```
sbrunner, line 620: Shouldn't we just remove the steps number?

sbrunner (approved):

sbrunner (commented):
.

PR #7604, opened 2021-10-29
Try fake ui tests

ger-benjamin:
Test to give a fake "UI test" signal to the PR, at least to know if it works.
Has it has to work with dependabot, I'll merge it.

See also this commit added after this PR https://github.com/camptocamp/ngeo/commit/416af9093c8256b8ca001e2f5da730bc5227cee9

See also: https://github.com/camptocamp/ngeo/pull/7606 (run it only with dependabot)

### DIFF
--- a/.github/workflows/fake_ui_tests.yaml
+++ b/.github/workflows/fake_ui_tests.yaml
@@ -1,0 +1,27 @@
+---
+# UI tests are not launched by dependabot (because it has no access
+# to the secrets. Run this check to send a valid "UI tests" signal.
+
+name: Fake UI tests
+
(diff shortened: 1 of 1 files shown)


sbrunner:
As I understand, this will drop the chromatic test on the Dependabot pull-request, And we need them!
Use the `get-chrmatic-on-dependabot` label for that!

ger-benjamin:
No, it was dropping the chromatic test only on PR from dependabot. (and these PR can't run chromatic)
What is "get-chrmatic-on-dependabot" ? I can't find any information about that, nowhere. And PR are not passing now...

PR #7527, opened 2021-10-08
Add desktop application canvas

sbrunner:
Create apisrc folder
Export desktop alt as a lib
Add some documentation

PR #7677, opened 2021-11-12
Panel resizing

vuilleumierc:
[GSGMF-1607](https://jira.camptocamp.com/browse/GSGMF-1607) second part (first part: https://github.com/camptocamp/ngeo/pull/7527)

To do:
- [x] Resize left panel (after https://stackoverflow.com/a/55202728/17220956)
- [x] Resize right panel
- [x] Collapse panels and add >> icon
- [x] Fix console error `Uncaught TypeError: panels is null`
- [x] Fix test (error on footer panel closing: `You try to close a panel that's not open`)
- [x] Remove SASS variables
- [x] Update reference image for desktop_alt 

Not addressed in this PR (-> https://jira.camptocamp.com/browse/GSGMF-1697)
- Resize map
-  Set minimum panel size
- Consume events


llienher:
One thing that I see as not really working right now is that the panels don't have a min or max values when we move it from the selector. In the desktop interface, there is a limit you can widen or reduce the panel from the button. Then if you want to get smaller, you collapse it from the button to close it entirely. If it is not too much work to add this. :-)

ger-benjamin (approved):
Address comment from the last meet and it will be fine for me:
 - Move todo point to a jira ticket
 - Set a minimal width for the panel
 - remove sass variable (or add this point in the jira ticket)

sbrunner (approved):

llienher (approved):

PR #7735, opened 2021-11-24
Display login message when authentication is required on private layers

llienher:
Probably not the cleanest implementation possible, but do the job with the following requirements:

- Switch to another tool panel should reset the login message to empty state
- Close the panel or collapse it should reset the login message to empty state
- If login is required will both broadcast an event for desktop the AngularJS way and use the store to change the store Observable for desktop_alt

Missing:

- Translation of the message value

NB: I had to fix the promises adding some catch() and the gmfUser naming in the permalink.js so it is a bit of a mess for this file...

### DIFF
--- a/src/auth/FormElement.ts
+++ b/src/auth/FormElement.ts
@@ -44,8 +44,8 @@ export type PasswordValidator = {
 
 @customElement('gmf-auth-form')
 export default class GmfAuthForm extends GmfBaseElement {
-  @property({type: String}) loginInfoMessage = '';
   @property({type: Object}) private passwordValidator: PasswordValidator = null;
+  @state() private loginInfoMessage = '';
   @state() private isLoading = false;
   @state() private disconnectedShown = false;
   @state() private resetPasswordShown = false;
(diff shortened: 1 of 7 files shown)


contribs/gmf/apps/desktop_alt/index.html.ejs
```diff
@@ -55,7 +55,9 @@
 
       <!-- Login -->
       <gmf-auth-button slot="tool-button"></gmf-auth-button>
-      <gmf-auth-panel slot="tool-panel-auth"></gmf-auth-panel>
+        <gmf-auth-panel slot="tool-panel-auth"
+          ng-prop-login_info_message="mainCtrl.loginInfoMessage">
```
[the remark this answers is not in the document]
llienher: Yes, I fixed it. :-)

src/auth/FormElement.ts
```diff
@@ -70,6 +70,11 @@ export default class GmfAuthForm extends GmfBaseElement {
           this.checkUserMustChangeItsPassword_();
           this.onUserStateUpdate_(user.getState());
         },
+      }),
+      user.getLoginMessage().subscribe({
+        next: (message) => {
+          this.loginInfoMessage = message;
+        },
```
sbrunner, line 77: line 47 =>
```
@state loginInfoMessage = '';
```
and remove loginInfoMessage in PanelElement...

sbrunner (approved):
Just one comment ;-)

src/auth/FormElement.ts
llienher: Done.

PR #7769, opened 2021-12-01
Complete collapse of data panel

vuilleumierc:
[GSGMF-1708](https://jira.camptocamp.com/browse/GSGMF-1708)
To do: 
- [x]  Move both toggle button (1px)


contribs/gmf/src/controllers/vars_desktop.scss
```diff
@@ -51,4 +52,7 @@ $notification-width: 12.5rem !default;
   --right-panel-width-googlestreetview: 25rem;
   --right-panel-width-mapillary: 25rem;
   --left-panel-width: #{$left-panel-width};
+  --panel-separator-width: #{$panel-separator-width};
+  --current-left-panel-separator-width: #{$panel-separator-width};
+  --left-panel-border-width: #{$border};
```
sbrunner, line 57: I think that good if you rename it to `--current-left-panel-border-width` then we know that it can't be set by the config...

sbrunner (commented):
It's happening something strange with the infobar, she became visible on the footer panel...

vuilleumierc:
> It's happening something strange with the infobar, she became visible on the footer panel...

Thanks for having a look, this should be fixed now.
@llienher also noticed it was still possible to resize the panel when it is closed by clicking on the collapse/toggle button -> I will try to fix that now

sbrunner:
It looks like that you need to rebase ...

vuilleumierc:
It's strange because I already did (the branch has the commit b7040c82aadad7fc20922da0be33b5b33e0285f5)
From the diff it looks like there's a problem with the infobar, but it is both in the reference image and in the PR image.
Do you understand why it's not passing? Should I replace the reference image?

sbrunner:
> Should I replace the reference image?$

Yes :-)

sbrunner:
In the new screenshot the handle is misplaced to 1px to the left...

llienher (approved):

PR #7799, opened 2021-12-08
Fix resize map

llienher:


vuilleumierc (approved):

sbrunner:
`flex-shrink` should be set to `0` on `#gmf-app-panel-separator-container`...

sbrunner:
The print mask display hide is delayed on mouse move ...

llienher:
> The print mask display hide is delayed on mouse move ...

@sbrunner Not related to this PR, seems to be the same on master now ...

sbrunner:
Yes, it can be added to the issue or in a new issue :-)

llienher:
@sbrunner I will not fix print stuff in this PR/issue, so we make a new one. :-)

sbrunner (approved):

PR #7807, opened 2021-12-10
Fix audit

sbrunner:
Use an older version of npm
Remove no more needed asdf
Use the latest version of c2csiutils


arnaud-morvan (approved):
LGTM

PR #7808, opened 2021-12-10
Correctly toggle print mask

llienher:
As we checked that the panel was not null in abstract controller, we did not set the value to false when the panel was closed so the mask in the print wasn't synced anymore with the correct value.

### DIFF
--- a/contribs/gmf/src/controllers/AbstractDesktopController.js
+++ b/contribs/gmf/src/controllers/AbstractDesktopController.js
@@ -168,12 +168,8 @@ export class AbstractDesktopController extends AbstractAPIController {
 
     panels.getActiveToolPanel().subscribe({
       next: (panels) => {
-        if (panels !== null) {
-          const newVal = panels.includes('print');
-          if (newVal != this.printPanelActive) {
-            this.printPanelActive = newVal;
-          }
-        }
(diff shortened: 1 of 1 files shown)


llienher:
@sbrunner Not working quite right now ...I just realized that you have to put the mouse on the map to trigger the mask update... 

It is triggered by onPointerMove_ method in the src/raster/component.js file, that I never heard about ... :-(

contribs/gmf/src/controllers/AbstractDesktopController.js
```diff
@@ -166,10 +166,13 @@ export class AbstractDesktopController extends AbstractAPIController {
      */
     this.routingPanelActive = false;
 
+    const $timeout = $injector.get('$timeout');
+
     panels.getActiveToolPanel().subscribe({
       next: (panels) => {
         panels === 'print' ? (this.printPanelActive = true) : (this.printPanelActive = false);
-        $scope.$apply();
+        $timeout(() => {}); // this triggered on DOM click, we call $timeout to force Angular diggest
```
sbrunner, line 174: We will make a fiesta when we will remove toes kind of code :-)

sbrunner (approved):

PR #7816, opened 2021-12-13
Fix this in WFSAttribute.read

arnaud-morvan:


sbrunner (approved):
I'm confident to you :-)

PR #7838, opened 2021-12-17
Merge ngeo/gmf

llienher:
Implements GSGMF-1654


src/sass/font.scss
```diff
@@ -47,3 +47,5 @@
 @import '~@fortawesome/fontawesome-free/scss/fontawesome.scss';
 @import '~@fortawesome/fontawesome-free/scss/regular.scss';
 @import '~@fortawesome/fontawesome-free/scss/solid.scss';
+@import '~gmf/sass/vars.scss';
+@import '~ngeo/sass/font.scss';
```
sbrunner, line 51: Why this?

llienher: It was in `contribs/gmf/src/sass/font.scss`

llienher: But I did not checked what I reported ... :-)

sbrunner: I think that better if we have two different files, the CSS generation is not strong then we wanted to avoid breaking it ...

llienher: Ok, I did a new file called `gmf-font.scss` that contains the old file from contrib, and fixed the import that used this.

src/message/displaywindowComponent.js
```diff
@@ -21,7 +21,7 @@
 
 import angular from 'angular';
 
-import 'ngeo/sass/font.scss';
+import 'ngeo/sass/gmf-font.scss';
```
sbrunner, line 24: C'est `gmf/sass/font.scss` pas `ngeo/sass/font.scss`...

.eslintrc.yaml
```diff
@@ -34,16 +33,6 @@ rules:
...
-  jsdoc/require-param-type: off
-  jsdoc/require-property-type: off
-  jsdoc/require-returns-type: off
-  jsdoc/check-tag-names: warn
-  # not compatible with jsdoc
-  jsdoc/check-types: off
-  jsdoc/no-undefined-types: warn
-  jsdoc/require-returns-description: warn
-  jsdoc/require-param-description: warn
-  jsdoc/require-property-description: warn
```
sbrunner, line 46: Do you verify that's checked by @typescript-eslint?

sbrunner (approved):

PR #7839, opened 2021-12-17
Fix query on unwanted panel

sbrunner:
Fix: https://jira.camptocamp.com/browse/GSGMF-1715


llienher (changes requested):
We can still activate the query when the drawing is active, we are not supposed to get WFS query result with this tool when we draw a geometry.

sbrunner:
@llienher fixed, thanks :-)

llienher (approved):
LGTM, thanks !

PR #7852, opened 2021-12-22
Fix print legend - iter on themes node

ger-benjamin:
For https://jira.camptocamp.com/browse/GSGMF-1702

Todo:
- [x] Generate first items based on the theme
- [x] More manual test
- [x] Check order
- [x] Check titles
- [x] Add comments
- [x] Fix typecheck
- [x] Adapt tests
- [x] test again and check CI


arnaud-morvan:
Maybe this is outside scope of this PR, but with a GMF layer with two WMS layers (with static legend image), I get two legend entries with the same image.

GMF Layer with 2 WMS layers

ger-benjamin:
@arnaud-morvan This bug is also in the current 2.6
I've fixed it in this PR, thanks !

sbrunner:
Layers name is a come separate list of layer, not an array ...

ger-benjamin:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/includes How do you have troubles ?

sbrunner:
But here it's contains
https://github.com/camptocamp/ngeo/pull/7852/files#diff-40a841aa51df586098f556d70e97b7f8d3ae30d81bf4dcd04eedd807b64a9a50R189

sbrunner:
And for me, it should be something like: `layerNames.split(',').every(NodeWms.layers.split(','))`

ger-benjamin:
As you can see, my brains is really confused between `includes` and `contains`...
I'll use `includes` as I wanted to do. It's safe enough and probably better if we have a wms layer `poste,police` (two wms in one)

ger-benjamin:
@sbrunner 
 - It's ready
 - I've sometimes a strange error in tests (in  `contribs/gmf/test/spec/services/treemanager.spec.js`, not a file I have touched, but I've modified the test data of the themes.json). I hope it's not too often. Now it looks working...
 - I have added some `@ts-ignore` in the test of the legend, It's probably possible to do better, but I'm out of time.

sbrunner (commented):
Looks good to me @llienher, can you give a second look and validate it?

llienher (approved):
Looks good to me !

sbrunner (approved):

PR #7880, opened 2022-01-03
Add options name for search options

llienher:
For instances using more than one search input, we need to support multiple configs in the vars like we do for the streetview component.

### DIFF
--- a/contribs/gmf/src/search/component.js
+++ b/contribs/gmf/src/search/component.js
@@ -1,6 +1,6 @@
 // The MIT License (MIT)
 //
-// Copyright (c) 2015-2021 Camptocamp SA
+// Copyright (c) 2015-2022 Camptocamp SA
 //
 // Permission is hereby granted, free of charge, to any person obtaining a copy of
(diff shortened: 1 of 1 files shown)


contribs/gmf/src/search/component.js
```diff
@@ -378,6 +392,13 @@ export class SearchController {
    * Called on initialization of the controller.
    */
   $onInit() {
+    const config = this.optionsName || 'gmfSearchOptions';
+
+    /**
+     * @type {import('gmf/options.js').gmfSearchOptions}
+     */
+    this.options = this.injector_.has(config) ? this.injector_.get(config) : null;
```
sbrunner, line 400: Why not simply having:
```suggestion
if (this.optionsName) {
   this.options = this.injector_.get(this.optionsName);
}
```

llienher: I guess there is actually no need to use if at all, and just load the config from the name with the injector...

sbrunner (approved):

PR #7881, opened 2022-01-03
Add timeout used to fetch the WMTS capabilities

sbrunner:
GSGMF-1694


llienher (approved):
Hopefully we won't be confused with seconds/milliseconds while maintaining this. :-)

PR #7904, opened 2022-01-06
Allows the placeholder to change dynamically with the search config

llienher:


contribs/gmf/src/search/component.js
```diff
@@ -293,10 +293,6 @@ export class SearchController {
      */
     this.searchActionCallback;
 
-    /**
-     * @type {string}
-     */
-    this.placeholder = this.options.placeholder || 'Search…';
     gettextCatalog.getString('Search…');
```
sbrunner, line 296: This should also be moved

sbrunner (approved):

PR #7918, opened 2022-01-10
Pin colors to fix master

sbrunner:
See also: https://twitter.com/louloulfx/status/1480491799961645056

### DIFF
--- a/package.json
+++ b/package.json
@@ -117,6 +117,7 @@
     "chokidar": "3.5.2",
     "chromatic": "6.3.3",
     "co": "4.6.0",
+    "colors": "1.4.1",
     "commander": "8.3.0",
     "console-control-strings": "1.1.0",
     "copy-webpack-plugin": "6.4.1",


sbrunner:
Fixed upstream

llienher (commented):
It's the broken version !!!! So 1.4.0 for the one we need.
https://diff.intrinsic.com/colors/1.4.0/1.4.1
https://diff.intrinsic.com/colors/1.4.0/1.4.2

By the way I guess npm did pulled 1.4.1 and 1.4.2 so 1.4.0 should be the latest version availlable. We only use karma and jscodeshift that seems to use the latest tag. Jscodeshift did a fix using the chalk lib instead https://github.com/facebook/jscodeshift/blob/7cf9969c1fed3dcf535aebbc9986c32be9214ffc/CHANGELOG.md#0131-2022-01-10.
Karma is still using latest, but if I'm not wrong, the latest should use 1.4.0.

PS: I know the PR is closed, but it wasn't when I started writing that, for reference I keep it here.

PR #7938, opened 2022-01-12
Move and rename map/component to /src folder

llienher:


src/map/oldComponent.js
```diff
@@ -0,0 +1,128 @@
...
+        );
+      }
+    },
+  };
+}
+
+// Register the directive in the module
+myModule.directive('ngeoMap', mapComponent);
+
+export default myModule;
```
sbrunner, line 128: Can't we remove this file?

vuilleumierc: In the end yes (= in this PR), but I wanted to keep it until all the logic has been moved to the gmfMap (the resize window part - by the way do you know if it's still used? it seems to be the case for the mobile interface but I could not confirm)

src/controllers/mobile.scss
```diff
@@ -58,13 +58,11 @@ main button {
 gmf-map {
   position: static;
   text-align: initial;
-  > div {
     position: absolute;
```
sbrunner, line 61: You will have two positions...

src/dropfile/component.js
```diff
@@ -56,13 +56,12 @@ const fileDrop = function (gmfExternalDataSourcesManager, gettextCatalog) {
...
-    link: function ($scope, element, attrs) {
-      // @ts-ignore: scope ......
-      if ($scope.ctrl.fileDropEnabled !== true) {
-        return;
-      }
+    link: function ($scope, element) {
+      // // @ts-ignore: scope ......
+      // if ($scope.ctrl.fileDropEnabled !== true) {
+      //   return;
+      // }
```
vuilleumierc, line 64: commented out because I get an "undefined" error for `$scope.ctrl`, but I wonder if we should drop this attribute (`fileDropEnabled`) altogether since the `gmf-file-drop-zone` component is now in the index.html and not in the map component?

ger-benjamin: I think we can drop it. It looks to be there to (and I think I remember than it's there to) disable the drop on map where we don't want this feature. If it's in the index, we can already choose if we want this function or not.

ger-benjamin (commented):
Do we need to keep the src/map/oldComponent.js ?

sbrunner (approved):

PR #8063, opened 2022-02-03
Generic solution for the footer panel height

sbrunner:


src/canvas/DesktopElement.ts
```diff
@@ -400,6 +416,19 @@ export default class GmfDesktopCanvas extends BaseElement {
...
       panels.getActiveFooterPanel().subscribe({
         next: (panel: string) => {
+          let height = '0';
+          if (panel) {
+            const styles = getComputedStyle(document.documentElement);
+            height = styles.getPropertyValue(`--footer-panel-height-${panel}`);
+            if (!height) {
+              height = styles.getPropertyValue(`--footer-panel-height`);
+            }
+            if (!height) {
```
llienher, line 426: same condition as line 423 ?

sbrunner: Yes, but between we have a `height = ...`

llienher (approved):
Keep as it is (ignore my previous feedback) ! :-)

PR #8088, opened 2022-02-08
Migration composant lidar

llienher:


src/download/service.ts
```diff
@@ -0,0 +1,48 @@
...
+import {saveAs} from 'file-saver';
+
+/**
+ * A function to start a download for a file.
+ *
+ * @param {string} content The content of the file to download.
+ * @param {string} fileName The name of the file to download.
+ * @param {string | undefined} opt_fileType The type of the file to download.
+ */
+export function download(content: string, fileName: string, opt_fileType: string | undefined): void {
```
sbrunner, line 32: Why not default export?

llienher: What is the benefit of calling this something else when you want to use it ? :/

llienher: Yes, thanks. :-)

src/controllers/vars_desktop.scss
```diff
@@ -59,4 +59,9 @@ $notification-width: 12.5rem !default;
   --footer-panel-height: 12.5rem;
   --footer-panel-height-lidarresult: 21.88rem;
   --current-footer-panel-height: 0;
+
+  // LIDAR profile
+  --app-margin: 0.62rem;
+  --lidarprofile-height: 21.88rem;
+  --profile-legend-width: 9.38rem;
```
sbrunner, line 66: --footer-panel-height-lidarresult  should be in this section why --app-margin is in this section?

llienher: the CSS I added used this variable, I wanted to keep the code as similar as the initial one for the beginning.

sbrunner: Yes, but it's a common variable already defined as SCSS variable then it should be 
```
--app-margin: #{$app-margin};
```
some lines above :-)

src/controllers/vars_desktop.scss
```diff
@@ -57,6 +57,11 @@ $notification-width: 12.5rem !default;
...
   --current-left-panel-border-width: #{$border};
   --footer-panel-height: 12.5rem;
-  --footer-panel-height-lidarresult: 21.88rem;
+  --footer-panel-height-lidar: 21.88rem;
   --current-footer-panel-height: 0;
+
+  // LIDAR profile
+  --app-margin: #{$app-margin};
+  --lidarprofile-height: 21.88rem;
+  --profile-legend-width: 9.38rem;
```
sbrunner, line 66:
```suggestion
  --current-footer-panel-height: 0;
  --app-margin: #{$app-margin};

  // LIDAR profile
  --footer-panel-height-lidar: 21.88rem;
  --lidarprofile-height: 21.88rem;
  --profile-legend-width: 9.38rem;
```
Better organisation :-)

src/controllers/vars_desktop.scss
sbrunner: Why don't you want to put `--footer-panel-height-lidar` after the LIDAR comment?

sbrunner: Thanks :-)

srcapi/store/line.ts
```diff
@@ -0,0 +1,67 @@
...
+   *
+   * @param line The line
+   */
+  setLine(line: OlGeomLineString): void {
+    this.line_.next(line);
+  }
+}
+
+const line = new LineModel();
+export default line;
```
sbrunner, line 67: Should be moved to lidar...

src/lidar/Manager.ts
```diff
@@ -277,7 +277,11 @@ export class LidarprofileManager {
       this.isPlotSetup_ = false;
     }
 
-    d3select('#gmf-lidarprofile-container .lidar-error').style('visibility', 'hidden');
+    const lidarContainerElement = document
+      .querySelector('gmf-desktop-canvas gmf-lidar-footer')
```
ger-benjamin, line 281:
```suggestion
      .querySelector('#gmf-lidar-footer')
```
I would use an id here. That would limit "not element found" errors.
Same below, do a git grep please

src/map/FeatureOverlayMgr.ts
```diff
@@ -0,0 +1,191 @@
...
+  /**
+   * @param {OlMap} map Map.
+   */
+  init(map: OlMap): void {
+    this.layer_.setMap(map);
+  }
+
+  /**
+   * @param {OlStyleStyleStyleLike} style
+   * Style.
```
ger-benjamin, line 164: I would like to see more comment, but this one is too much
```suggestion
```

ger-benjamin (commented):
Not tested, but looks good

PR #8171, opened 2022-02-28
Fix login panel spinner

llienher:


src/sass/base.scss
```diff
@@ -1,7 +1,7 @@
 /**
  * The MIT License (MIT)
  *
- * Copyright (c) Camptocamp SA
+// Copyright (c) 2022 Camptocamp SA
```
sbrunner, line 4: Shoulfn't be like this
```suggestion
 * Copyright (c) 2022 Camptocamp SA
```

llienher: Yes :-)

src/auth/FormElement.ts
```diff
@@ -95,6 +95,15 @@ export default class GmfAuthForm extends GmfBaseElement {
...
       }
+      .login-spinner {
+        float: left;
+        margin-right: 20px;
+      }
+
+      i.fa-spin {
+        fill: black;
+        width: 1.3rem;
+      }
```
ger-benjamin, line 106: It is not possible to move that in a specific exported variable, like with the `...GmfBaseElement.styles` (line 93) ? Because this css was in the "global" css before shadow dom. And now, it will be probably needed in multiple components, and we don't want to write the same code everytime, everywhere. (But maybe put it in another variable, because we don't want it in each component, like for the `...GmfBaseElement.styles`)

ger-benjamin (approved):
It is not possible to move it in a

src/auth/FormElement.ts
sbrunner: In the original file it was on the class `.login-spinner` then it should be related only to the login, and I think that better to don't have too many shared CSS, as I know it's hard to manage it...

src/auth/FormElement.ts
```diff
@@ -95,6 +95,15 @@ export default class GmfAuthForm extends GmfBaseElement {
       [hidden] {
         display: none !important;
       }
+      .login-spinner {
+        float: left;
+        margin-right: 20px;
+      }
+
+      i.fa-spin {
```
sbrunner, line 103: Not?
```suggestion
      .login-spinner.custom-spinner-connect {
```
With direct conversion, we will get that...

src/auth/FormElement.ts
ger-benjamin: Oh ok, in this case it's fine. But in the case it would be shared between 3 components or more, the question would be interesting.

src/auth/FormElement.ts
llienher: I guess we don't need any class to just do that now. It is encapsulated in the web-component.

PR #8187, opened 2022-03-01
Set the print panel closed when we reduce it

llienher:
Fix [GSGMF-1718](https://jira.camptocamp.com/browse/GSGMF-1718)


sbrunner (approved):

ger-benjamin (commented):
Thanks

PR #8189, opened 2022-03-01
Edition: Fix undefined boolean to false

ger-benjamin:
One more fix for GSGMF-1749.

Better code and avoid "sanitizing" boolean twice

I'll backport it manually in master in #8184


sbrunner (approved):
I'm confident to you :-)

ger-benjamin:
In fact, here there are currently no bug, but the code will be better now (avoid sanitizing twice the boolean). But in 2.4, that will fix a small issue with boolean (default to true, instead of false)

PR #8201, opened 2022-03-02
Catch error on WFS query promise (in snapping)

llienher:
Fix [GSGMF-1734](https://jira.camptocamp.com/browse/GSGMF-1734)

I don't know if we should better use console.info() for the message of this error when it is an aborted promise result...


sbrunner (approved):
Thanks :-)

PR #8208, opened 2022-03-04
Add missing reject in deferrableOnSuccessfulLogin

arnaud-morvan:
Fix https://jira.camptocamp.com/browse/GSGMF-1752


sbrunner (approved):
Nice thanks :-)

PR #8264, opened 2022-03-21
Do not overwrite metadata while restoring layertree state

arnaud-morvan:


llienher (approved):
Tested, looks good to me !

PR #8313, opened 2022-04-01
Display & hide disclaimers when a layers is visible/hidden

llienher:
Propertychange listener is not runned at the layer level but on the group activation, so using rootscope allows us to trigger the disclaimer code when a layer is enabled or disabled.


sbrunner (approved):
Nice :-)
Thanks

PR #8350, opened 2022-04-08
Fix password reset and remove theme reloading at password reset

llienher:


src/auth/FormElement.ts
```diff
@@ -333,9 +349,6 @@ export default class GmfAuthForm extends GmfBaseElement {
    * @private
    */
   checkUserMustChangeItsPassword_(): void {
```
sbrunner, line 351: Shouldn't be named as `handleUserMustChangeItsPassword_`?

sbrunner (approved):

src/auth/FormElement.ts
llienher: I will rename it, it does the check outside so the name is not accurate anymore.

PR #8385, opened 2022-04-13
Fix Mapillary bounding box (configurable value and buffer from map resolution)

llienher:
Fix [GSGMF-1791](https://jira.camptocamp.com/browse/GSGMF-1791)


sbrunner (approved):
Looks good, thanks :-)

PR #8418, opened 2022-04-22
Set grid-gutter-width variable in the CSS

llienher:
Fix [GSGMF-1808](https://jira.camptocamp.com/browse/GSGMF-1808)


arnaud-morvan (approved):
Looks good, thanks

PR #8420, opened 2022-04-25
Add bufferSize parameter to mapillary options in example

llienher:
Fix missing example: [GSGMF-1812](https://jira.camptocamp.com/browse/GSGMF-1812)

PS: mapillary key seems wrong for the local example, I don't know if there is somewhere the right one (probably for the Azure demo but I don't have it)...


arnaud-morvan (approved):
LGTM

PR #8439, opened 2022-04-28
Makes body element focusable for onKeyDown event

llienher:
Non interactive elements such as body or div don't takes the focus when we want to handle keyboard events on them ... So it seems like we need to add the element to the tab index as the keyboard events is not currently handled. Tab index lists the elements that support keyboard focus.

https://developer.mozilla.org/fr/docs/Web/HTML/Global_attributes/tabindex


arnaud-morvan (approved):
LGTM, thanks

PR #8454, opened 2022-05-02
Fix cypress config and ci workflow

llienher:
Some things that were buggy when starting doing automation of testbook.

Cypress `integrationFolder` is not supporting multiple paths so we need to trick it with glob in the `testFiles` parameter...

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -90,6 +90,8 @@ jobs:
       - run: make check-examples-checker
         if: github.event_name != 'pull_request_target'
       # Cypress tests
+      - run: make serve-gmf-apps &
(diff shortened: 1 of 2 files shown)


.github/workflows/main.yaml
```diff
@@ -90,6 +90,7 @@ jobs:
       - run: make check-examples-checker
         if: github.event_name != 'pull_request_target'
       # Cypress tests
+      - run: make serve-gmf-apps &
```
sbrunner, line 93:
```suggestion
      - run: make serve-gmf-apps &
        if: github.event_name != 'pull_request_target'
```

sbrunner (approved):

PR #8464, opened 2022-05-04
Fix the WebComponents source send

sbrunner:


llienher (approved):
Not tested, but looks good !

sbrunner:
@llienher I completely rework it...

llienher (approved):
Thanks !

PR #8469, opened 2022-05-05
Fix the theme loading message spinner in Auth panel

llienher:
Fix both browsers for [GSGMF-1822](https://jira.camptocamp.com/browse/GSGMF-1822)


sbrunner (approved):
I trust you :-)

PR #8502, opened 2022-05-10
Fix z-index of mobile buttons to not be overlapped by swipper element

llienher:
Fix [GSGMF-1829](https://jira.camptocamp.com/browse/GSGMF-1829)


sbrunner (approved):
Thanks :-)

PR #8539, opened 2022-05-17
Fix the i18n by removing the new lines

sbrunner:


.prettierignore
```diff
@@ -1,3 +1,5 @@
 *.min.js
 contribs/gmf/apps/*/index.html.ejs
 src/bootstrap-custom.css
+# New line barke the i18n
```
llienher, line 4:
```suggestion
# New line brake the i18n
```

llienher (approved):
Just a typo.

PR #8548, opened 2022-05-18
Generic LIDAR panel title translation key

sbrunner:


llienher (changes requested):
But it's the best canton :cry: 

Please amend your commit message and PR title as we don't want opinions in the history, otherwise I would just name all my fixes "Great, it's working now". Put something descriptive of the work like "Generic LIDAR panel title translation key", or really whatever that is descriptive...

llienher (approved):

PR #8681, opened 2022-06-21
Restore auto fill form

sbrunner:


sbrunner:
@llienher This time I think that's working, can you confirm?

src/auth/FormElement.ts
```diff
@@ -251,6 +247,7 @@ export default class GmfAuthForm extends GmfBaseElement {
                     type="password"
                     class="form-control"
                     name="oldpassword"
+                    autocomplte="current-password"
```
llienher, line 250:
```suggestion
                    autocomplete="current-password"
```

src/auth/FormElement.ts
```diff
@@ -259,6 +256,7 @@ export default class GmfAuthForm extends GmfBaseElement {
                     type="password"
                     class="form-control"
                     name="newpassword"
+                    autocomplte="new-password"
```
llienher, line 259:
```suggestion
                    autocomplete="new-password"
```

src/auth/FormElement.ts
```diff
@@ -267,6 +265,7 @@ export default class GmfAuthForm extends GmfBaseElement {
                     type="password"
                     class="form-control"
                     name="newpasswordconfirm"
+                    autocomplte="new-password"
```
llienher, line 268:
```suggestion
                    autocomplete="new-password"
```

src/auth/FormElement.ts
```diff
@@ -294,6 +293,7 @@ export default class GmfAuthForm extends GmfBaseElement {
                           autocomplete="off"
                           class="form-control"
                           name="otp"
+                          autocomplete="one-time-code"
```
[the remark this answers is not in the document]
sbrunner: Yes, it's part of recommendations :-)

src/auth/FormElement.ts
sbrunner: Oups, thanks :-)

llienher:
@sbrunner Both mobile and mobile_alt forms are broken and the field are not visible anymore.

llienher:
If I'm not wrong, mobile interface used directly the auth-form component, when desktop used the auth-panel that included the form inside, so the error could be there.

sbrunner:
@llienher I just fix it, probably they are not updated now...
and the CI is failing ...

llienher (approved):
Thanks ! Tested and looks fine ! :-)

PR #8723, opened 2022-07-04
Use the new Transifex client

sbrunner:


Makefile
```diff
@@ -348,23 +348,22 @@ contribs/gmf/build/angular-locale_%.js: package.json
...
-	echo "token =" >> $@
-
-.tx/config: .tx/config.mako .build/python-venv.timestamp
-	PYTHONIOENCODING=UTF-8 $(PY_VENV_BIN)/mako-render \
-		--var "tx_version=$(TX_VERSION)" --var "languages=$(L10N_LANGUAGES)" $< > $@
-
-contribs/gmf/apps/.tx/config: contribs/gmf/apps/.tx/config.mako .build/python-venv.timestamp
-	PYTHONIOENCODING=UTF-8 $(PY_VENV_BIN)/mako-render \
-		--var "tx_version=$(TX_VERSION)" --var "languages=$(L10N_LANGUAGES)" $< > $@
+	echo "token = 1/dc02578696187cc29e4e6486f8611fdbfe60b235" >> $@
```
arnaud-morvan, line 366: Should we commit this token in a public repository ?

Makefile
```diff
@@ -348,23 +348,22 @@ contribs/gmf/build/angular-locale_%.js: package.json
 	$(PY_VENV_BIN)/pip install -r requirements.txt
 	touch $@
 
+.build/applications.timestamp: ci/applications.yaml ci/applications-versions.yaml
+	c2cciutils-download-applications --applications-file=ci/applications.yaml --versions-file=ci/applications-versions.yaml || \
+		(echo You should install c2cciutils >= 1.3 && false)
```
arnaud-morvan, line 353:
```suggestion
		(echo 'You should install c2cciutils : pip3 install --user "c2cciutils>=1.3"'  && false)
```

arnaud-morvan: Note that if checks and/or dependencies vary between versions of c2cciutils, it will be hard to keep a version of c2cciutils on host that match with all projects.

Maybe we should use a venv per project.

But for me we should give information about how to install c2cciutils.

sbrunner: Done :-)

PR #8822, opened 2022-09-06
Be able to have second-level group active when added with FTS

llienher:
Implements [GSGMF-1892](https://jira.camptocamp.com/browse/GSGMF-1892)


sbrunner (approved):
Just add a setTimeout :-)

PR #8845, opened 2022-09-26
Fix external datasource url

arnaud-morvan:
https://jira.camptocamp.com/browse/GSGMF-1897

From WMS specification:
The OnlineResource element within the Service element may be used to refer to the web site of the service provider. There are other OnlineResource elements used for the URL prefix of each supported operation (see below).


arnaud-morvan:
Should we backport to 2.6 ?

arnaud-morvan:
Ouvrir une PR sur le changelog c2cgeoportal 
Vérifier pour le WMTS

sbrunner (approved):

PR #8879, opened 2022-10-20
Remove OL findIndex method with floor component

llienher:


sbrunner:
Desktop alt ref image should be updated :-)

PR #8941, opened 2022-11-23
Use a clearer name for the configuration of SCSS file order

jwkaltz:


buildtools/webpack.commons.js
```diff
@@ -223,7 +223,7 @@ module.exports = function (config) {
         });
         files.sort(
           get_comp(
-            config.fist_scss || [
+            config.scss_order || [
```
sbrunner, line 226: It's only the order of the first SCSS, we have the same for the last SCSS...

sbrunner (changes requested):

sbrunner:
#8942 can be a better and backward-compatible alternative...

PR #9033, opened 2023-03-09
Restore onSuccessulLogin method

llienher:


sbrunner:
I don't understand who it can/should work?
Your idea it to import and use the function directly?

llienher:
@sbrunner Yes, we import the function then do something like that in the controller constructor:

```
    setOnSuccessfulLoginFunction((user) => {
      console.log(user);
      return user;
    });
```

sbrunner (approved):

PR #9084, opened 2023-04-03
Handle error on failing CSV export in LiDAR panel

llienher:


sbrunner:
Why remove the error message?

ochriste (approved):

ochriste:
@sbrunner because the try/catch hide completly the error in case of error. 
so we either get ride of the try/catch completly, 
or we manually throw an error and pass the original error stack, so can can know whats wrong and debug this without spending 3 days setting up a projects in developement mode using the customer data

sbrunner:
Then we will no more have a user message, not problem for me :-)

PR #9089, opened 2023-04-14
Clear the WebGL rendering correctly

llienher:
In the context of multi-canvas use, the clear is not runned by default like in the OL example that was missing the clear command and was working because using only one single canvas (in such case it seems that the clear is done anyway). So Olivia is fixing the OL example and here is the fix for the project.


sbrunner:
Should be rebased :-)

sbrunner (approved):

sbrunner (approved):

PR #9091, opened 2023-04-18
Fix blurry WMS and WMTS layers

sbrunner:
Server part: https://github.com/camptocamp/c2cgeoportal/pull/10476

### DIFF
--- a/src/options.js
+++ b/src/options.js
@@ -1,6 +1,6 @@
 // The MIT License (MIT)
 //
-// Copyright (c) 2020-2022 Camptocamp SA
+// Copyright (c) 2020-2023 Camptocamp SA
 //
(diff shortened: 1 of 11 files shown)


src/options.js
```diff
@@ -657,6 +657,13 @@ export function buildStyle(styleDescriptor) {
  * @property {olTileGridOptions} tiles grid options.
  */
 
+/**
+ * Configuration options for the VectorTiles layers.
```
arnaud-morvan, line 661:
```suggestion
 * Configuration options for the WMS sources.
```

arnaud-morvan (approved):

src/options.js
sbrunner: Oups thanks :-)

PR #9132, opened 2023-05-25
End of support of version 2.4

sbrunner:


llienher:
@sbrunner Should not be <= 2.5 ? As 2.6 is marked as best effort, it should not be in the comparator.

PR #9140, opened 2023-06-16
Constraint map pixel on grid pixel

sbrunner:


llienher (approved):
I'm confident with your testing because I did not. :-)

PR #9146, opened 2023-06-22
Fix the center after the zoom

sbrunner:


src/controllers/AbstractAppController.js
```diff
@@ -171,6 +171,15 @@ export function AbstractAppController($scope, $injector, mobile) {
       this.options.map
     )
   );
+  // To fix the map position constraint, we need to set the center
+  let onModeEnd = false;
+  map.on('moveend', () => {
+    if (!onModeEnd) {
+      onModeEnd = true;
```
llienher, line 178: ModeEnd or MoveEnd ?

sbrunner: Ouops, thanks :-)

llienher (commented):
Is it that the onMoveEnd is always false anyways so it makes the use of it not necessary ? If it is set to false as soon it is set to true, so it's always checks the condition.

If I undestand correctly, we would need something like that ?
`   map.on('moveend', () => view.setCenter(view.getCenter()));`

sbrunner:
No, it's not always false this is to avoid infinite loop...

llienher (approved):

PR #9152, opened 2023-06-30
Only load the theme when a private layer is present when log in (flush mode)

llienher:


src/controllers/AbstractAppController.js
```diff
@@ -332,6 +332,13 @@ export function AbstractAppController($scope, $injector, mobile) {
     );
   });
 
+  let hasPrivateLayers = false;
+  user.getLoginMessage().next((loginMessage) => {
+    if (loginMessage == loginMessageRequired) {
+      this.hasPrivateLayers = true;
+    }
```
sbrunner, line 339:
```suggestion
    this.hasPrivateLayers = loginMessage == loginMessageRequired;
```

sbrunner (approved):

PR #9172, opened 2023-07-21
Restore the queryCountFirst=false functionality

sbrunner:
Fix for https://jira.camptocamp.com/browse/GEO-6051


srcapi/store/config.ts
```diff
@@ -277,6 +277,7 @@ export type ngeoQueryOptions = {
   /**
    * For WFS sources, should the number of features first be
    * requested with `resultType=hits` before requesting the actual features in an seconds request?
+   * Should be use on window result but not on grid result.
```
llienher, line 280:
```suggestion
   * Should be used on window result but not on grid result.
```

llienher (approved):
Just a typo, but looks good.

srcapi/store/config.ts
sbrunner: Thanks :-)

PR #9266, opened 2023-12-11
Be able to set the arcgisWMSInfoFormat per ogcServer

sbrunner:


srcapi/store/config.ts
```diff
@@ -1382,12 +1382,20 @@ export type gmfCustomCSS = {
...
+   * The WMS info format to be used with Arcgis OGC server.
+   */
+  arcgisWMSInfoFormat: string;
+};
 export type gmfDatasourceOptions = {
   /**
    * The WMS info format to be used with Arcgis OGC server.
    */
   arcgisWMSInfoFormat: string;
+  ogcServerSpecific: {[x: string]: ogcServerDatasourceOptions};
```
ochriste, line 1396: just to be sure, is it normal that one is ogcServerSpecific (Server without "s") while the property name in src/datasource/Manager.js is ogcServersSpecific (Servers with a "s") ?

sbrunner: Effectively thanks, fixed :-)

ochriste (approved):

llienher (approved):

PR #9294, opened 2024-02-08
Add came_from URL in the login, useful for 2FA

sbrunner:


src/auth/service.ts
```diff
@@ -228,7 +230,12 @@ export class AuthenticationService {
    */
   login(login: string, pwd: string, otp: string = undefined): Promise<void | User> {
     const url = `${this.baseUrl_}/${RouteSuffix.LOGIN}`;
-    const params = {'login': login, 'password': pwd};
+    const params = {
+      'login': login,
+      'password': pwd,
+      came_from:
+        user.getLoginMessage().value == loginMessageRequired ? this.originalUrl : window.location.href,
```
llienher, line 237: Tell me if I'm wrong, but does this.originalUrl can change somewhere else ? As I understand it is initialized as window.location.href. 

If the condition is true or false, the end result is window.location.href from this.originalUrl or window.location.href so the same value ?

llienher: I don't know if the commant will explain better than the full answer... :-)

llienher (approved):

src/auth/service.ts
llienher: I guess the full answer explains it.

Now the comment you wrote, not so sure. :-)

sbrunner: And with this comment?

llienher: Good, thanks !

PR #9541, opened 2024-11-26
Remove force center after view animation.

ger-benjamin:
For https://camptocamp.atlassian.net/browse/GSSCHWYZ-429
And https://github.com/camptocamp/GeoMapFish/issues/92

But will break (origin): https://camptocamp.atlassian.net/browse/GSGMF-2009
Who looks to be a issue on a particular GMF.

This workaround was making the map less blurry in case of map height with a not round number of pixels, but it cause the map zoom to "jump" if the view.contrainResolution is set to true.

My idea is to trash this function, and let the concerned GMF implement it on need.

Another possibility would be keep the function in the code but not active by default (let the user choose if he want to use it).

Or to add a options in the vars to activate it or not.

If we keep it we could also add a debounce to to not recenter too quickly.

[Examples](https://camptocamp.github.io/ngeo/refs/pull/9541/merge/examples/)
[Storybook](https://camptocamp.github.io/ngeo/refs/pull/9541/merge/storybook/)
[API help](https://camptocamp.github.io/ngeo/refs/pull/9541/merge/api/apihelp/apihelp.html)
[API documentation](https://camptocamp.github.io/ngeo/refs/pull/9541/merge/apidoc/)

### DIFF
--- a/src/controllers/AbstractAppController.js
+++ b/src/controllers/AbstractAppController.js
@@ -61,6 +61,7 @@ import gmfAuthenticationService from 'ngeo/auth/service';
 import ngeoMapFeatureOverlayMgr from 'ngeo/map/FeatureOverlayMgr';
 import storeMap from 'gmfapi/store/map';
 import user, {UserState, loginMessageRequired} from 'gmfapi/store/user';
+import {debounce} from 'gmf/misc/debounce2';
 
 AbstractAppController.$inject = ['$scope', '$injector', 'mobile'];
 /**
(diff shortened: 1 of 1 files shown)


sbrunner:
This code is required to don't get bury tiles, this should be activated by default, we can add an option to be able to disable it.

ger-benjamin:
I've tried several on different configuration to reproduce the "blurry effect", even with the commit before the original fix, but I was not able to see any difference. If we had a map quality issue with OpenLayers 7, there are good chance that it's fixed now. Therefor, I would still recommend to remove this function "setCenter" function.
And with this new code, it's now possible to remove this function at implementation if wanted/needed.

However, as I can't give guaranty that the blur effect is gone, I've added a debounce around the original fix. It allows to keep this fix, and avoid to recenter too frequently during the zoom (avoid the jumping-around effect).

sbrunner (approved):
Nice thanks :-)

maltaesousa:
Thanks @ger-benjamin for your tests! ping @gnerred

ger-benjamin:
Don't hesitate to open again the issue if needed (but I'm pretty confident on my fix).
You can take the new version of ngeo 2.8, 2.9 and master with the fix in some hours (tomorrow to be sure).

gnerred:
Thanks @ger-benjamin and @maltaesousa , I will test it asap.

gnerred:
I've upgraded our Cartoriviera instance to 2.8.1.170, and the problem is fixed. 

https://map.cartoriviera.ch 

Thanks a lot !

PR #10149, opened 2026-07-02
Add DebouncedImageWMS source to prevent duplicate GetMap requests on zoom

sbrunner:
## Description

When animated zoom ends, the custom center constraint (fixCenterOnGrid) in AbstractAppController snaps the view center to the WMTS grid. This triggers a second GetMap request with a nearly identical bbox as the one already fired during the animation.

## Solution

New `DebouncedImageWMS` source (subclass of `ol/source/ImageWMS`) that debounces loading when the view changes rapidly during animated zoom or pan:

1. On cache miss, starts a timer (configurable via `debounceDelay` in `gmfWMSSourceOptions`)
2. If the view stabilizes before the timer expires, a single request is made with the final view state
3. If the view changes again, the timer resets

## How to activate

Set `debounceDelay` in your app's Angular constant:
```js
module.constant('gmfWMSSourceOptions', { debounceDelay: 100 });
```

## Files changed

- `src/source/DebouncedImageWMS.js` — new source with debounce logic
- `src/map/LayerHelper.js` — conditionally use DebouncedImageWMS when debounceDelay is set
- `src/options.js` — document the new debounceDelay option

GSGEOC-463

See JIRA issue: [GSGEOC-463](https://jira.camptocamp.com/browse/GSGEOC-463).
[Examples](https://camptocamp.github.io/ngeo/refs/pull/10149/merge/examples/)
[Storybook](https://camptocamp.github.io/ngeo/refs/pull/10149/merge/storybook/)
[API help](https://camptocamp.github.io/ngeo/refs/pull/10149/merge/api/apihelp/apihelp.html)
[API documentation](https://camptocamp.github.io/ngeo/refs/pull/10149/merge/apidoc/)

[GSGEOC-463]: https://camptocamp.atlassian.net/browse/GSGEOC-463?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/map/LayerHelper.js
+++ b/src/map/LayerHelper.js
@@ -28,6 +28,7 @@ import olLayerImage from 'ol/layer/Image';
 import olLayerLayer from 'ol/layer/Layer';
 import olLayerVectorTile from 'ol/layer/VectorTile';
 import {isEmpty} from 'ol/obj';
+import DebouncedImageWMS from 'ngeo/source/DebouncedImageWMS';
 import olSourceImageWMS from 'ol/source/ImageWMS';
 import olSourceTileWMS from 'ol/source/TileWMS';
 import olSourceVectorTile from 'ol/source/VectorTile';
@@ -181,7 +182,8 @@ LayerHelper.prototype.createBasicWMSLayer = function (
   ) {
     options.hidpi = false;
   }
-  const source = new olSourceImageWMS(options);
+  const validDebounceDelay = options.debounceDelay && options.debounceDelay >= 0;
+  const source = validDebounceDelay ? new DebouncedImageWMS(options) : new olSourceImageWMS(options);
   if (opt_params) {
     source.updateParams(opt_params);
   }
--- a/src/options.js
+++ b/src/options.js
@@ -658,6 +658,10 @@ export function buildStyle(styleDescriptor) {
  *
  * @typedef {Object} gmfWMSSourceOptions
  * @property {number} ratio Ratio. 1 means image requests are the size of the map viewport, 2 means twice the width and height of the map viewport, and so on. Must be 1 or higher (non integer value make the layers blurry).
+ * @property {number} [debounceDelay] Delay in milliseconds before sending a WMS GetMap request when the view changes.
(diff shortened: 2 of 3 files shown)


arnaud-morvan (commented):
I suppose that in some way this follow analyse in https://camptocamp.atlassian.net/browse/GSAGG-29

I agree that a GetMap request before and after `fixCenterOnGrid` creates unnecessary load on WMS server, and it's a real good idea to avoid such behavior.

I find it a shame to have so much code duplication, and I'm not sure a debounce is the correct solution to fix such issue with `fixCenterOnGrid`.

I'm a little worried about maintaining the duplicated code, especially `createLoader` which may diverge in later versions of OL.

Would it be possible to avoid duplicated rendering directly in `fixCenterOnGrid` to avoid duplicating OL code in `DebouncedImageWMS` ?

arnaud-morvan:
Claude proposed : 

```typescript
  AbstractAppController.prototype.fixCenterOnGrid = function (view) {
    const finalize = debounce(() => {
      view.setCenter(view.getCenter());               // snap (isMoving=false)
      this.originalViewSetHint.call(view, ViewHint.ANIMATING, -1); // libère le renderer
    }, 50);
    view.setHint = (hint, value) => {
      if (hint === ViewHint.ANIMATING && value === -1) {
        finalize();          // on retient le -1, le renderer reste "en animation"
        return;
      }
      this.originalViewSetHint.call(view, hint, value);
    };
  };
```

But to be honest I do not understood.

arnaud-morvan:
After some inspection, cannot we just remove `fixCenterOnGrid`.

Is it really useful, still not understood why we do this. Could we have a look on the initial issue, was it really useful, and how we may work with it.

sbrunner:
the Claude patch is catastrophic, sometimes the map didn't zoom, and when she zoom we get a flash, it we completely remove the `fixCenterOnGrid` I'm pretty sure that the map is no more pixel perfect in all cases!

arnaud-morvan (approved):
Ok so we may give it a try, as far as this is purely optional.

